PR #2, opened 2022-04-06
updating readme with homebrew postgres

allegraholland:
added install directions for homebrew

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ mf health-checks # confirm your datawarehouse connection is working
 Run the tutorial:
 
 ```
-mf tutorial # optionally add `--skip-ds` if you have already confirmed your datawarehouse connection works
+mf tutorial # optionally add `--skip-dw` if you have already confirmed your datawarehouse connection works
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,6 +53,14 @@ If you’re new to MetricFlow, we suggest that you review our [Getting Started](
 
 Install MetricFlow:
 
+If you do not have postgres on your machine, first install postgres with homebrew:
+
+```
+brew install postgresql
+```
```
tlento, line 60: This only works on Macs that have Homebrew installed. I recommend something more like this:

```suggestion
If you do not have postgres on your machine, first install Postgres:
- Postgres provides [pre-built packages for download and installation](https://www.postgresql.org/download/)
- Mac users might prefer to use Homebrew: `brew install postgresql`
```

tlento (approved):
See inline suggestion - we shouldn't be assuming Homebrew is available in the generic user-facing instructions.

We might also move this down into troubleshooting or something, assuming the error is consistent, but I think that's a change for later.

PR #5, opened 2022-04-06
[Attribution] Remove staging

belindabennett:
Removing `staging` from previous staging repo.

### DIFF
--- a/attribution.md
+++ b/attribution.md
@@ -14,7 +14,7 @@ The following 3rd-party software packages may be used by or distributed with **g
 
 Date generated | Revision ID
 :------------: | :----------:
-04/06/22 | ac34affefc1ff90b740f42098731dad1a540f64b
+04/06/22 | dc975d95150a263d23c74a660f326e35925be662
 
 ---
 
--- a/attribution.md
+++ b/attribution.md
@@ -5,11 +5,11 @@ We're grateful to be part of the open source community, and to be able to build 
 
 ---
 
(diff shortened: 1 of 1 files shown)


marcodamore (approved):

attribution.md
```diff
@@ -5,11 +5,11 @@ We're grateful to be part of the open source community, and to be able to build
 
 ---
 
-# 3rd-Party Software for [git@github.com:transform-data/metricflow-staging.git]()
+# 3rd-Party Software for [git@github.com:transform-data/metricflow.git]()
 
 
 
-The following 3rd-party software packages may be used by or distributed with **git@github.com:transform-data/metricflow-staging.git**.  Any information relevant to third-party vendors listed below are collected using common, reasonable means.
+The following 3rd-party software packages may be used by or distributed with **git@github.com:transform-data/metricflow.git**.  Any information relevant to third-party vendors listed below are collected using common, reasonable means.
```
tlento, line 12: On line 17 there's an invalid commit sha (`ac34affefc1ff90b740f42098731dad1a540f64b`). We should probably replace it with this one:

`dc975d95150a263d23c74a660f326e35925be662`

I ran FOSSA on that and it seemed to be the same (a few dependencies got dropped out), so I'm comfortable with just this for now. We need to find a better process for managing this file, though.

tlento (commented):
Let me see if I can get FOSSA to spit out a not-broken markdown replacement for this now, because that will be more appropriate than hand-modifying the metadata.

PR #18, opened 2022-04-07
Fix non-deterministic failures in test_cross_element_names

### LINKED ISSUE #17: Validate unique names test fails in CI
**Describe the bug**
See #16 for the failure. This seems to happen a lot in CI and not at all locally.

**Steps To Reproduce**
1. Make a trivial change
2. Put up a PR
3. wait for failure
This is apparently not deterministic

**Expected behavior**
This test should pass

**Environment (please complete the following information):**

- MetricFlow Version [e.g. 1.0.0]

tlento:
This test was failing in CI sometimes, and began to fail quite persistently,
due to an issue with the way we were generating invalid models.

The core issue is that the invalid model element is being generated by
mutating an existing valid model to conform to the error. For cross-element
name checks we had to rename a dimension, but if you rename a primary
time dimension the model validator fails with the wrong error.

To sidestep this problem for the time being we simply restrict the mutated
dimension to one of type CATEGORICAL. Longer term we should create very simple
invalid models that match the validation error types, or models with parameterized
names so we can pass concrete values to the template to generate the errors we need.

Fixes #17

### DIFF
--- a/metricflow/test/model/validations/test_unique_valid_name.py
+++ b/metricflow/test/model/validations/test_unique_valid_name.py
@@ -1,7 +1,10 @@
+from typing import Tuple
 import pytest
 import copy
 
 from metricflow.model.model_validator import ModelValidator
+from metricflow.model.objects.data_source import DataSource
+from metricflow.model.objects.elements.dimension import Dimension, DimensionType
 from metricflow.model.validations.validator_helpers import ModelValidationException
 from metricflow.model.objects.user_configured_model import UserConfiguredModel
 from metricflow.model.validations.unique_valid_name import UniqueAndValidNameRule
@@ -11,6 +14,10 @@ from metricflow.test.test_utils import find_data_source_with
 
 def copied_model(simple_model__pre_transforms: UserConfiguredModel) -> UserConfiguredModel:  # noqa: D
     return copy.deepcopy(simple_model__pre_transforms)
+
+
+def _categorical_dimensions(data_source: DataSource) -> Tuple[Dimension, ...]:
+    return tuple(dim for dim in data_source.dimensions if dim.type == DimensionType.CATEGORICAL)
 
 
 """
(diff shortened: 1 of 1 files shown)


lebca:
> This test was failing in CI sometimes, and began to fail quite persistently, due to an issue with the way we were generating invalid models.
> 
> The core issue is that the invalid model element is being generated by mutating an existing valid model to conform to the error. For cross-element name checks we had to rename a dimension, but if you rename a primary time dimension the model validator fails with the wrong error.
> 
> To sidestep this problem for the time being we simply restrict the mutated dimension to one of type CATEGORICAL. Longer term we should create very simple invalid models that match the validation error types, or models with parameterized names so we can pass concrete values to the template to generate the errors we need.
> 
> Fixes #17

Can we move the categorical function into `test_utils.py`? Otherwise, LGTM. We could further improve by deliberately building a model and using specific names.

metricflow/test/model/validations/test_unique_valid_name.py
```diff
@@ -13,6 +16,10 @@ def copied_model(simple_model__pre_transforms: UserConfiguredModel) -> UserConfi
     return copy.deepcopy(simple_model__pre_transforms)
 
 
+def _categorical_dimensions(data_source: DataSource) -> Tuple[Dimension, ...]:
```
lebca, line 19: move to `test/test_utils.py`

lebca (approved):

plypaul (approved):
This works, but thinking that there are benefits to more deterministic test cases.

metricflow/test/model/validations/test_unique_valid_name.py
tlento: I prefer to leave this where it is and open a separate issue to rewrite this whole test to use purpose-built broken models instead of doing all of this mutation. As it is currently written the test is exceedingly difficult to understand, and for all I know there are more of these non-deterministic code gremlins lurking in the shadows and we've just been lucky.

PR #28, opened 2022-04-08
Fix the format of arguments sent to two instantiations of MeasureReference

lebca:
The wrong argument was sent to the `element_name` of `MeasureReference` in two instances within the TimeGranularitySolver object. The Reference was sent and not the string name.

Solution: send the string name of the Reference, instead of the Reference.


tlento (approved):
I don't understand why the typechecker ever allowed this....

PR #31, opened 2022-04-09
fix: bigquery read credential issue in __initialize_sql_client

zzsza:
### Context
When I run `mq health-checks` after specifying bigquery in `mf setup`, 
"json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 2 column 1 (char 2)" Error occurred

I checked the code and found that cred_file.readline() in cli_context.py returns "{". 

This issue seems to be due to the assumption of a specific form of Google Cloud Platform Service Account Key.

Usually, Service Account Key is configured as follows:

```
{
  "type": "service_account",
  "project_id": "project_id",
  "private_key_id": "private_key",
  "private_key": "-----BEGIN PRIVATE KEY-----~~~",
  "client_email": "mail",
  "client_id": "123123",
  "auth_uri": "~~~~~~~",
  "token_uri": "~~~~~~~",
  "auth_provider_x509_cert_url": "~~~~~~~~",
  "client_x509_cert_url": "~~~~~~~~~"
}
```

To fix this issue, we need to change to `cred_file.read()` instead of `cred_file.readline()`

**The [readline()](https://docs.python.org/3.8/tutorial/inputoutput.html#methods-of-file-objects) function reads only the first line, whereas the read() function is a function that reads all.**

Probably, when developing, it was assumed that the Service Account JSON file was saved as one line as follows:
```
{"type": "service_account", "project_id": "project_id", "private_key_id": "private_key", "private_key": "-----BEGIN PRIVATE KEY-----~~~", "client_email": "mail"}
```

If modify the code with read(), the code runs normally no matter how the Service Account JSON file is saved.

### Screenshots:
If I run the code before modifying it, I get something like this:
스크린샷 2022-04-10 오전 1 48 23

If I fix the code it runs fine:

스크린샷 2022-04-10 오전 1 47 57

## Types of Changes
- [x] Bug fix (non-breaking change that fixes an issue)

### DIFF
--- a/metricflow/cli/cli_context.py
+++ b/metricflow/cli/cli_context.py
@@ -88,7 +88,7 @@ class CLIContext:
                 if not pathlib.Path(path_to_creds).exists:
                     raise ValueError(f"`{path_to_creds}` does not contain the BigQuery credential file.")
                 with open(path_to_creds, "r") as cred_file:
-                    creds = cred_file.readline()
+                    creds = cred_file.read()
                 self._sql_client = BigQuerySqlClient(password=creds)
             elif dialect == SupportedSqlEngine.SNOWFLAKE.name:
                 host = not_empty(self.config.get_config_value(CONFIG_DWH_HOST), CONFIG_DWH_HOST, url)


plypaul (approved):
LGTM - thanks for the contribution @zzsza!

tlento:
Hi @zzsza ! It appears we forgot to ask you to sign our contributor license agreement. Could you please review the agreement (which has not changed since you made this contribution) and sign it here? You may sign by leaving a comment on this PR with something like the following:

>  I have read the CLA document, recognize that it covers this and future contributions, and hereby sign the CLA.

Agreement: https://github.com/transform-data/metricflow/blob/main/TransformCLA.md

Sorry for such a late follow-up, and thank you so much for your contribution!

zzsza:
I have read the CLA document, recognize that it covers this and future contributions, and hereby sign the CLA.

PR #33, opened 2022-04-12
Sync package version handling

WilliamDee:
## Context
We are using `pyproject.toml` to package the library to publish to pypi. Since the `.toml` wasn't able to read from another file and the cli code couldn't reach outside the top-level directory, we needed a way to sync the current version of the library without having to hardcode the version on both `pyproject.toml` and `metricflow/cli/__init__.py`. Instead, we can aggregate the current package version using `importlib.metadata.version` in the CLI. This allows us to keep the hardcoded version in 1 file which is inside `pyproject.toml`

## Changes
- Removed version in `metricflow/cli/__init__.py`
- Use `importlib.metadata` to get current package version in CLI
- Added test for version CLI command

### DIFF
--- a/metricflow/cli/__init__.py
+++ b/metricflow/cli/__init__.py
@@ -1,2 +1,1 @@
-__version__ = "0.91.9"
 PACKAGE_NAME = "metricflow"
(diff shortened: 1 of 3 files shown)


tlento (approved):
Nice!

PR #34, opened 2022-04-12
Adding Materialization checks to Validations

lebca:
Problem: Materializations did not have validation coverage

Solution: Added coverage within the `ModelValidator` for Materializations. In addition, we altered some of the attributes of Materialization to default to empty list instead of None to avoid returning null when returning a collection. In addition, some supplemental methods were added to other objects for easier usage of them.

### DIFF
--- a/metricflow/model/validations/materializations.py
+++ b/metricflow/model/validations/materializations.py
@@ -1,0 +1,116 @@
+import logging
+from typing import List
+
+from metricflow.errors.errors import UnableToSatisfyQueryError
+from metricflow.model.objects.materialization import Materialization
+from metricflow.model.objects.user_configured_model import UserConfiguredModel
+from metricflow.model.semantic_model import SemanticModel
+from metricflow.model.semantics.data_source_container import PydanticDataSourceContainer
+from metricflow.model.semantics.semantic_containers import DataSourceSemantics
+from metricflow.naming.linkable_spec_name import StructuredLinkableSpecName
+from metricflow.query.query_parser import MetricFlowQueryParser
+from metricflow.model.validations.validator_helpers import (
+    ModelValidationRule,
+    ValidationIssue,
+    ValidationError,
+    ValidationIssueType,
+    validate_safely,
+)
+from metricflow.specs import TimeDimensionReference
+
+logger = logging.getLogger(__name__)
+
+
+class ValidMaterializationRule(ModelValidationRule):
+    """Check that a materialization config is valid.
+
+    * Metrics listed are valid.
+    * Dimensions or identifiers are valid for the listed metrics.
+    * Primary time dimension is listed in the dimensions section.
+    """
+
+    @staticmethod
+    @validate_safely(whats_being_done="checking materialization is defined correctly")
+    def _validate_materialization(
+        materialization: Materialization,
+        primary_time_dimensions_reference: TimeDimensionReference,
+        mf_query_parser: MetricFlowQueryParser,
+    ) -> List[ValidationIssueType]:
+        issues: List[ValidationIssueType] = []
+
+        if not materialization.dimensions:
+            return [
+                ValidationError(
+                    model_object_reference=ValidationIssue.make_object_reference(
+                        materialization_name=materialization.name
+                    ),
+                    message=f"Materialization '{materialization.name}' does not have dimensions listed",
+                )
+            ]
+
+        try:
+            mf_query_parser.parse_and_validate_query(
+                metric_names=materialization.metrics,
+                group_by_names=materialization.dimensions,
+            )
+        except UnableToSatisfyQueryError as err:
+            issues.append(
+                ValidationError(
+                    model_object_reference=ValidationIssue.make_object_reference(
+                        materialization_name=materialization.name
+                    ),
+                    message=str(err),
+                )
+            )
+
+        # Primary dimension checks
+        mat_primary_time_dimension_names: List[str] = []
+        for mat_dimension_or_identifier_name in materialization.dimensions:
+            structured_spec = StructuredLinkableSpecName.from_name(mat_dimension_or_identifier_name)
+            if structured_spec.element_name == primary_time_dimensions_reference.element_name:
+                mat_primary_time_dimension_names.append(mat_dimension_or_identifier_name)
+
+        if len(mat_primary_time_dimension_names) == 0:
+            issues.append(
+                ValidationError(
+                    model_object_reference=ValidationIssue.make_object_reference(
+                        materialization_name=materialization.name
+                    ),
+                    message=f"Primary time dimension {primary_time_dimensions_reference.element_name} not listed"
+                    f" as a dimension in materialization {materialization.name}",
+                )
+            )
+
+        if len(mat_primary_time_dimension_names) > 1:
+            issues.append(
+                ValidationError(
+                    model_object_reference=ValidationIssue.make_object_reference(
+                        materialization_name=materialization.name
+                    ),
+                    message=f"Multiple primary time dimensions {mat_primary_time_dimension_names} listed in "
+                    f"materialization {materialization.name}",
+                )
+            )
+
+        return issues
+
+    @staticmethod
+    @validate_safely(whats_being_done="running model validation ensuring materializations are valid")
+    def validate_model(model: UserConfiguredModel) -> List[ValidationIssueType]:  # noqa: D
+        """Check that all of the metrics and dimensions listed in a materialization are valid."""
+        issues: List[ValidationIssueType] = []
+        ds_semantics = DataSourceSemantics(
+            model=model, configured_data_source_container=PydanticDataSourceContainer(model.data_sources)
+        )
+        primary_time_dimensions_reference = ds_semantics.primary_time_dimension_reference
+        mf_query_parser = MetricFlowQueryParser(
+            model=SemanticModel(user_configured_model=model),
+            primary_time_dimension_reference=primary_time_dimensions_reference,
+        )
+        for materialization in model.materializations:
+            issues += ValidMaterializationRule._validate_materialization(
+                materialization=materialization,
+                primary_time_dimensions_reference=primary_time_dimensions_reference,
+                mf_query_parser=mf_query_parser,
+            )
+        return issues
--- a/metricflow/test/model/validations/test_materializations.py
+++ b/metricflow/test/model/validations/test_materializations.py
@@ -1,0 +1,135 @@
+import logging
+
+from metricflow.model.objects.materialization import Materialization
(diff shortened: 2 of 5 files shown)


QMalcolm (approved):
This looks good to me. 

Something to note for anybody looking, the function which actually validations a given metric, `_validate_materialization` is a CHUNGUS of a validation. This is somewhat unwieldy to actually reason about when making changes and also means that not all parts of a given materialization might get validated if an exception gets raised. The solution is to refactor the `_validate_materialization` into distinct separate functions. However, this PR basically brings in the rule we had already written. Refactoring it is on the road map, just as a separate topic issue.

metricflow/time/time_granularity_solver.py
```diff
@@ -253,6 +268,80 @@ def adjust_time_range_to_granularity(
...
+    ) -> TimeGranularityValidationReference:
+        """Figure out how to transform the time granularity in a query."""
+
+        primary_time_dimension_spec: TimeDimensionSpec = self._get_primary_time_dimension()
+        min_time_granularity_for_querying = self._get_greatest_common_time_granularity(metrics=metric_specs)
+
+        requested_time_granularity = None
+        for time_dimension_spec in time_dimension_specs:
+            if primary_time_dimension_spec == time_dimension_spec:
+                if time_dimension_spec.time_granularity < min_time_granularity_for_querying:
```
tlento, line 311: I am confused about these nested conditionals. Why are we only checking the primary time dimension for the model? Also, I always get confused by this and it's something we should address. I think the issue here is that we've pre-materialized and now the lower granularity is effectively a request for a granularity we can no longer serve, is that correct? If so, we should probably find a way to document this more clearly, because in this location in the codebase it's not at all obvious why a lower granularity on the primary time dimension is an issue.

Also worth noting, if this is targeted at model validation, it should probably be separated out from the solution steps and this method should be targeted more directly at model validation outcomes. If this is targeted at finding a granularity transformation, I'd expect this logic to fail if any requested time dimension was incorrect.

metricflow/test/test_utils.py
```diff
@@ -63,3 +65,52 @@ def find_metric_with(model: UserConfiguredModel, function: Callable[[Metric], bo
...
     raise Exception("Unable to find a metric matching function criteria")
+
+
+def model_with_materialization(
+    initial_model: UserConfiguredModel, materializations: List[Materialization]
+) -> UserConfiguredModel:
+    """Returns a copy of the model but with the materialization added"""
+    model_copy = copy.deepcopy(initial_model)
+    # model_copy.materializations = []
+    model_copy.materializations.extend(materializations)
```
tlento, line 76: I don't like this mechanism. It makes it hard to reason about what's going on when something in the test is broken, because we have a model but then we mutate the instance to be a different model but the same model but it's also a different copy.

That said, we should not be trying to fix this here.

metricflow/test/model/validations/test_materializations.py
```diff
@@ -0,0 +1,280 @@
...
+    """Test to ensure the primary time dimension is used in materialization destination"""
+    model = copy.deepcopy(simple_model__pre_transforms)
+    func_for_materialization: Callable[[Materialization], bool] = (
+        lambda materialization: len(materialization.destinations) > 0
+    )
+    materialization_with_destinations, _ = find_materialization_with(model, func_for_materialization)
+    assert type(materialization_with_destinations) == Materialization
+
+    # We assume the primary time dimension is `ds`
+    primary_time = "ds"
```
tlento, line 192: So I get it, but this should probably be fixed at some point. At the very least, update the comment to reflect that the primary time dimension in the simple_model is, in fact, called "ds"

metricflow/test/model/validations/test_materializations.py
```diff
@@ -0,0 +1,280 @@
...
+                        Materialization(
+                            name="materialization_test_case",
+                            metrics=["revenue"],
+                            dimensions=["ds__hour"],
+                        )
+                    ],
+                )
+            )
+        )
+        == 2
```
tlento, line 176: Two? I cannot figure out where the second one comes in. What am I missing?

metricflow/test/model/validations/test_materializations.py
```diff
@@ -0,0 +1,280 @@
...
+                new_rollups.append(new_rollup_list)
+            destination.rollups = new_rollups
+
+    mat.dimensions = [
+        dim
+        for dim in mat.dimensions
+        if StructuredLinkableSpecName.from_name(dim).qualified_name_without_granularity
+        != primary_time.qualified_name_without_granularity
+    ]
+    mat.dimensions.append(primary_time.qualified_name)
```
tlento, line 250: Why do we need this validation in the first place? Is it because materialization dimension references are user-provided dundered strings? If so, ick, but I guess it's what we're doing right now so we have to roll with it. I would prefer to just not give users the opportunity to forget to append granularity. Can we somehow have a standard validation pass that does this?

Also, we probably don't want to special-case primary time there. We need every time dimension to have granularity specified, right?

metricflow/model/validations/materializations.py
```diff
@@ -0,0 +1,315 @@
...
+                                f"listed as dimensions in the materialization '{materialization.name}'",
+                            )
+                        )
+        return issues
+
+    @staticmethod
+    @validate_safely(whats_being_done="running model validation ensuring materialization rollups are valid")
+    def validate_model(model: UserConfiguredModel) -> List[ValidationIssueType]:  # noqa: D
+        issues = []
+        for materialization in model.materializations or []:
```
tlento, line 311: nit
```suggestion
        for materialization in model.materializations:
```

metricflow/test/model/validations/test_materializations.py
```diff
@@ -0,0 +1,280 @@
...
+            )
+        )
+        == 2
+    )
+
+
+def test_missing_primary_time_dimension_in_rollups(
+    simple_model__pre_transforms: UserConfiguredModel,
+) -> None:  # noqa: D
+    """Test to ensure the primary time dimension is used in materialization destination"""
```
tlento, line 183:
```suggestion
) -> None: 
    """Ensure validation enforces the primary time dimension is used in the materialization destination"""
```

metricflow/time/time_granularity_solver.py
```diff
@@ -253,6 +268,80 @@ def adjust_time_range_to_granularity(
 
         return TimeRangeConstraint(start_time=constraint_start, end_time=constraint_end)
 
+    def _get_primary_time_dimension(self) -> TimeDimensionSpec:  # noqa: D
+        for data_source in self._semantic_model.user_configured_model.data_sources:
+            for dimension in data_source.dimensions:
+                if dimension.type_params:
+                    if dimension.type_params.time_granularity:
+                        if dimension.type_params.is_primary:
+                            return TimeDimensionSpec.parse(dimension.name.element_name)
```
tlento, line 277: Let's not duplicate this dimension walk here. You can do the following, or you can just delete this method and inline the conversion to TimeDimensionSpec down below in the one place where you use this.

```suggestion
        primary_time_dimension_reference = self._semantic_model.data_source_semantics.primary_time_dimension_reference:
        return TimeDimensionSpec.parse(primary_time_dimension_reference.element_name)
```

metricflow/time/time_granularity_solver.py
```diff
@@ -253,6 +268,80 @@ def adjust_time_range_to_granularity(
...
+        # Shouldn't happen as validation should have caught this case.
+        raise RuntimeError("No primary time dimension found in the model.")
+
+    def _get_greatest_common_time_granularity(self, metrics: Sequence[MetricSpec]) -> TimeGranularity:
+        """Return the greatest common granularity for metrics"""
+        primary_time_dimension_reference = self._get_primary_time_dimension().reference
```
tlento, line 283:
```suggestion
        primary_time_dimension_reference = self.semantic_model.data_source_semantics.primary_time_dimension_reference
```

tlento (commented):
I'm on board with getting this validation in quickly, but we should have a plan for a follow-up pass on this logic because it is confusing.

I've noted a lot of smaller areas for cleanup inline, which I leave to you for either immediate update or a follow-up PR. Beyond those, I think the big three restructuring questions for our next pass at this are probably these:

1. Do we need this new public method in the TimeGranularitySolver at all, and if so, how can we make it something humans can reason about a bit more easily?
2. We need to streamline the validation logic itself, as @QMalcolm has already noted
3. Can we eliminate the "has a granularity" checking entirely and offload that to something that enforces all dimension references have a granularity? And if not, is it appropriate for us to keep treating primary time as special in this regard?

metricflow/model/validations/materializations.py
```diff
@@ -0,0 +1,315 @@
...
+                ):
+                    return dimension.name
+        raise RuntimeError(
+            f"No primary time dimension in the model. This should have been caught by "
+            f"{DataSourceTimeDimensionWarningsRule}"
+        )
+
+    @staticmethod
+    @validate_safely(whats_being_done="checking materialization is defined correctly")
+    def _validate_materialization(
```
plypaul, line 66: Actually, this can be greatly simplified by using `MetricSemantics.element_specs_for_metrics()`. Can you take a look at the function? If you do that, you won't need `DimensionAndIdentifierNameValidator` or `TimeGranularitySolver`.

metricflow/model/validations/materializations.py
```diff
@@ -0,0 +1,315 @@
...
+                            model_object_reference=ValidationIssue.make_object_reference(
+                                materialization_name=materialization.name
+                            ),
+                            message=f"Invalid dimension `{dimension_or_identifier_name}` for metric "
+                            f"`{metric_name}` in materialization `{materialization.name}`",
+                        )
+                    )
+            # Check dimensions listed in roll ups
+            destination: MaterializationDestination
+            for destination in materialization.destinations or []:
```
plypaul, line 145: Support for destinations is not in yet, so these should be removed.

plypaul (changes requested):
There are some large simplifications possible through some new methods, so I think we should use those instead.

metricflow/test/test_utils.py
QMalcolm: A lot of our model fixtures are scoped to the test session meaning that the model fixtures are persisted over the entirety of the test run (i.e. it doesn't get destroyed and recreated between tests). So we copy the model so that it doesn't affect the fixture for other tests. Another solution would be to switch the scope of the fixture to the default of `function` which means it would get destroyed and recreated between different test functions. I think we do it the current way to save time? I'm not wholly sure. 

Anyway, the model that is passed in is a model pre-transformations (i.e. it exists in the base state of a model which has been read from yaml files but hasn't been validated and transformed yet). So adding to the model is basically saying we need this for the state of this test, but not all other tests.

plypaul (approved):

PR #35, opened 2022-04-12
Removing duplicate MeasureReference definition

lebca:
There was a duplicate `MeasureReference` definition hiding in time granularity solver. It is deleted in this PR, and the usages of `MeasureReference` now point to our singular definition of `MeasureReference`.


tlento (approved):
Yikes. Thanks for fixing this.

PR #36, opened 2022-04-13
Add CLI Option to Display Plans

plypaul:
Users can now specify `--display-plans` to show the graphical version of the metric dataflow plan.


metricflow/test/dataflow_plan_to_svg.py
```diff
@@ -28,35 +29,46 @@ def add_nodes_to_digraph(node: DagNodeT, dot: graphviz.Digraph) -> None:
...
+    _render_via_graphviz(dag_graph=dag_graph, file_path_without_svg_suffix=random_file_path)
+    return random_file_path + ".svg"
+
+
+def _render_via_graphviz(dag_graph: DagGraphT, file_path_without_svg_suffix: str) -> None:
+    dot = graphviz.Digraph(comment=dag_graph.dag_id, node_attr={"shape": "box", "fontname": "Courier"})
+    # Not quite correct if there are shared nodes.
+    for sink_node in dag_graph.sink_nodes:
+        add_nodes_to_digraph(sink_node, dot)
+    dot.format = "svg"
```
andykram, line 73: Could we make the format configurable here? I find SVG kind of annoying to view.

metricflow/cli/main.py
```diff
@@ -213,6 +218,13 @@ def tutorial(ctx: click.core.Context, cfg: CLIContext, msg: bool, skip_dw: bool,
     default=False,
     help="In the query output, show the query that was executed against the data warehouse",
 )
+@click.option(
+    "--display-plans",
```
andykram, line 222: Can we please add this flag to the README and include an example output?

andykram (approved):

metricflow/cli/main.py
plypaul: There wasn't a section for CLI options in the README so I put it in the tutorial. Adding the image seems like it would take a lot of real estate.

WilliamDee: It would probably be better to add a example input/output in the docs somewhere rather than the README

metricflow/test/dataflow_plan_to_svg.py
plypaul: What are you running into? When using the option, a browser window pops up automatically and renders without any issues for me.

metricflow/cli/main.py
```diff
@@ -213,6 +218,13 @@ def tutorial(ctx: click.core.Context, cfg: CLIContext, msg: bool, skip_dw: bool,
...
+    is_flag=True,
+    required=False,
+    default=False,
```
WilliamDee, line 225: don't need the `default`, it defaults to False already via the `is_flag`

WilliamDee (approved):
1 nit, otherwise lgtm!

metricflow/test/dataflow_plan_to_svg.py
andykram: Oh got it. I haven't tried running the command. This default behavior should make viewing it easy.

metricflow/cli/main.py
andykram: To me an image demonstrates how truly powerful this feature is. It makes me want to try it out. That said, it's your call!

plypaul: That's fair, especially since users need to install another package for this to work. Let me try it out.

tlento (approved):
I like it! The inline image embed renders pretty nicely in the GitHub markdown preview, so I think it'll be ok.

We should add this to the main project docs repo as well - that way anybody who walks up and installs without ever looking in the MetricFlow repo will see this option as well.

metricflow/cli/main.py
plypaul: Updated.

PR #38, opened 2022-04-13
Added config-constructor method to MetricFlowEngine

WilliamDee:
## Context
Continuation of https://github.com/transform-data/metricflow/pull/37. An attempt to ease the process of initializing a functional `MetricFlowEngine` object. Instead of putting all the logic of constructing the MetricFlowEngine in `cli/` we simplify the constructor by making params that can be built via other params `Optional` and created a `from_config` method that allows user to pass in a `YamlFileHandler` which practically points to a config yaml file and initializes `MetricFlowEngine`

## Changes
- made `column_association_resolver`, `time_source`, `time_spine_source` optional as they can be built using the required params
- cleaned up some code in the `MetricFlowEngine.__init__`
- added `MetricFlowEngine.from_config` constructor method


metricflow/engine/metricflow_engine.py
```diff
@@ -244,25 +250,39 @@ class MetricFlowEngine(AbstractMetricFlowEngine):
...
         self,
         semantic_model: SemanticModel,
         sql_client: SqlClient,
-        column_association_resolver: ColumnAssociationResolver,
-        time_source: TimeSource,
-        time_spine_source: TimeSpineSource,
         system_schema: str,
+        column_association_resolver: Optional[ColumnAssociationResolver] = None,
+        time_source: Optional[TimeSource] = None,
+        time_spine_source: Optional[TimeSpineSource] = None,
```
tlento, line 272: Hmmm. Looks like we only use these overrides in testing. I'd prefer not to have them in the core initializer, but I can't come up with a nice way of handling the test overrides.

I think it's worth a docstring with param annotations on the initializer here as a short term hint to people to avoid direct calls to `MetricFlowEngine()` itself.

tlento (approved):
Nice, this makes the engine initialization in the CLI a lot easier to reason about as well.

Can you open a GitHub issue describing what we're doing here? Seems like this is the first cut at a proper API, or at least at providing a reasonable entry point for folks who want to do some development on top of MetricFlow.

PR #42, opened 2022-04-13
Removing time_format (legacy)

belindabennett:
## Context
MetricFlow no longer requires `time_format` for time dimensions. 

## Changes
This PR removes `time_format` from all sample models and tests.


nhandel (approved):
BYYYYEEEEE 👋

PR #44, opened 2022-04-13
Remove interfaces which are part of Transform

andykram:
Remove the Interfaces that are part of the Transform commercial offering.


rpourzand (approved):
LGTM!

PR #47, opened 2022-04-14
Added MetricFlowSimpleAPI

WilliamDee:
## Fixes
- Fix some datetime utils
- Fix docs/typing in `MetricFlowEngine`
- Added tests for `MetricFlowClient`

### Constructor
```python
MetricFlowClient(
  sql_client: SqlClient,
  user_configured_model: UserConfiguredModel,
  system_schema: str,
)
```
There is a factory method that will build the given parameters using the `config.yml` file
```python
MetricFlowClient.from_config()
```

### Usage
Assuming that you have ran `mf setup` and successfully setup your configurations
```python
from metricflow import MetricFlowClient

mf = MetricFlowClient.from_config()
metrics = mf.list_metrics()
```

#### Construct MetricFlowClient manually
**TODO:** Build out an easy path to building `SqlClient` and `SemanticModel`

### Doc
```python
MetricFlowClient.query(
    metrics: List[str],
    dimensions: List[str],
    limit: Optional[int] = None,
    start_time: Optional[str] = None,
    end_time: Optional[str] = None,
    where: Optional[str] = None,
    order: Optional[List[str]] = None,
    as_table: Optional[str] = None,
    sql_optimization_level: int = 4,
) -> MetricFlowQueryResult:
    """Makes a query for a metric.

    Args:
        metrics: Names of the metrics to query.
        dimensions: Names of the dimensions and identifiers to query.
        limit: Limit the result to this many rows.
        start_time: Get data for the start of this time range.
        end_time: Get data for the end of this time range.
        where: A SQL string using group by names that can be used like a where clause on the output data.
        order: metric and group by names to order by. A "-" can be used to specify reverse order e.g. "-ds"
        as_table: If specified, output the result data to this table instead of a result dataframe.
        sql_optimization_level: The level of optimization for the generated SQL. Pass integer from 0-4.

    Returns:
        MetricFlowQueryResult that contains the result and context of the query.
    """

MetricFlowClient.explain(
    metrics: List[str],
    dimensions: List[str],
    limit: Optional[int] = None,
    start_time: Optional[str] = None,
    end_time: Optional[str] = None,
    where: Optional[str] = None,
    order: Optional[List[str]] = None,
    as_table: Optional[str] = None,
    sql_optimization_level: int = 4,
) -> MetricFlowExplainResult:
    """Returns the plan for resolving a query.

    Args:
        metrics: Names of the metrics to query.
        dimensions: Names of the dimensions and identifiers to query.
        limit: Limit the result to this many rows.
        start_time: Get data for the start of this time range.
        end_time: Get data for the end of this time range.
        where: A SQL string using group by names that can be used like a where clause on the output data.
        order: metric and group by names to order by. A "-" can be used to specify reverse order e.g. "-ds"
        as_table: If specified, output the result data to this table instead of a result dataframe.
        sql_optimization_level: The level of optimization for the generated SQL. Pass integer from 0-4.

    Returns:
        MetricFlowExplainResult that contains the context of the query.
    """

MetricFlowClient.list_metrics() -> Dict[str, Metric]:
    """Retrieves a list of metric names.

    Returns:
        A dictionary with metric names as the key and the corresponding Metric object as the value.
    """

MetricFlowClient.list_dimensions(metric_names: List[str]) -> List[Dimension]:
    """Retrieves a list of all common dimensions for metric_names.

    "simple" dimensions are the ones that people expect from a UI perspective. For example, if "ds" is a time
    dimension at a day granularity, this would not list "ds__week".

    Args:
        metric_names: Names of metrics to get common dimensions from.

    Returns:
        A list of Dimension objects containing metadata.
    """

MetricFlowClient.list_materializations() -> List[Materialization]:
    """Retrieves a list of materialization names.

    Returns:
        A list of Materialization objects containing metadata.
    """


MetricFlowClient.get_dimension_values(
    metric_name: str, dimension_name: str, start_time: Optional[str] = None, end_time: Optional[str] = None
) -> List[str]:
    """Retrieves a list of dimension values given a [metric_name, dimension_name].

    Args:
        metric_name: Name of metric that contains the group_by.
        dimension_name: Name of group_by to get values from.
        start_time: Get data for the start of this time range.
        end_time: Get data for the end of this time range.

    Returns:
        A list of dimension values as string.
    """

MetricFlowClient.materialize(
    materialization_name: str, start_time: Optional[str] = None, end_time: Optional[str] = None
) -> SqlTable:
    """Builds a table containing metrics and dimensions from a materialization definition.

    This can be very expensive if a large time range is provided.

    Args:
        materialization_name: Name of materialization
        start_time: Materialized for the start of this time range.
        end_time: Materialized for the end of this time range.

    Returns:
        SqlTable object of the materialized table.
    """

MetricFlowClient.drop_materialization(materialization_name: str) -> bool:
    """Drops the table associated with a materialization definition.

    Args:
        materialization_name: Name of materialization to drop.

    Returns:
        True if a table has been drop, False if table doesn't exist.
    """

MetricFlowClient.validate_configs() -> Optional[Tuple[ValidationIssueType, ...]]:
    """Validate a model according to configured rules.

    Returns:
        Tuple of validation issues with the model provided.
    """
```


metricflow/api/simple_api.py
```diff
@@ -0,0 +1,259 @@
...
+from metricflow.model.model_validator import ModelValidator
+from metricflow.model.objects.user_configured_model import UserConfiguredModel
+from metricflow.model.semantic_model import SemanticModel
+from metricflow.model.validations.validator_helpers import ValidationIssueType
+from metricflow.protocols.sql_client import SqlClient
+from metricflow.sql.optimizer.optimization_levels import SqlQueryOptimizationLevel
+from metricflow.sql_clients.sql_utils import make_sql_client_from_config
+
+
+class MetricFlowSimpleAPI:
```
tlento, line 23: Let's get rid of the `SimpleAPI` appellation here. Simple because it leaves me wondering if there's a Complicated one someplace, and API because that's more rightly a reference to all of the classes of this sort.
 
I think something like `MetricFlowClient` might be the right name for this class. It's the in-code entrypoint to MetricFlow as a whole. All you need is a defined model and a warehouse and then you can issue requests via the methods defined below.

Another option is to define a Protocol called `MetricFlowClient` that has the base set of public method signatures specified, and then to name this something like `ConfigInitializedMetricFlowClient` and remove the optional parameters from the initializer. This streamlines the initializer here, makes it obvious  and allows other developers the option of implementing their own custom MetricFlowClient classes.

metricflow/api/simple_api.py
```diff
@@ -0,0 +1,259 @@
...
+            sql_client=self._sql_client,
+            system_schema=self._system_schema,
+        )
+
+    @property
+    def user_configured_model(self) -> UserConfiguredModel:  # noqa: D
+        return self._semantic_model.user_configured_model
+
+    @property
+    def mf_engine(self) -> MetricFlowEngine:  # noqa: D
```
tlento, line 48: Hm, can we call this just `engine` instead?

metricflow/test/api/conftest.py
```diff
@@ -0,0 +1,23 @@
...
+    sql_client: SqlClient,
+    simple_semantic_model: SemanticModel,
+    time_spine_source: TimeSpineSource,
+    mf_test_session_state: MetricFlowTestSessionState,
+) -> MetricFlowSimpleAPI:
+    """Fixture for MetricFlowSimpleAPI."""
+    return MetricFlowSimpleAPI(
+        sql_client=sql_client,
+        semantic_model=simple_semantic_model,
+        system_schema=mf_test_session_state.mf_system_schema,
```
tlento, line 22: Is it possible to have the fixture set up so that there is a test config in place and the config handler simply reads it in? That'd make this a bit better as a test of the client class mechanics and prevent us from exposing these parameter overrides.

metricflow/api/simple_api.py
```diff
@@ -0,0 +1,259 @@
+from typing import Dict, List, Optional
+
+from metricflow.configuration.config_handler import ConfigHandler
+from metricflow.configuration.constants import CONFIG_DWH_SCHEMA
+from metricflow.dataflow.sql_table import SqlTable
+from metricflow.engine.metricflow_engine import (
+    MetricFlowEngine,
+    MetricFlowExplainResult,
+    MetricFlowQueryRequest,
+    MetricFlowQueryResult,
```
tlento, line 10: Not now, but eventually these should probably be tidied up and moved in the api directory.

metricflow/api/simple_api.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+class MetricFlowSimpleAPI:
+    """Simple MetricFlow Python Interface."""
+
+    def __init__(
+        self,
+        sql_client: Optional[SqlClient] = None,
+        semantic_model: Optional[SemanticModel] = None,
+        system_schema: Optional[str] = None,
```
tlento, line 30: When will these be overridden? If they're only used for testing, can we remove these parameters and then put a mock config in place for the initializer to consume in test contexts?

I'd rather keep our exposure to parameter overrides to a minimum, because once these are out in the world they become part of the public API spec whether we want them there or not.

metricflow/api/simple_api.py
```diff
@@ -0,0 +1,259 @@
...
+    def user_configured_model(self) -> UserConfiguredModel:  # noqa: D
+        return self._semantic_model.user_configured_model
+
+    @property
+    def mf_engine(self) -> MetricFlowEngine:  # noqa: D
+        return self._mf
+
+    @property
+    def system_schema(self) -> str:  # noqa: D
+        return self._system_schema
```
tlento, line 53: Please add docstrings for all of these - this is going to be a public-facing developer API, so we should have full documentation of properties and what they're for.

metricflow/api/simple_api.py
```diff
@@ -0,0 +1,259 @@
...
+class MetricFlowSimpleAPI:
+    """Simple MetricFlow Python Interface."""
+
+    def __init__(
+        self,
+        sql_client: Optional[SqlClient] = None,
+        semantic_model: Optional[SemanticModel] = None,
+        system_schema: Optional[str] = None,
+    ):
+        """If parameters not passed, build via config."""
```
tlento, line 32: This can then be updated to explain how this is always initialized.

metricflow/api/simple_api.py
```diff
@@ -0,0 +1,259 @@
...
+
+        Args:
+            materialization_name: Name of materialization to drop.
+
+        Returns:
+            True if a table has been drop, False if table doesn't exist.
+        """
+        return self.mf_engine.drop_materialization(materialization_name=materialization_name)
+
+    def validate_configs(self) -> List[ValidationIssueType]:
```
tlento, line 252: Do we want `Tuple[ValidationIssueType, ...]` here instead? I think it might be better.

1. It simplifies a number of things, starting with the body of this method
2. I don't think we want people appending to the set of returned issues - if they have custom validation wrappers they can always do `mf.validate_configs + custom_mf.validate_configs` or similar.

metricflow/api/simple_api.py
```diff
@@ -0,0 +1,259 @@
...
+        """Validate a model according to configured rules.
+
+        Returns:
+            List of validation issues with the model provided.
+        """
+        issues = ModelValidator.validate_model(self.user_configured_model).issues
+        return [issue for issue in issues] if issues is not None else []
```
tlento, line 259: Ok, this is weird looking but I realize it's doing a cast from the internal tuple to a list. There are two things here:

1. The `None` handling seems silly. Might be worth making the issues property a non-Optional `Tuple[ValidationIssueType, ...]` in a separate PR and then stacking this one back on top of it (or just merging it first and then updating this to trunk) - I can't think of a situation where we want to differentiate between None and tuple() over there.
2. To cast an Iterator to List, use `list()` instead of a list comprehension. It's easier for readers to understand, and according to The Internet it is a lot faster in the runtime.
```suggestion
        issues = ModelValidator.validate_model(self.user_configured_model).issues
        return list(issue) if issues is not None else []
```

Of course, if we change the typehints as suggested we can simply `return ModelValidator.validate_model(...).issues`

tlento (changes requested):
I don't understand how GitHub action queues work but to facilitate some back and forth I'm going to push the "request changes" button now and see how that goes.

metricflow/api/simple_api.py
tlento: I take it back on list() performance, other parts of The Internet say the comprehension is now as fast or faster. So do whatever. We should just use Tuple anyway.

metricflow/api/simple_api.py
WilliamDee: This isn't intended to be always initialized via the configs, we would want to allow the user to build the SqlClient/Semantic models manually and pass it through as well.

metricflow/api/simple_api.py
WilliamDee: added

metricflow/api/simple_api.py
WilliamDee: these can be overridden by the user if they desire, it gives another option and doesn't restrict to using the config only

metricflow/test/api/conftest.py
WilliamDee: For now, it doesn't seem possible since the config only supports building via redshift/snowflake/bigquery and since we use sqlite for the testing this will be what we have to do in the meantime (unless we pass real creds via github secrets?).

metricflow/api/simple_api.py
WilliamDee: done

metricflow/api/simple_api.py
WilliamDee: changed to MetricFlowClient

metricflow/api/simple_api.py
WilliamDee: changed typing back to Tuple

metricflow/api/simple_api.py
```diff
@@ -0,0 +1,259 @@
...
+        self._sql_client = sql_client or make_sql_client_from_config(handler)
+        self._system_schema = system_schema or handler.get_value(CONFIG_DWH_SCHEMA)
+        self._semantic_model = semantic_model or SemanticModel(build_user_configured_model_from_config(handler))
+        self._mf = MetricFlowEngine(
+            semantic_model=self._semantic_model,
+            sql_client=self._sql_client,
+            system_schema=self._system_schema,
+        )
+
+    @property
```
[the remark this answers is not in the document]
WilliamDee: good point, not really. Removed.

metricflow/api/metricflow_client.py
```diff
@@ -247,13 +235,12 @@ def drop_materialization(self, materialization_name: str) -> bool:
         Returns:
             True if a table has been drop, False if table doesn't exist.
         """
-        return self.mf_engine.drop_materialization(materialization_name=materialization_name)
+        return self.engine.drop_materialization(materialization_name=materialization_name)
 
-    def validate_configs(self) -> List[ValidationIssueType]:
+    def validate_configs(self) -> Optional[Tuple[ValidationIssueType, ...]]:
```
tlento, line 240: Ok. I'm going to open an issue for this - I don't think validators should return Optional[Tuple[...]] - and then this callsite should get cleaned up along with the rest.

metricflow/test/api/conftest.py
tlento: I think this is an issue but it's outside the scope of this PR. We want a way to shim the config in tests so anything that does automatic access via the config object can be tested in its own right without us wiring in fully-formed instances as overrides like this.

For testing non-SQLite engines we could have something that reads warehouse credentials out of the environment (as we do today), so I don't think that's a huge blocker. The idea is basically to push the fixture work that initializes the SqlEngine etc. one level further out so it generates a config object.

metricflow/api/metricflow_client.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+class MetricFlowClient:
+    """MetricFlow Python Interface."""
+
+    def __init__(
+        self,
+        sql_client: Optional[SqlClient] = None,
+        semantic_model: Optional[SemanticModel] = None,
+        system_schema: Optional[str] = None,
```
tlento, line 29: This still bothers me a bit - I don't like exposing more than we need to since changing public APIs is quite difficult to do safely.

In thinking about this a bit more, it certainly makes sense that we'd want to expose the SqlClient and system schema as overrides, but I'm still not sold on the SemanticModel.  What's the use case for this?

I also think we might benefit from splitting this into an initializer with these elements required and a factory method that takes no arguments and builds those parameters from a config. Then developers get their ease of use and we can document the initializer as something that we don't want people to invoke directly. It also keeps the config fetching separate from the initialization of the object itself, which I think is desirable.

tlento (commented):
I'm still not sold on the initializer as it is but I think if you and @plypaul agree this is the API layout we want then we should proceed.

What do you think about splitting it up into a simple initializer with required parameters and a factory method for initializing from the config without any user input?

metricflow/api/metricflow_client.py
WilliamDee: These overrides are here as just an alternative where one could put the values directly into the constructor itself versus putting the values into the config file. For example, a user may want to make use of this client, but for whatever reason they don't want to use the config file or something. Then they would pass the required params to the initializer, which would directly allow them to use it without having to create/fill out the config file.

With that being said, I don't know how important this use case is, but we could technically simplify this further and instead of passing a SemanticModel object we can have them pass the `path_to_models_directory` as a str and we build the SemanticModel for them.

tlento: So I can definitely see scenarios for passing a custom SqlClient in - like if someone implements that interface for an engine we don't support inside MetricFlow itself - it's just having this open in the initializer with a magic fallback to the config that worries me.

I'd prefer to either:

1.  Make these parameters required and have the config mechanism handled by a factory method, then document that factory method  as the standard way to initialize one of these things
2. Remove these parameters and only support the config approach, and then provide a separate MetricFlowClient class that requires these overrides

I think option 1 is cleanest and simplest, because the config initialization is a separate convenience rather than something we'd ordinarily bake into the client directly.

WilliamDee: so, I drafted up 2 ways of going at this,
1. using a staticmethod in the class to instantiate
```python
@staticmethod
def from_config() -> MetricFlowClient:
    """Builds a MetricFlowClient via config yaml file."""
    handler = ConfigHandler()
    sql_client = make_sql_client_from_config(handler)
    user_configured_model = build_user_configured_model_from_config(handler)
    schema = handler.get_value(CONFIG_DWH_SCHEMA)

    return MetricFlowClient(
        sql_client=sql_client,
        user_configured_model=user_configured_model,
        system_schema=schema,
    )

def __init__(
    self,
    sql_client: SqlClient,
    user_configured_model: UserConfiguredModel,
    system_schema: str,
):
   pass
```
2. using a factory class
```python
class ConfigInitMetricFlowClient(MetricFlowClient):
    def __init__(self) -> None:
        handler = ConfigHandler()
        sql_client = make_sql_client_from_config(handler)
        user_configured_model = build_user_configured_model_from_config(handler)
        schema = handler.get_value(CONFIG_DWH_SCHEMA)
        super().__init__(
            sql_client=sql_client, user_configured_model=user_configured_model, system_schema=system_schema
        )


class MetricFlowClient:
    """MetricFlow Python Interface."""

    def __init__(
        self,
        sql_client: SqlClient,
        user_configured_model: UserConfiguredModel,
        system_schema: str,
    ):
    pass
```

Any preference here?

WilliamDee: implemented 1)

tlento: For posterity, we discussed separately and decided the static method was cleaner.

metricflow/api/metricflow_client.py
```diff
@@ -0,0 +1,269 @@
...
+from metricflow.sql_clients.common_client import not_empty
+from metricflow.sql_clients.sql_utils import make_sql_client_from_config
+
+
+class MetricFlowClient:
+    """MetricFlow Python Interface."""
+
+    @staticmethod
+    def from_config() -> MetricFlowClient:
+        """Builds a MetricFlowClient via config yaml file."""
```
tlento, line 31:
```suggestion
        """Builds a MetricFlowClient via config yaml file stored in default config location."""
```

metricflow/api/metricflow_client.py
```diff
@@ -0,0 +1,269 @@
...
+from metricflow.model.semantic_model import SemanticModel
+from metricflow.model.validations.validator_helpers import ValidationIssueType
+from metricflow.protocols.sql_client import SqlClient
+from metricflow.sql.optimizer.optimization_levels import SqlQueryOptimizationLevel
+from metricflow.sql_clients.common_client import not_empty
+from metricflow.sql_clients.sql_utils import make_sql_client_from_config
+
+
+class MetricFlowClient:
+    """MetricFlow Python Interface."""
```
tlento, line 27: Note: this is where we can get a little more descriptive about this being a simple version that uses string names in a manner similar to the CLI, or something like that. Not sure how much detail you want to provide.

Also, this isn't an interface.

```suggestion
    """MetricFlow Python client for running basic queries and other standard commands.
    
    Supports metadata lookups for dimensions and metrics, materialization management, and
    explain plan requests in addition to a standard query method.
    """
```

metricflow/api/metricflow_client.py
tlento: Filed #49

tlento (approved):
Looks great, thanks @WilliamDee !

metricflow/api/metricflow_client.py
WilliamDee: making it `designated` instead of `default` since the path can still be overridden via the ENV `MF_CONFIG_DIR`

PR #50, opened 2022-04-20
Removing Unnecessary Lone Identifiers Validation

lebca:
This validation added a nonblocking warning to the resulting validation issues.
Removing this warning is generally what customers want as the lone identifier doesn't cause any problems.

### DIFF
--- a/metricflow/model/model_validator.py
+++ b/metricflow/model/model_validator.py
@@ -4,7 +4,6 @@ from typing import List
 
 from metricflow.model.objects.user_configured_model import UserConfiguredModel
 from metricflow.model.parsing.dir_to_model import ModelBuildResult
-from metricflow.model.validations.common_identifiers import CommonIdentifiersRule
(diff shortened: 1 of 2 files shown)


QMalcolm:
What is the reasoning for removing the validation? To my knowledge we don't currently surface warnings to the customer unless there is also an error. I have a plan to condense warnings when there are errors, and begin surfacing warnings in a condensed manner when there aren't errors. Then a flag could optionally be supplied to see the warnings in a verbose manner.

QMalcolm:
As an alternative to deleting the validation in its entirety. could we instead just disable it for the time being by removing it from the list of validations to run, `VALIDATION_RULES`, in model_validator.py?

QMalcolm (approved):
Looks good! Thanks for taking care of this. I'll try to get us to a place as soon as possible where we can start running this rule again :slightly_smiling_face:

Issue #19, opened 2022-04-07
Re-write validation tests to rely on concrete model files


tlento:
**Describe the bug**
In #17 we encountered an issue where the model validation tests failed non-deterministically due to an in-place mutation of the simple model we use for most of our tests resulting in the wrong exception being thrown. The short term issue was addressed in #18, but this was a quick hack patch rather than a holistic solution.

**Steps To Reproduce**
1. Find all of the locations in test/model/validations where use copy.deepcopy
2. Observe that these are making local in-memory copies of the simple model and then mutating those copies to cause validation issues

**Expected behavior**
These should be pointing at a set of model files (or, alternatively, a constructed UserDefinedModel object) that fails validation in the specific manner expected by the test. We will decide on a path here and update this issue accordingly.

**Environment (please complete the following information):**

- MetricFlow Version [e.g. 1.0.0]

QMalcolm:
We should probably move this issue to dbt-semantic-interfaces

PR #51, opened 2022-04-21
Composite identifiers validation

lebca:
Adding Validation and Test for Composite Identifiers

MetricFlow enforces that each identifier is defined uniquely, one aspect of this is that the sub identifiers are the same.
This is not a problem when an identifier is only requested once in a model. But this can become a problem when a model requests the same identifier across multiple data sources of a model. Each time that identifier is requested, it must share the same sub-identifier structure.

Defaulting the issues attribute of ModelBuildResult to an empty tuple for more intuitive handling.

### DIFF
--- a/metricflow/model/parsing/dir_to_model.py
+++ b/metricflow/model/parsing/dir_to_model.py
@@ -37,7 +37,7 @@ logger = logging.getLogger(__name__)
 class ModelBuildResult:  # noqa: D
     model: Optional[UserConfiguredModel] = None
     # Issues found in the model.
-    issues: Optional[Tuple[ValidationIssueType, ...]] = None
+    issues: Tuple[ValidationIssueType, ...] = tuple()
 
 
 def parse_directory_of_yaml_files_to_model(
--- a/metricflow/model/validations/identifiers.py
+++ b/metricflow/model/validations/identifiers.py
@@ -1,8 +1,10 @@
 import logging
-from collections import OrderedDict
+from collections import defaultdict
 from dataclasses import dataclass
 from datetime import date
-from typing import List, MutableSet, Tuple, Sequence, Dict
+from typing import List, MutableSet, Tuple, Sequence, DefaultDict
+
+import more_itertools
 
 from metricflow.model.objects.data_source import DataSource
 from metricflow.model.objects.elements.identifier import Identifier, IdentifierType, CompositeSubIdentifier
@@ -165,48 +167,40 @@ class IdentifierConsistencyRule(ModelValidationRule):
     @validate_safely(whats_being_done="running model validation to ensure identifiers have consistent sub-identifiers")
     def validate_model(model: UserConfiguredModel) -> List[ValidationIssueType]:  # noqa: D
         issues: List[ValidationIssueType] = []
-        # The expected sub identifiers that an identifier has.
-        identifier_to_sub_identifiers: OrderedDict[str, Dict[str, Tuple[str, ...]]] = OrderedDict()
-        # The identifiers that have different sub-identifiers in different data sources.
-        identifiers_with_inconsistent_sub_identifiers = []
+        # build collection of sub-identifier contexts, keyed by identifier name
+        identifier_to_sub_identifier_contexts: DefaultDict[str, List[SubIdentifierContext]] = defaultdict(list)
+        all_contexts: List[SubIdentifierContext] = list(
+            more_itertools.flatten(
+                [
+                    IdentifierConsistencyRule._get_sub_identifier_context(data_source)
+                    for data_source in model.data_sources
+                ]
+            )
+        )
+        for context in all_contexts:
+            identifier_to_sub_identifier_contexts[context.identifier_reference.element_name].append(context)
 
-        identifier_contexts = []
-        issues = []
-        for data_source in model.data_sources or []:
-            sub_identifier_contexts = IdentifierConsistencyRule._get_sub_identifier_context(data_source)
-            for sub_identifier_context in sub_identifier_contexts:
-                identifier_contexts.append(sub_identifier_context)
+        # Filter out anything that has fewer than 2 distinct sub-identifier sets
+        invalid_sub_identifier_configurations = dict(
+            filter(
+                lambda item: len(set([context.sub_identifier_names for context in item[1]])) >= 2,
+                identifier_to_sub_identifier_contexts.items(),
+            )
+        )
 
-                identifier_reference = sub_identifier_context.identifier_reference
-                if identifier_reference.element_name not in identifier_to_sub_identifiers:
-                    identifier_to_sub_identifiers[identifier_reference.element_name] = {
-                        data_source.name: sub_identifier_context.sub_identifier_names
-                    }
+        # convert each invalid identifier configuration into a validation warning
+        for identifier_name, sub_identifier_contexts in invalid_sub_identifier_configurations.items():
+            issues.append(
+                ValidationWarning(
+                    model_object_reference=ValidationIssue.make_object_reference(
+                        data_source_name=sub_identifier_contexts[0].data_source_name,
+                        identifier_name=identifier_name,
+                    ),
+                    message=(
+                        f"Identifier '{identifier_name}' does not have consistent sub-identifiers "
+                        f"throughout the model: {list(sorted(sub_identifier_contexts, key=lambda x: x.sub_identifier_names))}"
+                    ),
+                )
+            )
 
-                elif (
-                    sub_identifier_context.sub_identifier_names
-                    not in identifier_to_sub_identifiers[identifier_reference.element_name].values()
-                ):
-                    identifier_to_sub_identifiers[identifier_reference.element_name][
-                        data_source.name
-                    ] = sub_identifier_context.sub_identifier_names
-                    identifiers_with_inconsistent_sub_identifiers.append(identifier_reference)
-
-        for identifier_reference in identifiers_with_inconsistent_sub_identifiers:
-            for identifier_context in identifier_contexts:
-                if identifier_context.identifier_reference.element_name == identifier_reference.element_name:
-                    issues.append(
-                        ValidationWarning(
-                            model_object_reference=ValidationIssue.make_object_reference(
-                                data_source_name=identifier_context.data_source_name,
-                                identifier_name=identifier_context.identifier_reference.element_name,
-                            ),
-                            message=(
-                                f"Identifier '{identifier_reference.element_name}' does not have consistent sub-identifiers "
-                                f"throughout the model: {identifier_to_sub_identifiers[identifier_reference.element_name]}. "
-                                f"Specifically, in data source '{identifier_context.data_source_name}', "
-                                f"it has conflicting sub-identifiers {list(identifier_context.sub_identifier_names)}."
-                            ),
-                        )
-                    )
         return issues
--- a/metricflow/model/validations/identifiers.py
+++ b/metricflow/model/validations/identifiers.py
@@ -2,7 +2,7 @@ import logging
 from collections import OrderedDict
 from dataclasses import dataclass
 from datetime import date
-from typing import List, MutableSet, Tuple, Sequence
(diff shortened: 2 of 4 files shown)


metricflow/model/validations/identifiers.py
```diff
@@ -119,3 +123,86 @@ def validate_model(model: UserConfiguredModel) -> List[ValidationIssue]:  # noqa
...
+
+                identifier_reference = sub_identifier_context.identifier_reference
+                if identifier_reference.element_name not in identifier_to_sub_identifiers:
+                    identifier_to_sub_identifiers[
+                        identifier_reference.element_name
+                    ] = sub_identifier_context.sub_identifier_names
+
+                elif (
+                    identifier_to_sub_identifiers[identifier_reference.element_name]
+                    != sub_identifier_context.sub_identifier_names
```
tlento, line 188: So these are order-dependent? Is that necessary? Assuming this is a direct port from the original validation, we should just go with it, as relaxing that restriction seems like a discussion beyond the scope of this PR.

metricflow/model/validations/identifiers.py
```diff
@@ -119,3 +123,86 @@ def validate_model(model: UserConfiguredModel) -> List[ValidationIssue]:  # noqa
...
+                        ValidationWarning(
+                            model_object_reference=ValidationIssue.make_object_reference(
+                                data_source_name=identifier_context.data_source_name,
+                                identifier_name=identifier_context.identifier_reference.element_name,
+                            ),
+                            message=(
+                                f"Identifier '{identifier_reference.element_name}' does not have consistent sub-identifiers "
+                                f"throughout the model. In data source '{identifier_context.data_source_name}', "
+                                f"it has sub-identifiers {list(identifier_context.sub_identifier_names)}."
+                            ),
```
tlento, line 205: It'd be nice to provide the sub-identifiers of the original in the error message as well, but it's a bit tricky since there could be more than one mismatch. Right now the user has to go look at the data source definition to find the original set of sub-identifiers, right? 

I'd rather have the original set of sub-identifiers repeated in every error then not have them at all, so if that information isn't included in the set of warnings anywhere let's add them somehow.

metricflow/model/validations/identifiers.py
```diff
@@ -119,3 +123,86 @@ def validate_model(model: UserConfiguredModel) -> List[ValidationIssue]:  # noqa
...
+                    identifier_to_sub_identifiers[identifier_reference.element_name]
+                    != sub_identifier_context.sub_identifier_names
+                ):
+                    identifiers_with_inconsistent_sub_identifiers.append(identifier_reference)
+
+        for identifier_reference in identifiers_with_inconsistent_sub_identifiers:
+            for identifier_context in identifier_contexts:
+                if identifier_context.identifier_reference.element_name == identifier_reference.element_name:
+                    issues.append(
+                        ValidationWarning(
```
tlento, line 196: `ValidationWarning` is non-blocking, right? Do we want to make this strict or is that a follow-up for later?

metricflow/test/model/validations/test_identifiers.py
```diff
@@ -68,6 +69,7 @@ def test_invalid_composite_identifiers() -> None:  # noqa:D
                                 type=DimensionType.TIME,
                                 type_params=DimensionTypeParams(
                                     is_primary=True,
+                                    time_format="YYYY-MM-DD",
```
tlento, line 72: I think these are a legacy holdover, and we've been removing them from the various models, so unless this is absolutely necessary for this PR to work correctly we should revert the time_format additions before merging.

metricflow/test/model/validations/test_identifiers.py
```diff
@@ -196,3 +200,42 @@ def test_composite_identifiers_ref_and_name() -> None:  # noqa:D
...
+        identifiers=[CompositeSubIdentifier(ref="sub_identifier1")],
+    )
+    bookings_source.identifiers = flatten_nested_sequence([bookings_source.identifiers, [identifier_bookings]])
+
+    identifier_listings = Identifier(
+        name=IdentifierReference(element_name="composite_identifier"),
+        type=IdentifierType.FOREIGN,
+        identifiers=[CompositeSubIdentifier(ref="sub_identifier2")],
+    )
+    listings_latest.identifiers = flatten_nested_sequence([listings_latest.identifiers, [identifier_listings]])
```
tlento, line 234: I know this is the current standard for these tests but I hate this copy-and-mutate approach. It makes the tests very hard to reason about. I guess we should leave it like this for the sake of consistency and just bulldoze it all whenever we take on #19

tlento (commented):
I have a couple of non-blocking questions inline. It's fine if the answers are that this is how we're proceeding in general, or that this is how we'll proceed for now and updates will come as follow-ups.

I'll leave this open for now so other reviewers have a chance to take a look.

metricflow/model/parsing/dir_to_model.py
```diff
@@ -37,7 +37,7 @@
 class ModelBuildResult:  # noqa: D
     model: Optional[UserConfiguredModel] = None
     # Issues found in the model.
-    issues: Optional[Tuple[ValidationIssueType, ...]] = None
+    issues: Tuple[ValidationIssueType, ...] = tuple()
```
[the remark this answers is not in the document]
lebca: Yes, will do.

metricflow/test/model/validations/test_identifiers.py
lebca: New commit removes these.

metricflow/model/validations/identifiers.py
lebca: @QMalcolm will be addressing the `ValidationWarning` behavior in a follow-up refactor, so let's leave this for now.

metricflow/model/validations/identifiers.py
lebca: This is a direct port from the original validation. We can re-address this later.

metricflow/model/validations/identifiers.py
lebca: New commit includes all the definitions of the offending identifiers, including in which data source they are defined.

metricflow/model/validations/identifiers.py
```diff
@@ -119,3 +123,90 @@ def validate_model(model: UserConfiguredModel) -> List[ValidationIssue]:  # noqa
...
+                                identifier_name=identifier_context.identifier_reference.element_name,
+                            ),
+                            message=(
+                                f"Identifier '{identifier_reference.element_name}' does not have consistent sub-identifiers "
+                                f"throughout the model: {identifier_to_sub_identifiers[identifier_reference.element_name]}. "
+                                f"Specifically, in data source '{identifier_context.data_source_name}', "
+                                f"it has conflicting sub-identifiers {list(identifier_context.sub_identifier_names)}."
+                            ),
+                        )
+                    )
```
tlento, line 211: This is very hard to understand because the assertion and message construction logic is now all tangled up with the loop iteration mechanics we're using to parse the data source model information. I think it'd be cleaner if we instead did something like:

```
Build mapping of identifiers to sub-identifiers, with a set of associated data sources
for every element in mapping:
    check for mismatched sub-identifiers
    add errors if necessary
```
This has the added benefit of allowing us to aggregate our error messages on a per-identifier basis. I'm thinking something like this, but actually, you know, working:
```suggestion
        # The expected sub identifiers that an identifier has.
        identifier_to_sub_identifiers: OrderedDict[str, DefaultDict[Tuple[str, ...], Set[str]]] = OrderedDict()

        for data_source in model.data_sources or []:
            sub_identifier_contexts = IdentifierConsistencyRule._get_sub_identifier_context(data_source)
            for sub_identifier_context in sub_identifier_contexts:
                identifier_reference = sub_identifier_context.identifier_reference
                if identifier_reference.element_name not in identifier_to_sub_identifiers:
                    sub_identifiers = collections.defaultdict(set)
                    sub_identifiers[sub_identifier_context.identifier_names].add(data_source.name)
                    identifier_to_sub_identifiers[identifier_reference.element_name] = sub_identifiers
                else:
                    identifier_to_sub_identifiers[identifier_reference.element_name][
                        sub_identifier_context.sub_identifier_names
                    ].add(data_source.name)

        for identifier_name, sub_identifiers in identifier_to_sub_identifiers.items():
            if len(sub_identifiers) < 2:
                continue
                
            first_data_source_name = sub_identifiers[0][0]
            issues.append(
                ValidationWarning(
                    model_object_reference=ValidationIssue.make_object_reference(
                        data_source_name=first_data_source_name,
                        identifier_name=identifier_name,
                    ),
                    message=(
                        f"Identifier '{identifier_name}' does not have consistent sub-identifiers "
                        f"throughout the model. We found {len(sub_identifiers)} sets of sub-identifiers, but there "
                        f"should be no more than 1! Here are all sub-identifier sets with associated data sources: "
                        f"{sub_identifiers}"
                    ),
                )
            )
```

tlento (commented):
At this point I think the logic in `validate_model` needs an overhaul, because I had to read it several times to understand what it's doing. I made a suggestion for simplification inline, which I think does the trick, but I can't be sure.

I'd rather clean this up before merging because it is rather confusing, but if that gets complicated let me know and we can merge as-is and do the cleanup as part of a separate issue.

metricflow/model/validations/identifiers.py
tlento: Worth noting, these loops can each be a separate function with a nice name as well, which would aid in readability. Might be worth doing the first one, so `validate_model` becomes:

```
identifier_to_sub_identifiers = IdentifierConsistencyRule._build_sub_identifier_mapping(model.data_sources or [])

for identifier_name, sub_identifiers in identifier_to_sub_identifiers.items():
    # etc.

return issues
```
You can do both of course but the second loop is both pretty straightforward and is doing the thing the so I don't think it buys you much, and anyway that's actually doing the validation work.

This probably isn't necessary, but it makes that first gnarly for loop a little more accessible to a reader.

tlento: Oh wait, we are working way too hard here. The data source names are in the sub-identifier contexts so we don't need to collect them again along the way, so we can do something more like aggregate -> filter -> convert.

Note - it'd be nicer to do some more work in the last block to print one line per sub-identifier list value with collections of data sources, instead of a list of all non-unique values, but I'm tired of writing code in GitHub comments.

```suggestion
        # build collection of sub-identifier contexts, keyed by identifier name
        identifier_to_sub_identifier_contexts: DefaultDict[str, List[SubIdentifierContext]] = collections.defaultdict(list)
        all_contexts: List[SubIdentifierContext]] = more_itertools.flatten([
            IdentifierConsistencyRule._get_sub_identifier_context(data_source)
            for data_source in model.data_sources
        ])
        for context in all_contexts:
            identifier_to_sub_identifier_contexts[
                context.identifier_reference.element_name
            ].append(context)
            
        # Filter out anything that has fewer than 2 distinct sub-identifier sets
        invalid_sub_identifier_configurations = dict(filter(
            lambda item: len(
                set([context.sub_identifier_names for context in item[1]])
            ) >= 2,
            identifier_to_sub_identifier_contexts.items()
        ))

        # convert each invalid identifier configuration into a validation warning 
        for identifier_name, sub_identifier_contexts in invalid_sub_identifier_configurations.items():
            issues.append(
                ValidationWarning(
                    model_object_reference=ValidationIssue.make_object_reference(
                        data_source_name=sub_identifier_contexts[0].data_source_name,
                        identifier_name=identifier_name,
                    ),
                    message=(
                        f"Identifier '{identifier_name}' does not have consistent sub-identifiers "
                        f"throughout the model: {list(sorted(sub_identifier_contexts, key=lambda x: x.sub_identifier_names)}"
                    ),
                )
            )
```

tlento (commented):
How's this? Maybe better?

metricflow/model/validations/identifiers.py
lebca: Added.

tlento (approved):
LGTM, thanks!

PR #60, opened 2022-05-02
Fix Dataframe Comparison in Tests

plypaul:
This PR fixes an issue where dataframes may not be compared correctly when there are NULL / NaN values in the dataframe.

### DIFF
--- a/metricflow/test/compare_df.py
+++ b/metricflow/test/compare_df.py
@@ -1,5 +1,34 @@
+import logging
+
 import math
 import pandas as pd
+
+
+logger = logging.getLogger(__name__)
+
+
+def _dataframes_contain_same_data(
+    actual: pd.DataFrame,
+    expected: pd.DataFrame,
+) -> bool:
+    """Compare all elements of the dataframe one by one for equality.
+
+    We use math.isclose() to avoid issues with float comparisons.
+    """
+    if expected.shape != actual.shape:
+        return False
+
+    for c in range(expected.shape[0]):
+        for r in range(expected.shape[1]):
+            # NaNs can't be compared for equality.
+            if pd.isna(expected.iloc[c, r]) and pd.isna(actual.iloc[c, r]):
+                pass
+            elif isinstance(expected.iloc[c, r], float) and isinstance(actual.iloc[c, r], float):
+                if not math.isclose(expected.iloc[c, r], actual.iloc[c, r]):
+                    return False
+            elif expected.iloc[c, r] != actual.iloc[c, r]:
+                return False
+    return True
 
 
 def assert_dataframes_equal(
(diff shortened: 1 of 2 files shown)


metricflow/test/compare_df.py
```diff
@@ -25,20 +48,7 @@ def assert_dataframes_equal(
             sort_by = list(sorted(actual.columns.tolist()))
             expected = expected.loc[:, sort_by].sort_values(sort_by).reset_index(drop=True)
             actual = actual.loc[:, sort_by].sort_values(sort_by).reset_index(drop=True)
-        equal = True
-        # Compare all of the elements of the dataframe one by one. If someone knows pandas magic, this can be improved.
-        for c in range(expected.shape[0]):
-            for r in range(actual.shape[1]):
-                # Floats need to be compared with an epsilon due to rounding error.
```
lebca, line 32: I think this comment is worth including in the new function.

metricflow/test/compare_df.py
```diff
@@ -25,20 +48,7 @@ def assert_dataframes_equal(
             sort_by = list(sorted(actual.columns.tolist()))
             expected = expected.loc[:, sort_by].sort_values(sort_by).reset_index(drop=True)
             actual = actual.loc[:, sort_by].sort_values(sort_by).reset_index(drop=True)
```
lebca, line 50: Does it make sense to include an `actual.shape == expected.shape` comparison here?

tlento: It's already there - line 23. It might be worth making that a little more obvious, I missed it too.

metricflow/test/compare_df.py
```diff
@@ -1,7 +1,30 @@
...
 
 
+logger = logging.getLogger(__name__)
+
+
+def _dataframes_contain_same_data(
+    actual: pd.DataFrame,
+    expected: pd.DataFrame,
+) -> bool:
+    """Compare all elements of the dataframe one by one, assuming both dataframes have the same shape."""
```
tlento, line 14: Per @lebca's comment:

```suggestion
    """Compare all elements of the dataframe one by one, assuming both dataframes have the same shape.
    
    We use math.isclose() to avoid issues with float comparisons
    """
```

metricflow/test/integration/test_cases/itest_metrics.yaml
```diff
@@ -173,18 +173,34 @@ integration_test:
 ---
 integration_test:
   name: same_measure_constrained_metric
-  description: Test metric with local dimensions
+  description: |
+    Query metrics based on the same measure, but where one is constrained. Uses a FULL OUTER JOIN in the check query to
```
tlento, line 177: Nice, thanks!

tlento (approved):

metricflow/test/compare_df.py
lebca: Because line 49 will fail if `actual` is bigger than `expected`?

tlento: Yes, and also if actual is smaller than expected. `equal` is initialized False, and if `expected.shape != actual.shape` none of this code executes. We immediately fall through to throwing the exception.

metricflow/test/compare_df.py
plypaul: Updated.

metricflow/test/compare_df.py
plypaul: Moved some things around for clarity.

PR #61, opened 2022-05-02
Add the ability to construct a MetricFlowClient with an absolute path to configs

WilliamDee:
## Context
Currently, when calling `MetricFlowConfig.from_config()` it defaults to using the default location of where the MetricFlow configs are `{HOME}/.metricflow/config.yml`. This restricts the user from being flexible to create config files elsewhere and dynamically use different configs on the fly. So in order to further enable users to setup MetricFlow in a more streamlined fashion, this PR will enhance the `from_config` constructor.

## Changes
- provide the ability to pass in `config_file_path` via `from_config`
  - this is an optional parameter and will default back to the original behaviour of using the default config path

## Example
```python
mf = MetricFlowClient.from_config("path/to/test_config.yml")  # will build using the test_config.yml file
mf = MetricFlowClient.from_config() # will act as normal and use the default config file

### DIFF
--- a/metricflow/api/metricflow_client.py
+++ b/metricflow/api/metricflow_client.py
@@ -1,9 +1,12 @@
 from __future__ import annotations
 
+import logging
+
 from typing import Dict, List, Optional, Tuple
 
 from metricflow.configuration.config_handler import ConfigHandler
 from metricflow.configuration.constants import CONFIG_DWH_SCHEMA
+from metricflow.configuration.yaml_handler import YamlFileHandler
 from metricflow.dataflow.sql_table import SqlTable
 from metricflow.engine.metricflow_engine import (
     MetricFlowEngine,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by plypaul — APPROVED** [review]

File: metricflow/api/metricflow_client.py [hunk 1]
```diff
@@ -27,9 +28,15 @@ class MetricFlowClient:
...
-    def from_config() -> MetricFlowClient:
-        """Builds a MetricFlowClient via config yaml file stored in the designated config location."""
-        handler = ConfigHandler()
+    def from_config(config_file_path: Optional[str] = None) -> MetricFlowClient:
+        """Builds a MetricFlowClient via config yaml file.
+
+        If config_file_path is not passed, it will use the default config location.
+        """
+        if config_file_path is not None:
+            handler = YamlFileHandler(yaml_file_path=config_file_path)
```
tlento, line 37: Nice. We might want to add a logger.debug() or something here to allow people to track which config path they're using when overrides are present. Or maybe just always do a logger.debug on the config file path, either way. Up to you if you want to add that or not.

tlento (approved):
Yay, thanks @WilliamDee!

PR #63, opened 2022-05-04
Add LOCAL_LINKED Property for Available Dimensions

plypaul:
This PR adds the `LOCAL_LINKED` property to indicate if an available dimension for a metric is the same as a local one, but linked with a local primary identifier. For example, if a data source has the primary identifier `listing` and dimension `capacity_latest`, then the dimension `listing__capacity_latest` for a measure in that data source would have the property `LOCAL_LINKED`.

This also updates the function for getting available dimensions for a metric to use properties, instead of a set of flags.

### DIFF
--- a/metricflow/engine/metricflow_engine.py
+++ b/metricflow/engine/metricflow_engine.py
@@ -22,6 +22,7 @@ from metricflow.execution.execution_plan import ExecutionPlan, SqlQuery
 from metricflow.execution.execution_plan_to_text import execution_plan_to_text
 from metricflow.execution.executor import SequentialPlanExecutor
 from metricflow.model.semantic_model import SemanticModel
+from metricflow.model.semantics.linkable_spec_resolver import LinkableElementProperties
 from metricflow.object_utils import pformat_big_objects, random_id
 from metricflow.plan_conversion.column_resolver import DefaultColumnAssociationResolver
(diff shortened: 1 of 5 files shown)


metricflow/engine/metricflow_engine.py
```diff
@@ -443,9 +444,13 @@ def simple_dimensions_for_metrics(self, metric_names: List[str]) -> List[Dimensi
             Dimension(name=dim.qualified_name)
             for dim in self._semantic_model.metric_semantics.element_specs_for_metrics(
                 metric_specs=[MetricSpec(element_name=mname) for mname in metric_names],
-                dimensions_only=True,
-                exclude_derived_time_granularities=True,
-                exclude_local_linked_primary_time=True,
+                without_any_property=frozenset(
```
courtneyholcomb, line 447: For my understanding here - are identifiers and dimensions the only things that might be returned from `element_specs_for_metrics`? So specifying without property `IDENTIFIER` is equivalent to specifying `dimensions_only`?

courtneyholcomb (approved):
Change looks awesome - cleans up so much code! Thank you for doing this!!

metricflow/engine/metricflow_engine.py
plypaul: That is correct.

PR #67, opened 2022-05-06
Add the ability to pass subclasses to parser

WilliamDee:
## Context
The `UserConfiguredModel` contains 3 core classes, `[DataSource, Materialization, Metric]`. However, there is no flexibility in allowing for adding extra (but not necessary for open source)attributes to the classes. This PR allows for passing custom subclasses of the 3 core classes. This would enable the ability to build custom classes that would be parsed by the parser.

## Example
```python3
class DataSourceWithDescription(DataSource):
   """DataSource class with description."""
   description: str

# Then to parse with this class
parse_yaml_files_to_model(files, data_source_class=DataSourceWithDescription)

```

### DIFF
--- a/metricflow/model/parsing/dir_to_model.py
+++ b/metricflow/model/parsing/dir_to_model.py
@@ -74,7 +74,12 @@ def parse_directory_of_yaml_files_to_model(
     return ModelBuildResult(model=model)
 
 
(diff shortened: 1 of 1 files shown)


plypaul (approved):
Sorry, forgot to approve earlier!

PR #68, opened 2022-05-06
Add Missing Type Annotation

plypaul:
This PR fixes a missing type annotation that should have been caught by the linter, but was not. Follow up item is to take a look at the linter configuration for this project.


WilliamDee (approved):
Also, I just noticed that the `api/` directory is missing an `__init__.py` file. Could we include that in this PR?

plypaul:
Added missing `__init__.py`.

PR #72, opened 2022-05-09
Use Warnings for Materialization Validations

plypaul:
Since the materialization validation was recently introduced, this changes the issue type to warning to that migrations are easier.

### DIFF
--- a/metricflow/model/validations/materializations.py
+++ b/metricflow/model/validations/materializations.py
@@ -1,21 +1,21 @@
 import logging
 from typing import List
 
-from metricflow.errors.errors import UnableToSatisfyQueryError
 from metricflow.model.objects.materialization import Materialization
 from metricflow.model.objects.user_configured_model import UserConfiguredModel
 from metricflow.model.semantic_model import SemanticModel
 from metricflow.model.semantics.data_source_container import PydanticDataSourceContainer
 from metricflow.model.semantics.semantic_containers import DataSourceSemantics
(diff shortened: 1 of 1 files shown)


metricflow/model/validations/materializations.py
```diff
@@ -53,9 +53,11 @@ def _validate_materialization(
                 metric_names=materialization.metrics,
                 group_by_names=materialization.dimensions,
             )
-        except UnableToSatisfyQueryError as err:
+        # TODO: Using broad exception clause until the query validation returns a list of errors.
+        except Exception as err:
             issues.append(
-                ValidationError(
+                ValidationIssue(
```
QMalcolm, line 59: Instead of using the base `ValidationIssue` and having to specify the type, we should be using the specific `Validation<Type>` error classes for consistency. We do have a specific [ValidationWarning](https://github.com/transform-data/metricflow/blob/main/metricflow/model/validations/validator_helpers.py#L110). However, in a side conversation it not within this PR it was mentioned that we intend to make this an error at some point. In which case I'd recommend using [ValidationFutureError](https://github.com/transform-data/metricflow/blob/c30eac0a4e1d02579c17e2a65cafd29f06192420/metricflow/model/validations/validator_helpers.py#L121)

QMalcolm (changes requested):
Small change request for consistency

QMalcolm (approved):
Looks good to me :shipit:

PR #75, opened 2022-05-10
Add Metadata to Model objects

WilliamDee:
## Context
Metadata field was missing from the Model objects and it should be surfaced as it contains useful data about the model objects. For example, for a metric, it would show which file config it was parsed from and the sliced raw yaml block.

## Changes
- Added `Metadata` to
  - `DataSource`
  - `Metric`
  - `Materialization`
  - `Measure`
- Fixed the parser to properly parse the metadata context for each object

### DIFF
--- a/metricflow/model/parsing/dir_to_model.py
+++ b/metricflow/model/parsing/dir_to_model.py
@@ -186,14 +186,13 @@ def parse_config_yaml(
             # retrieve last top-level key as type
             document_type = next(iter(config_document.keys()))
             object_cfg = config_document[document_type]
+            yaml_contents_by_line = config_yaml.contents.splitlines()
             if document_type == METRIC_TYPE:
-                results.append(parse(metric_class, ctx, object_cfg, config_yaml.file_path, config_yaml.contents))
+                results.append(parse(metric_class, object_cfg, config_yaml.file_path, yaml_contents_by_line))
             elif document_type == DATA_SOURCE_TYPE:
-                results.append(parse(data_source_class, ctx, object_cfg, config_yaml.file_path, config_yaml.contents))
+                results.append(parse(data_source_class, object_cfg, config_yaml.file_path, yaml_contents_by_line))
             elif document_type == MATERIALIZATION_TYPE:
-                results.append(
-                    parse(materialization_class, ctx, object_cfg, config_yaml.file_path, config_yaml.contents)
-                )
+                results.append(parse(materialization_class, object_cfg, config_yaml.file_path, yaml_contents_by_line))
             else:
                 errors.append(
                     str(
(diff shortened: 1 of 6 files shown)


metricflow/model/parsing/dir_to_model.py
```diff
@@ -237,19 +236,20 @@ def parse_config_yaml(
...
     _type: Type[Union[DataSource, Metric, Materialization]],
-    ctx: ParsingContext,
     yaml_dict: Dict[str, Any],
     filename: str,
-    contents: str,
+    contents_by_line: List[str],
 ) -> Any:
     """Parses a model object from (jsonschema-validated) yaml into python object"""
 
     #  Only - add MdoMetadata
```
tlento, line 245: Unrelated, but what is this comment referring to?

metricflow/model/parsing/dir_to_model.py
```diff
@@ -237,19 +236,20 @@ def parse_config_yaml(
 
 def parse(  # type: ignore[misc]
     _type: Type[Union[DataSource, Metric, Materialization]],
-    ctx: ParsingContext,
     yaml_dict: Dict[str, Any],
     filename: str,
-    contents: str,
+    contents_by_line: List[str],
```
tlento, line 241: I don't have enough context on the YAML parser to understand why this is necessary. Would you mind updating the commit message to include an explanation of this why we're making this change?

tlento (commented):
Are we expecting end users to provide the Metadata at any point, or is it an internal framework addition?

If this is always managed internal to the framework, we need to make it an Optional property in the model objects (or provide an empty object default), and we might consider calling it InternalMetadata or something (although that's much less important).

If this is something end users should provide, in whole or in part, we'll want to update the jsonschema accordingly, and we should probably designate it as optional anyway for now so we can roll it out without breaking existing models.

WilliamDee:
> Are we expecting end users to provide the Metadata at any point, or is it an internal framework addition?
> 
> If this is always managed internal to the framework, we need to make it an Optional property in the model objects (or provide an empty object default), and we might consider calling it InternalMetadata or something (although that's much less important).
> 
> If this is something end users should provide, in whole or in part, we'll want to update the jsonschema accordingly, and we should probably designate it as optional anyway for now so we can roll it out without breaking existing models.

Yup, changed it to Optional. This isn't something provided by the config, but when configs are being parsed it'll automatically insert the metadata. So this change shouldn't affect end users.

tlento (approved):
@WilliamDee you are too quick for me. :) This change looks fine, but if you could update the commit message as requested inline prior to merge that'd be great. Thanks!

metricflow/model/parsing/dir_to_model.py
WilliamDee: So initially, this wasn't even being used, but we do something similar in the old parser where we do a `contents.splitline()` to get the yaml lines by rows and get the file slice with the `context.start_line:context.end_line`. But it feels like just passing the splitted content already will avoid us from having to do a split every call to `parse` since it is recursive and it'll get very repetitive. I'll try to amend the commit message!

metricflow/model/parsing/dir_to_model.py
WilliamDee: To be frank, I'm not sure haha. I'll update this to be clearer.

PR #78, opened 2022-05-11
added description/owners to model objects

WilliamDee:
## Changes
Similar to https://github.com/transform-data/metricflow/pull/75 we will add addition metadata onto the model objects for extra details, specifically 
- description
- owners

### DIFF
--- a/metricflow/model/objects/data_source.py
+++ b/metricflow/model/objects/data_source.py
@@ -51,6 +51,7 @@ class DataSource(HashableBaseModel, ParseableObject):
     """Describes a data source"""
 
     name: str
+    description: Optional[str]
     sql_table: Optional[str]
     sql_query: Optional[str]
     dbt_model: Optional[str]
--- a/metricflow/model/objects/elements/dimension.py
+++ b/metricflow/model/objects/elements/dimension.py
@@ -35,6 +35,7 @@ class Dimension(HashableBaseModel, Element, ParseableObject):
     """Describes a dimension"""
 
     name: DimensionReference
+    description: Optional[str]
     type: DimensionType
     is_partition: bool = False
     type_params: Optional[DimensionTypeParams]
--- a/metricflow/model/objects/elements/identifier.py
+++ b/metricflow/model/objects/elements/identifier.py
@@ -31,6 +31,7 @@ class Identifier(HashableBaseModel, Element, ParseableObject):
     """Describes a identifier"""
 
     name: IdentifierReference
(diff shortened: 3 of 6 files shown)


metricflow/model/objects/data_source.py
```diff
@@ -51,6 +51,8 @@ class DataSource(HashableBaseModel, ParseableObject):
     """Describes a data source"""
 
     name: str
+    description: Optional[str]
+    owners: Optional[List[str]]
```
tlento, line 55: Two questions:

1. Do we need owners at this point? If not, do we really want to add this right now?
2. This would be better as an `Optional[List[Owner]]` instead of an `Optional[List[str]]`, although I guess the bigger issue is with our config spec (where we require a List[str] in the YAML instead of a List[struct] or whatever it's called) than whatever it is we do here.

tlento (commented):
What will we do with these properties? Also, are they already in the model parser today?

metricflow/model/objects/data_source.py
WilliamDee: 1. I'm assuming we would want it, tho it was originally planned to be an internal field, @plypaul felt it would be better to move it to oss
2. I agree it would be better to change it to a struct. however, if we do, it would break existing configs as we would have to do something like this
```yaml
owners:
  - email: test@gmail.com
  - email: test2@gmail.com
```

rather than the current version,
```yaml
owners: 
  - test@gmail.com
  - test2@gmail.com
```
As for the  model parser, it parses everything that exists in the model object, so by adding the additional fields, it would automatically be parsing it. (Look at https://github.com/transform-data/metricflow/pull/67 for reference)

tlento: Actually I got a bright idea and I'm experimenting with a way to make the array<string> -> array<struct> transition backwards compatible without adding a pile of internal technical debt, I just don't know if it works in our specific parsing/validation setup.

Separately, it appears there's some required/optional piece to this. I'll message you and if that works ok we can go as-is, but otherwise maybe we hold off on the owners field until we get that sorted out just so we can continue making progress.

tlento (approved):
Per offline discussion this looks like it'll be totally workable one way or another.

WilliamDee:
After abit of discussion, we decided to keep `owners` out of oss for now as we may want to change the structure of the `owners` field. Also, it isn't *that* important of a field as there are not much (if any) use cases in OSS. This means we can add it later on once we finalize how we want `owners` to look like.

PR #80, opened 2022-05-18
Update ModelValidator to allow for a configureable list of rules

QMalcolm:
This PR makes the ModelValidator configurable. Specifically it makes it so that the lists of rules that are run is configurable. This is important because it powers two important use cases. There are times when people only want to run a subset of rules (for instance when the only thing that was changed were the dimensions in a metric) or where custom validations are needed (for instance if someone extends the parsing schema so that all data sources have owners and they want to validate all data sources have owners).

# Example uses:
## Run validator with default rules
```
from metricflow.model.model_validator import ModelValidator

ModelValidator().validate_model(your_model)
```

## Run validator with only MetricMeasuresRule
```
from metricflow.model.model_validator import ModelValidator
from metricflow.model.validations.metrics import MetricMeasuresRule

ModelValidator(rules=[MetricMeasuresRule]).validate_model(your_model)
```

## Run the validator with default rules and custom DataSourceOwnersRule
```
from metricflow.model.model_validator import ModelValidator
from path.to.custom.rule.module import DataSourceOwnersRule

ModelValidator(rules=list(ModelValidator.DEFAULT_RULES) + [DataSourceOwnersRule]).validate_model(your_model)
```

### DIFF
--- a/metricflow/model/model_validator.py
+++ b/metricflow/model/model_validator.py
@@ -55,13 +55,12 @@ class ModelValidator:
         """
         self._rules = rules
 
-    @staticmethod
-    def validate_model(model: UserConfiguredModel) -> ModelBuildResult:
+    def validate_model(self, model: UserConfiguredModel) -> ModelBuildResult:
         """Validate a model according to configured rules."""
         model_copy = copy.deepcopy(model)
 
         issues: List[ValidationIssueType] = []
-        for validation_rule in ModelValidator.DEFAULT_RULES:
+        for validation_rule in self._rules:
             issues.extend(validation_rule.validate_model(model_copy))
             # If there are any fatal errors, stop the validation process.
             if any([x.level == ValidationIssueLevel.FATAL for x in issues]):
@@ -73,11 +72,10 @@ class ModelValidator:
 
         return ModelBuildResult(model=model_copy, issues=tuple(issues))
 
-    @staticmethod
-    def checked_validations(model: UserConfiguredModel) -> UserConfiguredModel:  # chTODO: remember checked_build
+    def checked_validations(self, model: UserConfiguredModel) -> UserConfiguredModel:  # chTODO: remember checked_build
         """Similar to validate(), but throws an exception if validation fails."""
         model_copy = copy.deepcopy(model)
-        build_result = ModelValidator.validate_model(model_copy)
+        build_result = self.validate_model(model_copy)
         if build_result.issues is not None:
             if any(
                 [
--- a/metricflow/model/model_validator.py
+++ b/metricflow/model/model_validator.py
@@ -1,6 +1,6 @@
 import copy
 import logging
-from typing import List
+from typing import List, Sequence
 
 from metricflow.model.objects.user_configured_model import UserConfiguredModel
 from metricflow.model.parsing.dir_to_model import ModelBuildResult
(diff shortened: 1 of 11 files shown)


metricflow/model/model_validator.py
```diff
@@ -32,7 +32,7 @@
 class ModelValidator:
     """A Validator that acts on UserConfiguredModel"""
 
-    VALIDATION_RULES: List[ModelValidationRule] = [
+    DEFAULT_RULES: List[ModelValidationRule] = [
```
tlento, line 35: Let's make this a Tuple so it can't be accidentally mutated in the runtime...

metricflow/model/model_validator.py
```diff
@@ -47,13 +47,21 @@ class ModelValidator:
         ValidMaterializationRule(),
     ]
 
+    def __init__(self, rules: List[ModelValidationRule] = DEFAULT_RULES) -> None:
```
tlento, line 50: .... and then this takes a `Tuple[ModelValidationRule, ...]`, or maybe a `Sequence[ModelValidationRule]`

metricflow/test/model/validations/test_configurable_rules.py
```diff
@@ -0,0 +1,31 @@
...
+        ],
+    )
+
+    # confirm that with the default configuration, an issue is raised
+    issues = ModelValidator().validate_model(model).issues
+    assert len(issues) == 1, f"ModelValidator with default rules had unexpected number of issues {issues}"
+
+    # confirm that with an empty configuration, no issue is raised
+    issues = ModelValidator(rules=[]).validate_model(model).issues
+    assert len(issues) == 0, f"Empty rules list configured ModelValidator returned issues {issues}"
```
tlento, line 26: This feels like it belongs in a different test case.

Also, do we want to allow empty validation runs? If not, let's add an assertion and then split this into a test case that makes sure the exception gets thrown correctly.

tlento (commented):
Nice! Left some inline comments.

One question: should there be some basic required validations we always run? Maybe that's a thing to address later, it's easy enough to add.

metricflow/model/model_validator.py
QMalcolm: Oooo, yes! Good call!

metricflow/test/model/validations/test_configurable_rules.py
```diff
@@ -21,11 +22,18 @@ def test_can_configure_model_validator_rules(simple_model__pre_transforms: UserC
...
+
+def test_cant_configure_model_validator_without_rules() -> None:  # noqa: D
+    with pytest.raises(ValueError):
+        ModelValidator(rules=[])
+
+    with pytest.raises(ValueError):
+        ModelValidator(rules=())
+
+    with pytest.raises(ValueError):
+        ModelValidator(rules=None)  # type: ignore
```
tlento, line 39: I think you can remove this - this case is not supported according to the API so there's no need for the added test.

tlento (approved):
Looks good to me!

metricflow/test/model/validations/test_configurable_rules.py
QMalcolm: Yea I mostly put it in because I expect someone to do something like `ModelValidator(rules=None)  # type: ignore` or similar as a work around something, I know I've done things like this myself in the past. In the case that someone does, it should and will throw the `ValueError` exception.

PR #82, opened 2022-05-23
Skip Hidden Directories and Files in the YAML Model Directory

plypaul:
The model directory contains the YAML files that define the semantic model. Currently, it parses all YAML files in the directory, but hidden files and directories should be skipped (e.g. `.git`).


tlento (approved):
Seems reasonable. 

I was actually just wondering if we did this, because we might want to use a hidden directory for system-managed model artifacts (like VERSION), but if we do that we should probably have a dedicated parsing function anyway.

PR #83, opened 2022-05-23
Validate Aggregation Time Dimension for Measures

plypaul:
This PR adds a validation rule that ensures that the aggregation time dimension set for a measure in a data source references a valid time dimension in the same data source.

### DIFF
--- a/metricflow/model/validations/agg_time_dimension.py
+++ b/metricflow/model/validations/agg_time_dimension.py
@@ -8,7 +8,7 @@ from metricflow.model.validations.validator_helpers import (
     ValidationIssueType,
     validate_safely,
     ValidationError,
-    ValidationIssue,
+    MeasureContext,
 )
 from metricflow.specs import TimeDimensionReference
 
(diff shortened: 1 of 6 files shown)


metricflow/model/validations/agg_time_dimension.py
```diff
@@ -0,0 +1,66 @@
...
+            model_object_reference = ValidationIssue.make_object_reference(
+                data_source_name=data_source.name, measure_name=measure.name
+            )
+            if not measure.agg_time_dimension:
+                issues.append(
+                    ValidationError(
+                        model_object_reference=model_object_reference,
+                        message=f"In data source '{data_source.name}', measure '{measure.name}' does not have an "
+                        f"aggregation time dimension set",
+                    )
```
tlento, line 50: Ah, because that's a pre-validation transformation step. We should think about ways to improve the parse -> transform -> validate -> transform flow, or at least the labeling of it, as it's kind of confusing to keep track of which state the model is in when one of these things gets called. That's an issue for later, for now maybe we can make it more obvious like this?
```suggestion
            if not measure.agg_time_dimension:
                issues.append(
                    ValidationError(
                        model_object_reference=model_object_reference,
                        message=f"In data source '{data_source.name}', measure '{measure.name}' does not have an "
                        f"aggregation time dimension set. This can happen if validation is called before the first " 
                        f"model transformation step.",
                    )
```

metricflow/model/validations/agg_time_dimension.py
```diff
@@ -0,0 +1,66 @@
...
+                )
+                continue
+            agg_time_dimension_reference = measure.checked_agg_time_dimension
```
tlento, line 53: This throws the assertion we need. Can we use that here instead of converting it to a validation error as above? 

I don't think we ever want a code path where someone runs the validators on an un-transformed model, because it'll lead to potentially huge numbers of validation failures. Seems better to indicate the order of operations flaw at the earliest possible moment in cases, like this one, where the only reason this value isn't set is because the model hasn't been processed correctly.

metricflow/model/validations/agg_time_dimension.py
```diff
@@ -0,0 +1,66 @@
...
+            if not AggregationTimeDimensionRule._time_dimension_in_model(
+                time_dimension_reference=agg_time_dimension_reference, data_source=data_source
+            ):
+                issues.append(
+                    ValidationError(
+                        model_object_reference=model_object_reference,
+                        message=f"In data source '{data_source.name}', measure '{measure.name}' has the aggregation "
+                        f"time dimension is set to '{agg_time_dimension_reference.element_name}', "
+                        f"which not a valid time dimension in the data source",
```
tlento, line 62: nit:
```suggestion
                        message=f"In data source '{data_source.name}', measure '{measure.name}' the aggregation "
                        f"time dimension is set to '{agg_time_dimension_reference.element_name}', "
                        f"which not a valid time dimension in the data source",
```

metricflow/model/validations/agg_time_dimension.py
plypaul: Updated.

metricflow/model/validations/agg_time_dimension.py
plypaul: Updated.

metricflow/model/validations/agg_time_dimension.py
plypaul: Updated.

PR #85, opened 2022-05-24
Data Warehouse Validation Runner and Data Source Data Warehouse Validations

QMalcolm:
To truly validate the model we have to check that what is said to exist via the model does actually exist in the data warehouse. Thus this PR introduces the `DataWarehouseModelValidator`. The `DataWarehouseModelValidator` has a method `run_tasks` which is handed a list of `DataWarehouseValidationTask`s that it runs through, compiling any errors that are run into.

Additionally, as a jumping off point, this PR adds the Data Source Data Warehouse Validations using the `DataWarehouseModelValidator` and it is now included as part of the cli command `mf validate-configs`

### Note:
Presently the the `DataWarehouseModelValidator` treats all tasks as just queries that we want to ensure run without issue. Thus we convert the queries to EXPLAIN queries (or set the engine connection to dry run mode) because we only care about the validity of the query. However, it's worth noting that sometime in the future it might be important to be able to validate the results of a query. If so, this architecture should be extensible allowing it to support such a use cases. I.e. it'd be extended to support different _types_ of `DataWarehosueValidationTask`s


metricflow/model/data_warehouse_model_validator.py
```diff
@@ -0,0 +1,120 @@
...
+    QUERY_TEMPLATE = "SELECT {columns_select} FROM {data_source}"
+    PARTITION_TEMPLATE = " WHERE {partition_col} IS NOT NULL"
+    WRAPPED_COL_TEMPLATE = "({column}) as col{unique_id}"
+
+    @staticmethod
+    def _wrap_col(col: str, id: int) -> str:  # noqa: D
+        return DataWarehouseTaskBuilder.WRAPPED_COL_TEMPLATE.format(column=col, unique_id=id)
+
+    @staticmethod
+    def _gen_query(data_source: DataSource, columns: List[str] = ["true"]) -> str:  # noqa: D
```
plypaul, line 33: Nit: elaborate on the query it's supposed to generate.

metricflow/model/data_warehouse_model_validator.py
```diff
@@ -0,0 +1,120 @@
...
+                DataWarehouseValidationTask(
+                    query_string=DataWarehouseTaskBuilder._gen_query(data_source=data_source),
+                    object_ref=ValidationIssue.make_object_reference(data_source_name=data_source.name),
+                    error_message=f"Unable to access data source `{data_source.name}` in data warehouse",
+                )
+            )
+        return tasks
+
+
+class DataWarehouseModelValidator:
```
plypaul, line 71: Nit: elaborating more on what's a data warehouse validation

metricflow/sql_clients/big_query.py
```diff
@@ -56,12 +57,31 @@ def __init__(self, project_id: str = "", password: Optional[str] = None, query_s
...
+    def _engine_specific_explain_implementation(
+        self, stmt: str, bind_params: SqlBindParameters
+    ) -> Optional[List[Exception]]:
+        """Overrides base `_engine_specific_explain_implementation` function for BigQuery specifics"""
+        _engine = self._create_bq_engine(
+            query_string="dry_run=true", project_id=self._project_id, password=self._password
+        )
+        exceptions: List[Exception] = []
+        with _engine.connect() as conn:
+            for s in stmt.split(";"):
```
plypaul, line 76: This is buggy - `;` can be in a string literal. Ditto for the other similar call site. Is this splitting necessary?

metricflow/model/data_warehouse_model_validator.py
```diff
@@ -0,0 +1,120 @@
...
+        columns_select = ", ".join(wrapped_cols)
+
+        query = DataWarehouseTaskBuilder.QUERY_TEMPLATE.format(
+            data_source=data_source_str, columns_select=columns_select
+        )
+
+        # If the data source has a partition, we need to include it in the
+        # query. This is because some data warehouses require the partition as
+        # part of the query for any table that has a partition defined
+        if data_source.partition:
```
plypaul, line 46: What happens if there are multiple partitions?

QMalcolm: What happens on the data warehouse side or what happens here when generating the query? 

My understanding is that the general rule on the data warehouse side is that one partition must be supplied (not all). Will that cover every data warehouse? Probably not, but time will tell and we'll tackle that when it comes up.

As far as generating the query, `data_source.partition` will only ever return one partition. So the query will be generated with one `WHERE` clause selection.

metricflow/sql_clients/big_query.py
QMalcolm: Doing some research we could use [sqlparse](https://pypi.org/project/sqlparse/) so that we can use `sqlparse.split()`

metricflow/model/data_warehouse_model_validator.py
plypaul: I was asking what happens if there are multiple partition dimensions defined in the data source. `data_source.partition` will return one partition, but that's only if the data source has been defined with one partition. From looking at the implementation and usages of that call, it seems like a legacy carry over.

There's also a difference between semantic partitions and physical partitions, but resolving that difference might be out of scope for this PR.

metricflow/sql_clients/big_query.py
QMalcolm: Paul and I talked offline. We don't want to encourage multiple statements in a statement to be queried / executed / explained. MetricFlow does not currently do these kind of compound statements. Thus we're going to drop the splitting.

metricflow/model/data_warehouse_model_validator.py
QMalcolm: Paul and I talked offline about this. Semantic partitions and data warehouse partitions might not always line up. I.e. a semantic partition might not actually be a partition in the data warehouse. It isn't guaranteed that any partition defined in the configs is a partition in the data warehouse. So our best bet is to include any dimension defined as a partition for a data source in the WHERE clause and hope one of them is a partition in the data warehouse.

metricflow/test/model/validations/test_data_warehouse_tasks.py
```diff
@@ -0,0 +1,68 @@
...
+                    ),
+                ],
+                mutability=Mutability(type=MutabilityType.IMMUTABLE),
+            ),
+        ],
+        metrics=[],
+        materializations=[],
+    )
+
+    tasks = DataWarehouseTaskBuilder.gen_data_source_tasks(model=model)
```
plypaul, line 40: Can you also add a call for `validate_data_sources()`?

QMalcolm: Part of me thinks that having a `validate_data_sources` test doesn't actually add value. Currently `validate_data_sources` is two lines, a call to `gen_data_source_tasks` and a call to `run_tasks` with those returned tasks. And both of these methods are currently tested. However, I guess a test specifically for `validate_data_sources` ensures that changes directly to `validate_data_sources` doesn't break it 🤷

plypaul: It - is `validate_data_sources` the main method that clients are expected to use?

QMalcolm: The `validate_data_sources` method is essentially a helper that generates a default set of validation tasks to run. We will eventually also have `validate_dimensions`, `validate_measures`, `validate_identifiers`, `validate_metrics`. All of these will essentially be two line functions (composing exposed `gen_<type>_tasks` on the `DataWarehouseTaskBuilder` and `run_tasks`). The helpers listed, `validate_data_sources` and the future ones, will be called by the CLI and likely Transform's internal servers. Others might call them, but it's more likely they'll compose a huge list of tasks using the different `gen_<type>_tasks` and their own custom tasks, and then running that list through `run_tasks`.

plypaul (approved):
LGTM

PR #89, opened 2022-06-01
Add Metric based Data Warehouse Validations

QMalcolm:
Perhaps the most important data warehouse validation is ensuring all metrics defined in the model will run in the data warehouse. This PR adds metric based data warehouse validations to the `DataWarhouseModelValidator`. Additionally, these validations are now run as part of `validate-configs`

### DIFF
--- a/metricflow/model/data_warehouse_model_validator.py
+++ b/metricflow/model/data_warehouse_model_validator.py
@@ -2,6 +2,7 @@ from collections import OrderedDict
 from dataclasses import dataclass, field
 from time import perf_counter
 from typing import List, Optional, OrderedDict as ODType
+from metricflow.engine.metricflow_engine import MetricFlowEngine
 
 from metricflow.model.objects.data_source import DataSource
 from metricflow.model.objects.user_configured_model import UserConfiguredModel
@@ -94,8 +95,9 @@ class DataWarehouseModelValidator:
     them (assuming the model has passed these validations before use).
     """
 
-    def __init__(self, sql_client: SqlClient) -> None:  # noqa: D
+    def __init__(self, sql_client: SqlClient, mf_engine: MetricFlowEngine) -> None:  # noqa: D
         self._sql_client = sql_client
+        self._mf_engine = mf_engine
 
     def run_tasks(
         self, tasks: List[DataWarehouseValidationTask], timeout: Optional[int] = None
--- a/metricflow/model/data_warehouse_model_validator.py
+++ b/metricflow/model/data_warehouse_model_validator.py
@@ -2,12 +2,14 @@ from collections import OrderedDict
 from dataclasses import dataclass, field
 from time import perf_counter
 from typing import List, Optional, OrderedDict as ODType
-from metricflow.engine.metricflow_engine import MetricFlowEngine
+from metricflow.engine.metricflow_engine import MetricFlowEngine, MetricFlowExplainResult, MetricFlowQueryRequest
 
 from metricflow.model.objects.data_source import DataSource
 from metricflow.model.objects.user_configured_model import UserConfiguredModel
+from metricflow.model.semantic_model import SemanticModel
 from metricflow.model.validations.validator_helpers import ValidationError, ValidationIssue, ValidationWarning
 from metricflow.protocols.sql_client import SqlClient
+from metricflow.sql.sql_bind_parameters import SqlBindParameters
 
 
 @dataclass
@@ -17,6 +19,7 @@ class DataWarehouseValidationTask:
     query_string: str
     error_message: str
     object_ref: ODType = field(default_factory=lambda: OrderedDict())
+    query_params: SqlBindParameters = field(default_factory=lambda: SqlBindParameters())
 
 
 class DataWarehouseTaskBuilder:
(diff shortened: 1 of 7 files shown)


QMalcolm:
Oof. Looks like I missed a borked test in `test_cli.py`. That's what I get for running `poetry run pytest ...` on specific files. On it.

metricflow/test/model/validations/test_data_warehouse_tasks.py
```diff
@@ -100,3 +138,76 @@ def test_validate_data_sources(sql_client: SqlClient) -> None:  # noqa: D
     assert len(issues) == 1
     assert issues[0].level == ValidationIssueLevel.ERROR
     assert "Unable to access data source `test_data_source2`" in issues[0].message
+
+
+def test_build_metric_tasks(basic_model: UserConfiguredModel, mf_engine: MetricFlowEngine) -> None:  # noqa: D
+    tasks = DataWarehouseTaskBuilder.gen_metric_tasks(model=basic_model, mf_engine=mf_engine)
+    assert len(tasks) == 1
+    assert tasks[0].query_string == (
```
plypaul, line 146: The query text may change, so to make updating tests easier, can you take a look at `assert_snapshot_text_equal` to do the comparison?

metricflow/test/model/validations/test_data_warehouse_tasks.py
```diff
@@ -1,52 +1,99 @@
...
+from metricflow.test.time.configurable_time_source import ConfigurableTimeSource
+from metricflow.time.time_granularity import TimeGranularity
 
 
-def test_build_data_source_tasks() -> None:  # noqa:D
-    dim_ref_1 = DimensionReference(element_name="dim1")
-    dim_ref_2 = DimensionReference(element_name="dim1")
-
+@pytest.fixture
+def basic_model() -> UserConfiguredModel:  # noqa: D
```
plypaul, line 28: We've been using YAML files to define models - can you take a look at `metricflow/test/fixtures/model_yamls`?

plypaul: Also `metricflow/test/fixtures/model_fixtures.py` would be relevant as well.

metricflow/model/data_warehouse_model_validator.py
```diff
@@ -139,3 +166,13 @@ def validate_data_sources(self, model: UserConfiguredModel, timeout: Optional[in
         """
         tasks = DataWarehouseTaskBuilder.gen_data_source_tasks(model=model)
         return self.run_tasks(tasks=tasks, timeout=timeout)
+
+    def validate_metrics(self, model: UserConfiguredModel, timeout: Optional[int] = None) -> List[ValidationIssue]:
+        """Generates a list of tasks for validating the metrics of the model and then runs them
+
+        :param model UserConfiguredModel: Model which to run data warehouse validations on
```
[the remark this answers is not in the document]
QMalcolm: Ahhh gotcha. I think I started using this pattern because of how things were commented in `base_sql_client_implementation.py` when I was writing the `dry_run` method. Looks like we might have some cleanup to do sometime to get `element_const.py`, `sql_client.py`, and `base_sql_client_implementation.py` comments converted sometime.

metricflow/test/model/validations/test_data_warehouse_tasks.py
QMalcolm: In an iteration of this testing file we had been using the `simple_semantic_model` in `model_fixtures.py`. I had reasons for refactoring away from it in this test suite, but I now forget what those were 🙃 In the commit where this happened I added the following as part of the commit message

> Additionally, these tests need an easy to use model which is always valid by default. Thus an entirely self referential model was added as a fixture.

It may be that the `simple_semantic_model` wasn't succeeding the validations we were checking in these tests for whatever reason. I'll double check to see if that's the case.

QMalcolm: Yea so I tried swapping out the `basic_model` I defined as a fixture for the `simple_user_configured_model` fixture defined in `model_fixtures.py` and without any changes that model has 9 data source validation issues and hits `Unkown element 'lux_or_normal_listing'` when generating metric queries. I believe the data source issues are because the data warehouse queried during testing is empty or has a different schema. The reason I built out the `basic_model` fixture was to make an entirely self referential model, i.e. it works against even an empty data warehouse.

QMalcolm: Paul and I chatted offline. I misread Paul's comments 🤦 We're going to formalize this model so that it's available to other tests as well as a fixture.

metricflow/test/fixtures/model_fixtures.py
```diff
@@ -249,3 +249,15 @@ def extended_date_semantic_model(mf_test_session_state: MetricFlowTestSessionSta
     )
     assert model_build_result.model
     return SemanticModel(model_build_result.model)
+
+
+@pytest.fixture(scope="session")
+def self_referential_model(template_mapping: Dict[str, str]) -> UserConfiguredModel:
```
plypaul, line 255: Curious what you mean by self-referential?

plypaul (approved):
LGTM aside from comment.

metricflow/test/fixtures/model_fixtures.py
QMalcolm: Paul and I came up with a name offline. The most applicable guidance is to name the model in regards to the tests it is relevant to. So the name will become `data_warehouse_validation_test_model`

PR #91, opened 2022-06-03
Lint model config yaml files during 'validate-configs' command

QMalcolm:


QMalcolm:
This needs more thought if this is actually necessary. The primary errors we want to deal with are syntax errors that break the parser like the following example
```
Flask caught error: Failed to parse yaml in file. Error: while parsing a block mapping
  in "<unicode string>", line 267, column 1:
    materialization:
    ^
expected <block end>, but found '<block mapping start>'
  in "<unicode string>", line 283, column 2:
     Optional destinations:
```
The linter won't actually take care of this. We should instead perhaps focus on adding more context (like file name) to the error message, and maybe also moving parsing exceptions to ValidationIssues.

PR #92, opened 2022-06-03
Add Method to Cancel Running Queries

plypaul:
This PR add a method to `SqlClient` to enable cancelling of running queries. This could be useful if a large query is submitted by mistake and the user wants to cancel running it.

### DIFF
--- a/metricflow/cli/main.py
+++ b/metricflow/cli/main.py
@@ -1,3 +1,6 @@
+import logging
+import signal
+
 import click
 import datetime as dt
 import jinja2
@@ -38,6 +41,8 @@ from metricflow.telemetry.models import TelemetryLevel
 from metricflow.telemetry.reporter import TelemetryReporter, log_call
 from metricflow.dag.dag_visualization import display_dag_as_svg
 
+logger = logging.getLogger(__name__)
+
 pass_config = click.make_pass_decorator(CLIContext, ensure=True)
 _telemetry_reporter = TelemetryReporter(report_levels_higher_or_equal_to=TelemetryLevel.USAGE)
 _telemetry_reporter.add_python_log_handler()
@@ -62,8 +67,30 @@ def cli(cfg: CLIContext, verbose: bool) -> None:  # noqa: D
         )
 
         click.echo(
-            f"💡 Please update to version {result.available_version}, released {result.release_date} by running:\n\t$ pip install --upgrade {PACKAGE_NAME}",
+            f"💡 Please update to version {result.available_version}, released {result.release_date} by running:\n"
+            f"\t$ pip install --upgrade {PACKAGE_NAME}",
         )
+
+    # Cancel queries submitted to the DW if the user precess CTRL + c / process is terminated.
+    # Note: docs unclear on the type for the 'frame' argument.
+    def exit_signal_handler(signal_type: int, frame) -> None:  # type: ignore
+        if signal_type == signal.SIGINT:
+            click.echo("Got SIGINT")
+        elif signal_type == signal.SIGTERM:
+            click.echo("Got SIGTERM")
+        else:
+            # Shouldn't happen since this should ony be registered for SIGINT / SIGTERM.
+            click.echo(f"Got signal {signal_type}")
+
+        if (
+            signal_type in (signal.SIGINT, signal.SIGTERM)
+            and cfg.sql_client.sql_engine_attributes.cancel_submitted_queries_supported
+        ):
+            logger.info("Cancelling submitted queries")
+            cfg.sql_client.cancel_submitted_queries()
+
+    signal.signal(signal.SIGINT, exit_signal_handler)
+    signal.signal(signal.SIGTERM, exit_signal_handler)
 
 
 @cli.command()
--- a/metricflow/protocols/sql_client.py
+++ b/metricflow/protocols/sql_client.py
@@ -137,6 +137,11 @@ class SqlClient(Protocol):
         """Close the connections / engines used by this client."""
         raise NotImplementedError
 
+    @abstractmethod
+    def cancel_submitted_queries(self) -> None:  # noqa: D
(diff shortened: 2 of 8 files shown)


metricflow/cli/main.py
```diff
@@ -62,9 +67,32 @@ def cli(cfg: CLIContext, verbose: bool) -> None:  # noqa: D
...
         )
 
+    # Cancel queries submitted to the DW if the user precess CTRL + c / process is terminated.
+    # Note: docs unclear on the type for the 'frame' argument.
+    def exit_signal_handler(signal_type: int, frame) -> None:  # type: ignore
+        if signal_type == signal.SIGINT:
+            click.echo("Got SIGINT")
+        elif signal_type == signal.SIGTERM:
+            click.echo("Got SIGTERM")
+            signal.valid_signals()
```
WilliamDee, line 81: do we need this? checked the docs it says 
```
Return the set of valid signal numbers on this platform. 
This can be less than range(1, NSIG) if some signals are reserved by the system for internal use
```
doesn't seem like we're using the return value

metricflow/cli/main.py
```diff
@@ -62,9 +67,32 @@ def cli(cfg: CLIContext, verbose: bool) -> None:  # noqa: D
...
+        else:
+            # Shouldn't happen since this should ony be registered for SIGINT / SIGTERM.
+            click.echo(f"Got signal {signal_type}")
+
+        if (
+            signal_type in (signal.SIGINT, signal.SIGTERM)
+            and cfg.sql_client.sql_engine_attributes.cancel_submitted_queries_supported
+        ):
+            logger.info("Cancelling submitted queries")
```
WilliamDee, line 90: where would this log to?

WilliamDee (approved):
2 small things, otherwise lgtm

metricflow/test/sql_clients/test_sql_client.py
```diff
@@ -128,3 +134,30 @@ def test_dry_run_of_bad_query_raises_exception(sql_client: SqlClient) -> None:
...
+) -> None:
+    if (
+        not sql_client.sql_engine_attributes.sleep_supported
+        or not sql_client.sql_engine_attributes.cancel_submitted_queries_supported
+    ):
+        logger.info(
+            f"Skipping test {request.node.fspath}::{request.node.name} since the sql_client does not support features "
+            f"required for this test."
+        )
+        return
```
tlento, line 150: Hm. Maybe call `pytest.skip()` here instead of a logger.info() and return statement? That way the results show a skip rather than a fake pass for engines that don't support this - having code that "passes" in the relevant target tests only to find out much later that it was being bypassed is a deeply frustrating dev experience.

tlento (commented):
I feel like there has to be a better-supported way of doing this but I've actually spent a fair bit of time searching for one and have turned up exactly nothing. Every engine does it differently and SqlAlchemy doesn't abstract any of this away as far as I can tell (probably because every engine does it differently). Sigh.

I left some mechanical comments inline to go along with Will's questions.

metricflow/cli/main.py
plypaul: Will remove - this was left from testing.

metricflow/cli/main.py
plypaul: This would log to a file in `~/.metricflow/logs`.

metricflow/protocols/sql_client.py
```diff
@@ -137,6 +137,16 @@ def close(self) -> None:  # noqa: D
         """Close the connections / engines used by this client."""
         raise NotImplementedError
 
+    @abstractmethod
+    def sleep(self, sleep_time_ms: int) -> None:  # noqa: D
```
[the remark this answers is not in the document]
plypaul: Do you have thoughts on how the helper could detect a Snowflake client?

metricflow/sql_clients/snowflake.py
```diff
@@ -160,13 +164,21 @@ def engine_connection(self, engine: sqlalchemy.engine.Engine) -> Iterator[sqlalc
...
+                sessions.append(row[0])
+            assert len(sessions) == 1
+            session = sessions[0]
+            with self._known_sessions_ids_lock:
+                self._known_session_ids.add(session)
             yield conn
-        finally:
-            conn.close()
+            with self._known_sessions_ids_lock:
+                self._known_session_ids.remove(session)
```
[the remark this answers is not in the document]
plypaul: I didn't see any when I looked.

metricflow/test/sql_clients/test_sql_client.py
plypaul: The issue with the skip annotation is that it can't make use of the built in fixtures.

plypaul: We could make that happen with a global variable, but that doesn't seem great.

plypaul: Actually, it seems like `pytest.xfail` can help accomplish similar behavior.

metricflow/protocols/sql_client.py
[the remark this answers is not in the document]
tlento: You have the attributes accesible from the sql client fixture, so you should be able to rely on the engine enumeration value in there.

metricflow/test/sql_clients/test_sql_client.py
tlento: You can use pytest.skip inside the method. It's just a function call, behaves just like the decorator but it's invoked imperatively.

If the xfail decorator will work outside the function that's way better though.

plypaul: Nice, didn't realize you could call `pytest.skip` inside the method. I had seen docs for using `pytest.xfail` as a function call, so I was referring to doing that. I haven't seen a way of using `pytest.xfail` as a decorator for this scenario.

PR #94, opened 2022-06-07
update numpy and pyjwt

alliehowe:
- updated numpy from 1.22.0 to 1.22.2 to fix https://security.snyk.io/vuln/SNYK-PYTHON-NUMPY-2321964
- updated pyjwt from 2.3.0 to 2.4.0 to fix https://github.com/advisories/GHSA-ffqj-6fqr-9h24


pyproject.toml
```diff
@@ -37,6 +37,7 @@ click = "^7.1.2"
 "ruamel.yaml" = "^0.17.21"
 python-Levenshtein = "^0.12.2"
 rudder-sdk-python = "^1.0.3"
+PyJWT = ">=2.4.0"
```
tlento, line 40: This is a transitive dependency - we don't actually use it anywhere. Here's where it's coming from:

image

Looks like these are test environment requirements (see https://github.com/snowflakedb/snowflake-connector-python/blob/main/tested_requirements/requirements_39.reqs#L11 for an example - the readme in there indicates these are for testing alone) - the snowflake-sqlalchemy dependency comes from the snowflake-connector so it's the same deal.

tlento: Ok, looks like if we update line 22 to be 2.7.8 this should get resolved along the way.

poetry.lock
```diff
@@ -2002,19 +2002,22 @@ six = [
...
-    {file = "snowflake_connector_python-2.7.6-cp38-cp38-manylinux_2_12_x86_64.manylinux2010_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl", hash = "sha256:f59ed0c3223548a25aca9bcacf2c5cde04ad5e15366f0cf9e7b20dc34554b1cf"},
-    {file = "snowflake_connector_python-2.7.6-cp38-cp38-win_amd64.whl", hash = "sha256:8c82eda884967ef00d1b7aa59c925c83368f886bfa447c8fa2a81dced03dea70"},
-    {file = "snowflake_connector_python-2.7.6-cp39-cp39-macosx_10_14_universal2.whl", hash = "sha256:22437ee26baaa83cabcebbe9fbc9c95b36bdca441db61822588a622e32bcbecd"},
-    {file = "snowflake_connector_python-2.7.6-cp39-cp39-manylinux_2_12_x86_64.manylinux2010_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl", hash = "sha256:8dbaf2dfed6c7ba4fd90c9f0ec604f86a2e070e4219c182f79f44ca225749c18"},
-    {file = "snowflake_connector_python-2.7.6-cp39-cp39-win_amd64.whl", hash = "sha256:d94eea3047c6cf6d5e4bb8a4b9f961f77f87e25671e7cb6257c0459419a9b194"},
+    {file = "snowflake-connector-python-2.7.8.tar.gz", hash = "sha256:9cfb07b048bcb1fe646d03f7480ccb6a877e9c45067302e90bcff7fd72f6bc2f"},
```
tlento, line 2005: Ooof. Check out line 1872. For some reason the pyjwt entry isn't getting updated there, and I have no idea why. :(

tlento (commented):
Sadly, I don't know if this worked. Somewhere around line 1872 in the poetry lock file there's still a reference to the wrong jwt version.

I'll see if I can dig in a bit more into transitive dependencies - I'm pretty sure the FOSSA dependency list I was looking at only provided first level dependencies, so maybe that entry is getting added by something else first. 

If we can't figure it out quickly we can add your original override back in so you can make progress on eliminating the open vulnerability, and I'll open an issue to clean it up later.

poetry.lock
alliehowe: Oh good catch, bummer :( I did poetry install with a fresh env to test this and did get version 2.3.0 like this says

alliehowe: I did `poetry update` and then `poetry install` in a fresh env and that worked!

alliehowe: check out my latest commit and lmk what you think

tlento (approved):
Sweet, seems like the poetry update did the trick! Thanks @alliehowe

PR #97, opened 2022-06-08
Expose Integration Test Case Repository

plypaul:
This PR changes the integration test case repository defined in the test module  from an internal variable name to a public variable name so that it can be imported and used for other testing purposes.


QMalcolm (approved):
Sweet! Thank you for doing this 🙏 Looks good to me!

PR #100, opened 2022-06-08
chore: mark snapshots as generated

rexledesma:
As the title. The generated files are pretty overwhelming for review.

#### Test Plan

Stack on https://github.com/transform-data/metricflow/pull/99

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+metricflow/test/snapshots/* linguist-generated=true


plypaul (approved):
Nice, this will make PRs faster to load! Though in many cases, the reviewer should be looking at the generated queries to check that it's doing what's it's supposed to, or if a change is causing unexpected changes.

PR #101, opened 2022-06-08
docs(sql-client): add instructions to run tests against local postgres

rexledesma:
As the title. 

#### Test Plan
In a separate window:

```
make postgresql
```

In the main window:

```
make test-postgresql
```

All tests fail until #99 lands.


CONTRIBUTING.md
```diff
@@ -19,6 +19,8 @@ Welcome to the MetricFlow developer community, we're thrilled to have you aboard
         - Mac users might prefer to use Homebrew: `brew install postgresql`
     - SQLite:
         - You likely have this installed already, but if it is missing [SQLite provides pre-built packages for download and installation](https://www.sqlite.org/download.html)
+    - Docker:
```
plypaul, line 22: Can you add that it's only needed if you're doing tests with PostgreSQL?

plypaul (approved):
Thanks for the helpful docs!

PR #102, opened 2022-06-09
Update DW Validation Model To Use Appropriate Time Types

plypaul:
The data warehouse validation tests check the SQL that is generated for metrics. In the previous setup, the data source was defined using a data source that should have SQL that does not change between SQL engines. However, this created issues as the time dimension was defined using a string, but MF expects time dimensions to be the appropriate time dimension type for the engine. This PR updates the model so that the time dimension is of the correct type, and also updates the tests so that different SQL query snapshots are produced for different engines.


QMalcolm (approved):
Nice catch! LGTM 🚢 

Is there a good way to go about catching this issue in the future. I though the test runs would have caught it 🙃

plypaul:
I see what was up - looks like we forgot to run DW engine tests on that PR.

Issue #56, opened 2022-04-29
Typo in displayed command.


AlessandroLollo:
https://github.com/transform-data/metricflow/blob/883081b951ed1abf8b5446e0552101254d94793e/metricflow/cli/main.py#L142

The displayed command should be `mf query --metrics transactions --dimensions ds --order ds --explain --display-plans`

vvlu603:
Closing, the command + tutorial were fixed in both of the merges -- thank you!

PR #104, opened 2022-06-09
Removing extra space in the MetricFlow tutorial in Step 7 of the README.md file 

vvlu603:
**Issue opened here**: https://github.com/transform-data/metricflow/issues/56 

**Why:** The tutorial section has a Step 7 with the original code `mf query --metrics transactions --dimensions ds --order ds --explain -- display-plans` (note the extra space prior to the `display-plans`). This was throwing an error. 

**What was changed:** Removing the space prior to `display-plans` to fix the command, now showing up as `mf query --metrics transactions --dimensions ds --order ds --explain --display-plans`


plypaul (approved):
Thanks for the fix!

PR #106, opened 2022-06-09
Clean up docstrings to ensure all follow google docstring formatting

QMalcolm:
We used to use a lot of reST style doc strings in our infrastructure. After cutting things over to metricflow we started using google style docstrings for documenting classes / methods / etc. However, not everything got converted over, there were still some stragglers. This PR simply makes sure any docstrings following the old reST style now follow the google style.


plypaul (approved):
🎉

PR #109, opened 2022-06-10
Add the ability to pass custom rules to ModelTransformer

WilliamDee:
## Context
Currently, we have a set of rules that gets ran for transforming the model pre/post validation. However, this isn't extensible in a way that we can't customize it for certain scenarios. So similarly to the `ModelValidator` changes, we would want to make it so we can pass custom rules that can be ran during transformation without having to run the preset rules all the time.

## Changes
- Enable the ability to pass custom build `ModelTransformRule` into the `rules` param to run those custom rules instead of the default rules

### DIFF
--- a/metricflow/model/model_transformer.py
+++ b/metricflow/model/model_transformer.py
@@ -1,11 +1,14 @@
 import copy
 import logging
 
+from typing import Sequence
+
 from metricflow.model.objects.user_configured_model import UserConfiguredModel
 from metricflow.model.transformations.boolean_measure import BooleanMeasureAggregationRule
 from metricflow.model.transformations.identifiers import CompositeIdentifierExpressionRule
 from metricflow.model.transformations.names import LowerCaseNamesRule
 from metricflow.model.transformations.proxy_measure import CreateProxyMeasureRule
+from metricflow.model.transformations.transform_rule import ModelTransformRule
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


metricflow/model/model_transformer.py
```diff
@@ -16,24 +19,35 @@ class ModelTransformer:
     Generally used to make it more convenient for the user to develop their model.
     """
 
+    DEFAULT_PRE_VALIDATION_RULES: List[ModelTransformRule] = [LowerCaseNamesRule()]
+
+    DEFAULT_POST_VALIDATION_RULES: List[ModelTransformRule] = [
+        CreateProxyMeasureRule(),
+        BooleanMeasureAggregationRule(),
+        CompositeIdentifierExpressionRule(),
+    ]
```
tlento, line 28: Let's make these tuples so they can't be mutated in the runtime.

metricflow/model/model_transformer.py
```diff
@@ -16,24 +19,35 @@ class ModelTransformer:
...
+
     @staticmethod
-    def pre_validation_transform_model(model: UserConfiguredModel) -> UserConfiguredModel:
+    def pre_validation_transform_model(
+        model: UserConfiguredModel, rules: List[ModelTransformRule] = []
```
tlento, line 32: Update the typing here and on line 45 to Sequence or Tuple.

metricflow/model/model_transformer.py
WilliamDee: done

plypaul (approved):

PR #111, opened 2022-06-13
Add DuckDB Support for Local Tests

plypaul:
This PR adds support to use DuckDB for running local unit tests. Currently, SQLite is used, but this presents limitations as queries that use full outer joins or date functions can't be tested. DuckDB has support for those features, so now all tests can be run without remote connections.

### DIFF
--- a/.github/workflows/cd-sql-engine-tests.yaml
+++ b/.github/workflows/cd-sql-engine-tests.yaml
@@ -109,6 +109,39 @@ jobs:
           MF_SQL_ENGINE_URL: ${{ secrets.MF_BIGQUERY_URL }}
           MF_SQL_ENGINE_PASSWORD: ${{ secrets.MF_BIGQUERY_PWD }}
 
+  sqlite-tests:
+    environment: DW_INTEGRATION_TESTS
+    name: SQLite Tests
+    if: ${{ github.event.action != 'labeled' || github.event.label.name == 'run_mf_sql_engine_tests' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check-out the repo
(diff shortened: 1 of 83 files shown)


tlento (approved):
This seems reasonable to me - it's certainly a big step up from SQLite in terms of capabilities. I do wonder about the following things:

1. Does DuckDB provide us with a meaningful improvement over, say, a test that makes it easy to operate against a local MySQL instance?
2. If the answer to 1 is no, are there other reasons to support DuckDB? Do they justify the extra package dependencies and stuff? If so, we should add CLI support as well so people can run `mf query` and whatnot against DuckDB.
3. Do we want to continue supporting SQLite at all if we have a readily available in-memory local client with full support for the features we need to test? Shall we open an issue to remove SQLite once we're comfortable with DuckDB in this role?

plypaul:
> Does DuckDB provide us with a meaningful improvement over, say, a test that makes it easy to operate against a local MySQL instance?

Aside from the ease of not having to setup a MySQL instance, I don't see any additional benefit to using DuckDB.

> If the answer to 1 is no, are there other reasons to support DuckDB? Do they justify the extra package dependencies and stuff? If so, we should add CLI support as well so people can run mf query and whatnot against DuckDB.

I'd say the ease in testing makes the dependencies worthwhile. Yeah, CLI support sounds useful but will need to handle that in a follow up.

> Do we want to continue supporting SQLite at all if we have a readily available in-memory local client with full support for the features we need to test? Shall we open an issue to remove SQLite once we're comfortable with DuckDB in this role?

Yeah, for simplicity, we should remove `SQLite` support in a follow up.

PR #112, opened 2022-06-13
Test CLI without mock MF engine

WilliamDee:
## Context
Currently, we test the CLI without any integration with the MF engine as we mock all the returned results and just check that the CLI returns properly and displays the proper results. This is an _okay_ approach, but we would like to test with the real thing rather than mocking an integral component to the CLI engine.

## Changes
- removed the mocked `MetricFlowEngine`, `SemanticModel`, `UserConfiguredModel`
- Test the CLI with the `simple_user_configured_model` model
- Added a test for `health_checks` command


tlento (approved):
🎉

PR #113, opened 2022-06-14
Add 'to_json' helper for 'ValidationIssue' classes

QMalcolm:
MetricFlow is used in other projects. Sometimes MetricFlow data might need to be returned over an API. For `ValidationIssues` to be returnable they need to be jsonify-able. We could force any downstream project to implement this themselves. Instead it's easier for everyone if it's just built into the `ValidaitonIssue` classes. I myself am running into this for an internal API.

If you have a sequence of `ValidaitonIssue` classes then you could do the follwoing
```python
issues # the sequence of validation issues
json_issues = [issue.to_json for issue in issues]
json.dumps({"issues": json_issues})
```

This is necessary because by default some of the variables of the ValidaitonIssue classes aren't JSON serializable, specifically `model_object_reference` for all issues and `error_date` of the `FutureError` type issue. We could have alternatively overwritten the base `__dict__` method. However, the base `__dict__` has its own uses which require `model_object_reference` and `error_date` to retain their base object types.


plypaul (commented):
Could the class be made into a Pydantic class? (i.e. `class ValidationIssue(BaseModel)`. If so, then `.json()` would come for free.

tlento (commented):
+1 on the Pydantic question. If we decide not to go that route this looks fine to me.

Pro on Pydantic: you get to_dict and to_json, among other niceties, and if you use our HashableBaseModel subclass you get a hash function as well (which is based on the to_json method)
Con: BaseModel initializers are very slow. If our inputs are always pre-validated we won't want to pay that cost, and instead we'd use the `.construct()` method, which is a little weird, and I don't know if it's type checkable.

tlento:
I just realized runtime cost only comes up when both of the following are true:

1. We run validation
2. We have validation issues

Given that, let's move to Pydantic - it automatically includes any new properties, so it's a little easier to maintain, and it's more consistent with how we're dealing with serializable objects elsewhere in the repo. If we're constructing tons of these things just for fun to the point where valid models are super slow due to this change we're probably doing something else wrong.

QMalcolm:
This PR is currently waiting on #119 so that we can better jsonify the context of ValidationIssues (what use to be `model_object_reference`)

QMalcolm:
#119 has been merged, and this PR has been rebased off of main

plypaul (approved):

PR #119, opened 2022-06-22
Improve Validation Issue Context

QMalcolm:
Historically we have used model_object_reference (an OrderedDict) to give context to ValidationIssues. However, this has some drawbacks. First the OrderedDict could be over specified (i.e. what if both a dimension and a measure name are specified?). Second it's not easily JSON serializable. Thus we've moved to something more concrete, ValidationContext and sub-classes thereof. This new class, and sub-classes, provides a hierarchy of contexts which all provide the same interface for generating human readable context strings. Additionally, ValidationContexts are of our specifically JSON serializable 'BaseModel' class. The one hickup is that in most cases we want to generate the ValidaitonContext from an available 'Metadata' object for 'DataSources', 'Metrics', and 'Materializations'. It'd be useful at some point to guarantee the 'Metadata' object on these objects to simply the code somewhat.

From a CLI prospective, `mf validate-configs` with errors used to look like:
```
mf validate-configs
✖ Errors found when building model from configs
• ERROR: Found measure with name MeasureReference(element_name='amount') in multiple data sources with names (['salesforce_oppportunities'])
• FATAL: In data source `salesforce_oppportunities`, can't use name `amount` for a measure when it was already used for a measure
```

and now they look like:
```
$ mf validate-configs                                               
✖ Errors found when building model from configs
• ERROR: with measure `amount` in data source `salesforce_oppportunities` in file `opportunities.yaml` on line #2 - Found measure with name MeasureReference(element_name='amount') in multiple data sources with names (['salesforce_oppportunities'])
• FATAL: with measure `amount` in data source `salesforce_oppportunities` in file `opportunities.yaml` on line #2 - In data source `salesforce_oppportunities`, can't use name `amount` for a measure when it was already used for a measure
```

### DIFF
--- a/metricflow/model/validations/metrics.py
+++ b/metricflow/model/validations/metrics.py
@@ -5,12 +5,11 @@ from metricflow.errors.errors import ParsingException
 from metricflow.model.objects.metric import Metric, MetricType, CumulativeMetricWindow
 from metricflow.model.objects.user_configured_model import UserConfiguredModel
 from metricflow.model.validations.validator_helpers import (
+    MetricContext,
     ModelValidationRule,
-    ValidationIssue,
     ValidationIssueType,
     ValidationError,
     ValidationFatal,
-    ModelObjectType,
     validate_safely,
 )
 
@@ -38,7 +37,9 @@ class MetricMeasuresRule(ModelValidationRule):
             if measure_in_metric.element_name not in valid_measure_names:
                 issues.append(
                     ValidationFatal(
-                        model_object_reference=ValidationIssue.make_object_reference(
+                        context=MetricContext(
+                            file_name=metric.metadata.file_slice.filename if metric.metadata else None,
+                            line_number=metric.metadata.file_slice.start_line_number if metric.metadata else None,
                             metric_name=metric.name,
                         ),
                         message=f"Invalid measure {measure_in_metric.element_name} in metric {metric.name}",
@@ -74,9 +75,10 @@ class CumulativeMetricRule(ModelValidationRule):
             if metric.type_params.window and metric.type_params.grain_to_date:
                 issues.append(
                     ValidationError(
-                        model_object_reference=ValidationIssue.make_object_reference(
-                            object_type=ModelObjectType.METRIC,
-                            object_name=metric.name,
+                        context=MetricContext(
+                            file_name=metric.metadata.file_slice.filename if metric.metadata else None,
+                            line_number=metric.metadata.file_slice.start_line_number if metric.metadata else None,
+                            metric_name=metric.name,
                         ),
                         message="Both window and grain_to_date set for cumulative metric. Please set one or the other",
                     )
@@ -88,7 +90,9 @@ class CumulativeMetricRule(ModelValidationRule):
                 except ParsingException:
                     issues.append(
                         ValidationError(
-                            model_object_reference=ValidationIssue.make_object_reference(
+                            context=MetricContext(
+                                file_name=metric.metadata.file_slice.filename if metric.metadata else None,
+                                line_number=metric.metadata.file_slice.start_line_number if metric.metadata else None,
                                 metric_name=metric.name,
                             ),
                             message=traceback.format_exc(),
--- a/metricflow/model/validations/unique_valid_name.py
+++ b/metricflow/model/validations/unique_valid_name.py
@@ -1,13 +1,21 @@
 import re
-from collections import OrderedDict
 from typing import Dict, Tuple, List, Optional
 
 from metricflow.model.objects.data_source import DataSource
 from metricflow.model.objects.user_configured_model import UserConfiguredModel
 from metricflow.model.validations.validator_helpers import (
+    DataSourceContext,
+    DimensionContext,
+    IdentifierContext,
+    MaterializationContext,
+    MeasureContext,
+    MetricContext,
     ModelValidationRule,
-    ValidationIssue,
+    ValidationContext,
+    ValidationError,
+    ValidationFatal,
     ValidationIssueLevel,
+    ValidationIssueType,
     validate_safely,
 )
 from metricflow.specs import ElementReference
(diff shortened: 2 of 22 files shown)


QMalcolm:
The branch `main` was fixed, I have now rebased this branch off of `main`.

metricflow/model/validations/metrics.py
```diff
@@ -38,8 +37,10 @@ def _validate_metric_measure_references(
             if measure_in_metric.element_name not in valid_measure_names:
                 issues.append(
                     ValidationFatal(
-                        model_object_reference=ValidationIssue.make_object_reference(
-                            metric_name=metric.name,
+                        context=MetricContext(
+                            file_name=metric.metadata.file_slice.filename if metric.metadata else None,
+                            line_number=metric.metadata.file_slice.start_line_number if metric.metadata else None,
+                            materialization_name=metric.name,
```
tlento, line 43: materialization_name?

metricflow/model/validations/validator_helpers.py
```diff
@@ -39,82 +40,117 @@ class ModelObjectType(Enum):
...
+    """The context class for vaidation issues involving measures"""
+
+    measure_name: str
+
+    def context_str(self) -> str:
+        """Human readable stringified representation of the context"""
+        return f"with measure `{self.measure_name}` in data source `{self.data_source_name}` {ValidationContext.context_str(self)}"
+
+
+ContextType = Union[
```
tlento, line 122: Do we need this? Everything inherits from ValidationContext, so we could just use that.

The other alternative, if you want to differentiate between a concrete ValidationContext and a general validation context object, is to define this as a Protocol instead of a custom union type, which feels very clumsy.

Note this is mainly an aesthetic comment. I don't think it makes much practical difference and it is easy enough to change.

metricflow/model/validations/metrics.py
```diff
@@ -74,9 +75,10 @@ def _validate_cumulative_sum_metric_params(metric: Metric) -> List[ValidationIss
             if metric.type_params.window and metric.type_params.grain_to_date:
                 issues.append(
                     ValidationError(
-                        model_object_reference=ValidationIssue.make_object_reference(
-                            object_type=ModelObjectType.METRIC,
-                            object_name=metric.name,
+                        context=MetricContext(
+                            file_name=metric.metadata.file_slice.filename if metric.metadata else None,
+                            line_number=metric.metadata.file_slice.start_line_number if metric.metadata else None,
+                            materialization_name=metric.name,
```
tlento, line 81: materialization_name?

metricflow/model/validations/metrics.py
```diff
@@ -88,8 +90,10 @@ def _validate_cumulative_sum_metric_params(metric: Metric) -> List[ValidationIss
                 except ParsingException:
                     issues.append(
                         ValidationError(
-                            model_object_reference=ValidationIssue.make_object_reference(
-                                metric_name=metric.name,
+                            context=MetricContext(
+                                file_name=metric.metadata.file_slice.filename if metric.metadata else None,
+                                line_number=metric.metadata.file_slice.start_line_number if metric.metadata else None,
+                                materialization_name=metric.name,
```
tlento, line 96: materialization_name?

metricflow/test/model/validations/helpers.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def materialization_with_guaranteed_meta(
+    name: str,
+    metrics: List[str],
+    dimensions: List[str],
+    description: str = "adhoc materialization",
+    metadata: Metadata = default_meta(),
+    destinations: List[MaterializationDestination] = [],
+    destination_table: Optional[SqlTable] = None,
```
tlento, line 35: I don't love these shadow factory functions but I have no better ideas to offer unless you don't necessarily care about the metadata values themselves.

metricflow/model/validations/unique_valid_name.py
```diff
@@ -57,64 +56,125 @@ def check_valid_name(name: str, data_source_name: Optional[str] = None) -> List[
 
     @staticmethod
     @validate_safely(whats_being_done="checking data source sub element names are unique")
-    def _validate_data_source_elements(data_source: DataSource) -> List[ValidationIssue]:
-        issues = []
-        name_and_type_tuples: List[Tuple[ElementReference, str]] = []
+    def _validate_data_source_elements(data_source: DataSource) -> List[ValidationIssueType]:
+        issues: List[ValidationIssueType] = []
+        name_and_type_tuples: List[Tuple[ElementReference, str, ContextType]] = []
```
tlento, line 61: I feel like I'm on book 4 of the increasingly inappropriately named Hitchiker's Guide to the Galaxy trilogy....

It seems `name_and_type_tuples` would benefit from a new name.

metricflow/model/validations/validator_helpers.py
```diff
@@ -39,82 +40,121 @@ class ModelObjectType(Enum):
     METRIC = "metric"
 
 
+class ValidationContext(BaseModel):
+    """The base context class for vaidation issues"""
```
tlento, line 44: nit:
```suggestion
    """The base context class for validation issues"""
```

metricflow/model/validations/validator_helpers.py
```diff
@@ -39,82 +40,121 @@ class ModelObjectType(Enum):
...
+    def context_str(self) -> str:
+        """Human readable stringified representation of the context"""
+        return f"with materialization `{self.materialization_name}` {ValidationContext.context_str(self)}"
+
+
+class MetricContext(ValidationContext):
+    """The context class for vaidation issues involving metrics"""
+
+    metric_name: str
+
```
tlento, line 76: So above you are passing "materialization_name" and that doesn't blow up at either typecheck time or in the runtime because by default Pydantic will simply drop that key/value pair on the floor.

metricflow/model/validations/validator_helpers.py
```diff
@@ -39,82 +40,121 @@ class ModelObjectType(Enum):
     METRIC = "metric"
 
 
+class ValidationContext(BaseModel):
+    """The base context class for vaidation issues"""
+
```
tlento, line 45: We might want to add a Config here that sets the `extra` property to `Extra.forbid` or equivalent. See https://pydantic-docs.helpmanual.io/usage/model_config/#options

tlento (commented):
The MetricContext instances are not all properly initialized.

metricflow/model/validations/metrics.py
QMalcolm: Woops thats a copy paste error on my part. Good catch

metricflow/model/validations/metrics.py
QMalcolm: Yep

metricflow/model/validations/metrics.py
QMalcolm: yarp

metricflow/test/model/validations/helpers.py
QMalcolm: I don't necessarily care about the metadata values themselves, so I would love any ideas 😄

tlento (approved):
Sweet! I don't think you need to worry about the commit tree fixup unless it really bothers you since the new commits have good messages. We're humans, we don't expect a perfect commit tree. :)

PR #128, opened 2022-06-27
Refactor Validation Issue Structuring

QMalcolm:
This PR started as the second half to #133, in that we wanted to be able to deserialize JSON-ified ValidationIssues into ValidationIssue objeccts. In doing this work we found blocking problems with the structure of `ValidationIssues` and `ValidationContext` which stopped the magical parsing of Pydantic `BaseModel` from working. This in turn lead to a refactor to the structure of`ValidationIssues` and `ValidationContexts` so that both serialization and de-serialization of `ValidationIssues` is simple.

# Highlights
• The different `ValidationContexts` now contain a `FileContext` instead of inheriting from `FileContext`.
• `ValidationIssue` is now an `ABC` which the different issue types implement, making them distinct.
• There is a new `ModelValidationResults` class which should be the primary source of serialization and deserialization for validation issues
• The different Validators `ModelValiator`, `DataWarehouseModelValidator`, `ConfigLinter` now return `ModelValidationResults` instead of a list/tuple of ValidationIssues

# Good to knows

## Serialization and de-serialization
Serialization and de-serialization should primarily happen with `ModelValidationResults` objects.
```Python
# serialization
serialized = model_validation_results.json()

# de-serialization
deserialized_mvr = ModelValidationResults.parse_raw(serialized)
```

## ModelValidationResults creation and helpers
```Python
# Create a ModelValidationResults instance from a sequence of issues
new_mvr = ModelValidationResults.from_issue_sequence(list_of_issues)

# Get a list of issues from a ModelValidationResults instance
validation_issues_list = new_mvr.all_issues

# Check if a ModelValidationResults has blocking issues
new_mvr.has_blocking_issues # True if object has errors or fatals, otherwise false

# Merge multiple ModelValidationResults objects together
big_mvr = ModelValidationResults.merge(mvr1, mvr2, mvr3, ...)
```

### DIFF
--- a/metricflow/model/parsing/config_linter.py
+++ b/metricflow/model/parsing/config_linter.py
@@ -6,6 +6,7 @@ from metricflow.model.parsing.validation import METRIC_TYPE, DATA_SOURCE_TYPE, M
 
 from metricflow.model.validations.validator_helpers import (
     FileContext,
+    ModelValidationResults,
     ValidationError,
     ValidationIssue,
     ValidationWarning,
@@ -102,7 +103,7 @@ class ConfigLinter:  # noqa: D
 
         return issues
 
-    def lint_dir(self, dir_path: str) -> List[ValidationIssue]:  # noqa: D
+    def lint_dir(self, dir_path: str) -> ModelValidationResults:  # noqa: D
         issues: List[ValidationIssue] = []
         for root, _dirs, files in os.walk(dir_path):
             for file in files:
@@ -111,4 +112,4 @@ class ConfigLinter:  # noqa: D
                 file_path = os.path.join(root, file)
                 issues += self.lint_file(file_path, file)
 
-        return issues
+        return ModelValidationResults.from_issues_sequence(issues)
--- a/metricflow/model/parsing/config_linter.py
+++ b/metricflow/model/parsing/config_linter.py
@@ -4,7 +4,12 @@ import yaml
 from yamllint import config, linter, rules
 from metricflow.model.parsing.validation import METRIC_TYPE, DATA_SOURCE_TYPE, MATERIALIZATION_TYPE
 
-from metricflow.model.validations.validator_helpers import ValidationContext, ValidationIssue, ValidationIssueLevel
+from metricflow.model.validations.validator_helpers import (
+    FileContext,
+    ValidationError,
+    ValidationIssue,
+    ValidationWarning,
+)
 
 WARNING = "warning"
 ERROR = "error"
@@ -80,14 +85,21 @@ class ConfigLinter:  # noqa: D
                 if problem.rule == rules.key_duplicates.ID:
                     self.add_additional_key_duplicates_info(problem=problem)
 
-                level = ValidationIssueLevel.ERROR if problem.level == ERROR else ValidationIssueLevel.WARNING
-                issues.append(
-                    ValidationIssue(
-                        level=level,
-                        context=ValidationContext(file_name=file_name, line_number=problem.line),
-                        message=problem.desc,  # type: ignore[misc]
+                if problem.level == ERROR:
+                    issues.append(
+                        ValidationError(
+                            context=FileContext(file_name=file_name, line_number=problem.line),
+                            message=problem.desc,  # type: ignore[misc]
+                        )
                     )
-                )
+                else:
+                    issues.append(
+                        ValidationWarning(
+                            context=FileContext(file_name=file_name, line_number=problem.line),
+                            message=problem.desc,  # type: ignore[misc]
+                        )
+                    )
+
         return issues
 
     def lint_dir(self, dir_path: str) -> List[ValidationIssue]:  # noqa: D
--- a/metricflow/model/validations/validator_helpers.py
+++ b/metricflow/model/validations/validator_helpers.py
@@ -42,7 +42,7 @@ class ModelObjectType(Enum):
     METRIC = "metric"
 
 
-class ValidationContext(BaseModel):
+class FileContext(BaseModel):
     """The base context class for validation issues"""
 
     file_name: Optional[str]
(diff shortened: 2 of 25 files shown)


QMalcolm:
As a side note this file should maybe get split into a few separate files soon given the intensity of the classes which have emerged. Additionally, some refactoring to make the different parts of `validator_helpers` DRY-er is likely in our future 🙃

plypaul (commented):
Pydantic objects have the `.parse_raw()` method - does that work for this case?

QMalcolm:
> Pydantic objects have the `.parse_raw()` method - does that work for this case?

Not sure, but I'll find out!

metricflow/model/validations/validator_helpers.py
```diff
@@ -62,6 +65,59 @@ def context_str(self) -> str:
...
+                materialization_name=context_dict["materialization_name"],
+            )
+        elif "metric_name" in keys:
+            return MetricContext(
+                file_name=file_name,
+                line_number=line_number,
+                metric_name=context_dict["metric_name"],
+            )
+        else:
+            return ValidationContext(file_name=file_name, line_number=line_number)
```
tlento, line 119: Why is all of this necessary? Pydantic provides parse_obj for this purpose. If the model is serialized using the Pydantic `.json()` family of methods, it should be deserializable using `.parse_raw()` or `.parse_obj()`

The primary issue here seems to be the tagged union model, where you don't actually know the subclass type without inspecting the JSON value. There are two options that I think would be a bit more palatable:

1. Define a Pydantic model for the encapsulating object that includes a properly typed entry for the issue, and call parse_obj from that. It's entirely possible that my earlier suggestion to move away from a Union type alias was, for this reason, absolutely the wrong call, because you may find you need a Union type for Pydantic to work properly here.
2. Streamline this logic to call `parse_obj` once you've done your subclass detection, and then define this as a custom "validator" and override `__get_validators__` in the ValidationIssue class. This will allow callers to use the standard Pydantic APIs and still get correct results.

See https://pydantic-docs.helpmanual.io/usage/types/#classes-with-__get_validators__ for details. Also I have an example in #129

tlento (commented):
I don't think we want to be hand-rolling custom parsers all over the place, as that defeats the purpose of defining these as Pydantic types.

I don't know if my earlier not-Pydantic-parsing-aware suggestion to move away from the Union type alias is what forced you into this position. If so, maybe we go back to that, at least at the point where things are getting serialized and deserialized.

If you need to do custom parsing regardless, I have a second suggestion inline for how to do that in a manner that allows Pydantic to do its magic.

metricflow/model/validations/validator_helpers.py
tlento: Oooh, this looks potentially useful:

https://pydantic-docs.helpmanual.io/usage/types/#nested-discriminated-unions

QMalcolm: It might be the Seattle heat + time of day, but I'm struggling to put together. I think I'm most of the way there, but I'm not sure how `__get_validators__` actually gets called in the end. Once I have `__get_validators__` that returns the method which calls the correct `<class>.parse_obj`, how do I then go from a json string to the parsed object class? Calling just `ValidationIssue(json_string)` doesn't do it, and `ValidationIssue.from_raw(json_string)` seems to skip `__get_validators__` which makes sense. But then to what do I hand the json_string to be parsed? 😖

QMalcolm:
Converted to a draft as for the time being this is a work in progress

metricflow/model/validations/validator_helpers.py
```diff
@@ -229,6 +229,12 @@ class ModelValidationResults(BaseModel):
     errors: List[ValidationError] = []
     fatals: List[ValidationFatal] = []
 
+    class Config:
```
plypaul, line 232: There's a `FrozenBaseModel` that you could inherit from instead.

metricflow/test/model/validations/test_validator_helpers.py
```diff
@@ -20,25 +21,76 @@
...
+def file_context_dict() -> Dict[str, Union[str, int]]:  # noqa: D
     return {
         "file_name": "foo.py",
         "line_number": 1337,
     }
 
 
 @pytest.fixture
-def base_issue_dict(base_context_dict: ValidationContextJSON) -> ValidationIssueJSON:  # noqa: D
+def data_source_context_dict(file_context_dict: Dict[str, Union[str, int]]) -> Dict[str, Any]:  # noqa: D
```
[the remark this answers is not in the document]
QMalcolm: Can you clarify which tests you are talking about? Is this in reference to all the tests in this file in general or perhaps tests that use this fixture? The comment seems to be linked to a fixture for tests instead of a test itself.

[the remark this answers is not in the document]
plypaul: Sorry, I was referring to use tests that use this fixture.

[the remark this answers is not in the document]
tlento: It might be worth some assertions for configuration bits that we want to make sure we get right (like round trip serialization/deserialization on tagged union types, if we end up using those), but otherwise I agree with Paul - we don't need to say "given a valid JSON input we will deserialize it correctly".

tlento (commented):
Oooh, nice improvements! 

I don't have much to add to Paul's comments, saw his review about halfway through mine. I particularly agree about the tests and not switching on isinstance checks - your idea of using the level property instead is much better.

metricflow/model/validations/validator_helpers.py
```diff
@@ -160,19 +180,92 @@ def as_readable_str(self, with_level: bool = True) -> str:
...
+        """Constructs a ModelValidationResults class from a list of ValidationIssues"""
+
+        warnings: List[ValidationWarning] = []
+        future_errors: List[ValidationFutureError] = []
+        errors: List[ValidationError] = []
+        fatals: List[ValidationFatal] = []
+
+        for issue in issues:
+            if issue.level == ValidationIssueLevel.WARNING:
+                warnings.append(issue)
```
plypaul, line 226: I'm surprised that there's no type issue here. `issue` is of type `ValidationIssueType` but warnings is a `List[ValidationWarning]`.

metricflow/model/parsing/config_linter.py
```diff
@@ -80,17 +86,24 @@ def lint_file(self, file_path: str, file_name: str) -> List[ValidationIssue]:  #
...
-                issues.append(
-                    ValidationIssue(
-                        level=level,
-                        context=ValidationContext(file_name=file_name, line_number=problem.line),
-                        message=problem.desc,  # type: ignore[misc]
+                if problem.level == ERROR:
+                    issues.append(
+                        ValidationError(
+                            context=FileContext(file_name=file_name, line_number=problem.line),
+                            message=problem.desc,  # type: ignore[misc]
```
plypaul, line 93: Context on why `# type: ignore[misc]`?

metricflow/model/validations/validator_helpers.py
QMalcolm: I think it's because issue `ValidationIssueType` is a `Union[ValidationWarning, ValidationFutureError, ValidationError, ValidationFatal` and the `level` of each of those is hardcoded. I.e. a `ValidationWarning` will always and only have a `level` value of `ValidationIssueLevel.WARNING`. Thus mypy can do it's correct magic.

metricflow/model/parsing/config_linter.py
QMalcolm: The [YamlLint package isn't currently typed](https://github.com/adrienverge/yamllint/blob/40cab7f999b7d8f4f143f22f6932126274238e0d/yamllint/linter.py#L33) 🙃 I've debated about spending some time contributing to the package to make it typed, but time is a lacking. So `problem.desc` has a type of `Any`. I could either do an `assert isinstance(problem.desc, str)`, try to stringify it `str(probelm.desc)`, or just ignore it because we know it's always a string. In this instance because `problem.desc` is actually always a string, I decided to just ignore the type problem.

metricflow/test/model/validations/test_validator_helpers.py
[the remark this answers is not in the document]
QMalcolm: Sounds good. I think I removed all the relevant tests. Also I believe the [test_jsonifying_and_reloading_model_validation_results_is_equal](https://github.com/transform-data/metricflow/blob/089d9c8a8afbc48666c40d228d532780933d2ee8/metricflow/test/model/validations/test_validator_helpers.py#L123) covers the round trip serialization/deserialization on tagged union types.

metricflow/model/validations/validator_helpers.py
```diff
@@ -160,19 +180,92 @@ def as_readable_str(self, with_level: bool = True) -> str:
...
+            elif issue.level == ValidationIssueLevel.FUTURE_ERROR:
+                future_errors.append(issue)
+            elif issue.level == ValidationIssueLevel.ERROR:
+                errors.append(issue)
+            elif issue.level == ValidationIssueLevel.FATAL:
```
plypaul, line 231: Can you add the else condition using `assert_values_exhausted`?

plypaul (approved):
Nice update! LGTM but left a comment.

metricflow/model/validations/validator_helpers.py
QMalcolm: Can do

QMalcolm:
Gonna fixup some commits and will then merge

PR #129, opened 2022-06-27
Move object parsing to Pydantic's BaseModel.parse_obj

tlento:
The original model parsing logic had a custom parsing method for
extracting the values provided by the user via their YAML configs.
This was in place because certain object value types, such as the
WhereClauseConstraint, expect user-provided values to be provided
as strings rather than object (key/value) types.

This parsing logic was broken in several subtle ways - the recursive
callback was not conforming to the type signature of the method,
for one thing, which causes quite a bit of confusion for readers.
More importantly, a hack in place to work around nested lists
of strings - which was needed for the materialization config spec -
would throw exceptions if anyone ever created a value with a nested
list of object types.

Rather than attempt to disentangle the parsing issues, this PR
moves us onto Pydantic's already-fully-functional parse_obj method by
replacing the ParseableField interface with a mixin allowing for an
input parser override for the Pydantic model objects that require it.
This is done by overriding the base "validator" method on that model
object, which Pydantic uses to both parse and validate the input
used to initialize the model object. For more details, please refer to
the Pydantic docs:
https://pydantic-docs.helpmanual.io/usage/types/#classes-with-__get_validators__

Since Pydantic expects this method to accept `Any` and return the class
type, this naturally allows us to extend our custom parsing to cases
where we need to add handling for string or object types, which is a
feature we will require in the future. We use the abstract class mixin
as both a convenience for providing the common Pydantic override and
an interface tag for allowing us to keep track of all special parsing
cases, which will make it easier to manage deprecation of backwards
compatibility shims for model parsing.


tlento:
Superseded by #151

PR #135, opened 2022-06-30
Enable struct parsing for metric measures

tlento:
Our Metric model object takes in MetricTypeParams that allows for various
input measure name references. These are all typed as strings or
lists of strings, since the current MetricFlow framework only requires
a pointer to the original measure model object.

Upcoming changes will add constraint expressions to the measure
pointers stored inside of the Metric model object, and as such we
will need to attach additional properties to each of the measures
referenced in the MetricTypeParams.

This PR converts the internal types of these measure representations
from strings to a new model object, the MetricInputMeasure, and updates
the parser to enable users to pass in either the existing string name
representation of the measure or a structured representation. This
paves the way for adding new properties, such as measure-specific
constraints, in metric definitions.


tlento:
Superseded by #152

PR #137, opened 2022-07-01
Add basic inference rule functionality

serramatutu:
# What does this PR do?

It adds the base classes required for implementing inference rules with minimal effort. Those classes make it easy for custom rules to return signals about columns, associated with confidence scores and reasons. For example, a custom rule could return a signal as follows:
```python
InferenceSignal(
    column=SqlColumn.from_names(db="db", schema="schema", table="table", column="id")
    type=InferenceSignalType.PRIMARY_IDENTIFIER,
    confidence=InferenceSignalConfidence.FOR_SURE,
    reason="The column is named `id`"
)
```

A heuristic that reasons about a collection of signals could then use confidence values to make conclusions about a column, giving well-described reasons to the end user. It could also prevent false positives by detecting conflicting signals (see `InferenceSignal.conflict`).

Apart from that, it also makes it very easy for an `InferenceRunner` to determine which contexts are required by a rule to run without failure by using introspection from type annotations in the `process` method.

Example:
```python
class CtxType1(InferenceContext):
    pass

class CtxType2(InferenceContext):
    pass

class TestRule1(InferenceRule):
    def process(self, ctx1: CtxType1):
        pass

class TestRule2(InferenceRule):
    def process(self, ctx2: CtxType2):
        pass

class TestRule1And2(InferenceRule):
    def process(self, ctx1: CtxType1, ctx2: CtxType2):
        pass

print(TestRule1().required_contexts)  # (CtxType1,)
print(TestRule2().required_contexts)  # (CtxType2,)
print(TestRule1And2().required_contexts)  # (CtxType1, CtxType2)
```

Finally, added a basic rule that produces signals based on a Regex matches to column names.

### DIFF
--- a/metricflow/inference/rule/base.py
+++ b/metricflow/inference/rule/base.py
@@ -1,13 +1,14 @@
 from __future__ import annotations
+from abc import ABC, abstractmethod
 
 from dataclasses import dataclass
 from enum import Enum
 import logging
-import inspect
-from typing import List, Optional, Tuple, Type
+from typing import List, Tuple, Type
 
 from metricflow.dataflow.sql_column import SqlColumn
 from metricflow.inference.context.base import InferenceContext
+from metricflow.inference.context.data_warehouse import DataWarehouseInferenceContext
 
 
 logger = logging.getLogger(__name__)
@@ -79,10 +80,10 @@ class InferenceSignal:
     confidence: InferenceSignalConfidence
 
 
-RequiredContextsType = Tuple[Optional[Type[InferenceContext]], ...]
+RequiredContextsType = Tuple[Type[InferenceContext], ...]
 
 
-class InferenceRule:
+class InferenceRule(ABC):
     """Implements some sort of heuristic that produces signals about columns.
 
     An inference rule produces zero or more `InferenceSignal` instances about whatever
@@ -91,49 +92,12 @@ class InferenceRule:
     Concrete implementations should aim to be short and modularized. It is preferred to
     compose multiple small rules that each produce a signal type than to make one large
     rule with complex logic to produce a bunch of signals.
+
+    It currently only accepts DataWarehouseInferenceContext as input, but this will probably
+    be generalized later.
     """
 
-    _required_contexts: RequiredContextsType
-
-    def __init_subclass__(cls) -> None:
-        """Initializes `cls._required_contexts` from `cls.process` type annotations."""
-        parameters = inspect.signature(cls.process).parameters
-        context_types: List[Optional[Type[InferenceContext]]] = []
-        for param in list(parameters.values())[1:]:  # exclude self
-            if param.annotation == inspect.Parameter.empty:
-                logger.warning(
-                    f"`{cls.__name__}.process`'s parameters must all be type annotated. "
-                    "Unanottated types will be provided None contexts, which might trigger further runtime errors."
-                )
-                context_types.append(None)
-                continue
-
-            if isinstance(param.annotation, str):
-                annotation_str = param.annotation
-            else:
-                annotation_str = param.annotation.__name__
-
-            annotation_type = InferenceContext.known_subclasses.get(annotation_str, None)
-            if annotation_type is None:
-                logger.warning(
-                    "`InferenceRule.process` arguments must be subclasses of `InferenceContext`. "
-                    "Some calls may have None arguments, which may result in runtime errors."
-                )
-            context_types.append(annotation_type)
-
-        cls._required_contexts = tuple(context_types)
-
-        super().__init_subclass__()
-
-    @property
-    def required_contexts(self) -> RequiredContextsType:
-        """The context types this rule requires to run successfully."""
-        return self._required_contexts
-
-    # NOTE: this is not an @abstractmethod since we don't expect children to implement the same signature
-    def process(self) -> List[InferenceSignal]:
-        """The actual rule implementation that returns a list of signals based on the input contexts.
-
-        All input context instances will be provided in the same order as specified in `required_contexts`.
-        """
+    @abstractmethod
+    def process(self, warehouse: DataWarehouseInferenceContext) -> List[InferenceSignal]:
+        """The actual rule implementation that returns a list of signals based on the input contexts."""
         raise NotImplementedError
--- a/metricflow/inference/rule/base.py
+++ b/metricflow/inference/rule/base.py
@@ -4,7 +4,7 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from enum import Enum
 import logging
-from typing import List, Tuple, Type
+from typing import List, Optional, Tuple, Type
 
 from metricflow.dataflow.sql_column import SqlColumn
 from metricflow.inference.context.base import InferenceContext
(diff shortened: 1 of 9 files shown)


metricflow/inference/rule/base.py
```diff
@@ -0,0 +1,139 @@
...
+
+        super().__init_subclass__()
+
+    @property
+    def required_contexts(self) -> RequiredContextsType:
+        """The context types this rule requires to run successfully."""
+        return self._required_contexts
+
+    # NOTE: this is not an @abstractmethod since we don't expect children to implement the same signature
+    def process(self) -> List[InferenceSignal]:
```
serramatutu, line 134: I like making the superclass infer the `required_contexts` from class annotations because it makes subclassing very easy and idiomatic, as in the PR description. However, this violates the principle that subclasses must keep the subclass interface. 

I wanna know your opinions on this. Maybe this is just too much magic? I kinda like it :)

serramatutu: Btw, MyPy complains about this, so you need to add `# type: ignore` to each subclass implementation (which I hate). I haven't found a solution for this, though...

metricflow/inference/rule/rules.py
```diff
@@ -0,0 +1,46 @@
...
+        pattern: regex Pattern to match against columns' full names, i.e `<db>.<schema>.<table>.<column>`
+        signal_type: the `InferenceSignalType` to produce whenever the pattern is matched
+        confidence: the `InferenceSignalConfidence` to produce whenever the pattern is matched
+        """
+        self.pattern = pattern
+        self.signal_type = signal_type
+        self.confidence = confidence
+
+    def process(self, warehouse: DataWarehouseInferenceContext) -> List[InferenceSignal]:  # type: ignore
+        """Try to match all columns' full names with the matching function or pattern.
```
serramatutu, line 32: Just realized I need to update docs since this does not `match` the full name, but `searches` it.

metricflow/test/inference/rule/test_base.py
```diff
@@ -0,0 +1,96 @@
...
+
+    # make sure we're emitting warnings when users try to use contexts that don't
+    # inherit from InferenceContext or if the process method is unannotated
+    with patch("metricflow.inference.rule.base.logger.warning") as warning_mock:
+
+        class TestError(InferenceRule):
+            def process(self, ctx: CtxType1, ctx2: int, ctx3):  # type: ignore
+                pass
+
+        print(TestError().required_contexts)
```
serramatutu, line 94: oops

metricflow/test/inference/rule/test_rules.py
```diff
@@ -0,0 +1,20 @@
+import re
+from metricflow.dataflow.sql_column import SqlColumn
+from metricflow.inference.context.data_warehouse import DataWarehouseInferenceContext
+from metricflow.inference.rule.base import InferenceSignalConfidence, InferenceSignalType
+from metricflow.inference.rule.rules import ColumnRegexMatcherRule
+
+
+def test_column_regex_matcher(warehouse_ctx: DataWarehouseInferenceContext):  # noqa: D
+    rule = ColumnRegexMatcherRule(
+        pattern=re.compile(r"test_column$"),  # have to pass in
```
serramatutu, line 10: will also remove this comment

metricflow/inference/rule/base.py
```diff
@@ -0,0 +1,139 @@
...
+
+    column: the target column for this signal
+    type: the type of the signal
+    reason: a human-readable string that explains why this signal was produced. It may
+        eventually reach the user's eyeballs.
+    confidence: the confidence that this signal is correct.
+    """
+
+    column: SqlColumn
+    type: InferenceSignalType
```
tlento, line 77: In thinking about this a bit further I think you want a hierarchy around InferenceSignalType that makes sense as a return.

Fundamentally, you're trying to figure out how to populate field sets in a DataSource. That means you care about, basically, three things: Dimensions, Identifiers, and Measures

So you can have essentially three concrete classes that are something like `InferredColumnType` and each of those can then encapsulate their own hierarchy of inferred attributes, and manage consistency checking internally. They'd implement a shared `InferredColumnType` interface, so you can operate on them interchangeably.

If you ever have something that's shared across those class divisions - like maybe there's a signal that something maybe shouldn't be included in the dimension/identifier/measure set but if the user does want it the inference should remain - you can define another cross-cutting class that essentially extends the functionality across all `InferredColumnType` base classes.

That gives you a much cleaner, more extensible model, full typechecking support, and will naturally dictate the boundaries of what's acceptable in terms of return types for various inference rules. The internals of which class of thing to return might be a little more difficult on the implementer, but everything else will likely be cleaner, and you can smooth out those rough edges as needed.

tlento (commented):
So I think I see what you're trying to do here, and it seems good to me. If I'm right, you want:

1. A generic framework that can rely on some contextual data retrieved from somewhere to produce an inference for evaluation by whatever will write out the detected model
2. Some way of allowing people to specify which sorts of contextual data their inference generator requires
3. The ability to share logic across rules of similar type

The main issue with the core implementation laid out here is it does everything in a mostly untyped manner - the core API is effectively untyped, requirements are identified based on runtime lookups of type inspection, and logic is shared by coercing inputs to strings and running regular expressions (in the one concrete example you've provided, anyway).

I've put some design suggestions inline for you before I pack it in for the long weekend. They're not the only way to do this, necessarily, but they're things we do elsewhere in the framework and so they'll look familiar to developers working here, and you can probably poke around and see some examples of sufficiently similar things to get you started.

I'll be happy to answer any questions next week, although given how subtle and difficult some of this stuff can get it might be more efficient for us to jump on a call. I'll leave that up to you to decide, feel free to grab time with me at your convenience.

metricflow/inference/rule/defaults.py
```diff
@@ -0,0 +1,78 @@
...
+    # This is the default regex pattern that is used to determine if columns are identifiers.
+    # It simply matches column names ending with "id", case insensitive.
+    #
+    # We searched for words ending with "id" just to assess the chance of this resulting in a
+    # false positive. Our guess is most of those words would rarely, if ever, be used as column names.
+    # Therefore, not adding a mandatory "_" before "id" would benefit the product by matching names
+    # like "userid", despite the rare "squid", "mermaid" or "android" matches.
+    #
+    # See: https://www.thefreedictionary.com/words-that-end-in-id
+    ANY_IDENTIFIER_REGEX_PATTERN = re.compile(r"id$", flags=re.IGNORECASE)
```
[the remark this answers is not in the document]
serramatutu: I know it is broad, but my rationale is that it's better to match stuff like `userid` and `customerid` at the expense of having a couple of (hopefully rare) false positives like `is_livid` and `grid` than to not match them and have the user type them in everytime. 

It will be very difficult to get a solution that gets everything right everytime, but IMO that's a tolerable compromise (given we make sure the user reviews everything to catch those misses).

tlento (commented):
Ok, I'm back here to cover the changes in prep for #163. This seems pretty reasonable for now, given we're in prototype mode, so I will come back and approve once I see how it works in practice over there. :)

tlento (approved):
This has all really come together nicely. Great stuff! Thanks again for such a well-crafted set of PRs, and for your patience on the back and forth. Love the commit structure and messaging!

PR #138, opened 2022-07-01
Run the full test suite for engine-specific tests

tlento:
Engine-specific tests currently only run for things under
metricflow/test/integration, but that does not cover all of the
engine-specific test cases. We need to run them on any test that
might generate engine-specific plan files, and also any test that
targets the SqlClient classes themselves.


plypaul (approved):

tlento:
Weird, the CI deployment seems to be running the old file...... I guess we just merge this and hope for the best?

PR #139, opened 2022-07-01
Remove SQLite Support

plypaul:
This PR removes SQLite support as local tests are now handled with DuckDb.


tlento (approved):
This is so much better than my approach to fixing that test.

You've got some rebase issues to resolve here before merging, though.

PR #140, opened 2022-07-01
Fix CLI test for SQLite

tlento:
SQLite booleans render as 1 and 0, not True and False, so one of our
CLI test cases was failing for SQLite and only SQLite.


tlento:
I prefer #139 to this

PR #145, opened 2022-07-07
Fix where clause parsing for empty functions

tlento:
The parser we use for the WHERE clause returns a JSON-like structure
with dict types representing key/value pairs for things like literals
and function definitions. For a function taking arguments, the key
would be the name of the function and the value would be a list of
arguments, which themselves could be strings, lists, or dicts.

For any function taking no arguments, such as CURRENT_DATE(), or one
taking keywords, such as CAST(x AS DOUBLE), the parsed output would
represent the keywords (or empty set of arguments) as an empty dict.

Since we recursively call a helper that asserts that the dict contains
exactly one key, our parsing logic throws an error on such input.

This commit simply moves that assertion where it belongs - into the
first call into the recursive processing function, the one for the
top-level where clause, because at that level we expect exactly one
key: the index into the comparison, or else the function name.
The test suite for this function has also been expanded.

### DIFF
--- a/metricflow/model/objects/constraints/where.py
+++ b/metricflow/model/objects/constraints/where.py
@@ -63,6 +63,8 @@ class WhereClauseConstraint(HashableBaseModel, ParseableField):
 
         where = parsed["where"]
         if isinstance(where, dict):
+            if not len(where.keys()) == 1:
+                raise ConstraintParseException(f"expected parsed constraint to contain exactly one key; got {where}")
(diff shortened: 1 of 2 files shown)


QMalcolm (approved):
It's so nice that all we had to do was move the check. Thank you for figuring this out so quickly 🙏 Looks good to me

belindabennett (approved):
tests lgtm but i'm not as familiar with the other parts!

PR #147, opened 2022-07-08
Refactor query generation for data warehouse tasks to happen in the query runner

QMalcolm:
We were running into an issue where the generation of queries for metrics were taking multiple minutes. This was problematic because when used in a server setting, this meant that the timeout wasn't being respected. This PR moves the generation of queries for data warehouse validations to be delayed until actually running the task.

### DIFF
--- a/metricflow/model/data_warehouse_model_validator.py
+++ b/metricflow/model/data_warehouse_model_validator.py
@@ -1,15 +1,16 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from time import perf_counter
-from typing import List, Optional
-
+from functools import partial
 from math import floor
+from time import perf_counter
+from typing import Callable, List, Optional, Tuple
 
 from metricflow.dataset.dataset import DataSet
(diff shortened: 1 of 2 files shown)


tlento (approved):
These callback mechanisms are always a little gnarly to deal with due to the added layer of indirection, but given the issue you describe this seems reasonable for now. 

I wonder if async dispatch would be help, and if it does, whether or not it would be any easier to reason about. That's a much (much) bigger change, of course, so iI'm not suggesting we try to make it here.

QMalcolm:
> These callback mechanisms are always a little gnarly to deal with due to the added layer of indirection, but given the issue you describe this seems reasonable for now.
> 
> I wonder if async dispatch would be help, and if it does, whether or not it would be any easier to reason about. That's a much (much) bigger change, of course, so I'm not suggesting we try to make it here.

I agree. This isn't my favorite outcome, but it is the most straightforward small solution. I definitely want to revisit this in the future to make it easier to reason about, but yea, we should wait until it makes sense to take on that work.

PR #149, opened 2022-07-11
Add tests for metric config parsing

tlento:
## Add tests for metric config parsing

The core MetricFlow model parser converts user-defined YAML files
into MetricFlow model objects via a combination of standard,
jsonschema-backed YAML parsing and Pydantic BaseModel object
parsing. This is managed by a set of custom parsers.

Although all of these paths are probably tested indirectly via
test model parsing in various other test cases, the concrete attribute
parsing is not directly tested anywhere. This makes the expected output
of the core parsing logic a bit harder to reason through.

This is the first of three commits to add direct tests for the core parser
logic, which will make updates and changes safer and easier to reason about.
This commit adds test coverage for metric parsing. All tests have YAML
inline instead of saved in a file, as the file and directory reading logic
is outside the scope of the metric parser itself and the metric config spec
is simple enough to comfortably fit inside of each test function.

## Add tests for materialization config parsing

Second of three commits adding test coverage on core parsing logic.
This one covers all of the materialization object types, especially
the one object implementing custom string input parsing.

Also of note for materializations is the nested list construct for
the `rollups` attribute, which presents something of an edge case.

Please see the recent commit titled "Add tests for metric config parsing"
for more background on the reasoning behind this set of changes.

## Add tests for data source config parsing

Third of three, this adds test coverage for bare data source parsing
with various options enabled. There are no error states to check at
the moment because data source properties are all fairly simple.
Note this includes a base test for the metadata parsing for the measure
objects as well, since those are defined in the datasource and have
metadata propagated into them as part of that parsing process.

See commit titled "Add tests for metric config parsing" for more context
on this change.

## Add test for Metadata parsing from simple model files

A recent commit enabled Metadata parsing for model objects (see
PR #75), but the particulars of how the metadata was parsed into
the object were not tested. Since the metadata parsing itself is
currently handled by a custom extraction in the dir_to_model.py
module, we add some tests to ensure upcoming adjustments to the
parser keep the metadata parsing consistent with current behavior.

Note these tests provided detailed and limited scope coverage for
the specific case of parsing YAML files into the UserDefinedModel.
This is by design, as upcoming changes will be focused mainly
on internal parsing logic rather than expanding inputs. We can
broaden out the behavior of these tests if we ever allow for things
like YAML bytestream inputs in place of file paths for YAML file
locations.

### DIFF
--- a/metricflow/model/parsing/dir_to_model.py
+++ b/metricflow/model/parsing/dir_to_model.py
@@ -57,7 +57,7 @@ def parse_directory_of_yaml_files_to_model(
         dirs[:] = [d for d in dirs if not d.startswith(".")]
 
         for file in files:
-            if not (file.endswith(".yaml") or file.endswith(".yml")):
+            if not SafeLineLoader.is_valid_yaml_file_ending(file):
                 continue
             # Skip hidden files
             if file.startswith("."):
--- a/metricflow/model/parsing/yaml_loader.py
+++ b/metricflow/model/parsing/yaml_loader.py
@@ -34,3 +34,8 @@ class SafeLineLoader(yaml.SafeLoader):
         )  # change to 1-indexed
 
         return mapping
+
+    @staticmethod
+    def is_valid_yaml_file_ending(filename: str) -> bool:
+        """Checks if YAML file name ends with one of the supported suffixes"""
+        return filename.endswith(".yaml") or filename.endswith(".yml")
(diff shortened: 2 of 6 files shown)


tlento:
* main
  * **PR #149** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/149" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/transform-data/metricflow/149?utm_source=stack-comment).

tlento:
Worth noting - these are pretty detailed. They were immensely helpful in the parser refactor, but might be overkill as general purpose constructs. Given that other upcoming features depend on messing about with parsing and model semantics I'd like to leave these in place, but we can prune out the ones that are more "model implementation checks" and leave behind only those that are definitely valuable.

PR #150, opened 2022-07-11
Streamline YAML loading and parsing context construction

tlento:
Metricflow's YAML file loading uses two different loader types,
depending on whether we want file parsing context metadata included
in the payload or not. These were invoked via the one-line
yaml.load_all calls with the relevant parser (although sometimes we'd
use the wrong one and load context data in only to remove it later).

This causes problems with our context-aware parsing, because PyYAML
will only correctly populate the `name` field on its output nodes if
the input stream is a file-like object with the `name` property set
correctly. See the [relevant PyYAML source](https://github.com/yaml/pyyaml/blob/8cdff2c80573b8be8e8ad28929264a913a63aa33/lib/yaml/reader.py#L72-L85
) for details.

This PR streamlines the parser by centralizing our logic for converting 
an input string into a file-like object, and requesting the context-aware
loader when needed. 

We take advantage of this to streamline our context processing, making
both file name and input contents available through the ParsingContext
object instead of wiring them through custom callbacks.

This enables us to move onto Pydantic's parsing abstractions for 
things like populating the metadata field in a more re-usable manner.

### DIFF
--- a/metricflow/model/objects/common.py
+++ b/metricflow/model/objects/common.py
@@ -50,7 +50,13 @@ class FileSlice(HashableBaseModel):  # noqa: D
     end_line_number: int
 
 
-class YamlConfigFile(HashableBaseModel):  # noqa: D
+class YamlConfigFile(HashableBaseModel):
+    """Serializable container for customer model YAML contents
+
+    The serialization support is included here for scenarios where persisting the contents in non-filesystem storage
+    services is necessary or desirable.
+    """
+
     filepath: str
     contents: str
     url: Optional[str]
--- a/metricflow/test/model/parsing/test_metadata_parsing.py
+++ b/metricflow/test/model/parsing/test_metadata_parsing.py
@@ -110,3 +110,5 @@ def _assert_measure_metadata_is_valid(measures: Sequence[Measure]) -> None:
             f"Metadata: {measure.metadata}"
         )
         last_end_number = measure.metadata.file_slice.end_line_number
+        # if measure.name == "bookers":
+        #     raise ValueError(measure)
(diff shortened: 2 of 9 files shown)


tlento:
* main
    * **PR #150** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/150" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/transform-data/metricflow/150?utm_source=stack-comment).

metricflow/model/objects/common.py
```diff
@@ -50,7 +50,13 @@ class FileSlice(HashableBaseModel):  # noqa: D
...
 
 
-class YamlConfigFile(HashableBaseModel):  # noqa: D
+class YamlConfigFile(HashableBaseModel):
+    """Serializable container for customer model YAML contents
+
+    The serialization support is included here for scenarios where persisting the contents in non-filesystem storage
+    services is necessary or desirable.
+    """
+
```
QMalcolm, line 59: Thank you for adding a docstring to this class 🙏

QMalcolm (approved):
Thank you for disentangling this work. Not having to pass the context everywhere, makes things easier to reason about.

metricflow/test/model/parsing/test_metadata_parsing.py
```diff
@@ -102,3 +110,5 @@ def _assert_measure_metadata_is_valid(measures: Sequence[Measure]) -> None:
             f"Metadata: {measure.metadata}"
         )
         last_end_number = measure.metadata.file_slice.end_line_number
+        # if measure.name == "bookers":
```
QMalcolm, line 113: Apparently I did hit "add to review" on this before I submitted my review.
If this is commented out because it's no longer the correct thing to check, can we delete this? If this is commented out because, it's broken but we plan on bringing it back, can we add context about what is making this check an issue currently?

tlento: 🤦  Oops. Will remove.

PR #151, opened 2022-07-11
Move object level parsing into Pydantic's BaseModel APIs

tlento:
The original model parsing logic had a custom parsing method for
extracting the values provided by the user via their YAML configs.
This was in place because certain object value types, such as the
WhereClauseConstraint, expect user-provided values to be provided
as strings rather than object (key/value) types.

This parsing logic was broken in several subtle ways - the recursive
callback was not conforming to the type signature of the method,
for one thing, which causes quite a bit of confusion for readers.
More importantly, a hack in place to work around nested lists
of strings - which was needed for the materialization config spec -
would throw exceptions if anyone ever created a value with a nested
list of object types.

Rather than attempt to disentangle the parsing issues, this PR
moves us onto Pydantic's already-fully-functional parse_obj method by
replacing the ParseableField interface with a mixin allowing for an
input parser override for the Pydantic model objects that require it.
This is done by overriding the base "validator" method on that model
object, which Pydantic uses to both parse and validate the input
used to initialize the model object.

Since Pydantic expects this method to accept `Any` and return the class
type, this naturally allows us to extend our custom parsing to cases
where we need to add handling for string or object types, which is a
feature we will require in the future. We use the abstract class mixin
as both a convenience for providing the common Pydantic override and
an interface tag for allowing us to keep track of all special parsing
cases, which will make it easier to manage deprecation of backwards
compatibility shims for model parsing.

A similar treatment, albeit one somewhat different in implementation
specifics, was applied to our metadata parsing.


tlento:
* main
      * **PR #151** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/151" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/transform-data/metricflow/151?utm_source=stack-comment).

metricflow/test/model/parsing/test_metadata_parsing.py
```diff
@@ -39,6 +42,25 @@ def test_metric_metadata_parsing(simple_user_configured_model: UserConfiguredMod
...
+    so the YAML contents might or might not match.
+    """
+    assert len(multi_hop_join_semantic_model.user_configured_model.metrics) > 0
+    for metric in multi_hop_join_semantic_model.user_configured_model.metrics:
+        assert (
+            metric.metadata is not None
+        ), f"Metadata should always be parsed out of the model, but None found for metric: {metric}!"
+        _assert_metadata_filename_is_valid(metric.metadata)
+
+
```
QMalcolm, line 63: At what point in the future do we decide on if we are doing this? Personally I'm on the side of "metrics created by proxy should get the metadata of the measure, and maybe something denoting it was created by proxy"

QMalcolm (approved):
This is so clean. I'm so happy we're moving to using Pydantic's built in object parsing. That `_get_validators_` magic is so convenient for this kind of stuff

metricflow/test/model/parsing/test_metadata_parsing.py
tlento: We should decide this week and then we can open an issue to finalize the fixes. I don't want to push that across the MetricFlow/Transform boundary along with these changes since it affects parsing output, and we'll likely want to make Metadata required instead of optional in the model objects requesting it. But I also don't want this hanging in space forever, because if we really want Metadata required when set it's easier to do that now than later.

PR #152, opened 2022-07-11
Switch metric measure references from string name to model object

tlento:
## Switch metric measure references from string name to model object

Our Metric model object takes in MetricTypeParams that allows for various
input measure name references. These are all typed as strings or
lists of strings, since the current MetricFlow framework only requires
a pointer to the original measure model object.

Upcoming changes will add constraint expressions to the measure
pointers stored inside of the Metric model object, and as such we
will need to attach additional properties to each of the measures
referenced in the MetricTypeParams.

This commit simply introduces a new model object type to represent the
measure inputs for a metric, and converts the string types to use the
new object. All other changes are simply to manage the fallout of the
related type change. Subsequent commits will update the jsonschema and
enable user input of object-type representations and add new properties
as appropriate.

## Enable YAML object parsing for metric input measures

Now that the internals of the framework rely on the MetricInputMeasure
objects, we can allow users to specify either the legacy string name
representation of the measure or the full object representation.

This sets us up to add properties, such as constraints, to the
measures defined as inputs to metrics in the model.

PR #153, opened 2022-07-12
Allow specification of constraints on metric input measures

tlento:
One core feature request from customers is the ability to specify constrained input measures for metric definitions. For example, a user might want to define a metric that looks like this:

```
metric:
  name: instant_bookings_rate
  type: ratio
  type_params:
    numerator:
      name: bookings
      constraint: is_instant
      alias: bookings_with_is_instant_constraint
    denominator:
      name: bookings
```

Today, this is only possible by defining a separate data source with a `sql_query` property that applies the `WHERE is_instant` constraint in the source query, and then defining an `instant_bookings` measure on that data source, and using it in this ratio metric. This is silly - users shouldn't have to define new data sources for every possible constraint they care about for a given set of measures. It's also pretty nasty in the case where the constraint is based on a dimensional join.

This PR adds the ability to specify constrained measures in exactly the manner illustrated above. We chose this approach over composite metrics (so having an `instant_bookings` metric and a `bookings` metric, and then a metric ratio derived from them) primarily because it is more flexible, and we have additional feature requests pending that will require constrained measures in other contexts. However, future changes may allow for metric composition.

One note on aliasing - the alias is necessary, and must be user-provided for expr type metrics, because end users have to be able to refer to the metric in their expr. For example:

```
metric:
  name: instant_bookings_expr
  type: expr
  type_params:
    measures:
      - name: bookings
        constraint: is_instant
         alias: bookings_with_is_instant_constraint
      - name: bookings
    expr: "bookings_with_is_instant_constraint / CAST(NULLIF(bookings, 0) AS DOUBLE)"
```

The implementation of alias handling here was chosen because it has the smallest blast radius for refactoring, but future structural improvements (e.g., making our MeasureSpec more flexible and doing lookups via references rather than specs)  are likely in order.

A future PR will add validation to enforce that measure aliases follow the same constraints as measure names, because subsequent joins can cause problems with alias/measure name collisions.


tlento:
* main
          * **PR #153** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/153" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/transform-data/metricflow/153?utm_source=stack-comment).

metricflow/dataflow/dataflow_plan.py
```diff
@@ -343,10 +344,23 @@ class AggregatedMeasuresOutput(Generic[SourceDataSetT], BaseOutput[SourceDataSet
 
 
 class AggregateMeasuresNode(Generic[SourceDataSetT], AggregatedMeasuresOutput[SourceDataSetT]):
-    """A node that aggregates the measures by the associated group by elements."""
+    """A node that aggregates the measures by the associated group by elements.
 
-    def __init__(self, parent_node: BaseOutput) -> None:  # noqa: D
+    In the event that one or more of the aggregated input measures has an alias assigned to it, any output query
```
plypaul, line 349: Given the alias behavior, I'm thinking if we should rename this node. Can't come up with a name that isn't awkward though.

metricflow/plan_conversion/instance_converters.py
```diff
@@ -421,6 +450,54 @@ def transform(self, instance_set: InstanceSet) -> InstanceSet:  # noqa: D
...
+        combinded via joins.
+        """
+        self.metric_input_measure_specs = metric_input_measure_specs
+
+    def _alias_measure_instances(self, measure_instances: Tuple[MeasureInstance, ...]) -> Tuple[MeasureInstance, ...]:
+        """Update all measure instances with aliases, if any are found in the input spec set"""
+        aliased_instances: List[MeasureInstance] = []
+        aliased_input_specs = [spec for spec in self.metric_input_measure_specs if spec.alias]
+        for instance in measure_instances:
+            matches = [spec for spec in aliased_input_specs if spec.measure_spec == instance.spec]
```
plypaul, line 472: Could be good to assert that there's at most 1 match.

plypaul (commented):
Nice, generally looks good to me. With the change to some of the dataflow plan building functions and node behavior changes, there could be some downstream effects on internal engines. Did you try doing a sync to see what breaks? There's likely updates necessary for persisting aggregated measures for a metric.

metricflow/plan_conversion/instance_converters.py
tlento: I thought about this, but the reality is there could be more than 1 match because the measure_spec will be the same across all of the possibilities.

So this will work "correctly" in the end, but it may result in the wrong aliases being used.

Actually, now that I write it out, that's a disaster. I'll add the assertion for now since it should be safe with where this is called at the moment, and prioritize merging the new spec type into the base MeasureSpec after this launch sprint since that should ensure a single match in all cases.

metricflow/dataflow/dataflow_plan.py
tlento: I can't find this comment anywhere. WTF GitHub?

Anyway, yeah, it's tough, because MaybeAliasedAggregateMeasuresNode is super awkward, and AliasedAggregateMeasuresNode is also awkward and not exactly correct.

My personal preference is to find a way to have a type bound to the relevant operations so that the aliased or not behavior is either invisible to the caller (because they just call the common interface method and get the right value) or else is breaking in the type system (because anywhere they expect a non-aliased thing they cannot accept one of these), but I couldn't figure out how to make that happen without a pretty major set of changes. Something for us to powwow about later, perhaps.

tlento (commented):
> Did you try doing a sync to see what breaks? There's likely updates necessary for persisting aggregated measures for a metric.

I have not done it yet, that's the next step. Would be great to pair on that so we can come up with a complete test suite for those scenarios.

In theory it should all just work with today's use cases, but aliased partial computations are a concern.

plypaul (approved):
LGTM - though yeah, chatting about the implications before merge would be beneficial.

metricflow/test/fixtures/model_yamls/simple_model/metrics.yaml
```diff
@@ -3,21 +3,21 @@ metric:
...
 metric:
-    name: "average_booking_value"
-    description: "average booking value metric"
-    owners:
-        - support@transformdata.io
-    type: measure_proxy
-    type_params:
-        measures:
-            - average_booking_value
+  name: "average_booking_value"
```
tlento, line 13: Ooof, I think the metricflow model extension auto-formatted this whole file.

metricflow/test/fixtures/model_yamls/simple_model/metrics.yaml
```diff
@@ -350,4 +349,56 @@ metric:
   type_params:
     measures:
       - current_account_balance_by_user
----
\ No newline at end of file
+---
```
tlento, line 352: Below here is all the new stuff

metricflow/specs.py
```diff
@@ -346,10 +346,19 @@ class MetricInputMeasureSpec(SerializableDataclass):
...
     """
 
     measure_spec: MeasureSpec
     constraint: Optional[SpecWhereClauseConstraint] = None
+    alias: Optional[str] = None
+
+    @property
+    def post_aggregation_spec(self) -> MeasureSpec:
+        """Return a MeasureSpec instance representing the post-aggregation spec state for the underlying measure"""
+        return MeasureSpec(element_name=self.alias) if self.alias else self.measure_spec
```
WilliamDee, line 361: what happens here for measures with non-additive-dimension? or is this what you mean by this approach being brittle and having to pass everything through the dataflow plan?

WilliamDee (approved):
saw 1 potential issue that may occur for when interacting with semi-additive metrics, but otherwise looks good!

metricflow/model/objects/metric.py
```diff
@@ -27,10 +27,15 @@ class MetricType(ExtendedEnum):
...
+    """Provides a pointer to a measure along with metric-specific processing directives
+
+    If an alias is set, this will be used as the string name reference for this measure after the aggregation
+    phase in the SQL plan.
+    """
 
     name: str
-    # TODO add constraint handling property
+    constraint: Optional[WhereClauseConstraint]
+    alias: Optional[str]
```
QMalcolm, line 38: Would there be an issue if to two `MetricInputMeasure`s have the same alias? If so, we need to either add a new validation rule or update the unique naming validation rule to account for this. In the same vain, I assume there'd be an issue if an alias conflicts with an existing measure name?

QMalcolm (commented):
I know I wasn't requested for review, but... I took a look. I think this is going to require some validation improvements

metricflow/model/objects/metric.py
tlento: Yes. There are a bunch of subtleties with the alias attribute. I'm planning to add those validations in a separate PR.

metricflow/specs.py
tlento: Oh wow, I missed this. The new MeasureSpec needs to inherit the properties of the old one, just with a new name. I'll fix that here but I think we need to start thinking about factory methods for these "modified copy" cases so this logic isn't spread out all over the codebase.

metricflow/model/objects/metric.py
QMalcolm: I'm a dumb dumb, you put that at the end of your PR description 🤦 my bad

WilliamDee (approved):

PR #154, opened 2022-07-12
add option to remove SQL descriptions

WilliamDee:
## Context
When using `--explain` to get the SQL, it can be cluttering with all the inline description in the SQL itself.
## PR Changes
- added a new flag `--hide-sql-descriptions` to optionally remove the inline descriptions on the returned SQL

### DIFF
--- a/metricflow/engine/metricflow_engine.py
+++ b/metricflow/engine/metricflow_engine.py
@@ -144,6 +144,17 @@ class MetricFlowExplainResult:
 
         return sql_query
 
+    @property
+    def rendered_sql_without_descriptions(self) -> SqlQuery:
+        """Return the SQL query without the inline descriptions."""
+        sql_query = self.rendered_sql
+        return SqlQuery(
+            sql_query="\n".join(
+                filter(lambda line: not line.strip().startswith("--"), sql_query.sql_query.split("\n"))
+            ),
+            bind_parameters=sql_query.bind_parameters,
+        )
+
 
 class AbstractMetricFlowEngine(ABC):
     """Query interface for clients"""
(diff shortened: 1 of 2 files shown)


metricflow/engine/metricflow_engine.py
```diff
@@ -144,6 +144,15 @@ def rendered_sql(self) -> SqlQuery:
 
         return sql_query
 
+    @property
+    def rendered_sql_without_descriptions(self) -> SqlQuery:
+        """Return the SQL query without the inline descriptions."""
+        sql_query = self.rendered_sql
+        return SqlQuery(
+            sql_query="\n".join(filter(lambda line: "--" not in line, sql_query.sql_query.split("\n"))),
```
tlento, line 152: Careful. This:

```
SELECT
  ,metric_time  -- ds or date, depending on the measure
  ,measure_one
  ...
```

Produces this:

```
SELECT
  , measure_one
  ...
```

You should test it, but I think this works better, although it'll leave end-of-line comments:

```suggestion
            sql_query="\n".join(filter(lambda line: line.lstrip().startswith("--"), sql_query.sql_query.split("\n"))),
```

tlento: Note we could try to get fancy with string splitting or other detection to remove end-of-line comments as well, but we'd have to be careful about this:

```
constraint: 'user_flags NOT IN ("--help", "-h")'
```
SQL is not a regular language, which means getting that right requires the AST, so I'd go with the simpler startswith approach. It should be good enough, anyway.

tlento (changes requested):
There's a subtle flaw here that needs fixing.

Unrelated to anything, I'm going to try requesting changes to see what happens in my Graphite review queue, although every time I try request changes in GitHub I end up regretting it.

tlento (approved):

PR #156, opened 2022-07-13
reorder the results to identifiers,dimensions,metrics

WilliamDee:
## Context
Current ordering of columns on the tables coming out of MetricFlow is `metrics,dimensions,identifiers`. The more common practice is to do `identifiers,dimensions,metrics`.

## Changes
- Reorder the returned columns to be more consistent with most common ordering in current practice


metricflow/plan_conversion/select_column_gen.py
```diff
@@ -33,11 +33,11 @@ def as_tuple(self) -> Tuple[SqlSelectColumn, ...]:
         """Return all select columns as a tuple."""
         return tuple(
             # Preferred order is to have metrics first.
-            self.metric_columns
+            self.identifier_columns
```
tlento, line 36: That comment is now LIES! LIES I tell you!

Uh, sorry, maybe the iced coffee was a mistake. Please remove that comment.

tlento (commented):
Technically, this change looks good to me, but won't it break things downstream?

I can imagine materializations will be affected. Certainly anybody doing janky unsupported things like streaming CLI output through a bunch of unix pipes in a cron script is going to have a bad time here, but I'm ok with that because it's 2022 and nobody but me should be getting up to such things anymore.

Anyway, I think it'd be good to check in with @plypaul to see what other output interfaces we might be busting up here before we merge this, he has a more complete knowledge of the ways these datasets get emitted.

metricflow/plan_conversion/select_column_gen.py
```diff
@@ -33,11 +33,11 @@ def as_tuple(self) -> Tuple[SqlSelectColumn, ...]:
...
             + self.measure_columns
             + self.dimension_columns
             + self.time_dimension_columns
```
nhandel, line 39: I just noticed the time dimension and measure here also. per @belindabennett's suggestions, can we order these by:
time dimension, 
identifiers, 
dimensions, 
metrics, 
measures

plypaul:
This should be fine as the materialize call `MetricFlowEngine` drops the table first. In addition, internal use cases also check the columns and the order, so that should be okay as well.

metricflow/plan_conversion/select_column_gen.py
```diff
@@ -32,12 +32,12 @@ def merge(self, other_set: SelectColumnSet) -> SelectColumnSet:
     def as_tuple(self) -> Tuple[SqlSelectColumn, ...]:
         """Return all select columns as a tuple."""
         return tuple(
-            # Preferred order is to have metrics first.
-            self.metric_columns
-            + self.measure_columns
-            + self.dimension_columns
-            + self.time_dimension_columns
+            # Preferred order is to have time_dimensions first.
```
tlento, line 35: If we leave a comment here it should be something more like:

```suggestion
            # This order was chosen to match the column sequence data consumers typically prefer.
```

This tells people of the future that this order was chosen for a reason, which implies any change to the ordering needs justification (either a better reason or else that the order is not correct).

tlento (approved):

PR #157, opened 2022-07-13
Make mutability optional on configs

WilliamDee:
## Context
We don't use mutability in the OSS, but we still require it in the model configs.

## Changes
- Make mutability a default to fully mutable unless stated otherwise

### DIFF
--- a/metricflow/model/objects/data_source.py
+++ b/metricflow/model/objects/data_source.py
@@ -61,7 +61,9 @@ class DataSource(HashableBaseModel, ModelWithMetadataParsing):
     measures: Sequence[Measure] = []
     dimensions: Sequence[Dimension] = []
 
-    mutability: Mutability
+    mutability: Mutability = Mutability(
+        type=MutabilityType.FULL_MUTATION, type_params=MutabilityTypeParams(update_cron="0 0,12 * * *")
+    )
 
     origin: DataSourceOrigin = DataSourceOrigin.SOURCE
     metadata: Optional[Metadata]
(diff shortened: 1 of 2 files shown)


metricflow/model/objects/data_source.py
```diff
@@ -61,7 +61,7 @@ class DataSource(HashableBaseModel, ModelWithMetadataParsing):
     measures: Sequence[Measure] = []
     dimensions: Sequence[Dimension] = []
 
-    mutability: Mutability
+    mutability: Mutability = Mutability(type=MutabilityType.FULL_MUTATION, type_params=None)
```
tlento, line 64: We currently define all of our defaults in our model objects, but our dev tooling efforts around IDE integration are most likely going to be based on the jsonschema. This seems a bit off to me, but for now let's leave it as-is.

Longer term we might consider getting to a place where we can generate the jsonschema from our Pydantic objects. We have type mismatches between the jsonschema and the pydantic object itself which makes that very difficult at the moment, but it's something to think about.

tlento (approved):
Ok, I think this works as-is since we define no defaults in jsonschema today.

metricflow/model/parsing/schemas.py
```diff
@@ -245,7 +245,7 @@
         "constraint": {"$ref": "constraint_schema"},
     },
     "additionalProperties": False,
-    "required": ["name", "mutability"],
+    "required": ["name"],
```
tlento, line 248: Oh, good find, didn't realize it was required here.

tlento (approved):

PR #159, opened 2022-07-14
Update config parsing to optionally return issues instead of raising exceptions

QMalcolm:
The follower are examples prior to this PR and with this PR for configs which have two issues. 1) A metric without a name 2) a time dimension missing type params. Note that prior to this PR only one of the issues gets caught (thats because its the first one hit), whereas both are identified with the updates from this PR. I admit that the JSON schema validations aren't the best still, but they are a lot easier to look at than before.

Previously
```
poetry run mf validate-configs
(To see warnings and future-errors, run again with flag `--show-all`)
✔ 🎉 Successfully linted config YAML files (FATALS: 0, ERRORS: 0, FUTURE_ERRORS: 0, WARNINGS: 0)
⠙ Building model from configs
ERROR: An error occurred when attempting to build the semantic model
Log file: /Users/tqm/.metricflow/logs/metricflow.log
Traceback (most recent call last):
  File "/Users/tqm/Developer/transform-data/metricflow/metricflow/model/parsing/dir_to_model.py", line 141, in parse_config_yaml
    validate_config_structure(config_yaml)
  File "/Users/tqm/Developer/transform-data/metricflow/metricflow/model/parsing/validation.py", line 65, in validate_config_structure
    raise exceptions.ValidationError("\n".join(errors))
jsonschema.exceptions.ValidationError: {'name': 'ds', 'type': 'time', 'expr': 'CAST(created_date as DATETIME)'} is not valid under any of the given schemas

Failed validating 'anyOf' in schema['properties']['dimensions']['items']:
    {'$id': 'dimension_schema',
     'additionalProperties': False,
     'anyOf': [{'not': {'$ref': '#/definitions/is-time-dimension'}},
               {'required': ['type_params']}],
     'definitions': {'is-time-dimension': {'properties': {'type': {'enum': ['TIME',
                                                                            'time']}},
                                           'required': ['type']}},
     'properties': {'description': {'type': 'string'},
                    'display_name': {'type': 'string'},
                    'expr': {'type': ['string', 'boolean']},
                    'is_partition': {'type': 'boolean'},
                    'name': {'pattern': '(?!.*__).*^[a-z][a-z0-9_]*[a-z0-9]$',
                             'type': 'string'},
                    'owners': {'items': {'type': 'string'},
                               'type': 'array'},
                    'tier': {'type': ['string', 'integer']},
                    'type': {'enum': ['CATEGORICAL',
                                      'TIME',
                                      'categorical',
                                      'time']},
                    'type_params': {'$ref': 'dimension_type_params_schema'}},
     'required': ['name', 'type'],
     'type': 'object'}

On instance['dimensions'][0]:
    {'expr': 'CAST(created_date as DATETIME)', 'name': 'ds', 'type': 'time'}
'name' is a required property

Failed validating 'required' in schema:
    {'$id': 'metric',
     'additionalProperties': False,
     'properties': {'constraint': {'type': 'string'},
                    'description': {'type': 'string'},
                    'display_name': {'type': 'string'},
                    'locked_metadata': {'$ref': 'locked_metadata'},
                    'name': {'pattern': '(?!.*__).*^[a-z][a-z0-9_]*[a-z0-9]$',
                             'type': 'string'},
                    'owners': {'items': {'type': 'string'},
                               'type': 'array'},
                    'tier': {'type': ['string', 'integer']},
                    'type': {'enum': ['MEASURE_PROXY',
                                      'RATIO',
                                      'EXPR',
                                      'CUMULATIVE',
                                      'measure_proxy',
                                      'ratio',
                                      'expr',
                                      'cumulative']},
                    'type_params': {'$ref': 'metric_type_params'},
                    'where_constraint': {'type': 'string'}},
     'required': ['name', 'type', 'type_params'],
     'type': 'object'}

On instance:
    {'constraint': "stage_name = '6 - Closed Won' \n",
     'owners': ['support@transformdata.io'],
     'type': 'measure_proxy',
     'type_params': {'measures': ['opps']}}

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/tqm/Developer/transform-data/metricflow/metricflow/engine/utils.py", line 23, in build_user_configured_model_from_config
    model = parse_directory_of_yaml_files_to_model(models_path).model
  File "/Users/tqm/Developer/transform-data/metricflow/metricflow/model/parsing/dir_to_model.py", line 74, in parse_directory_of_yaml_files_to_model
    model = parse_yaml_files_to_model(yaml_config_files)
  File "/Users/tqm/Developer/transform-data/metricflow/metricflow/model/parsing/dir_to_model.py", line 103, in parse_yaml_files_to_model
    objects = parse_config_yaml(  # parse config file
  File "/Users/tqm/Developer/transform-data/metricflow/metricflow/model/parsing/dir_to_model.py", line 223, in parse_config_yaml
    raise ParsingException(
metricflow.errors.errors.ParsingException: Failed to parse YAML file '/Users/tqm/Developer/transform-data/metricflow/config-templates/salesforce/opportunities.yaml' - YAML file did not conform to metric spec.
Error: {'name': 'ds', 'type': 'time', 'expr': 'CAST(created_date as DATETIME)'} is not valid under any of the given schemas

Failed validating 'anyOf' in schema['properties']['dimensions']['items']:
    {'$id': 'dimension_schema',
     'additionalProperties': False,
     'anyOf': [{'not': {'$ref': '#/definitions/is-time-dimension'}},
               {'required': ['type_params']}],
     'definitions': {'is-time-dimension': {'properties': {'type': {'enum': ['TIME',
                                                                            'time']}},
                                           'required': ['type']}},
     'properties': {'description': {'type': 'string'},
                    'display_name': {'type': 'string'},
                    'expr': {'type': ['string', 'boolean']},
                    'is_partition': {'type': 'boolean'},
                    'name': {'pattern': '(?!.*__).*^[a-z][a-z0-9_]*[a-z0-9]$',
                             'type': 'string'},
                    'owners': {'items': {'type': 'string'},
                               'type': 'array'},
                    'tier': {'type': ['string', 'integer']},
                    'type': {'enum': ['CATEGORICAL',
                                      'TIME',
                                      'categorical',
                                      'time']},
                    'type_params': {'$ref': 'dimension_type_params_schema'}},
     'required': ['name', 'type'],
     'type': 'object'}

On instance['dimensions'][0]:
    {'expr': 'CAST(created_date as DATETIME)', 'name': 'ds', 'type': 'time'}
'name' is a required property

Failed validating 'required' in schema:
    {'$id': 'metric',
     'additionalProperties': False,
     'properties': {'constraint': {'type': 'string'},
                    'description': {'type': 'string'},
                    'display_name': {'type': 'string'},
                    'locked_metadata': {'$ref': 'locked_metadata'},
                    'name': {'pattern': '(?!.*__).*^[a-z][a-z0-9_]*[a-z0-9]$',
                             'type': 'string'},
                    'owners': {'items': {'type': 'string'},
                               'type': 'array'},
                    'tier': {'type': ['string', 'integer']},
                    'type': {'enum': ['MEASURE_PROXY',
                                      'RATIO',
                                      'EXPR',
                                      'CUMULATIVE',
                                      'measure_proxy',
                                      'ratio',
                                      'expr',
                                      'cumulative']},
                    'type_params': {'$ref': 'metric_type_params'},
                    'where_constraint': {'type': 'string'}},
     'required': ['name', 'type', 'type_params'],
     'type': 'object'}

On instance:
    {'constraint': "stage_name = '6 - Closed Won' \n",
     'owners': ['support@transformdata.io'],
     'type': 'measure_proxy',
     'type_params': {'measures': ['opps']}}

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/tqm/Developer/transform-data/metricflow/metricflow/cli/utils.py", line 198, in wrapper
    func(*args, **kwargs)
  File "/Users/tqm/Developer/transform-data/metricflow/metricflow/telemetry/reporter.py", line 149, in wrapped
    return func(*args, **kwargs)
  File "/Users/tqm/Developer/transform-data/metricflow/metricflow/cli/main.py", line 721, in validate_configs
    user_model = cfg.user_configured_model
  File "/Users/tqm/Developer/transform-data/metricflow/metricflow/cli/cli_context.py", line 111, in user_configured_model
    self._user_configured_model = build_user_configured_model_from_config(self.config)
  File "/Users/tqm/Developer/transform-data/metricflow/metricflow/engine/utils.py", line 27, in build_user_configured_model_from_config
    raise ModelCreationException from e
metricflow.errors.errors.ModelCreationException: An error occurred when attempting to build the semantic model
```
now
```
$ poetry run mf validate-configs    
(To see warnings and future-errors, run again with flag `--show-all`)
✔ 🎉 Successfully linted config YAML files (FATALS: 0, ERRORS: 0, FUTURE_ERRORS: 0, WARNINGS: 0)
✖ Breaking issues found when building model from configs (FATALS: 0, ERRORS: 2, FUTURE_ERRORS: 0, WARNINGS: 0)
• ERROR: in file `/Users/tqm/Developer/transform-data/metricflow/config-templates/salesforce/opportunities.yaml` on line #1 - YAML document did not conform to metric spec.
Error: {'name': 'ds', 'type': 'time', 'expr': 'CAST(created_date as DATETIME)', '__parsing_context__': <metricflow.model.parsing.yaml_loader.ParsingContext object at 0x120578490>} is not valid under any of the given schemas

Failed validating 'anyOf' in schema['properties']['dimensions']['items']:
    {'$id': 'dimension_schema',
     'additionalProperties': False,
     'anyOf': [{'not': {'$ref': '#/definitions/is-time-dimension'}},
               {'required': ['type_params']}],
     'definitions': {'is-time-dimension': {'properties': {'type': {'enum': ['TIME',
                                                                            'time']}},
                                           'required': ['type']}},
     'properties': {'__parsing_context__': {},
                    'description': {'type': 'string'},
                    'display_name': {'type': 'string'},
                    'expr': {'type': ['string', 'boolean']},
                    'is_partition': {'type': 'boolean'},
                    'name': {'pattern': '(?!.*__).*^[a-z][a-z0-9_]*[a-z0-9]$',
                             'type': 'string'},
                    'owners': {'items': {'type': 'string'},
                               'type': 'array'},
                    'tier': {'type': ['string', 'integer']},
                    'type': {'enum': ['CATEGORICAL',
                                      'TIME',
                                      'categorical',
                                      'time']},
                    'type_params': {'$ref': 'dimension_type_params_schema'}},
     'required': ['name', 'type'],
     'type': 'object'}

On instance['dimensions'][0]:
    {'__parsing_context__': <metricflow.model.parsing.yaml_loader.ParsingContext object at 0x120578490>,
     'expr': 'CAST(created_date as DATETIME)',
     'name': 'ds',
     'type': 'time'}
• ERROR: in file `/Users/tqm/Developer/transform-data/metricflow/config-templates/salesforce/opportunities.yaml` on line #62 - YAML document did not conform to metric spec.
Error: 'name' is a required property

Failed validating 'required' in schema:
    {'$id': 'metric',
     'additionalProperties': False,
     'properties': {'__parsing_context__': {},
                    'constraint': {'type': 'string'},
                    'description': {'type': 'string'},
                    'display_name': {'type': 'string'},
                    'locked_metadata': {'$ref': 'locked_metadata'},
                    'name': {'pattern': '(?!.*__).*^[a-z][a-z0-9_]*[a-z0-9]$',
                             'type': 'string'},
                    'owners': {'items': {'type': 'string'},
                               'type': 'array'},
                    'tier': {'type': ['string', 'integer']},
                    'type': {'enum': ['MEASURE_PROXY',
                                      'RATIO',
                                      'EXPR',
                                      'CUMULATIVE',
                                      'measure_proxy',
                                      'ratio',
                                      'expr',
                                      'cumulative']},
                    'type_params': {'$ref': 'metric_type_params'},
                    'where_constraint': {'type': 'string'}},
     'required': ['name', 'type', 'type_params'],
     'type': 'object'}

On instance:
    {'__parsing_context__': <metricflow.model.parsing.yaml_loader.ParsingContext object at 0x12057a430>,
     'constraint': "stage_name = '6 - Closed Won' \n",
     'owners': ['support@transformdata.io'],
     'type': 'measure_proxy',
     'type_params': {'__parsing_context__': <metricflow.model.parsing.yaml_loader.ParsingContext object at 0x12057a4f0>,
                     'measures': ['opps']}}
```


metricflow/model/parsing/schemas.py
```diff
@@ -1,5 +1,7 @@
 from jsonschema import RefResolver, Draft7Validator
 
+from metricflow.model.parsing.yaml_loader import PARSING_CONTEXT_KEY
```
tlento, line 3: Per outside discussion, I think we can probably set an internal standard that keys surrounded by (or at least starting with) double-underscore are for internal use only. It's consistent with a lot of other usages as well.

It might be good to document that. Maybe we can put it over wherever we have PARSING_CONTEXT_KEY. I'm not sure if it should be a random module-level constant, but I guess there it is. Moving it seems like a lot of bother for not much benefit at this point.

metricflow/model/parsing/dir_to_model.py
```diff
@@ -134,11 +138,11 @@ def parse_config_yaml(
     data_source_class: Type[DataSource] = DataSource,
     metric_class: Type[Metric] = Metric,
     materialization_class: Type[Materialization] = Materialization,
-) -> List[Union[DataSource, Metric, Materialization]]:
+) -> Tuple[List[Union[DataSource, Metric, Materialization]], List[ValidationIssueType]]:
```
tlento, line 161: Normally when you have a pair of lists in a return like this they line up. So I'd expect this to return a 1:1 mapping of results to issues, or possibly a result or an issue in a given index location.

But what this does is it interleaves them in the 1:1 case - if there's an issue I generally won't get a whatever - but it can also spawn a bunch of issues. This means there's no way for me to map issues back to whatever I'm seeing in the return output.

Would we be better off making this return a `List[Tuple[Union[DataSource etc.], [List[ValidationIssueType]]]]`? Then every entry has one or both of an object and a list of validation issues.

Alternatively, maybe we're better off with a more descriptive data structure that has `result` and `errors` and is documented as having one or the other but not both.

metricflow/model/parsing/dir_to_model.py
```diff
@@ -171,23 +177,20 @@ def parse_config_yaml(
...
-                    errors.append(
-                        str(
-                            ParsingException(
-                                f"Unsupported version {version} in config document.",
-                                config_filepath=config_yaml.filepath,
-                            )
+                    issues.append(
+                        ValidationError(
+                            context=FileContext(file_name=ctx.filename, line_number=ctx.start_line),
+                            message=f"Unsupported version {version} in config document.",
```
tlento, line 203: See like this doesn't continue or skip or anything, so we'll get results offset from errors by one per document with VERSION_KEY set to the wrong thing.

metricflow/model/parsing/dir_to_model.py
```diff
@@ -206,41 +209,28 @@ def parse_config_yaml(
...
+        context = FileContext(file_name=ctx.filename, line_number=ctx.start_line) if ctx is not None else None
+        issues.append(ValidationError(context=context, message=str(e)))
     except Exception as e:
-        raise ParsingException(
-            message=str(e),
-            ctx=ctx,
-            config_filepath=config_yaml.filepath,
-        ) from e
+        context = FileContext(file_name=ctx.filename, line_number=ctx.start_line) if ctx is not None else None
+        issues.append(ValidationError(context=context, message=str(e)))
```
tlento, line 263: This obliterates the stack traces, which will make debugging really painful for us if we ever hit a model validation issue that resolves to an issue in the parsing/validation stack itself.

Let's either store the stack traces or else collect exceptions and re-throw in the end.

metricflow/model/parsing/dir_to_model.py
```diff
@@ -82,13 +84,21 @@ def parse_directory_of_yaml_files_to_model(
...
+        if apply_post_transformations:
+            model = ModelTransformer.post_validation_transform_model(model)
+    except Exception as e:
+        transformation_issue_results = ModelValidationResults(errors=[ValidationError(message=str(e))])
+        build_issues = ModelValidationResults.merge([build_issues, transformation_issue_results])
 
-    if apply_post_transformations:
-        model = ModelTransformer.post_validation_transform_model(model)
+    if raise_issues_as_exceptions and build_issues.has_blocking_issues:
+        raise ModelValidationException(build_issues.all_issues)
```
tlento, line 98: So if we can store the original exceptions we can throw this from those, and we'd be able to collect all of the stack traces. Here it just has string pointers to things happening... somewhere.

I've had mixed experiences with this kind of global "collect issues and maybe throw later" implementation pattern - I think we might want to have cleaner separation between what constitutes an exception and what constitutes a user error, and throw exceptions for the former and collect issues for the latter. I'm not sure how to go about achieving this, honestly, because I don't know where our parsing error boundary ends and the user input error boundary begins, but it's something to think about as we tinker further here.

In the meantime, preserving stack trace information is probably sufficient.

metricflow/model/parsing/dir_to_model.py
```diff
@@ -220,38 +219,47 @@ def parse_config_yaml(
...
+                    results.append(materialization_class.parse_obj(object_cfg))
+                else:
+                    issues.append(
+                        ValidationError(
+                            context=FileContext(file_name=ctx.filename, line_number=ctx.start_line),
+                            message=f"Invalid document type: {document_type}. Expected {DOCUMENT_TYPES}.",
+                        )
+                    )
+            # catches exceptions from jsonschema validator
+            except exceptions.ValidationError as e:
```
tlento, line 240: Oh I see, this thing is from jsonschema. Got it.

metricflow/model/parsing/dir_to_model.py
```diff
@@ -220,38 +219,47 @@ def parse_config_yaml(
...
+            except ParsingException as e:
+                context = FileContext(file_name=ctx.filename, line_number=ctx.start_line)
+                issues.append(ValidationError(context=context, message=str(e)))
+            # general exception for a given document. Basicially we don't want an exception
+            # on one document to halt checking the rest of the documents
+            except Exception as e:
+                context = FileContext(file_name=ctx.filename, line_number=ctx.start_line)
+                issues.append(ValidationError(context=context, message=str(e)))
+    # If a runtime error occured, we still want this to break things
+    except RuntimeError:
```
tlento, line 258: Interesting. Is this establishing our boundary between user input issues and internal errors?

I don't know if our error hierarchy is this disciplined, though. Even with Python - ValueError and AssertionErrors are not RuntimeError subclasses. I suggest we lose the global Exception -> issue conversion and instead refine around what user input might throw.

metricflow/cli/main.py
```diff
@@ -36,7 +36,7 @@
 from metricflow.configuration.config_builder import YamlTemplateBuilder
 from metricflow.dataflow.dataflow_plan_to_text import dataflow_plan_as_text
 from metricflow.engine.metricflow_engine import MetricFlowQueryRequest, MetricFlowExplainResult, MetricFlowQueryResult
-from metricflow.engine.utils import path_to_models
+from metricflow.engine.utils import model_build_result_from_config, path_to_models
```
tlento, line 39: Your commit message has a typo in it. Github won't let me make that comment any other way so I'm putting it here.

metricflow/model/parsing/dir_to_model.py
```diff
@@ -196,48 +219,47 @@ def parse_config_yaml(
...
-    except ParsingException:
+            # catches exceptions from *.parse_obj calls
+            except ParsingException as e:
+                context = FileContext(file_name=ctx.filename, line_number=ctx.start_line)
+                issues.append(ValidationError(context=context, message=str(e)))
+            # general exception for a given document. Basicially we don't want an exception
+            # on one document to halt checking the rest of the documents
+            except Exception as e:
+                context = FileContext(file_name=ctx.filename, line_number=ctx.start_line)
+                issues.append(ValidationError(context=context, message=str(e)))
```
tlento, line 256: These are identical, let's consolidate them.
```suggestion
            except (ParsingException, Exception) as e:
                # convert errors with *.parse_obj or general exceptions for a given document into issues
                context = FileContext(file_name=ctx.filename, line_number=ctx.start_line)
                issues.append(ValidationError(context=context, message=str(e)))
```

tlento (commented):
Any thoughts on stack trace preservation for ease of debugging? str(Exception) is just str(Exception.args).

QMalcolm:
> Any thoughts on stack trace preservation for ease of debugging? str(Exception) is just str(Exception.args).

I've been debating adding more properties to ValidationIssue. Specifically because some errors are rather large, and and much of it isn't *actually* needed. I.e. perhaps adding an `extra_details` property which is then available if a flag like `--verbose` is specified

metricflow/model/parsing/dir_to_model.py
QMalcolm: Yea 🙃 I should have realized names like `ValidationError` would eventually have collisions like this. But I didn't, and so here we are 😬

metricflow/model/parsing/dir_to_model.py
QMalcolm: It was not intended that the two lists be related in that manner. I.e. I wasn't intending to ensure a one-to-one matching. Additionally it's possible that an issue exists for an object that doesn't because creation of the object from the config failed. I'll go the route of having a more descriptive data structure that has `results` and `errors`, it'll be somewhat similar to `ModelBuildResults`

metricflow/cli/main.py
QMalcolm: Fixed 👍

metricflow/model/parsing/dir_to_model.py
QMalcolm: Yes indeed. Basically internal errors aren't actually a validation error, or shouldn't be. If we swallow them, they might be harder to realize. Ideally if there is an internal logic error, it should be raised an exception and hopefully caught during unit tests or sanity testing changes.

metricflow/model/parsing/schema_validator.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+    if not validator.is_type(instance, "object"):
+        return
+
+    extras = set(custom_find_additional_properties(instance, schema))
+
+    if validator.is_type(aP, "object"):
+        for extra in extras:
+            for error in validator.descend(instance[extra], aP, path=extra):
+                yield error
```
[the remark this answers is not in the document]
tlento: Also, if taking over control flow at this point is intentional, that should probably be documented.

tlento (commented):
Everything but the last commit (the custom jsonschema validator) looks good to me. 

I think it's worth separating that one into its own PR, honestly, it'll be easier for people to find it and we can do a bit more discussion on that without blocking the rest of these improvements. Or maybe it's fine to put a pattern in each schema and just use the standard Draft7Validator - how many new schemas will we create?

I wish there was a way to share core properties in all schemas, haven't found one yet.

QMalcolm:
> Everything but the last commit (the custom jsonschema validator) looks good to me.
> 
> I think it's worth separating that one into its own PR, honestly, it'll be easier for people to find it and we can do a bit more discussion on that without blocking the rest of these improvements. Or maybe it's fine to put a pattern in each schema and just use the standard Draft7Validator - how many new schemas will we create?
> 
> I wish there was a way to share core properties in all schemas, haven't found one yet.

Because this PR includes the merge of the old `validate_config_structure` into `parse_config_yaml`, we have to ignore the `__parsing_context__` key in objects somehow. I originally did this by simply included it as a property in all schemas, which was a sticking point, which is why we ended up adding the custom validator instead, because it was desired to ignore the key programmatically and not explicitly say "hey you can add this to your configs". I am happy to instead do it as a `patternProperty`, but I don't know how that gets passed the originally raised concern. It seems though I copied an out of date version of jsonschema and the included Draft7Validator.

All that said, I'm happy to drop the custom validator and instead manually add the it as a `patternProperty` to every schema if that is acceptable (which it sounds like it is).

metricflow/model/validations/validator_helpers.py
```diff
@@ -155,6 +154,7 @@ class ValidationIssue(ABC, BaseModel):
 
     message: str
     context: Optional[ValidationContext] = None
+    extra_detail: Optional[str]
```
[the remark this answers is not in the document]
QMalcolm: List[str] works from now. We're gonna want a simple way to stringify whatever extra_detail is. But what that looks like doesn't seem important for this PR. If it was a dict, perhaps just `str(extra_details)` would be good enough 🤷 But again, I think `List[str]` is good enough for now.

metricflow/model/parsing/schema_validator.py
```diff
@@ -0,0 +1,105 @@
...
+        "not": _validators.not_,
+        "pattern": _validators.pattern,
+        "patternProperties": _validators.patternProperties,
+        "properties": _validators.properties,
+        "propertyNames": _validators.propertyNames,
+        "required": _validators.required,
+        "type": _validators.type,
+        "uniqueItems": _validators.uniqueItems,
+    },
+    type_checker=_types.draft7_type_checker,
```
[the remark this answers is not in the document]
QMalcolm: Yes, looks like I either copied an old version Draft7Validator or incompletely copied Draft7Validator somehow

metricflow/model/parsing/schema_validator.py
```diff
@@ -0,0 +1,105 @@
...
+        "additionalItems": _validators.additionalItems,
+        "additionalProperties": customAdditionalProperties,
+        "allOf": _validators.allOf,
+        "anyOf": _validators.anyOf,
+        "const": _validators.const,
+        "contains": _validators.contains,
+        "dependencies": _validators.dependencies,
+        "enum": _validators.enum,
+        "exclusiveMaximum": _validators.exclusiveMaximum,
+        "exclusiveMinimum": _validators.exclusiveMinimum,
```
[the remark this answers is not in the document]
QMalcolm: No it was not not intentional. Things were supposed to be on par with whatever Draft7Validator is doing currently

metricflow/model/parsing/schema_validator.py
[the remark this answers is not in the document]
QMalcolm: The only difference for this method in comparison to the original was the call to `custom_find_additonal_properties`, so yes I agree, and the control flow change is unintentional.

metricflow/model/parsing/schema_validator.py
```diff
@@ -0,0 +1,105 @@
...
+        if "patternProperties" in schema:
+            patterns = sorted(schema["patternProperties"])
+            if len(extras) == 1:
+                verb = "does"
+            else:
+                verb = "do"
+            error = "%s %s not match any of the regexes: %s" % (
+                ", ".join(map(repr, sorted(extras))),
+                verb,
+                ", ".join(map(repr, patterns)),
```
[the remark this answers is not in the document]
QMalcolm: This was copied from the original, although somehow an out of date version. Happy to update to the the current version which I somehow didn't copy

metricflow/model/parsing/schema_validator.py
```diff
@@ -0,0 +1,105 @@
...
+            )
+            yield ValidationError(error)
+        else:
+            error = "Additional properties are not allowed (%s %s unexpected)"
+            yield ValidationError(error % extras_msg(extras))
+
+
+SchemaValidator = create(
```
tlento, line 67: If you use `extend()` here you only need to specify `validators = {"additionalProperties": customAdditionalProperties}`

According to the docblock that will override the Draft7 default for that one validator, while keeping everything else as it is defined in the package, and will give you what amounts to a customized Draft7Validator instance.

tlento (approved):
Oh I see, the jsonschema validator stack got updated. We're like a year out of date. And it looks like they did yield from because it's more concise but there's no way to be sure.

I think we should unblock this at this point and then open an issue to update the jsonschema package and clean everything up later. We can punt the decision about whether to keep the custom validator or replace it with a utility function to then. I'll document what I"m talking about in the issue.

One change you might want to make before merging is to use extend() instead of create() and pass our custom validator through to clobber the one Draft7 is using. That way we get whatever Draft7 is using but updated with our own validator. Then when we update the package we only need to clean up our custom validator implementation - the Draft7 inheritance should continue to work as-is.

metricflow/model/parsing/schema_validator.py
QMalcolm: Nice call, that makes things more straight forward!

PR #161, opened 2022-07-15
Update Postgres snapshot files

tlento:
We don't currently run Postgres tests on PRs or merge or via our
label or, really, at all. So we keep forgetting to update
the snapshots. The first commit should mostly be changes caused by the metric
time change in #59, and the second should be due to the column reordering in #156.

### DIFF
--- a/metricflow/test/snapshots/test_dataflow_to_sql_plan.py/SqlQueryPlan/PostgresSqlClient/test_cumulative_metric__plan0.sql
+++ b/metricflow/test/snapshots/test_dataflow_to_sql_plan.py/SqlQueryPlan/PostgresSqlClient/test_cumulative_metric__plan0.sql
@@ -1,31 +1,33 @@
 -- Compute Metrics via Expressions
 SELECT
-  subq_5.txn_revenue AS trailing_2_months_revenue
-  , subq_5.ds__month
+  subq_4.txn_revenue AS trailing_2_months_revenue
+  , subq_4.ds__month
 FROM (
   -- Aggregate Measures
   SELECT
-    SUM(subq_4.txn_revenue) AS txn_revenue
-    , subq_4.ds__month
+    SUM(subq_3.txn_revenue) AS txn_revenue
+    , subq_3.ds__month
   FROM (
-    -- Join Self Over Time Range
+    -- Pass Only Elements:
+    --   ['txn_revenue', 'ds__month']
     SELECT
-      subq_1.txn_revenue AS txn_revenue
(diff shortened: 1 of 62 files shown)


tlento:
@plypaul please review with some care to make sure these really are expected, I only took a quick pass.

tlento:
* main
  * **PR #161** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/161" target="_blank">Graphite</a>  👈
    * **PR #162** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/162" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/transform-data/metricflow/161?utm_source=stack-comment).

metricflow/test/snapshots/test_dataflow_to_sql_plan.py/SqlQueryPlan/PostgresSqlClient/test_cumulative_metric__plan0.sql
```diff
@@ -1,31 +1,33 @@
 -- Compute Metrics via Expressions
 SELECT
-  subq_5.txn_revenue AS trailing_2_months_revenue
-  , subq_5.ds__month
+  subq_4.txn_revenue AS trailing_2_months_revenue
```
plypaul, line 3: I think there's a bug in how cumulative metrics are generated with `metric_time` - pending further investigation.

plypaul: Actually, this is because the test is not using `metric_time`:

https://github.com/transform-data/metricflow/blob/1c498ddf147eb3e1d9d3b6fe983425824e9af459/metricflow/test/plan_conversion/test_dataflow_to_sql_plan.py#L769

Although it seems like we should throw an error if you try to select a cumulative metric with a window / grain to date without it.

tlento: That's probably reasonable. I guess the question is if we should throw an error if they manually select the true column `metric_time` would target - like I'm assuming `ds` in this case is what `metric_time` would resolve to in practice.

Let's file a separate issue for that so we can get the Postgres work unstuck.

WilliamDee (approved):
ah, forgot to update pg when I made my reordering column change

PR #162, opened 2022-07-16
Add PostgreSQL tests to SQL Engine test github action

tlento:
Not testing PostgreSQL means we forget to do important things
like keep the PostgreSQL plans up to date or just generally run
the test suite against it.

This PR adds it to the action so any time we trigger a full
SQL engine test run PostgreSQL will be included.

### DIFF
--- a/.github/workflows/cd-sql-engine-tests.yaml
+++ b/.github/workflows/cd-sql-engine-tests.yaml
@@ -107,6 +107,55 @@ jobs:
           MF_SQL_ENGINE_URL: ${{ secrets.MF_BIGQUERY_URL }}
           MF_SQL_ENGINE_PASSWORD: ${{ secrets.MF_BIGQUERY_PWD }}
 
+  postgres-tests:
+    name: PostgreSQL Tests
+    if: ${{ github.event.action != 'labeled' || github.event.label.name == 'run_mf_sql_engine_tests' }}
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgres
+        env:
+          POSTGRES_USER: postgres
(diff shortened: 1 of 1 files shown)


tlento:
* main
  * **PR #161** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/161" target="_blank">Graphite</a> 
    * **PR #162** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/162" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/transform-data/metricflow/162?utm_source=stack-comment).

.github/workflows/cd-sql-engine-tests.yaml
```diff
@@ -107,6 +107,55 @@ jobs:
           MF_SQL_ENGINE_URL: ${{ secrets.MF_BIGQUERY_URL }}
           MF_SQL_ENGINE_PASSWORD: ${{ secrets.MF_BIGQUERY_PWD }}
 
+  postgres-tests:
```
tlento, line 110: These have no special environment dependencies, so they fire immediately on adding the label.

Do we want to move this to the blocking test suite that runs on every PR? There are no external dependencies that GitHub can't handle on its own, we don't need any secrets, and PostgreSQL is pretty fast, so it might be nice to have a "regular" engine test running to make sure we don't forget plan file updates and whatnot.

plypaul: Yeah, it would be fine to add this, but DuckDB is also a "regular" engine.

plypaul (approved):

PR #163, opened 2022-07-18
Add classes for running and rendering inference results

serramatutu:
This PR adds classes for coercing inference results as definitive results, rendering inference results and running it as a cohesive unit.

For SOLVING (coercing results):
- Added base `InferenceSolver` class.
- Added `WeightedTypeTreeInferenceSolver` concrete implementation. It uses assigns weights to each node in the column type tree, and progresses deeper through the path with most weight. If there is a weight bifurcation (which means two paths are equally likely (according to a threshold), it assumes the rules produced confusing results and returns the deepest common ancestor.

For RENDERING:
- Added base `InferenceRenderer` class.
- Added `StreamInferenceRenderer` concrete implementation which prints inference results/reasons as human-readable text to a `TextIO`. This is mainly going to be used by the CLI module to explain what was produced and why.
- Added `ConfigFileInferenceRenderer` concrete implementation which writes inference results to actual Metricflow YAML config files.

For RUNNING:
- Added `InferenceRunner` class. It exposes a `run()` function that will act as a main function for inference.
- In the future, the CLI will use it to set up all necessary classes based on input. It can also be used to easily write custom inference scripts.

Example usage of `InferenceRunner`:

```python
from metricflow.configuration.config_handler import ConfigHandler
from metricflow.dataflow.sql_table import SqlTable
from metricflow.sql_clients.sql_utils import make_sql_client_from_config

from metricflow.inference.context.snowflake import SnowflakeInferenceContextProvider
from metricflow.inference.rule.defaults import DEFAULT_RULESET
from metricflow.inference.solver.weighted_tree import WeightedTypeTreeInferenceSolver
from metricflow.inference.renderer.stream import StreamInferenceRenderer
from metricflow.inference.renderer.config_file import ConfigFileRenderer
from metricflow.inference.runner import InferenceRunner


provider = SnowflakeInferenceContextProvider(
    client=make_sql_client_from_config(ConfigHandler()),
    tables=[
        SqlTable.from_string("DB.SCHEMA.TABLE1"),
        SqlTable.from_string("DB.SCHEMA.TABLE2"),
        SqlTable.from_string("DB.SCHEMA.TABLE3"),
    ],
    max_sample_size=10000
)

runner = InferenceRunner(
    context_providers=[provider],
    ruleset=DEFAULT_RULESET,
    solver=WeightedTypeTreeInferenceSolver(),
    renderers=[
        StreamInferenceRenderer.stdout(),
        ConfigFileRenderer(".testing/configs", overwrite=True),
    ]
)

runner.run()
```


metricflow/inference/models.py
```diff
@@ -0,0 +1,134 @@
...
+        NOTE: for practical reasons, a node is considered a descendant of itself.
+        """
+        return other == self or other in self.ancestors
+
+
+# This is kinda horrible but there's no way of instancing the tree with type safety without
+# hardcoding it. Having some magic that dynamically assigns attributes could work, but then
+# we lose IDE autocompletion and static checking
+class _TreeNodes:  # noqa: D
+    root = InferenceSignalNode(None, "UNKNOWN")
```
serramatutu, line 68: Btw I don't really love this code but I just wanted to get something going. I know this model is gonna fall short for some future necessities (like adding type-specific parameters, e.g target table/column for foreign keys).

I'll refactor this soon, but wanted to get something rolling for now.

tlento:
Neat, I'll take a look tomorrow sometime.

One high level note: builder patterns are not common in Python because they rarely provide meaningful value over named parameters. With the context you've provided here it doesn't seem to add much, and it sets users up to get into chained error states by not calling all of the right methods.

Do you expect to be passing around partly configured InferenceRunner instances a lot? If so it'd be good to document that in the InferenceRunner docstring and associated commit message, otherwise the need for the builder won't be apparent at this stage.

metricflow/inference/renderer/config_file.py
```diff
@@ -0,0 +1,97 @@
...
+    def __init__(self, dir_path: Union[str, Path], overwrite=False) -> None:
+        """Initializes the renderer.
+
+        dir_path: The path to the config directory
+        overwrite: If set to False, will raise error if the directory exists
+        """
+        dir_path = os.path.abspath(dir_path)
+
+        if not overwrite and os.path.exists(dir_path):
+            raise ValueError("ConfigFileRender.override is False but path exists.")
```
serramatutu, line 29: For some reason I wrote `override` instead of `overwrite`. Will change.

metricflow/inference/renderer/stream.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def render(self, results: List[InferenceResult]) -> None:
+        """Write the results to the configured TextIO as human-readable text."""
+        reason_delim = "\n  - "
+        for i, result in enumerate(results):
+            reasons_str = "No reason found for this result."
+            if len(result.reasons) > 0:
+                reasons_str = reason_delim + reason_delim.join(result.reasons)
+            lines = [
+                f"{result.column.sql} (#{i + 1})\n" f"  type: {result.type_node.name}\n" f"  reasons: ",
```
serramatutu, line 33: missing comma here. will fix also.

metricflow/inference/rule/rules.py
```diff
@@ -30,18 +33,19 @@ def __init__(
         self.matcher = matcher
         self.type_node = type_node
         self.confidence = confidence
+        self.match_reason = match_reason
 
     def process(self, warehouse: DataWarehouseInferenceContext) -> List[InferenceSignal]:  # type: ignore
         """Try to match all columns' names with the matching function.
```
serramatutu, line 39: need to update this doc.

metricflow/inference/rule/defaults.py
```diff
@@ -59,6 +90,23 @@ def primary_identifier_rule() -> ColumnMatcherRule:
...
+        The returned rule will match columns with all unique values if their types are either
+        INTEGER or STRING.
+
+        It will always produce a UNIQUE_IDENTIFIER signal with HIGH confidence.
+        """
+        return ColumnMatcherRule(
+            matcher=RuleDefaults._unique_identifier_matcher,
+            type_node=InferenceSignalType.ID.UNIQUE,
+            confidence=InferenceSignalConfidence.HIGH,
+            match_reason="Column values are unique.",
```
serramatutu, line 109: add `is INTEGER or STRING` to this doc.

metricflow/inference/rule/base.py
```diff
@@ -1,123 +1,11 @@
 from __future__ import annotations
 from abc import ABC, abstractmethod
+from typing import List
 
-from dataclasses import dataclass
```
serramatutu, line 4: These were moved to `metricflow/inference/models.py` since they're common classes to all inference steps and don't really belong to `rules` only.

metricflow/inference/solver/weighted_tree.py
```diff
@@ -0,0 +1,163 @@
...
+
+        A parent node's weight is the sum of all its children plus its own weight. Complimentary children get
+        excluded from the parent's sum.
+
+        root: the root to start assigning cumulative weights from.
+        weights: the output dictionary to assign the cumulative weights to
+        non_complimentary_weights: similar to weights, but the weight of each node excludes the weights
+            of all of its complimentary children
+        signals_by_type: a dictionary that maps signal type nodes to signals
+        """
```
tlento, line 74: I like to make the function signatures tell the callers what to expect even if they don't read the docstring, because most people won't - and here maybe you set input defaults to defaultdict(int) (or maybe not, depends on what you want from the typing):

```suggestion
    def _get_cumulative_weights_for_root(
        self,
        root: InferenceSignalNode,
        output_weights: Dict[InferenceSignalNode, int],
        output_parent_only_weights: Dict[InferenceSignalNode, int],
        signals_by_type: Dict[InferenceSignalNode, List[InferenceSignal]],
    ) -> Dict[InferenceSignalNode, int]:
        """Get a dict of cumulative weights, starting at `root`.

        A parent node's weight is the sum of all its children plus its own weight. Children tagged as only
        applying to their parent are excluded from their grand-parent's sum.

        root: the root to start assigning cumulative weights from.
        output_weights: the output dictionary to assign the cumulative weights to
        output_parent_only_weights: similar to output_weights, but the weight of each node excludes the weights
            of all of its parent-only children
        signals_by_type: a dictionary that maps signal type nodes to signals
        """
```

metricflow/inference/solver/weighted_tree.py
```diff
@@ -0,0 +1,163 @@
...
+            )
+
+        weights[root] = sum(self._weighter_function(signal.confidence) for signal in signals_by_type[root])
+        non_complimentary_weights[root] = sum(
+            self._weighter_function(signal.confidence)
+            for signal in signals_by_type[root]
+            if not signal.is_complimentary
+        )
+
+        weights[root] += sum(non_complimentary_weights[child] for child in root.children)
```
tlento, line 90: These aren't order dependent, right? If not, swapping like this (with renames as needed) would be easier to understand:
```suggestion
        weights[root] = sum(self._weighter_function(signal.confidence) for signal in signals_by_type[root])
        weights[root] += sum(non_complimentary_weights[child] for child in root.children)

        non_complimentary_weights[root] = sum(
            self._weighter_function(signal.confidence)
            for signal in signals_by_type[root]
            if not signal.is_complimentary
        )
```

metricflow/inference/solver/weighted_tree.py
```diff
@@ -0,0 +1,163 @@
...
+
+            if children_weight_total == 0:
+                break
+
+            next_node = None
+            for child in node.children:
+                if node_weights[child] / children_weight_total >= self._weight_percent_threshold:
+                    next_node = child
+                    reasons += reasons_by_type[child]
+                    break
```
tlento, line 145: Careful - first one in wins, but with `self._weight_percent_threshold` in [0.5, 1] you can end up with 2 equally good matches. See above for the simplest, most robust fix - I'm not recommending you do it here because your chosen weights can lead to weirdness on things like thresholds of 0.6 or 0.8 or 0.75.

metricflow/inference/solver/weighted_tree.py
```diff
@@ -0,0 +1,163 @@
...
+    ) -> None:
+        """Initialize the solver.
+
+        weight_percent_threshold: a number between 0.5 and 1. If a node's weight corresponds to a percentage
+            above this threshold with respect to its siblings' total weight sum, the solver will progress deeper
+            into the type tree, entering that node. If not, it stops at the parent.
+        weighter_function: a function that returns a weight given a confidence score. It will be used
+            to assign integer weights to each node in the type tree based on its input signals.
+        """
+        assert weight_percent_threshold >= 0.5 and weight_percent_threshold <= 1
```
tlento, line 48: This should solve the boundary issue below:
```suggestion
        weight_percent_threshold: a number in (0.5, 1]. If a node's weight corresponds to a percentage
            above this threshold with respect to its siblings' total weight sum, the solver will progress deeper
            into the type tree, entering that node. If not, it stops at the parent.
        weighter_function: a function that returns a weight given a confidence score. It will be used
            to assign integer weights to each node in the type tree based on its input signals.
        """
        assert weight_percent_threshold > 0.5 and weight_percent_threshold <= 1, f"weight_percent_threshold is {weight_percent_threshold}, but it must be > 0.5 and <= 1!"
```

tlento (commented):
OK! This is pretty slick. It's a large PR, and as such I've left a ton of inline comments. Thank you for splitting out your commits and providing solid commit messages on each, it helped a lot.

Pretty much everything I put in here is about documentation, readability, and code accessibility. I think there are design tradeoffs we might consider for later revisions, but nothing needs to be overhauled here. The core looks really strong, especially for a first cut at what is really a beta prototype. I can definitely see how we can extend this to other input types, or swap out the scoring model for something a bit more involved, or whatever else.

I'm leaving this up for added reviewer commentary for now. If you prefer to make fixes in follow-up PRs that's fine, just let me know. Otherwise, please avoid rebasing until they're pushed to the PR, because GitHub loses that reviewer placeholder info if you rebase. It's so annoying.

tlento (approved):
I glanced over my commentary and apart from one item about the boundary condition inside the weighted tree solver (the > 0.5 thing) this is ready to roll.

Thanks for all of your effort on this. It's really great stuff. I'm excited about the possibilities. :)

metricflow/inference/models.py
```diff
@@ -0,0 +1,153 @@
...
+
+        NOTE: for practical reasons, a node is considered a descendant of itself.
+        """
+        return other == self or other in self.ancestors
+
+
+# This is kinda horrible but there's no way of instancing the tree with type safety without
+# hardcoding it. Having some magic that dynamically assigns attributes could work, but then
+# we lose IDE autocompletion and static checking
+class _TreeNodes:  # noqa: D
```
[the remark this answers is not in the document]
serramatutu: 👍

metricflow/inference/renderer/config_file.py
```diff
@@ -0,0 +1,97 @@
...
+
+        if os.path.isfile(dir_path):
+            raise ValueError("ConfigFileRenderer `dir_path` is a file.")
+
+        self.dir_path = dir_path
+
+    def _get_filename_for_table(self, table: SqlTable) -> str:
+        return os.path.abspath(os.path.join(self.dir_path, f"{table.table_name}.yaml"))
```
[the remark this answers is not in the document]
serramatutu: Good point! I'll change it to `table.sql` so we get `db.schema.table.yaml` as a filename.

metricflow/inference/rule/defaults.py
```diff
@@ -1,77 +1,255 @@
...
+    """
+
+    type_node = InferenceSignalType.ID.PRIMARY
+    confidence = InferenceSignalConfidence.FOR_SURE
+    complimentary_signal = False
+    match_reason = "Column name matches `(table_name?)(_?)id`"
+
+    def match_column(self, props: ColumnProperties) -> bool:  # noqa: D
+        col_lower = props.column.column_name.lower()
+        table_lower = props.column.table_name.lower().rstrip("s")
```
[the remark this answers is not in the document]
serramatutu: Yeah, you have a great point. I think removing the `s` as a general rule might fail, but _at least_ we get the majority of cases (I hope).

PR #164, opened 2022-07-19
Add SQL Reserved Keyword Semantic Rule

QMalcolm:
This PR adds a rule which ensures when element `names` are used for sql generation (i.e. the element `expr` or `ref` aren't used) that they aren't SQL reserved keywords.

It should be known that this rule should not be a catch all. Data warehouse validations should catch these issues, but we currently don't have identifier and measure data warehouse validations (coming soon), and data warehouse validations are slow in comparison to semantic validation rules. Thus this rule is intended to catch words that are likely reserved keywords in all engines and to fail fast.

### DIFF
--- a/metricflow/model/validations/reserved_keywords.py
+++ b/metricflow/model/validations/reserved_keywords.py
@@ -1,0 +1,157 @@
+from typing import List
+from metricflow.instances import DataSourceElementReference
+
+
+from metricflow.model.objects.data_source import DataSource
+from metricflow.model.objects.user_configured_model import UserConfiguredModel
+from metricflow.model.validations.validator_helpers import (
+    DataSourceContext,
+    DataSourceElementContext,
+    DataSourceElementType,
+    FileContext,
+    ModelValidationRule,
+    ValidationError,
+    ValidationIssueType,
+    validate_safely,
+)
+
+# A non-exaustive tuple of reserved keywords
+# This list was created by running an intersection of keywords for redshift,
+# postgres, bigquery, and snowflake
+RESERVED_KEYWORDS = (
+    "AND",
+    "AS",
+    "CREATE",
+    "DISTINCT",
+    "FOR",
+    "FROM",
+    "FULL",
+    "HAVING",
+    "IN",
+    "INNER",
+    "INTO",
+    "IS",
+    "JOIN",
+    "LEFT",
+    "LIKE",
+    "NATURAL",
+    "NOT",
+    "NULL",
+    "ON",
+    "OR",
+    "RIGHT",
+    "SELECT",
+    "UNION",
+    "USING",
+    "WHERE",
+    "WITH",
+)
+
+
+class ReservedKeywordsRule(ModelValidationRule):
+    """Check that any element that ends up being selected by name (instead of expr) isn't a commonly reserved keyword.
+
+    Note: This rule DOES NOT catch all keywords. That is because keywords are
+    engine specific, and semantic validations are not engine specific. I.e. if
+    you change your underlying data warehouse engine, semantic validations
+    should still pass, but your data warehouse validations might fail. However,
+    data warehouse validations are slow in comparison to semantic validation
+    rules. Thus this rule is intended to catch words that are reserved keywords
+    in all supported engines and to fail fast. E.g., `USER` is a reserved keyword
+    in Redshift but not in all other supported engines. Therefore if one is
+    using Redshift and sets a dimension name to `user`, the config would pass
+    this rule, but would then fail Data Warehouse Validations.
+    """
+
+    @staticmethod
+    @validate_safely(whats_being_done="checking that data source sub element names aren't reserved sql keywords")
+    def _validate_data_source_sub_elements(data_source: DataSource) -> List[ValidationIssueType]:
+        issues: List[ValidationIssueType] = []
+
+        for dimension in data_source.dimensions:
+            if dimension.name.upper() in RESERVED_KEYWORDS:
+                issues.append(
+                    ValidationError(
+                        context=DataSourceElementContext(
+                            file_context=FileContext.from_metadata(data_source.metadata),
+                            data_source_element=DataSourceElementReference(
+                                data_source_name=data_source.name, element_name=dimension.name
+                            ),
+                            element_type=DataSourceElementType.DIMENSION,
+                        ),
+                        message=f"'{dimension.name}' is an SQL reserved keyword, and thus cannot be used as a dimension 'name'.",
+                    )
+                )
+
+        for identifier in data_source.identifiers:
+            if identifier.is_composite:
+                msg = "'{name}' is an SQL reserved keyword, and thus cannot be used as a sub-identifier 'name'"
+                names = [sub_ident.name for sub_ident in identifier.identifiers if sub_ident.name is not None]
+            else:
+                msg = "'{name}' is an SQL reserved keyword, and thus cannot be used as an identifier 'name'"
+                names = [identifier.name]
+
+            for name in names:
+                if name.upper() in RESERVED_KEYWORDS:
+                    issues.append(
+                        ValidationError(
+                            context=DataSourceElementContext(
+                                file_context=FileContext.from_metadata(data_source.metadata),
+                                data_source_element=DataSourceElementReference(
+                                    data_source_name=data_source.name, element_name=identifier.name
+                                ),
+                                element_type=DataSourceElementType.IDENTIFIER,
+                            ),
+                            message=msg.format(name=name),
+                        )
+                    )
+
+        for measure in data_source.measures:
+            if measure.name.upper() in RESERVED_KEYWORDS:
+                issues.append(
+                    ValidationError(
+                        context=DataSourceElementContext(
+                            file_context=FileContext.from_metadata(data_source.metadata),
+                            data_source_element=DataSourceElementReference(
+                                data_source_name=data_source.name, element_name=measure.name
+                            ),
+                            element_type=DataSourceElementType.MEASURE,
+                        ),
+                        message=f"'{measure.name}' is an SQL reserved keyword, and thus cannot be used as an measure 'name'.",
+                    )
+                )
+
+        return issues
+
+    @classmethod
+    @validate_safely(whats_being_done="checking that data_source sql_tables are not sql reserved keywords")
+    def _validate_data_sources(cls, model: UserConfiguredModel) -> List[ValidationIssueType]:
+        """Checks names of objects that are not nested."""
+        issues: List[ValidationIssueType] = []
+        set_keywords = set(RESERVED_KEYWORDS)
+
+        for data_source in model.data_sources:
+            if data_source.sql_table is not None:
+                set_sql_table_path_parts = set([part.upper() for part in data_source.sql_table.split(".")])
+                keyword_intersection = set_keywords.intersection(set_sql_table_path_parts)
+
+                if len(keyword_intersection) > 0:
+                    issues.append(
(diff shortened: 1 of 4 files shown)


metricflow/model/validations/reserved_keywords.py
```diff
@@ -0,0 +1,174 @@
...
+                                        data_source_name=data_source.name, element_name=identifier.name
+                                    ),
+                                    element_type=DataSourceElementType.IDENTIFIER,
+                                ),
+                                message=msg.format(name=name),
+                            )
+                        )
+
+        for measure in data_source.measures:
+            if measure.expr is None and measure.name.upper() in RESERVED_KEYWORDS:
```
tlento, line 127:
```suggestion
            if measure.name.upper() in RESERVED_KEYWORDS:
```

tlento (commented):
We clearly need this but I also wish we didn't, like it seems like we should be escaping the parts of the model we can control.

Anyway, some issues that spring to mind:

1. How are we going to manage the rollout? Do we know if this is going to break any existing models? Here on the OSS side we'll mostly need to just announce things to people, I guess, although that's fairly unsatisfying. I suppose folks can always hold off on upgrading until it's convenient for them.
2. What are your thoughts on engine-specific stuff like `user`?
3. There are some edge cases called out inline, we should figure those out

metricflow/model/validations/reserved_keywords.py
```diff
@@ -0,0 +1,174 @@
...
+    DataSourceElementType,
+    FileContext,
+    ModelValidationRule,
+    ValidationError,
+    ValidationIssueType,
+    validate_safely,
+)
+
+# A non-exaustive tuple of reserved sql keywords
+RESERVED_KEYWORDS = (
```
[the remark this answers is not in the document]
QMalcolm: I forget exactly which source I pulled this list from. It was one of: [PSQL](https://www.postgresql.org/docs/current/sql-keywords-appendix.html), [IBM](https://www.ibm.com/docs/en/psfa/7.2.1?topic=keywords-sql-common-reserved-words), or [SQL Server](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/reserved-keywords-transact-sql?view=sql-server-ver16), but I am unsure which. Additionally, I trimmed down the list to try and only have words that would be likely reserved in all engines. And I think I must have accidentally removed `USER` 🙃 I'm trying to _not_ make this rule engine dependent. We're also adding identifier and measure data warehouse validations in the next two weeks which will also catch these errors. The reason why we are adding this rule instead of just relying on data warehouse validations is because data warehouse validations are slow. So we're using this rule to catch the most common words to fail the model quickly.

metricflow/model/validations/reserved_keywords.py
```diff
@@ -0,0 +1,174 @@
...
+class ReservedKeywordsRule(ModelValidationRule):
+    """Check that any element that ends up being selected by name (instead of expr) isn't a reserved sql keyword."""
+
+    @staticmethod
+    @validate_safely(whats_being_done="checking that data source sub element names aren't reserved sql keywords")
+    def _validate_data_source_sub_elements(data_source: DataSource) -> List[ValidationIssueType]:
+        issues: List[ValidationIssueType] = []
+
+        for dimension in data_source.dimensions:
+            if dimension.expr is None and dimension.name.upper() in RESERVED_KEYWORDS:
```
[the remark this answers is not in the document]
QMalcolm: Ah I didn't realize we did aliasing to the element name. Okay I'll do less escapes and just always check the name.

metricflow/model/validations/reserved_keywords.py
```diff
@@ -0,0 +1,174 @@
...
+            if identifier.expr is None:
+                msg = "'{name}' is an SQL reserved keyword, and thus cannot be used as an identifier 'name'"
+                names = [identifier.name]
+                if identifier.is_composite:
+                    msg = "'{name}' is an SQL reserved keyword, and thus cannot be used as a sub-identifier 'name'"
+                    names = [
+                        sub_ident.name
+                        for sub_ident in identifier.identifiers
+                        if sub_ident.expr is None and sub_ident.name
+                    ]
```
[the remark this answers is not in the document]
QMalcolm: We don't need to account for `ref`. My understanding is that `ref`s are references to other identifiers. Those other identifiers are checked separately with this rule, and I don't think we should create multiple validation errors for what is really one issue.

metricflow/model/validations/reserved_keywords.py
```diff
@@ -0,0 +1,174 @@
...
+    @classmethod
+    @validate_safely(whats_being_done="checking that data_source sql_tables are not sql reserved keywords")
+    def _validate_data_sources(cls, model: UserConfiguredModel) -> List[ValidationIssueType]:
+        """Checks names of objects that are not nested."""
+        issues: List[ValidationIssueType] = []
+        set_keywords = set(RESERVED_KEYWORDS)
+
+        for data_source in model.data_sources:
+            if data_source.sql_table is not None:
+                set_sql_table_path_parts = set([part.upper() for part in data_source.sql_table.split(".")])
```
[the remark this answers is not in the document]
QMalcolm: So I looked into SqlTable. It does do the parsing, but then there isn't a straightforward way to get it back as a list. Having all the parsing logic in one spot though does make sense. I'll just add a helper to SqlTable to get the list. I think specifically my hold up was that the `db` property of SqlTable is optional, and I didn't want to have `None` in the list and I also didn't want the logic to optionally include `db` in the list to live in the rule.

metricflow/model/validations/reserved_keywords.py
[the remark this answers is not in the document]
QMalcolm: Alright so I updated the list by taking an intersection of PSQL, BigQuery, Redshift, and Snowflake keywords. It results in a smaller list. However, this also means we won't be blocking words when they aren't reserved for a specific engine (i.e. no false failures)

QMalcolm:
To circle back to the three specific questions you raised

**1. How are we going to manage the rollout? Do we know if this is going to break any existing models? Here on the OSS side we'll mostly need to just announce things to people, I guess, although that's fairly unsatisfying. I suppose folks can always hold off on upgrading until it's convenient for them.**

Now that the list of reserved keywords has been slimmed down to the intersection of postgres/redshift/bigquery/snowflake keywords this shouldn't break anyones current config. If it did, it means that their config was already broken. Someone could be running with a bad config, i.e. maybe a specific dimension/identifier/measure isn't being used. The alternative is that we roll this out as a `FUTURE-ERROR`.

**2. What are your thoughts on engine-specific stuff like user?**
Engine specific stuff should be left to data warehouse validations. Semantic validations are unaware of the engine, and I think it'd be good to keep it that way. If we start introducing engine specific semantic validations, I think things will get muddy and tangled quickly.

**3. There are some edge cases called out inline, we should figure those out**
I believe all the edge cases are false successes (i.e. they pass when they shouldn't). I think this is okay, in that if something doesn't get caught here it will be caught by data warehouse validations. The goal of this rule isn't to catch everything, it's to catch the common reserved words.

metricflow/model/validations/reserved_keywords.py
```diff
@@ -0,0 +1,162 @@
...
+class ReservedKeywordsRule(ModelValidationRule):
+    """Check that any element that ends up being selected by name (instead of expr) isn't a commonly reserved keyword.
+
+    Note: This rule DOES NOT catch all keywords. That is because keywords are
+    engine specific, and semantic validations are not engine specific. I.e. if
+    you change your underlying data warehouse engine, semantic validations
+    should still pass, but your data warehouse validations might fail. However,
+    data warehouse validations are slow in comparison to semantic validation
+    rules. Thus this rule is intended to catch words that are reserved keywords
+    in all supported engines and to fail fast. I.e. `USER` is a reserved keyword
```
tlento, line 61: e.g. means "for example" and i.e. means "that is"

```suggestion
    in all supported engines and to fail fast. E.g., `USER` is a reserved keyword
```

My inner pedant is satisfied. Thanks for your patience.

metricflow/model/validations/reserved_keywords.py
```diff
@@ -0,0 +1,162 @@
...
+                        ),
+                        message=f"'{dimension.name}' is an SQL reserved keyword, and thus cannot be used as a dimension 'name'.",
+                    )
+                )
+
+        for identifier in data_source.identifiers:
+            if identifier.is_composite:
+                msg = "'{name}' is an SQL reserved keyword, and thus cannot be used as a sub-identifier 'name'"
+                names = [
+                    sub_ident.name for sub_ident in identifier.identifiers if sub_ident.expr is None and sub_ident.name
```
tlento, line 91:
```suggestion
                    sub_ident.name for sub_ident in identifier.identifiers if sub_ident.name
```

metricflow/model/validations/reserved_keywords.py
```diff
@@ -0,0 +1,162 @@
...
+    def _validate_data_sources(cls, model: UserConfiguredModel) -> List[ValidationIssueType]:
+        """Checks names of objects that are not nested."""
+        issues: List[ValidationIssueType] = []
+        set_keywords = set(RESERVED_KEYWORDS)
+
+        for data_source in model.data_sources:
+            if data_source.sql_table is not None:
+                set_sql_table_path_parts = set(
+                    [part.upper() for part in SqlTable.from_string(data_source.sql_table).parts_tuple]
+                )
```
tlento, line 140: Seems like you should be able to just do this:

```suggestion
                set_sql_table_path_parts = set(
                    SqlTable.from_string(data_source.sql_table).parts_tuple
                )
```

metricflow/test/model/validations/test_reserved_keywords.py
```diff
@@ -0,0 +1,84 @@
...
+    identifier.identifiers = [
+        CompositeSubIdentifier(name=random_keyword()),
+        CompositeSubIdentifier(name=random_keyword()),
+        CompositeSubIdentifier(expr="SELECT TRUE AS col1"),  # shouldn't error
+    ]
+
+    issues = ReservedKeywordsRule.validate_model(model)
+    assert len(issues) == 2
+    assert issues[0].level == ValidationIssueLevel.ERROR
+    assert issues[1].level == ValidationIssueLevel.ERROR
```
tlento, line 84: I'd prefer to have this in its own test case. It's a little confusing as written because the mutation puts it back in a good state.

tlento (approved):
Ok, this all makes sense, thanks for the various improvements!

I have a few minor nits annotated in line and one big rant about SQL keyword escaping, but I think this is ready for approval. Merge when ready!

metricflow/model/validations/reserved_keywords.py
tlento: Actually this one is kind of important.

metricflow/model/validations/reserved_keywords.py
QMalcolm: So `.parts_tuple` doesn't guarantee that the parts are all in upper case. And we're storing the reserved keywords all in upper case. A quick check
```Python
>>> parts = ['select', 'where', 'from']
>>> reserved_keywords = set(['SELECT', 'WHERE', 'FROM'])
>>> lower = set(parts)
>>> upper = set([part.upper() for part in parts])
>>> reserved_keywords.intersection(lower)
set()
>>> reserved_keywords.intersection(upper)
{'SELECT', 'WHERE', 'FROM'}
```
Thus, ensuring things are in upper case is necessary

QMalcolm:
Pushing a cleaned up version of the branch (Fixup commits have been fixup'd into the appropriate commits)

Issue #9, opened 2022-04-06
Simplify Time Constraint SQL


tlento:
**Describe the bug**
Time Constraints are rendered in SQL as follows:

```
WHERE (
  subq_166.ds >= CAST('2022-02-26' AS TIMESTAMP)
) AND (
  subq_166.ds <= CAST('2022-03-28' AS TIMESTAMP)
)
```

Change them to:
```
WHERE
  subq_166.ds BETWEEN CAST('2022-02-26' AS TIMESTAMP) AND CAST('2022-03-28' AS TIMESTAMP)
```

patclrk:
This project is awesome! Looking to contribute by taking a crack at this issue. 

I've got the SQL working but I broke many of the integration tests. It looks like these tests are pulling from `snapshots`. Changing the `.sql` files for each SqlClient seems straightforward but I'm not sure about the `.xml` files. Are the `.xml` files generated somewhere?

Cheers!

tlento:
Hi @patclrk , thanks for diving in!

Those .xml files are generated (as are the .sql files). In theory you should be able to regenerate all snapshot files by running this command:

`poetry run pytest --overwrite-snapshots metricflow/tests`

Then you can commit those changes and add them to your PR. I typically have all plans change in the same commit with the thing that modified them so I don't have broken tests in the middle of the PR, but sometimes that gets too crazy.

I've also taken the liberty of assigning this issue to you, but feel free to put it back up for grabs if you find you don't have the time to keep going.

tlento:
Also you'll need to run the tests once per SQL client type (SQLite, Redshift, Snowflake, BigQuery). If you don't have access to all of those let us know and we'll work out a way to provide you with a patch for the plan files.

PR #168, opened 2022-07-21
Add SqlBetweenExpr and simplify time_range_comparison_expr rendering logic

WilliamDee:
## Context
Github issue: https://github.com/transform-data/metricflow/issues/9
Rendering time range comparisons can be simplified down from 
```sql
... WHERE start_time <= time_dim AND time_dim <= end_time ...
```
to
```sql
... WHERE time_dim BETWEEN start_time AND end_time ...
```

## Changes
- Adds `SqlBetweenExpression` to render BETWEEN clauses
- Replaced time range comparison expr to use BETWEEN rather than equality comparisons.
- Updated snapshots for redshift/snowflake/duckdb/bigquery/postgres


metricflow/sql/sql_exprs.py
```diff
@@ -1144,3 +1149,70 @@ def matches(self, other: SqlExpressionNode) -> bool:  # noqa: D
...
+    def description(self) -> str:  # noqa: D
+        return "BETWEEN operator"
+
+    @property
+    def column_arg(self) -> SqlExpressionNode:  # noqa: D
+        return self._column_arg
+
+    @property
+    def left_value(self) -> SqlExpressionNode:  # noqa: D
+        assert isinstance(
```
plypaul, line 1186: Actually, other expressions should be possible. e.g. `WHERE ds BETWEEN some_date_column AND some_other_date_column`

metricflow/sql/render/expr_renderer.py
```diff
@@ -255,3 +256,13 @@ def visit_ratio_computation_expr(self, node: SqlRatioComputationExpression) -> S
...
         )
+
+    def visit_between_expr(self, node: SqlBetweenExpression) -> SqlExpressionRenderResult:  # noqa: D
+        rendered_column_arg = self.render_sql_expr(node.column_arg)
+        rendered_left_value = self.render_sql_expr(node.left_value)
+        rendered_right_value = self.render_sql_expr(node.right_value)
+
+        return SqlExpressionRenderResult(
+            sql=f"{rendered_column_arg.sql} BETWEEN {rendered_left_value.sql} AND {rendered_right_value.sql}",
+            execution_parameters=SqlBindParameters(),
```
plypaul, line 267: You'll want to combine the bind parameters from the arguments and put them here.

metricflow/sql/sql_exprs.py
WilliamDee: removed this assert!

metricflow/sql/render/expr_renderer.py
WilliamDee: added execution parameters

metricflow/sql/sql_exprs.py
```diff
@@ -1144,3 +1149,64 @@ def matches(self, other: SqlExpressionNode) -> bool:  # noqa: D
         if not isinstance(other, SqlRatioComputationExpression):
             return False
         return self._parents_match(other)
+
+
+class SqlBetweenExpression(SqlExpressionNode):
+    """A BETWEEN clause like `column BETWEEN val1 AND val2`"""
+
+    def __init__(  # noqa: D
+        self, column_arg: SqlExpressionNode, left_value: SqlExpressionNode, right_value: SqlExpressionNode
```
tlento, line 1158: Shall we call these `min_value` and `max_value` instead of `left_value` and `right_value`?

I guess at this level we don't necessarily know if someone has done something silly and entered a config expr which will return FALSE regardless of input (e.g., `num_entries BETWEEN 10 AND 2`), so maybe `left` and `right` are less confusing overall.

tlento (approved):
I have convinced myself that "left_value" and "right_value" are good names, but if you think "min_value" and "max_value" or "expected_min_value" and "expected_max_value" or something similar would be better, please do feel free to make the change and rerun the tests before merging.

metricflow/sql/sql_exprs.py
WilliamDee: I did some searching, theres a bunch of variety. Examples,

```
- start_expr | end_expr
- start_value | end_value
- begin_expr | end_expr
```

I'm not the best at naming, but either of those feels fine to me (leaning towards `start_expr/end_expr`). Lemme know your thoughts

tlento: start_expr/end_expr seems pretty reasonable to me!

Issue #10, opened 2022-04-06
ALL TIME CONSTRAINT should not be rendered as a WHERE clause


tlento:
**Describe the bug**
When a query does not specify a time constraint we define a TimeRangeConstraint using ALL_TIME_BEGIN and ALL_TIME_END, and then construct a WHERE clause with this constraint. So we end up with something like:

```
SELECT <columns>
FROM <data_source>
WHERE primary_time >= CAST('1700-01-01' AS TIMESTAMP) AND primary_time <= CAST('2300-12-31' AS TIMESTAMP)
```

This is silly. We should just not have any primary_time filters in the rendered SQL.

**Steps To Reproduce**
Generate the SQL query output for any query with no time constraint applied on the primary time dimension.

tlento:
Note - to really make this a good first issue we need better steps to reproduce, but I don't have that flow worked out yet.

PR #169, opened 2022-07-21
do not render an ALL TIME time constraint WHERE clause

WilliamDee:
## Context
GIthub Issue: https://github.com/transform-data/metricflow/issues/10

Rendering the ALL TIME time range constraint is not important and clogs the output plans while providing not much input value.

## Changes
- Do not render a WHERE TimeConstraint if the time constraint ends up being the ALL TIME time range

### DIFF
--- a/metricflow/query/query_parser.py
+++ b/metricflow/query/query_parser.py
@@ -318,7 +318,12 @@ class MetricFlowQueryParser:
                 f"{time_constraint_start.isoformat()}"
             )
 
-        time_constraint = TimeRangeConstraint(start_time=time_constraint_start, end_time=time_constraint_end)
+        time_constraint: Optional[TimeRangeConstraint] = TimeRangeConstraint(
+            start_time=time_constraint_start, end_time=time_constraint_end
+        )
+        if time_constraint == TimeRangeConstraint.all_time():
+            # If the time constraint is all time, just ignore and not render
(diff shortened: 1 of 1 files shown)


plypaul:
Looks like the test snapshots need an update.

metricflow/query/query_parser.py
```diff
@@ -318,7 +318,12 @@ def _parse_and_validate_query(
                 f"{time_constraint_start.isoformat()}"
             )
 
-        time_constraint = TimeRangeConstraint(start_time=time_constraint_start, end_time=time_constraint_end)
+        time_constraint: Optional[TimeRangeConstraint] = TimeRangeConstraint(
+            start_time=time_constraint_start, end_time=time_constraint_end
+        )
+        if time_constraint == TimeRangeConstraint.all_time():
```
plypaul, line 324: Might be easier to understand if you update the series of conditionals above.

plypaul (approved):
LGTM but left a comment.

metricflow/query/query_parser.py
WilliamDee: I think we can simplify the above conditionals by creating a `TimeRangeConstraint` with whatever is provided then just do an intersection with ALL TIME, but we would lose the warnings messages.

PR #170, opened 2022-07-22
Disallow metric_time as an element name

tlento:
With the advent of the metric_time query construct it has effectively
become a reserved keyword for Metricflow, as we cannot resolve queries
that are requesting `metric_time` if there is some other time dimension
also named `metric_time` in the user model.

This commit adds a general MetricFlowReservedKeywords enumeration and
adds a validation check to assert that none of the checked names
match a value in the enum.

### DIFF
--- a/metricflow/model/validations/unique_valid_name.py
+++ b/metricflow/model/validations/unique_valid_name.py
@@ -36,7 +36,7 @@ from metricflow.time.time_granularity import TimeGranularity
 class MetricFlowReservedKeywords(enum.Enum):
     """Enumeration of reserved keywords with helper for accessing the reason they are reserved"""
 
-    METRIC_TIME = "METRIC_TIME"
+    METRIC_TIME = "metric_time"
 
     @staticmethod
     def get_reserved_reason(keyword: MetricFlowReservedKeywords) -> str:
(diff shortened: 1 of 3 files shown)


metricflow/model/validations/unique_valid_name.py
```diff
@@ -24,10 +27,29 @@
...
+from metricflow.object_utils import assert_values_exhausted
 from metricflow.specs import ElementReference
 from metricflow.time.time_granularity import TimeGranularity
 
 
+@enum.unique
+class MetricFlowReservedKeywords(enum.Enum):
+    """Enumeration of reserved keywords with helper for accessing the reason they are reserved"""
+
+    METRIC_TIME = "METRIC_TIME"
```
plypaul, line 39: This is also defined in `DataSet.metric_time_dimension_name()`

metricflow/model/validations/unique_valid_name.py
```diff
@@ -60,6 +82,14 @@ def check_valid_name(  # noqa: D
                     f"({TimeGranularity.list_names()})",
                 )
             )
+        if name.upper() in {reserved_name.value for reserved_name in MetricFlowReservedKeywords}:
```
plypaul, line 85: We've been going with lower case as per `LowerCaseNamesRule`

plypaul (approved):
LGTM but left some comments

metricflow/model/validations/unique_valid_name.py
tlento: Will change. We use upper case in the SQL reserved keywords checks, so I followed that because this is a reserved keyword, but now that I think of it we always reference it lower case ourselves since it's an element name.

metricflow/model/validations/unique_valid_name.py
tlento: Not anymore!

PR #173, opened 2022-07-22
Refactor data warehouse validations to use 'sql_query_plan_renderer'

QMalcolm:
This change is two fold. First generating queries for measures and identifiers is fair bit harder than the queries we'd been generating for data_sources and dimensions, which was going to increase the complexity of the task builder greatly. So we wanted to begin using the tools the rest of MetricFlow uses for generating queries, no need to reinvent the wheel. Additionally, by using the `sql_query_plan_renderer` we ensure validation queries are similar to what MetricFlow would render using the same elements.

**TLDR**: This makes future work on data warehouse validations easier and makes validations consistent with the rest of MetricFlow.

### DIFF
--- a/metricflow/model/data_warehouse_model_validator.py
+++ b/metricflow/model/data_warehouse_model_validator.py
@@ -35,7 +35,7 @@ from metricflow.plan_conversion.column_resolver import DefaultColumnAssociationR
 from metricflow.plan_conversion.dataflow_to_sql import DataflowToSqlQueryPlanConverter
 from metricflow.plan_conversion.time_spine import TimeSpineSource
 from metricflow.protocols.sql_client import SqlClient
-from metricflow.specs import DimensionSpec
+from metricflow.specs import DimensionSpec, LinkableInstanceSpec
 from metricflow.sql.sql_bind_parameters import SqlBindParameters
 
 
@@ -79,6 +79,11 @@ class DataWarehouseValidationTask:
 
 class DataWarehouseTaskBuilder:
     """Task builder for standard data warehouse validation tasks"""
+
+    @staticmethod
+    def _remove_identifier_link_specs(specs: Tuple[LinkableInstanceSpec, ...]) -> Tuple[LinkableInstanceSpec, ...]:
+        """For the purposes of data warehouse validation, specs with identifier_links are unnecesary"""
+        return tuple(spec for spec in specs if not spec.identifier_links)
 
     @staticmethod
     def _data_source_node(render_tools: QueryRenderingTools, data_source: DataSource) -> BaseOutput[DataSourceDataSet]:
@@ -176,7 +181,7 @@ class DataWarehouseTaskBuilder:
 
             data_source_sub_tasks: List[DataWarehouseValidationTask] = []
             dataset = render_tools.converter.create_sql_source_data_set(data_source)
-            data_source_specs = (
+            data_source_specs = DataWarehouseTaskBuilder._remove_identifier_link_specs(
                 dataset.instance_set.spec_set.dimension_specs + dataset.instance_set.spec_set.time_dimension_specs
             )
             for spec in data_source_specs:
--- a/metricflow/model/data_warehouse_model_validator.py
+++ b/metricflow/model/data_warehouse_model_validator.py
@@ -1,17 +1,24 @@
 from __future__ import annotations
+from copy import deepcopy
 
 from dataclasses import dataclass, field
 from functools import partial
 from math import floor
 from time import perf_counter
 from typing import Callable, List, Optional, Tuple
+from metricflow.dataflow.builder.source_node import SourceNodeBuilder
+from metricflow.dataflow.dataflow_plan import BaseOutput, FilterElementsNode
+from metricflow.dataset.convert_data_source import DataSourceToDataSetConverter
+from metricflow.dataset.data_source_adapter import DataSourceDataSet
 
 from metricflow.dataset.dataset import DataSet
 from metricflow.engine.metricflow_engine import MetricFlowEngine, MetricFlowExplainResult, MetricFlowQueryRequest
 from metricflow.instances import DataSourceElementReference, DataSourceReference, MetricModelReference
 from metricflow.model.objects.data_source import DataSource
+from metricflow.model.objects.elements.dimension import Dimension, DimensionType
 from metricflow.model.objects.metric import Metric
 from metricflow.model.objects.user_configured_model import UserConfiguredModel
+from metricflow.model.semantic_model import SemanticModel
 from metricflow.model.validations.validator_helpers import (
     DataSourceContext,
     DataSourceElementContext,
@@ -24,8 +31,40 @@ from metricflow.model.validations.validator_helpers import (
     ValidationIssue,
     ValidationWarning,
 )
+from metricflow.plan_conversion.column_resolver import DefaultColumnAssociationResolver
+from metricflow.plan_conversion.dataflow_to_sql import DataflowToSqlQueryPlanConverter
+from metricflow.plan_conversion.time_spine import TimeSpineSource
 from metricflow.protocols.sql_client import SqlClient
+from metricflow.specs import DimensionSpec
 from metricflow.sql.sql_bind_parameters import SqlBindParameters
+
+
+@dataclass
+class QueryRenderingTools:
+    """General tools that data warehosue validations use for rendering the validation queries
+
+    This is necessary for the validation steps that generate raw partial queries against the individual data sources
+    (e.g., selecting a single dimension column).
+    """
+
+    semantic_model: SemanticModel
+    source_node_builder: SourceNodeBuilder
+    converter: DataSourceToDataSetConverter
+    time_spine_source: TimeSpineSource
+    plan_converter: DataflowToSqlQueryPlanConverter
+
+    def __init__(self, model: UserConfiguredModel, system_schema: str) -> None:  # noqa: D
+        self.semantic_model = SemanticModel(user_configured_model=model)
+        self.source_node_builder = SourceNodeBuilder(semantic_model=self.semantic_model)
+        self.time_spine_source = TimeSpineSource(schema_name=system_schema)
+        self.converter = DataSourceToDataSetConverter(
+            column_association_resolver=DefaultColumnAssociationResolver(semantic_model=self.semantic_model)
+        )
+        self.plan_converter = DataflowToSqlQueryPlanConverter(
+            column_association_resolver=DefaultColumnAssociationResolver(self.semantic_model),
+            semantic_model=self.semantic_model,
+            time_spine_source=self.time_spine_source,
+        )
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


metricflow/model/data_warehouse_model_validator.py
```diff
@@ -41,86 +76,82 @@ class DataWarehouseValidationTask:
...
+            ]
+
+        render_tools = QueryRenderingTools(model=model, system_schema=system_schema)
+
         tasks: List[DataWarehouseValidationTask] = []
-        for index, data_source in enumerate(model.data_sources):
+        for data_source in model.data_sources:
+            source_node = cls._data_source_node(render_tools=render_tools, data_source=data_source)
+            spec = DimensionSpec.from_name(name=f"validation_dim_for_{data_source.name}")
+            filter_elements_node = FilterElementsNode(parent_node=source_node, include_specs=[spec])
```
tlento, line 130: Ok, so this is doing `SELECT 1 FROM {data_source.sql_query or data_source.sql_table}` for every data source, which essentially says "yes, we can execute against that rendered inner query".

metricflow/model/data_warehouse_model_validator.py
```diff
@@ -130,46 +161,60 @@ def gen_dimension_tasks(cls, model: UserConfiguredModel) -> List[DataWarehouseVa
...
-                            data_source_element=DataSourceElementReference(
-                                data_source_name=data_source.name, element_name=dimension.name
-                            ),
-                            element_type=DataSourceElementType.DIMENSION,
+                            data_source=DataSourceReference(data_source_name=data_source.name),
                         ),
-                        error_message=f"Unable to query `{dim_to_query}` in data warehouse for dimension "
-                        f"`{dimension.name}` on data source `{data_source.name}`.",
+                        error_message=f"Failed to query dimensions in data warehouse for data source `{data_source.name}`",
+                        on_fail_subtasks=data_source_sub_tasks,
```
tlento, line 215: I realize this is essentially a port, so it's fine here, but why do we need this, exactly? It's an extra query that simply asserts we can select all of the dimensions at once in addition to selecting them one at a time.

Is this to make sure someone hasn't somehow created a duplicate column reference? Wouldn't our regular model validation catch that?

metricflow/model/data_warehouse_model_validator.py
```diff
@@ -41,86 +76,82 @@ class DataWarehouseValidationTask:
...
-        wrapped_cols = [DataWarehouseTaskBuilder._wrap_col(col, index) for index, col in enumerate(columns)]
-        columns_select = ", ".join(wrapped_cols)
-
-        query = DataWarehouseTaskBuilder.QUERY_TEMPLATE.format(
-            data_source=data_source_str, columns_select=columns_select, unique_id=id
+    def renderize(
+        sql_client: SqlClient, plan_converter: DataflowToSqlQueryPlanConverter, plan_id: str, nodes: FilterElementsNode
+    ):
+        """Generates a sql query plan and returns the rendered sql and execution_parameters"""
+        sql_plan = plan_converter.convert_to_sql_query_plan(
```
tlento, line 99: Looking at the old class and what we're doing in here it seems like we're running a lot of fairly simple queries. Shall we add a `limit 1` here to speed things up on the final stage? For the ones that just grab a column this will be a lot faster, and for anything aggregating a metric it'll have to compute the full set and then truncate the result.

Check out the OrderByLimitNode for this - you can pass an empty list for order by specs and then it'll just process the limit. You might need to play around with the node ordering to avoid selecting other columns in the inner query (also you need full optimization, otherwise we will select everything in the data source by name).

This is probably best left for a separate PR, since the existing queries do not use the LIMIT statement and this is a refactor, but it's a thought.

metricflow/test/model/validations/test_data_warehouse_tasks.py
```diff
@@ -152,26 +126,20 @@ def test_validate_dimensions(  # noqa: D
     issues = dw_validator.validate_dimensions(model)
     # One isssue is created for the short circuit query failure, and another is
     # created for the sub task checking the specific dimension
-    print("\nModelValidationResults", issues)
```
tlento, line 155: LOL.

metricflow/test/model/validations/test_data_warehouse_tasks.py
```diff
@@ -152,26 +126,20 @@ def test_validate_dimensions(  # noqa: D
...
     assert len(issues.all_issues) == 2
-    assert (
-        f"Failed to query dimensions in data warehouse for data source `{model.data_sources[0].name}`"
-        in issues.all_issues[0].message
-    )
-    assert (
-        f"Unable to query `doesnt_exist` in data warehouse for dimension `doesnt_exist` on data source `{model.data_sources[0].name}`"
-        in issues.all_issues[1].message
-    )
```
tlento, line 164: The issues all generate the same basic message anyway regardless of the failure cause, right? If not, we could convert these to more generic error message assertions, or scan the full set for each failing data source, but if so removing this makes more sense.

metricflow/model/data_warehouse_model_validator.py
```diff
@@ -24,10 +31,38 @@
...
+from metricflow.plan_conversion.dataflow_to_sql import DataflowToSqlQueryPlanConverter
+from metricflow.plan_conversion.time_spine import TimeSpineSource
 from metricflow.protocols.sql_client import SqlClient
+from metricflow.specs import DimensionSpec, TimeDimensionSpec
 from metricflow.sql.sql_bind_parameters import SqlBindParameters
 
 
+@dataclass
+class QueryRenderingTools:
+    """General tools that data warehosue validations use for rendering the validation queries"""
```
tlento, line 44: Let's expand on this a bit, because it looks like a mini MetricFlowEngine container.

```suggestion
    """General tools that data warehosue validations use for rendering the validation queries
    
    This is necessary for the validation steps that generate raw partial queries against the individual data sources
    (e.g., selecting a single dimension column).
    """
```

metricflow/model/data_warehouse_model_validator.py
```diff
@@ -41,86 +76,82 @@ class DataWarehouseValidationTask:
...
+                Dimension(name=f"validation_dim_for_{data_source.name}", type=DimensionType.CATEGORICAL, expr="1")
+            ]
+
+        render_tools = QueryRenderingTools(model=model, system_schema=system_schema)
+
         tasks: List[DataWarehouseValidationTask] = []
-        for index, data_source in enumerate(model.data_sources):
+        for data_source in model.data_sources:
+            source_node = cls._data_source_node(render_tools=render_tools, data_source=data_source)
+            spec = DimensionSpec.from_name(name=f"validation_dim_for_{data_source.name}")
```
tlento, line 129: In this case I think this is fine, but we might want to use the same process as recommended below and extract this spec by name instead.

metricflow/model/data_warehouse_model_validator.py
```diff
@@ -130,46 +161,60 @@ def gen_dimension_tasks(cls, model: UserConfiguredModel) -> List[DataWarehouseVa
...
-            # generate the subtasks
-            data_source_tasks: List[DataWarehouseValidationTask] = []
-            data_source_columns: List[str] = []
-            for dimension in data_source.dimensions:
-                dim_to_query = dimension.expr if dimension.expr is not None else dimension.name
-                data_source_columns.append(dim_to_query)
-
-                data_source_tasks.append(
+        for data_source in model.data_sources:
+            if len(data_source.dimensions) > 0:
```
tlento, line 168: I don't think we need this. Crafting the source node is pretty cheap, and everything else will get skipped by the for loop if the dimension set is empty.

metricflow/model/data_warehouse_model_validator.py
```diff
@@ -130,46 +161,60 @@ def gen_dimension_tasks(cls, model: UserConfiguredModel) -> List[DataWarehouseVa
...
+
+                data_source_sub_tasks: List[DataWarehouseValidationTask] = []
+                data_source_specs: List[DimensionSpec] = []
+                for dimension in data_source.dimensions:
+                    if dimension.type == DimensionType.TIME:
+                        spec = TimeDimensionSpec.from_name(name=dimension.name)
+                    else:
+                        spec = DimensionSpec.from_name(name=dimension.name)
+                    data_source_specs.append(spec)
+                    filter_elements_node = FilterElementsNode(parent_node=source_node, include_specs=[spec])
```
tlento, line 179: Let's use the existing converters for initializing these dimension specs - the process might change in the future and cause validation to fail in some real world cases but not others, so this will protect us from that.
```suggestion
                data_source_sub_tasks: List[DataWarehouseValidationTask] = []
                dataset = render_tools.converter.create_sql_source_dataset(data_source)
                data_source_specs = dataset.instance_set.spec_set.dimension_specs + dataset.instance_set.spec_set.time_dimension_specs
                for spec in data_source_specs:
                    filter_elements_node = FilterElementsNode(parent_node=source_node, include_specs=[spec])
```

tlento (approved):

tlento:
Duh, I forgot to add an overall comment:

This looks great! Inline change suggestions are recommended for pre-merge, the broader stuff should probably be addressed (or not!) in follow-up PRs.

metricflow/model/data_warehouse_model_validator.py
QMalcolm: So the goal is to do as few queries as possible. So for things like dimensions/measures/identifiers we try to get all the dimensions/measures/identifiers on one data source in one query. If the query for all dimensions/measures/identifiers fail then we do the sub queries for the individual dimensions/measures/identifiers. However, if the query for all dimensions/measures/identifiers on a data source passes, we can skip the individualized query. We do the on fail subtask because the initial query for all of one element on a data source fails, it'll only tell us about the first element that failed when there may be multiple that are problematic.

metricflow/model/data_warehouse_model_validator.py
```diff
@@ -130,46 +161,60 @@ def gen_dimension_tasks(cls, model: UserConfiguredModel) -> List[DataWarehouseVa
...
-            data_source_tasks: List[DataWarehouseValidationTask] = []
-            data_source_columns: List[str] = []
-            for dimension in data_source.dimensions:
-                dim_to_query = dimension.expr if dimension.expr is not None else dimension.name
-                data_source_columns.append(dim_to_query)
-
-                data_source_tasks.append(
+        for data_source in model.data_sources:
+            if len(data_source.dimensions) > 0:
+                source_node = cls._data_source_node(render_tools=render_tools, data_source=data_source)
```
tlento, line 169: Never mind, I see what I missed. You might want this to unnest the below, but it's optional:
```suggestion
            if not data_source.dimensions:
                continue

            source_node = cls._data_source_node(render_tools=render_tools, data_source=data_source)
```

metricflow/test/model/validations/test_data_warehouse_tasks.py
QMalcolm: Nothing to see here 🙃 totally didn't accidentally leave that in sometime on a previous PR

metricflow/model/data_warehouse_model_validator.py
tlento: Ah, got it, that makes sense. Thanks!

metricflow/model/data_warehouse_model_validator.py
QMalcolm: So when we actually go about running the query we use the engine's implementation of dry run (https://github.com/transform-data/metricflow/blob/47ad03a9cfd12a5de36228784480cbb46589c0b0/metricflow/model/data_warehouse_model_validator.py#L301). For most engines this means that an `EXPLAIN` is appended to the front of the query. I think the exception is BigQuery which instead appends `dry_run` to the data warehouse query URL.

metricflow/model/data_warehouse_model_validator.py
QMalcolm: The issue is that if there are no dimensions it'll try rendering the query with an empty node list, which raises an exception. This guard was added specifically because I ran into the exception while developing. Will add a comment to that effect.

tlento: Yeah, I figured it out eventually. Comment might be nice, or the if not: continue idiom is less jarring for people.

PR #174, opened 2022-07-22
Add Measure based Data Warehouse Validations

QMalcolm:
We haven't been validating model measures against the data warehouse. This PR remedies that.

### DIFF
--- a/metricflow/model/data_warehouse_model_validator.py
+++ b/metricflow/model/data_warehouse_model_validator.py
@@ -226,6 +226,73 @@ class DataWarehouseTaskBuilder:
             )
         return tasks
 
+    @classmethod
+    def gen_measure_tasks(
+        cls, model: UserConfiguredModel, sql_client: SqlClient, system_schema: str
+    ) -> List[DataWarehouseValidationTask]:
+        """Generates a list of tasks for validating the measures of the model
+
+        The high level tasks returned are "short cut" queries which try to
+        query all the measures for a given data source. If that query fails,
+        one or more of the measures is incorrectly specified. Thus if the
+        query fails, there are subtasks which query the individual measures
+        on the data source to identify which have issues.
+        """
+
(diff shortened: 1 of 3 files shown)


metricflow/model/data_warehouse_model_validator.py
```diff
@@ -217,6 +217,72 @@ def gen_dimension_tasks(
...
+        for data_source in model.data_sources:
+            if len(data_source.measures) > 0:
+                source_node = cls._data_source_node(render_tools=render_tools, data_source=data_source)
+
+                data_source_sub_tasks: List[DataWarehouseValidationTask] = []
+                data_source_specs: List[MeasureSpec] = []
+                for measure in data_source.measures:
+                    spec = MeasureSpec.from_name(name=measure.name)
+                    data_source_specs.append(spec)
+                    filter_elements_node = FilterElementsNode(parent_node=source_node, include_specs=[spec])
```
tlento, line 245: As in previous:
```suggestion
            # optional unnest
            if not data_source.measures:
                continue

            source_node = cls._data_source_node(render_tools=render_tools, data_source=data_source)
            data_source_sub_tasks: List[DataWarehouseValidationTask] = []
            dataset = render_tools.converter.create_sql_source_dataset(data_source)
            data_source_specs =  dataset.instance_set.spec_set.measure_specs
            for spec in data_source_specs:
                filter_elements_node = FilterElementsNode(parent_node=source_node, include_specs=[spec])
```

tlento (approved):
This looks good, with the one suggested change inside.

As a note, what this does is query the measure *column* defined in the data source. It does not attempt to process the measure definition in any way, because the spec doesn't contain that information.

So this simply asserts that the measure column name is valid. It does not assert that the measure aggregation will run, as the aggregation happens in a much later step.

I think this has some utility as-is so approving here, but if you want the full aggregation phases you'll need to go after the engine. That said, I think those are handled by the metric tests, so I'm not sure you need it.

PR #177, opened 2022-07-23
Release 0.110.1

tlento:
Minor fixes to package metadata errors in 0.110.0


tlento:
Note - won't be merging until Monday, can pick up other small fixes and deploy them then if appropriate.

tlento:
Renamed the branch without thinking, abandoning.

PR #179, opened 2022-07-25
Followup: Reorder fields 

WilliamDee:
## Context
This is an amendment PR that fixes a few stragglers that were missed during this PR https://github.com/transform-data/metricflow/pull/156

There were cases in the visitor like this
```python3
select_columns=tuple(metric_select_columns) + non_metric_select_column_set.as_tuple()
```
which just adds the `metric_select_columns`  and the `non_metric_select_column_set.as_tuple()`. This leads to the ordering being this (metric, others…). Since the ordering is sorted in the `as_tuple()` method. So in order to make the ordering the way we wanted it to be, we have to combine the columns sets and use `as_tuple()` on the entire thing.

## Changes
- Ensure that we are correctly ordering the select_columns using `SelectColumnSet.as_tuple` method rather than appending the `SqlSelectColumn` in the visitor which doesn't respect the ordering.

### DIFF
--- a/metricflow/plan_conversion/dataflow_to_sql.py
+++ b/metricflow/plan_conversion/dataflow_to_sql.py
@@ -975,11 +975,15 @@ class DataflowToSqlQueryPlanConverter(Generic[SqlDataSetT], DataflowPlanNodeVisi
             )
         output_instance_set = output_instance_set.transform(AddMetrics(metric_instances))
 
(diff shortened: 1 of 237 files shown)


tlento (approved):
Nice, thanks for the quick fixup!

Please update the commit message and PR description with the added context from Slack. Thanks!

PR #182, opened 2022-07-27
Multi-Thread Semantic Validations

QMalcolm:
This might be a premature improvement, but it is a matter of time before it becomes absolutely necessary. Semantic validations are CPU bound. They don't depend on any external information. As we add more semantic validation rules and as configs grow larger, semantic validations become slower. Additionally, semantic validations don't depend on one another and they don't alter the underlying model. Thus the straight forward solution is to allow more CPU threads to be thrown at them. On a small config, there is virtually no difference in runtime currently. On exceedingly  large configs, time gains are seen (and to different degrees based on how many threads there are).

### DIFF
--- a/metricflow/model/model_validator.py
+++ b/metricflow/model/model_validator.py
@@ -52,7 +52,7 @@ class ModelValidator:
         ReservedKeywordsRule(),
     )
 
-    def __init__(self, rules: Sequence[ModelValidationRule] = DEFAULT_RULES, max_workers: int = 4) -> None:
+    def __init__(self, rules: Sequence[ModelValidationRule] = DEFAULT_RULES, max_workers: int = 1) -> None:
         """Constructor.
 
         Args:
--- a/metricflow/model/model_validator.py
+++ b/metricflow/model/model_validator.py
@@ -1,3 +1,4 @@
+from concurrent.futures import ThreadPoolExecutor, as_completed
 import copy
 import logging
 from typing import List, Sequence
@@ -51,11 +52,12 @@ class ModelValidator:
         ReservedKeywordsRule(),
     )
 
-    def __init__(self, rules: Sequence[ModelValidationRule] = DEFAULT_RULES) -> None:
+    def __init__(self, rules: Sequence[ModelValidationRule] = DEFAULT_RULES, max_workers: int = 4) -> None:
         """Constructor.
 
         Args:
             rules: List of validation rules to run. Defaults to DEFAULT_RULES
+            max_workers: sets the max number of rules to run against the model concurrently
         """
 
         # Raises an error if 'rules' is an empty sequence or None
@@ -63,14 +65,17 @@ class ModelValidator:
             raise ValueError("ModelValidator 'rules' must be a sequence with at least one ModelValidationRule.")
 
         self._rules = rules
+        self._executor = ThreadPoolExecutor(max_workers=max_workers)
 
     def validate_model(self, model: UserConfiguredModel) -> ModelBuildResult:
         """Validate a model according to configured rules."""
         model_copy = copy.deepcopy(model)
 
         issues: List[ValidationIssueType] = []
-        for validation_rule in self._rules:
-            issues.extend(validation_rule.validate_model(model_copy))
+
+        futures = [self._executor.submit(validation_rule.validate_model, model_copy) for validation_rule in self._rules]
+        for future in as_completed(futures):
+            issues.extend(future.result())
 
         return ModelBuildResult(model=model_copy, issues=ModelValidationResults.from_issues_sequence(issues))
 
--- a/metricflow/model/model_validator.py
+++ b/metricflow/model/model_validator.py
@@ -1,4 +1,4 @@
-from concurrent.futures import ThreadPoolExecutor, as_completed
+from concurrent.futures import ProcessPoolExecutor, as_completed
 import copy
 import logging
 from typing import List, Sequence
@@ -65,7 +65,7 @@ class ModelValidator:
             raise ValueError("ModelValidator 'rules' must be a sequence with at least one ModelValidationRule.")
 
         self._rules = rules
-        self._executor = ThreadPoolExecutor(max_workers=max_workers)
+        self._executor = ProcessPoolExecutor(max_workers=max_workers)
 
     def validate_model(self, model: UserConfiguredModel) -> ModelBuildResult:
         """Validate a model according to configured rules."""
--- a/metricflow/cli/main.py
+++ b/metricflow/cli/main.py
@@ -713,6 +713,13 @@ def _data_warehouse_validations_runner(
 @click.option(
     "--verbose-issues", is_flag=True, default=False, help="If specified, prints any extra details issues might have"
 )
+@click.option(
+    "--semantic-validation-workers",
+    required=False,
+    type=int,
+    default=1,
+    help="Optional. Uses the number of workers specified to run the semantic validations. Should only be used for exceptionally large configs",
+)
 @pass_config
 @exception_handler
 @log_call(module_name=__name__, telemetry_reporter=_telemetry_reporter)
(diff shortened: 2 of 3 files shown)


plypaul:
So threads in Python don't actually improve compute throughput because of the GIL. For parallelism with better multi-core utilization, you would need to use the `multiprocessing` module or similar, but that can be more difficult to do.

tlento:
I was going to ask about asyncio but that only really applies to warehouse queries (also, we should probably look into using asyncio for warehouse validations, I *think* it'll work but I admit I'm deeply familiar with async queries in Hack and basically a total n00b when it comes to asyncio in Python). 

TIL: 

> CPython implementation detail: In CPython, due to the [Global Interpreter Lock](https://docs.python.org/3.9/glossary.html#term-global-interpreter-lock), only one thread can execute Python code at once (even though certain performance-oriented libraries might overcome this limitation). If you want your application to make better use of the computational resources of multi-core machines, you are advised to use [multiprocessing](https://docs.python.org/3.9/library/multiprocessing.html#module-multiprocessing) or [concurrent.futures.ProcessPoolExecutor](https://docs.python.org/3.9/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor). However, threading is still an appropriate model if you want to run multiple I/O-bound tasks simultaneously.

Also, TIL:

https://docs.python.org/3.9/library/concurrent.futures.html#processpoolexecutor

@QMalcolm does this work with the ProcessPoolExecutor? If so, is that even more faster?

QMalcolm:
> I was going to ask about asyncio but that only really applies to warehouse queries (also, we should probably look into using asyncio for warehouse validations, I _think_ it'll work but I admit I'm deeply familiar with async queries in Hack and basically a total n00b when it comes to asyncio in Python).
> 
> TIL:
> 
> > CPython implementation detail: In CPython, due to the [Global Interpreter Lock](https://docs.python.org/3.9/glossary.html#term-global-interpreter-lock), only one thread can execute Python code at once (even though certain performance-oriented libraries might overcome this limitation). If you want your application to make better use of the computational resources of multi-core machines, you are advised to use [multiprocessing](https://docs.python.org/3.9/library/multiprocessing.html#module-multiprocessing) or [concurrent.futures.ProcessPoolExecutor](https://docs.python.org/3.9/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor). However, threading is still an appropriate model if you want to run multiple I/O-bound tasks simultaneously.
> 
> Also, TIL:
> 
> https://docs.python.org/3.9/library/concurrent.futures.html#processpoolexecutor
> 
> @QMalcolm does this work with the ProcessPoolExecutor? If so, is that even more faster?

@tlento In regards to data warehouse validations, using asyncio is on the road map. Hopefully will be happening soon. The current hold up I believe is that snowflake's connector doesn't support asyncio https://github.com/snowflakedb/snowflake-connector-python/issues/38#issuecomment-1005082501 😞

QMalcolm:
> So threads in Python don't actually improve compute throughput because of the GIL. For parallelism with better multi-core utilization, you would need to use the `multiprocessing` module or similar, but that can be more difficult to do.

@plypaul you are absolutely correct. Good catch. On @tlento recommendation, I looked into the [ProcessPoolExecutor](https://docs.python.org/3.9/library/concurrent.futures.html#processpoolexecutor) which uses `multiprocessing` under the hood. In testing, on sufficiently large configs it's faster. However, on small configs it's slower. I'm sure we could do something smart like detecting model size and determine whether threads should be used. However, that's gonna take a fair bit of trial and error to find the line. Instead for the time being I think it's worthwhile to allow users to specify multiprocessing via a flag. By default the threads would be `1` and the `ProcessPoolExecutor` would be skipped, but if they specified a number greater than `1` then the `ProcessPoolExecutor` would be used. Thoughts?

tlento:
> The current hold up I believe is that snowflake's connector doesn't support asyncio https://github.com/snowflakedb/snowflake-connector-python/issues/38#issuecomment-1005082501 😞

Yeah, I forgot about that. Lucas has an idea for a possible work-around.

> I think it's worthwhile to allow users to specify multiprocessing via a flag. By default the threads would be 1 and the ProcessPoolExecutor would be skipped, but if they specified a number greater than 1 then the ProcessPoolExecutor would be used. Thoughts?

Seems reasonable to me, although it's maybe asking kind of a lot for people to decide how many threads to run for validation.

QMalcolm:
> > The current hold up I believe is that snowflake's connector doesn't support asyncio [snowflakedb/snowflake-connector-python#38 (comment)](https://github.com/snowflakedb/snowflake-connector-python/issues/38#issuecomment-1005082501) 😞
> 
> Yeah, I forgot about that. Lucas has an idea for a possible work-around.
> 
> > I think it's worthwhile to allow users to specify multiprocessing via a flag. By default the threads would be 1 and the ProcessPoolExecutor would be skipped, but if they specified a number greater than 1 then the ProcessPoolExecutor would be used. Thoughts?
> 
> Seems reasonable to me, although it's maybe asking kind of a lot for people to decide how many threads to run for validation.

Yea I think long term we'll want to intelligently determine how many threads to use. If this is good enough for now, I think we should move forward. If we don't think this is valuable until we can intelligently determine thread use, then I'm okay with holding off on this for now.

plypaul:
When you say the `ProcessPoolExecutor` is slower on small configs, how much slower is it? If it's not too much lower, I would say to avoid the complexities of branching. Also since each validation will run in a separate process, checking that there's no other shared state? Like each validation only depends on the model?

tlento:
I'm happy with whatever you all decide. After looking at the code I have a preference for avoiding branching (whether auto-detected or done via a flag), but @QMalcolm you have the best sense of what might be painful to users here so whatever you choose is most likely the best call. I'll leave this for @plypaul to approve once his questions are resolved.

QMalcolm:
> When you say the `ProcessPoolExecutor` is slower on small configs, how much slower is it? If it's not too much lower, I would say to avoid the complexities of branching. Also since each validation will run in a separate process, checking that there's no other shared state? Like each validation only depends on the model?

To address the first question: for a small config, as an example the config specified in [metricflow/config-templates/salesforce/](https://github.com/transform-data/metricflow/tree/main/config-templates/salesforce), takes something like 0.084 seconds for the default rules when not using the ProcessPoolExecutor. This same config takes 1.614 seconds using the ProcessPoolExecutor with 4 workers. Thus the setup and teardown for the the processes for the rules is taking like ~1.53 seconds. Thus it's really only beneficial to start using the ProcessPoolExecutor when the validations are taking ~2 seconds or more.

For the second question. Validation rules are only handed the model. We've intended it to be that the rules are isolated and not depend on one another because we realized that at some point in the future we'd need to / want to multi thread them. I'm fairly certain we've stuck to that. The metric and materialization validations are perhaps the most complex as they make calls out to the framework to see if something is well defined. But to my knowledge these don't require access to the config file or other os provided items and instead everything needed is derived from the model itself. Additionally, model validations do not apply any transformations to the model.

QMalcolm:
Alright, here to follow up with more data. The runtime for the same config using a `ProcessPoolExecutor` with one worker is ~1.326 seconds. Which is ~1.242 seconds slower than running without an executor for the small config. This is still 2 orders of magnitude different and thus technically a relatively large difference, however objectively small. I'll drop [8251816](https://github.com/transform-data/metricflow/pull/182/commits/8251816598d6f070a41cf6dfc97b6aadec2e290d) to remove the undesirable branching logic

metricflow/model/model_validator.py
```diff
@@ -51,26 +52,30 @@ class ModelValidator:
         ReservedKeywordsRule(),
     )
 
-    def __init__(self, rules: Sequence[ModelValidationRule] = DEFAULT_RULES) -> None:
+    def __init__(self, rules: Sequence[ModelValidationRule] = DEFAULT_RULES, max_workers: int = 4) -> None:
```
plypaul: I thought we were defaulting to 1?

plypaul (approved):
LGTM

metricflow/model/model_validator.py
QMalcolm: Huh this comment was out of date as soon as it was made 🤔 After dropping the commit that added the conditional logic (which was also the commit that made the default 1), I added a separate commit that set the default as 1. I wonder if github somehow showed you the changes without that commit?

PR #183, opened 2022-07-27
Semi-Additive Measure

WilliamDee:
This PR adds the ability to specific dimensions that are non additive in a measure, otherwise known as a semi-additive measure. This is used when we can aggregate a measure on certain dimensions, but not on other dimensions. A way to denote these measures would be as followed in the config:
```yaml
data_source:
   ...
   measures:
      ...
      non_additive_dimensions:
         - name: <dimension_name> # Currently only TimeDimension are allowed to be specified as a non_additive dimension
           window_choice: # one of ["min"/"max"] 
           window_groupings: # <optional list of identifiers to join on>
```

The underlying sql rendering here is filtering the data source down to singular values via an aggregate function for the non-additive dimension then joining it back to the data set.

For example if we look at the following config,
```yaml
data_source: 
  measures:
    - name: average_account_balance_specified
      agg: avg
      non_additive_dimensions:
        - name: statement_date
          window_choice: min
``` 
this would be rendered out as
```sql
SELECT AVG(average_account_balance_specified), ...
FROM
   <data_set_1> ds1
INNER JOIN
   SELECT MIN(statement_date) AS statement_date
   FROM 
   <data_set_1> ds2
ON
  ds1.statement_date = ds2.statement_date
...
```
In addition, one can specify additional group by identifiers for that right join statement.
```yaml
data_source: 
  measures:
    - name: average_account_balance_specified
      agg: avg
      non_additive_dimensions:
        - name: statement_date
           window_choice: min
           window_grouping:
              - customer_id
``` 
we would get this,
```sql
SELECT AVG(average_account_balance_specified), ...
FROM
   <data_set_1> ds1
INNER JOIN
   SELECT MIN(statement_date) AS statement_date, customer_id
   FROM 
   <data_set_1> ds2
   GROUP BY customer_id
ON
  ds1.statement_date = ds2.statement_date AND ds1.customer_id = ds2.customer_id
...
```


metricflow/dataflow/builder/source_node.py
```diff
@@ -21,19 +29,68 @@ def create_from_data_sets(self, data_sets: Sequence[DataSourceDataSet]) -> Seque
...
             else:
-                for time_dimension_reference in aggregation_time_dimensions:
-                    source_nodes.append(
-                        MetricTimeDimensionTransformNode[DataSourceDataSet](
+                for time_dimension_reference, measures in agg_time_dim_to_measures.items():
+                    measures_for_data_source = set(measures)
+
+                    # Build all source nodes from non-additive measures
+                    for non_additive_measure in non_additive_measures:
+                        if non_additive_measure not in measures_for_data_source:
```
tlento, line 52: There's a lot of cross-checking gymnastics in this PR. If we make the non-additive attributes part of the MeasureSpec instead of keeping them separated in DataSourceSemantics, does that simplify the other things?

I think it should work to do that although it is a larger refactor underneath this to update the various direct calls to the MeasureSpec initializer.

metricflow/dataflow/builder/source_node.py
```diff
@@ -21,19 +29,68 @@ def create_from_data_sets(self, data_sets: Sequence[DataSourceDataSet]) -> Seque
...
+                            identifier_specs=window_groupings,
+                            time_dimension_spec=time_dimension_spec,
+                            agg_by_function=non_additive_params.window_choice,
+                        )
+
+                        source_nodes.append(
+                            MetricTimeDimensionTransformNode[DataSourceDataSet](
+                                parent_node=semi_additive_join_node,
+                                aggregation_time_dimension_reference=time_dimension_reference,
+                            )
```
tlento, line 80: Ok this seems right to me but I'll defer to @plypaul on whatever he might turn up here.

From a code structure perspective, I think it'll be easier to understand the block contained in this inner for loop if a lot of this was moved into a helper function to initialize and return a MetricTimeDimensionTransformNode.

tlento (commented):
Looks promising!

metricflow/specs.py
```diff
@@ -315,7 +315,13 @@ def dimension_reference(self) -> DimensionReference:  # noqa: D
         return DimensionReference(element_name=self.element_name)
 
 
+# How to avoid circular import here
+from metricflow.model.objects.elements.measure import NonAdditiveDimensionParams
```
WilliamDee, line 319: Is there another way around this?

metricflow/dataflow/dataflow_plan.py
```diff
@@ -673,8 +739,10 @@ def __init__(  # noqa: D
         self,
         parent_node: BaseOutput[SourceDataSetT],
         include_specs: Sequence[InstanceSpec],
+        hide_description: bool = False,
```
plypaul, line 742: Instead of hiding the description, what about a different description that better describes the operation You were looking to filter out measures that have different additive properties?

metricflow/model/objects/elements/measure.py
```diff
@@ -44,6 +45,28 @@ def can_limit_dimension_values(self) -> bool:
...
 
+class NonAdditiveDimensionParams(HashableBaseModel):
+    """Describes the params for specifying non-additive dimensions in a measure.
+
+    NOTE: Currently, only TimeDimensions are supported for this filter
+    """
+
+    name: str
+    window_choice: AggregationType
+    window_groupings: List[str] = []
```
plypaul, line 56: For a follow up, we'll want to have a rule to validate these items.

metricflow/model/semantics/semantic_containers.py
```diff
@@ -142,14 +143,17 @@ def __init__(  # noqa: D
...
+        self._measure_non_additive_dimension: Dict[MeasureReference, NonAdditiveDimensionParams] = {}
         self._dimension_index: Dict[DimensionReference, List[DataSource]] = defaultdict(list)
         self._linkable_reference_index: Dict[LinkableElementReference, List[DataSource]] = defaultdict(list)
         self._entity_index: Dict[Optional[str], List[DataSource]] = defaultdict(list)
         self._identifier_ref_to_entity: Dict[IdentifierReference, Optional[str]] = {}
         self._data_source_names: Set[str] = set()
 
         self._configured_data_source_container = configured_data_source_container
-        self._data_source_to_aggregation_time_dimensions: Dict[DataSourceReference, List[TimeDimensionReference]] = {}
+        self._data_source_to_aggregation_time_dimensions: Dict[
```
plypaul, line 154: This nesting is getting a little complicated - maybe introduce a helper object?

metricflow/dataflow/builder/source_node.py
```diff
@@ -16,24 +26,78 @@ class SourceNodeBuilder:
...
+            LinklessIdentifierSpec.from_element_name(name) for name in non_additive_dimension.window_groupings
+        )
+        return SemiAdditiveJoinNode[DataSourceDataSet](
+            parent_node=filter_non_additive_measure_node,
+            identifier_specs=window_groupings,
+            time_dimension_spec=time_dimension_spec,
+            agg_by_function=non_additive_dimension.window_choice,
+        )
+
     def create_from_data_sets(self, data_sets: Sequence[DataSourceDataSet]) -> Sequence[BaseOutput[DataSourceDataSet]]:
```
plypaul, line 53: Mind updating the comment to describe how semi-additive measures are handled?

plypaul (commented):
Generally looks good, but it would be great to see some test cases that demonstrate how the node gets rendered into a SQL query.

metricflow/specs.py
WilliamDee: Yup, it errors out with circular import if I put this to the top where the imports should be. Putting it where I put it in this PR works, but I didn't think it was a good idea to put it there unless it's actually okay.

metricflow/dataflow/dataflow_plan.py
WilliamDee: replaced with `replace_description` where we would pass a string to replace the current description! For example, here when we filter only semi additive measures it would say,
```
<FilterElementsNode>
                <!-- description = Filter all Semi-additive Measures -->
                <!-- node_id = pfe_6 -->
```

metricflow/dataflow/builder/costing.py
```diff
@@ -153,3 +154,7 @@ def visit_metric_time_dimension_transform_node(  # noqa: D
         self, node: MetricTimeDimensionTransformNode[SourceDataSetT]
     ) -> DefaultCost:
         return DefaultCost.sum([x.accept(self) for x in node.parent_nodes])
+
+    def visit_semi_additive_join_node(self, node: SemiAdditiveJoinNode[SourceDataSetT]) -> DefaultCost:  # noqa: D
+        # TODO: FIX
+        return DefaultCost.sum([x.accept(self) for x in node.parent_nodes])
```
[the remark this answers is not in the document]
WilliamDee: followed that, but did this 
```python
parent_costs = [x.accept(self) for x in node.parent_nodes]
 # Add number of joins to the cost.
node_cost = DefaultCost(num_joins=1)
return DefaultCost.sum(parent_costs + [node_cost])
```
to accomodate for the join. Is this correct?

metricflow/model/objects/elements/measure.py
```diff
@@ -44,6 +45,28 @@ def can_limit_dimension_values(self) -> bool:
...
+    @property
+    def bucket_hash(self) -> int:
+        """Returns the hash value used for grouping equivalent params."""
+        attributes = {
+            "window_choice": self.window_choice.name,
+            "name": self.name,
+        }
+        if self.window_groupings:
+            attributes["window_groupings"] = str(sorted(list(self.window_groupings)))
+        return hash(json.dumps(attributes, sort_keys=True))
```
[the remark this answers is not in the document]
WilliamDee: agreed to use sha1 through discussions!

metricflow/test/integration/test_cases/itest_semi_additive_measure.yaml
```diff
@@ -0,0 +1,64 @@
...
+  check_query: |
+    SELECT
+      a.user AS user
+      , SUM(a.current_account_balance_by_user) AS current_account_balance_by_user
+    FROM (
+      SELECT
+        ds
+        , user_id AS user
+        , account_balance AS current_account_balance_by_user
+      FROM (
```
plypaul, line 17: You can remove the `SELECT *` and just do `{{ source_schema }}.fct_accounts`. Same for the other cases where you do `SELECT *`.

metricflow/test/integration/test_cases/itest_semi_additive_measure.yaml
```diff
@@ -0,0 +1,64 @@
...
+        user_id AS user
+        , MAX(ds) AS ds
+      FROM (
+        SELECT * FROM {{ source_schema }}.fct_accounts
+      )
+      GROUP BY
+        user_id
+    ) b
+    ON
+      (a.ds = b.ds) AND (a.user = b.user)
```
plypaul, line 32: Guessing this might be from generated SQL - parenthesis are in there because we haven't put in operator precedence logic, but since people will be reading these, you can remove the parenthesis.

plypaul (commented):
Almost ready to approve, but seeing if you saw my comment about the nested `Dict` structure?

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -506,17 +501,43 @@ def build_aggregated_measures(
...
+            additive_measure_specs = grouped_measures_by_additiveness.additive_measures
+            grouped_semi_additive_measures = grouped_measures_by_additiveness.grouped_semi_additive_measures
+
+            # Build output node for additive measures
+            if additive_measure_specs:
+                logger.info(
+                    f"Building aggregated measures for {data_source} with additive measures: {additive_measure_specs}"
+                )
+                output_nodes.append(
+                    self._build_aggregated_measures_for_single_data_source(
```
plypaul, line 514: In retrospect, `build_aggregated_measures_for_single_data_source` is better named to something like `build_aggregated_measures_from_measure_source_node` now. What do you think?

plypaul (approved):
Great work on this feature! Left some nits.

metricflow/model/semantics/element_group.py
```diff
@@ -0,0 +1,25 @@
+import collections
+
+from typing import Dict, Generic, List, TypeVar
+
+ElementX = TypeVar("ElementX")
+ElementY = TypeVar("ElementY")
+
+
+class ElementGrouper(Generic[ElementX, ElementY]):
+    """Groups an Element Pair such that we have (ElementX, List[ElementY])."""
```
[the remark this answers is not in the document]
WilliamDee: no reason, going to do `Element pair` instead

metricflow/model/objects/elements/measure.py
```diff
@@ -44,6 +43,24 @@ def can_limit_dimension_values(self) -> bool:
         return self in (AggregationType.SUM, AggregationType.SUM_BOOLEAN)
 
 
+class NonAdditiveDimensionParams(HashableBaseModel):
```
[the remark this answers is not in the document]
WilliamDee: changed

metricflow/dataflow/builder/source_node.py
```diff
@@ -1,39 +1,110 @@
...
+        self,
+        parent_node: BaseOutput[DataSourceDataSet],
+        non_additive_dimension: NonAdditiveDimensionParams,
+        filter_specs: Sequence[InstanceSpec],
+    ) -> SemiAdditiveJoinNode[DataSourceDataSet]:
+        """Builds a SemiAdditiveJoinNode given measures and non-additive dimension attributes."""
+        filter_non_additive_measure_node = FilterElementsNode[DataSourceDataSet](
+            parent_node=parent_node,
+            include_specs=filter_specs,
+            replace_description="Filter all Semi-additive Measures",
```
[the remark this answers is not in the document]
WilliamDee: Made it similar to the original description in `FilterElementsNode` ie., `Pass only Semi-Additive Measures`

metricflow/test/fixtures/model_yamls/simple_model/metrics.yaml
```diff
@@ -330,3 +330,24 @@ metric:
   type_params:
     numerator: bookings
     denominator: booking_value
+---
+metric:
+  name: "total_account_balance_first_day"
+  description: "total_account_balance_first_day"
+  owners:
+    - bobby@transformdata.io
```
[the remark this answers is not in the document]
WilliamDee: Changed all addresses in this file to support@transformdata.io

metricflow/dataflow/dataflow_plan.py
```diff
@@ -410,6 +417,65 @@ def displayed_properties(self) -> List[DisplayedProperty]:  # noqa: D
         ]
 
 
+class SemiAdditiveJoinNode(Generic[SourceDataSetT], BaseOutput[SourceDataSetT]):
+    """A node that performs a row filter by aggregating a given non-additive dimension."""
```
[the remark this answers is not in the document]
WilliamDee: Changed to this, please review as I'm not too familiar with the data lingo here.

```
A node that performs a row filter by aggregating a given non-additive dimension.

    This is designed to filter a dataset down to singular non-additive time dimension values by aggregating
    the time dimension with MAX/MIN then joining back to the original dataset on that aggregated value.
    Additionally, an optional sequence of identifiers can be passed in to group by and join on during the filtering.

    For example, if we have a data set that includes "account_balances,user,timestamp", we can build a
    "latest_account_balance_by_user" data set using this node by filtering the data set such that for each user,
    we get the MAX(date) and return the account_balance corresponding to that date.
```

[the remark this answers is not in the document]
plypaul: That looks good, but how about a small table with the columns account_balances, user, timestamp shown before and after the?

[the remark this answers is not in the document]
WilliamDee: Added!

PR #186, opened 2022-07-28
Enable custom Pydantic model deserialization

tlento:
The initial cutover to using Pydantic-native parsing via input
validator overrides only allowed for dict-type or class instance inputs
on a per-model object basis. For example, the SqlTable class would
only accept strings or SqlTable objects, and every individual
class implementing our PydanticCustomInputParser base was responsible
for handling dict or calling class type inputs.

The issue with this object-by-object approach is any time someone
either uses direct initialization or, more importantly, uses Pydantic
to serialize and deserialize a model, they will encounter failures
whenever a value needs to be parsed by pydantic through one of these
custom override classes unless every single one of them implements
the relevant instance conditions.

This change centralizes that logic by moving it into the
PydanticCustomInputParser and passing the values along as appropriate.

### DIFF
--- a/metricflow/dataflow/sql_table.py
+++ b/metricflow/dataflow/sql_table.py
@@ -16,15 +16,10 @@ class SqlTable(PydanticCustomInputParser, FrozenBaseModel):
         """Parses a SqlTable from string input found in a user-provided model specification
 
         Raises a ValueError on any non-string input, as all user-provided specifications of table identifiers
-        should be strings conforming to the expectations defined in the from_string method. The lone exception
-        is for SqlTable instance inputs, which could arise if we are initializing Materialization objects internally
-        without direct access to user input.
+        should be strings conforming to the expectations defined in the from_string method.
         """
         if isinstance(input, str):
             return SqlTable.from_string(input)
-        elif isinstance(input, SqlTable):
-            # This is internally constructed, pass it through and ignore it in error messaging
-            return input
         else:
             raise ValueError(
                 f"SqlTable inputs from model configs are expected to always be of type string, but got type "
--- a/metricflow/model/objects/base.py
+++ b/metricflow/model/objects/base.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import json
 import os
 from abc import ABC, abstractmethod
-from typing import Any, ClassVar, Generic, TypeVar
+from typing import Any, Callable, ClassVar, Generic, Iterator, TypeVar
 
 from pydantic import BaseModel, root_validator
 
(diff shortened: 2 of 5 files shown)


metricflow/dataflow/sql_table.py
```diff
@@ -16,18 +16,13 @@ def _from_yaml_value(cls, input: PydanticParseableValueType) -> SqlTable:
...
-        should be strings conforming to the expectations defined in the from_string method. The lone exception
-        is for SqlTable instance inputs, which could arise if we are initializing Materialization objects internally
-        without direct access to user input.
+        should be strings conforming to the expectations defined in the from_string method.
         """
         if isinstance(input, str):
             return SqlTable.from_string(input)
-        elif isinstance(input, SqlTable):
-            # This is internally constructed, pass it through and ignore it in error messaging
-            return input
```
QMalcolm, line 27: This was added in 15352c70f641b3e09680b790fe51a47399d6ff81. It looks like we added it for cases where we internally build a model. Do we expect to not need to internally build the model or is there different reasoning?

metricflow/dataflow/sql_table.py
```diff
@@ -16,18 +16,13 @@ def _from_yaml_value(cls, input: PydanticParseableValueType) -> SqlTable:
...
         else:
             raise ValueError(
-                f"SqlTable inputs from model configs are expected to always be of type string, but got type "
+                f"SqlTable inputs from model configs are expected to always be of type string or dict, but got type "
```
QMalcolm, line 25: It was updated to say `string or dict` but this method will only handle strings, are dicts handled somewhere else?

metricflow/model/objects/base.py
```diff
@@ -104,20 +104,53 @@ def extract_metadata_from_parsing_context(cls, values: PydanticParseableValueTyp
...
+    This class will NOT allow for custom handling of dict or object type inputs, assuming that they should be
+    processed and validated in the standard Pydantic way. Any subclass needing to mutate the inputs from a dict
+    or object type in a subclass-specific way should still be able to do so via the standard Pydantic approach
+    of adding a method decorated with @validator or @root_validator, which will work internally to initialization
+    and validation of that model object itself.
+    """
 
     @classmethod
-    def __get_validators__(cls):  # type: ignore[no-untyped-def]
+    def __get_validators__(cls) -> Iterator[Callable]:
```
QMalcolm, line 121: Wooo types!

metricflow/dataflow/sql_table.py
QMalcolm: Ahh I see, [metricflow/model/objects/base.py#145](https://github.com/transform-data/metricflow/blob/4536401243884fc45cd970b0b9395b10444d96b0/metricflow/model/objects/base.py#L145) handles the dict case

metricflow/dataflow/sql_table.py
QMalcolm: To answer my own question, [metricflow/model/objects/base.py#147](https://github.com/transform-data/metricflow/blob/4536401243884fc45cd970b0b9395b10444d96b0/metricflow/model/objects/base.py#L147) handles the case wherein the input is an instance of the desired class itself.

QMalcolm (approved):
Thank you for this wonderful improvement! It's nice that we don't have to handle the dict/cls case individually in each object anymore 🎉

metricflow/dataflow/sql_table.py
tlento: Oh that's a mistake actually, will revert before merge.

PR #187, opened 2022-07-28
Make validation issue strings plain text by default

tlento:
The original implementation of ValidationIssue.as_readable_string
included a reference to click formatting that color-coded the
issue level. This looks great in the CLI, but in contexts where
that gets converted to plain text it inserts a bunch of unicode
sequences into the header of the error message, making the exception
traces harder to read.

This splits that behavior into a CLI-focused method and base method.
Local testing confirms the CLI output is unchanged, while error logs
no longer contain the unicode markers.

### DIFF
--- a/metricflow/model/validations/validator_helpers.py
+++ b/metricflow/model/validations/validator_helpers.py
@@ -160,24 +160,26 @@ class ValidationIssue(ABC, BaseModel):
 
         raise NotImplementedError
 
-    @classmethod
-    def build_validation_header_msg(cls, level: ValidationIssueLevel) -> str:
-        """Builds the header message with colour."""
-        return click.style(level.name, bold=True, fg=ISSUE_COLOR_MAP[level])
-
-    def as_readable_str(self, verbose: bool = False) -> str:
+    def as_readable_str(self, verbose: bool = False, prefix: Optional[str] = None) -> str:
         """Return a easily readable string that can be used to log the issue."""
+        prefix = prefix or self.level.name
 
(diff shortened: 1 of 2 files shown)


QMalcolm:
Waiiiit just noticed something

metricflow/model/validations/validator_helpers.py
```diff
@@ -197,7 +199,7 @@ class ValidationFutureError(ValidationIssue, BaseModel):
     def level(self) -> ValidationIssueLevel:  # noqa: D
         return ValidationIssueLevel.FUTURE_ERROR
 
-    def as_readable_str(self, verbose: bool = False) -> str:
+    def as_readable_str(self, verbose: bool = False, prefix: Optional[str] = None) -> str:
         """Return a easily readable string that can be used to log the issue."""
         return (
             f"{super().as_readable_str(verbose=verbose)}"
```
QMalcolm, line 205: We now have to pass the prefix in the call to `super().as_readable_str` so that when `as_cli_formatted_str` gets called, future errors get the coloration.
```suggestion
            f"{super().as_readable_str(verbose=verbose, prefix=prefix)}"
```

QMalcolm (changes requested):

metricflow/model/validations/validator_helpers.py
tlento: Duh. Nice catch, thanks!

QMalcolm (approved):
Sweet sweet!

PR #191, opened 2022-08-01
Add CLI for data source inference

serramatutu:
This PR adds a CLI interface for data source inference.

Main functionalities are:
- Main function in `metricflow/cli/main.py` for using all parameters to construct an inference pipeline.
- New `InferenceProgressReporter` class for reporting begin/end of inference pipeline steps. This can be useful for custom reporting in scripts but is also used to peek into the inference pipeline in a generic and non-intrusive way.
- New `custom_click_types` module with a couple of CLI helper functions to make our lives easier:
  - `MutuallyExclusiveOption`: class for making click options mutually exclusive
  - `ListParamType`: class for making list parameters like `1,2,3,4` that are automatically coerced to useful lists like `[1,2,3,4]`. See docs and implementation for more details.

### Help guide
image

### Usage

inference-cli

NOTE: Table and schema names were censored for obvious reasons.

### DIFF
--- a/metricflow/cli/custom_click_types.py
+++ b/metricflow/cli/custom_click_types.py
@@ -1,4 +1,4 @@
-from typing import Callable, Generic, List, Optional, TypeVar
+from typing import Any, Callable, Dict, Generic, List, Optional, TypeVar
 import click
 
 T = TypeVar("T")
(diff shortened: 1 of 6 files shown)


metricflow/cli/custom_click_types.py
```diff
@@ -0,0 +1,119 @@
+"""This module contains custom helper click types."""
+
+from typing import Any, Callable, Dict, Generic, List, Optional, TypeVar
+import click
+
+T = TypeVar("T")
+
+
+class ListParamType(click.ParamType, Generic[T]):
```
plypaul, line 9: One issue that I found with `List` is that it's not covariant, so I've generally been trying to use `Sequence` or `Tuple`.

plypaul (commented):
Tagging @WilliamDee to see if he could review the Click / CLI changes.

metricflow/cli/main.py
```diff
@@ -721,5 +733,181 @@ def validate_configs(cfg: CLIContext, dw_timeout: Optional[int] = None, skip_dw:
...
+            tables = [SqlTable(schema_name=schema, table_name=table_name.upper()) for table_name in tables_strs]
+
+        if len(tables) == 0:
+            click.echo("Schema has no tables.")
+            return
+
+    provider = SnowflakeInferenceContextProvider(client=cfg.sql_client, tables=tables, max_sample_size=max_sample_size)
+
+    # set up the solver
+    def solver_weighter_function(confidence: InferenceSignalConfidence) -> int:
```
WilliamDee, line 877: CURIOUS QUESTION (no need to make changes here): is it necessary to allow the user to put custom weights?

metricflow/cli/main.py
```diff
@@ -721,5 +733,181 @@ def validate_configs(cfg: CLIContext, dw_timeout: Optional[int] = None, skip_dw:
...
+
+    if cfg.sql_client.sql_engine_attributes.sql_engine_type is not SupportedSqlEngine.SNOWFLAKE:
+        click.echo(
+            "Data Source Inference is currently only supported for Snowflake. "
+            "We will add support for all the other warehouses before it becomes a "
+            "stable feature. Stay tuned!"
+        )
+        return
+
+    if tables is None and schema is None:
```
WilliamDee, line 857: If you want this to have a sorta XOR operation there's a helper `assert_exactly_one_arg_set` in `metricflow/object_utils.py`. Could possibly be useful here

metricflow/cli/main.py
```diff
@@ -721,5 +733,181 @@ def validate_configs(cfg: CLIContext, dw_timeout: Optional[int] = None, skip_dw:
...
+    def warehouse() -> Iterator[None]:  # noqa: D
+        yield
+
+    @staticmethod
+    @contextlib.contextmanager
+    def table(table: SqlTable, index: int, total: int) -> Iterator[None]:  # noqa: D
+        with _get_spin_context_manager_factory(f"🔍 Querying `{table.sql}` ({index + 1} out of {total})")():
+            yield
+
+    rules = staticmethod(_get_spin_context_manager_factory("🤔 Processing inference rules"))  # type: ignore
```
WilliamDee, line 764: I'm assuming the `type:  ignore` is needed due to the style of staticmethod construction here? If so, we should definition just make it through the decorator as normal. The type ignore is not great and I think the style of `staticmethod(...)` is considered unpythonic

WilliamDee (commented):
Few comments on the CLI changes

metricflow/cli/main.py
serramatutu: People shouldn't really need to, but I wanted to make the CLI as flexible as possible. If we didn't add that option, they'd have to implement a custom script just for that, which is much more painful and inconvenient than just doing `--solver weights 1,2,3,4`.

metricflow/cli/main.py
serramatutu: that is good to know :)

metricflow/cli/main.py
serramatutu: Will do.

metricflow/cli/main.py
serramatutu: they might still need custom scripts if they want to tune/add/remove individual rules because that would be a mess to add to the CLI.

WilliamDee (approved):

PR #199, opened 2022-08-10
Separate `*Reference` classes out of specs.py

WilliamDee:
## Context
Currently, all the `*Reference` and `*Spec` all lives together in the `specs.py` file. We should split these out as they have different functionality. This helps with code organization by not cluttering a file with unrelated code.

## Changes
- Removed all `*Reference` classes out of `specs.py` -> `references.py`
- Moved an import in the middle of the file in `specs.py` back to the top of the file. Wasn't possible due to circular import prior to the move.


plypaul (approved):

plypaul:
Chatted about the fact that while commit that includes this update looks good, but the PR has commits that it shouldn't have.

PR #200, opened 2022-08-10
Relax Jinja2 minimum version requirement

tlento:
The shift to `^3.0.0` was meant to allow downstream Flask `1.1.x`
users to be able to update Flask to `2.x.x`.

Unfortunately, Flask `1.1.x` does not allow Jinja2 `3.x.x` versions,
so we must do this.


tlento:
See Jinja setup on flask 1.1.x here:

https://github.com/pallets/flask/blob/1.1.x/setup.py#L57

plypaul (approved):

jpreillymb (approved):

Issue #204, opened 2022-08-12
CLI MF Infer Inference solver could not determine a type for this column


edlouth:
**Describe the bug**
When running `mf infer` against a snowflake database we get lots of FIXMEs like:
```
# FIXME: company_name
#   - Inference solver could not determine a type for this column
```

**Expected behavior**
Would expect these columns to appear as dimensions

**Environment (please complete the following information):**

- MetricFlow CLI Version 0.111.0

tlento:
Interesting. Based on that name, yes, I agree, but the inference for dimensions is mostly driven by distributional attributes and things like column types. I have two hypotheses:

1. Something about those columns in your snowflake DB are causing them to fall through to unknown in the inference engine, like maybe a type we aren't accounting for (or maybe a bug in our type matching logic).
2. The default confidence filter is too high, and so a lot of things we think are "maybe dimensions" are getting unclassified

@edlouth Two questions:

1. Are there common column types for these fields? Or are they all VARCHAR?
2. Do you get better results if you pass `--solver-threshold 0.5`?

@serramatutu is there an easy way for a user to get a log stream of the inference choices and how things were chosen? I don't see anything in the CLI but I know the API can do this.

edlouth:
1. All these columns are typed as `VARCHAR(16777216)` in Snowflake
2. I ran infer with `--solver-threshold 0.6` and there was no improvement

If it helps at all we get lots of rows like this in the `requirements.txt` file

```
  company_name
    type: UNKNOWN
    reasons:  -- 
    problems: 
    - Inference solver could not determine a type for this column
 ```

tlento:
@edlouth thanks for the info, that problem output and typing info is very helpful. @serramatutu is looking into this so I expect we'll have a fix soon.

tlento:
@edlouth we released the attached improvement for this last week (in `v0.111.1`). Hopefully this improves things for you. 

I'll leave this open for a bit longer in case you're still hitting the same issue. Please do let us know if this is working better for you!

PR #210, opened 2022-08-19
[Inference] Classify low cardinality strings as dimensions

tlento:
#204 highlighted an issue where string type columns were not being classified as dimensions.

This PR is an attempt at improving our detection there. The core of it is a new rule, which classifies a column in the warehouse as a dimension if it is both a STRING type AND relatively low cardinality (below 40% unique).

It's not clear if this is a huge improvement. It should reduce the false negative rate, which appears to be burdensome on early users, but it could result in a higher than expected false positive rate where string foreign key fields, or string fields which should be omitted (like the "name" field for a customer_organization table, where you might have a lot of duplicate entries for franchise customers or whatever). However, this rule appears to be useful, even if we need to tweak the thresholds somewhat.

As a matter of future work, we might consider making these cardinality thresholds adjustable on the CLI, although figuring out how to do so cleanly is a bit of a challenge.

### DIFF
--- a/metricflow/inference/context/snowflake.py
+++ b/metricflow/inference/context/snowflake.py
@@ -21,10 +21,11 @@ class SnowflakeInferenceContextProvider(DataWarehouseInferenceContextProvider):
         """Get the correspondent InferenceColumnType from Snowflake's returned type string.
 
         See for reference: https://docs.snowflake.com/en/sql-reference/sql/show-columns.html
+        For string types: https://docs.snowflake.com/en/sql-reference/data-types-text.html#data-types-for-text-strings
         """
-        str_dict = {
+        type_str = type_str.upper()
+        type_mapping = {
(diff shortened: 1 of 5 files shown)


tlento:
Yes, I tested it on our internal warehouse. We had a "name" field in one of the tables that was a string type and happened to have a matching cardinality ratio. Before:

```
# FIXME: Unreviewed inferred config file
# FIXME: name
#   - Inference solver could not determine a type for this column
data_source:
  name: org_metrics
```

After:

```
# FIXME: Unreviewed inferred config file
data_source:
  name: org_metrics
  # other stuff here...
  dimensions:
  - name: name
    type: categorical
```

tlento:
We might want to tune some of these "weaker" inferences to have a FIXME inline that people can remove or whatever, but that's a longer term thing.

Thanks for the speedy review!

PR #211, opened 2022-08-19
Generate an explicit json schema from the parsing logic

jack-transform:
Changes
- Adds functionality to generate an explicit json schema from the one defined throughout the metricflow parsing code.

- Adds a Makefile command to generate this explicit schema in `models/parsing/schemas`

The generated schema is similar to the one used for parsing but flattened into a single json file. Each subschema in the schema store becomes a definition, while each top level schema is a property that references it's corresponding definition. Some subschemas have their own definition sections, so for simplicity they are lifted into the single top level definitions section. 

With this change, we can easily reference schemas across tagged versions. e.g. `https://github.com/transform-data/metricflow/blob/v0.X.Y/metricflow/model/parsing/schemas/metricflow.json`. 

 The next step is a Github action that runs when the `models/parsing` folder is changed; this will make sure that the generated json schema matches the schema defined in the parsing code. 

Notes
-  I wasn't really sure where to put this code and made a choice fairly arbitarily so let me know if there's a better structure for it. 
- I decided to just use schemas-internal since that's what we use for parsing everywhere but it's an easy change to generate both.

### DIFF
--- a/metricflow/model/parsing/schemas/metricflow.json
+++ b/metricflow/model/parsing/schemas/metricflow.json
@@ -1,0 +1,728 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "definitions": {
+        "composite_sub_identifier_schema": {
+            "$id": "composite_sub_identifier_schema",
+            "additionalProperties": false,
+            "properties": {
+                "expr": {
+                    "type": [
+                        "string",
+                        "boolean"
+                    ]
+                },
+                "name": {
+                    "type": "string"
+                },
+                "ref": {
+                    "type": "string"
+                }
+            },
+            "type": "object"
+        },
+        "data_source": {
+            "$id": "data_source",
+            "additionalProperties": false,
+            "properties": {
+                "dbt_model": {
+                    "type": "string"
+                },
+                "description": {
+                    "type": "string"
+                },
+                "dimensions": {
+                    "items": {
+                        "$ref": "#/definitions/dimension_schema"
+                    },
+                    "type": "array"
+                },
+                "display_name": {
+                    "type": "string"
+                },
+                "identifiers": {
+                    "items": {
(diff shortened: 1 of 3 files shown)


QMalcolm (approved):
This looks great! And to clarify, although there may be other applications, the motivation for this is having an explicit schema enables creating an auto-complete VSCode tool?

tlento:
I have not read any of this PR yet, but I think at some point we should either get the "internal" bits out of open source or else merge them into the proper open source schema files, because having to deal with both and make the kind of decisions you were forced into here is annoying.

For now, basing on "internal" seems correct, since that's the end stage of the model schema, and the parser operates on it as well.

jack-transform:
> 



> This looks great! And to clarify, although there may be other applications, the motivation for this is having an explicit schema enables creating an auto-complete VSCode tool?

Yup! I think this will be helpful as we continue down the line but that's the impetus for it.

metricflow/model/parsing/schemas/metricflow.json
```diff
@@ -0,0 +1,690 @@
...
+                }
+            },
+            "type": "object"
+        },
+        "data_source": {
+            "$id": "data_source",
+            "additionalProperties": false,
+            "properties": {
+                "constraint": {
+                    "$ref": "constraint_schema"
```
tlento, line 28: This schema doesn't exist. I thought @QMalcolm removed it. Did you run this on an old branch?

tlento (approved):
LGTM, assuming a re-run gets rid of that stray constraint schema entry.

metricflow/model/parsing/schemas/metricflow.json
jack-transform: ah, i must have. good catch

jack-transform: Alright fixed it

PR #213, opened 2022-08-22
Serialize validation inputs and outputs for use in ProcessPoolExecutor

tlento:
We recently forked model validation into separate threads via calls
to ProcessPoolExecutor, which pickles the inputs and outputs to the
job. This can apparently cause idiosyncratic behavior when combined
with Pydantic objects, as a consistent local test failure suggests.
This manifests with the following error snippet, but does not occur
in CI runs:

`AttributeError: Can't get attribute '_Pydantic_DataSourceReference_140501776576224' on <module 'pydantic.dataclasses'`

Rather than mess around with internal pickling and unpickling this
commit changes our approach to use Pydantic's serialization support
to pass raw data across. This will be slower per thread, but
more robust to different configurations.

### DIFF
--- a/metricflow/model/validations/validator_helpers.py
+++ b/metricflow/model/validations/validator_helpers.py
@@ -357,11 +357,23 @@ class DimensionInvariants:
 class ModelValidationRule(ABC):
     """Encapsulates logic for checking the values of objects in a model."""
 
-    @staticmethod
+    @classmethod
     @abstractmethod
-    def validate_model(model: UserConfiguredModel) -> List[ValidationIssueType]:
+    def validate_model(cls, model: UserConfiguredModel) -> List[ValidationIssueType]:
         """Check the given model and return a list of validation issues"""
         pass
+
+    @classmethod
+    def validate_model_serialized_for_multiprocessing(cls, serialized_model: str) -> str:
+        """Validate a model serialized via Pydantic's .json() method, and return a list of JSON serialized issues
+
+        This method exists because our validations are forked into parallel processes via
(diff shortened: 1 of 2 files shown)


tlento:
* main
  * **PR #213** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/213" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/transform-data/metricflow/213?utm_source=stack-comment).

QMalcolm (approved):
Odd that we have to do this. But if we have to then we have to, it is what it is. One small thing to fix, the docstring of `validate_model_serialized_for_multiprocessing` has an issue. Otherwise, this looks golden 👍

QMalcolm:
And of course I didn't hit "add comment to review" first 🙃

metricflow/model/validations/validator_helpers.py
```diff
@@ -357,12 +357,24 @@ class DimensionInvariants:
...
-    def validate_model(model: UserConfiguredModel) -> List[ValidationIssueType]:
+    def validate_model(cls, model: UserConfiguredModel) -> List[ValidationIssueType]:
         """Check the given model and return a list of validation issues"""
         pass
 
+    @classmethod
+    def validate_model_serialized_for_multiprocessing(cls, serialized_model: str) -> str:
+        """Validate a model serialized via Pydantic's .json() method, and return a list of JSON serialized issues
+
+        This method exists because our warehouse validations are forked into parallel processes via
```
QMalcolm, line 370: Comment information fix: these are considered `semantic validations` not `warehouse validations` as these validation rules aren't run against the data warehouse.

PR #217, opened 2022-08-23
Github Action to check JSON schema consistency

jack-transform:
Adds a Github action that is triggered when models/parsing is edited. 

The action regenerates the JSON schema using the latest changes in `models/parsing`. If the generated schema doesn't match the existing json schema in `/models/parsing/schemas/metricflow.json`, the action fails. 

This is a follow up to the work in https://github.com/transform-data/metricflow/pull/211

I'm not sure a great way to test Github actions so I tested against a forked repo to avoid spamming metricflow's Github Action log; a failure looks like this: https://github.com/jack-transform/metricflow-jckelly-test/runs/7968193985?check_suite_focus=true

### DIFF
--- a/.github/workflows/ci-schema-consistency.yaml
+++ b/.github/workflows/ci-schema-consistency.yaml
@@ -3,8 +3,7 @@ on:
   workflow_dispatch:
 
   pull_request:
-    # run these jobs when a PR is opened, edited, reopened, or updated (synchronize)
-    # edited = title, body, or the base branch of the PR is modified
+    # run these jobs when a PR is opened, reopened, or updated (synchronize)
     # synchronize = commit(s) pushed to the pull request
     types:
       - opened
@@ -47,7 +46,7 @@ jobs:
           if [[ $(git diff) ]]
           then
             git diff
-            echo 'Failure: JSON schema and generated schema do not match.'
+            echo 'Failure: JSON schema and generated schema do not match. Run make json_schema and commit the updated schema changes before merge.'
             exit 1
           else
             echo 'Success: JSON schema and generated schema match.'
--- a/.github/workflows/ci-schema-consistency.yaml
+++ b/.github/workflows/ci-schema-consistency.yaml
@@ -1,0 +1,55 @@
+name: JSON Schema Consistency Check
+on:
+  workflow_dispatch:
+
+  pull_request:
+    # run these jobs when a PR is opened, edited, reopened, or updated (synchronize)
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-schema-consistency.yaml
```diff
@@ -0,0 +1,55 @@
...
+      
+      - name: Generate JSON Schema
+        run: python3 metricflow/model/parsing/explicit_schema.py
+
+      - name: Schema Consistency Check
+        run: |
+          if [[ $(git diff) ]]
+          then
+            git diff
+            echo 'Failure: JSON schema and generated schema do not match.'
```
tlento, line 50: Should we add instructions for how to resolve this?
```suggestion
            echo 'Failure: JSON schema and generated schema do not match. Run python3 metricflow/model/parsing/explicit_schema.py and commit the updated schema changes before merge.'
```

I'm fine either way, I'm not sure how much we want developers to do this on their own vs handling it via post-merge cleanup PRs.

tlento (approved):
Nice! Let me know if you want to make this commit blocking.

.github/workflows/ci-schema-consistency.yaml
```diff
@@ -0,0 +1,55 @@
+name: JSON Schema Consistency Check
+on:
+  workflow_dispatch:
+
+  pull_request:
+    # run these jobs when a PR is opened, edited, reopened, or updated (synchronize)
+    # edited = title, body, or the base branch of the PR is modified
```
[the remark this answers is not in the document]
jack-transform: Good call, thanks.

.github/workflows/ci-schema-consistency.yaml
jack-transform: Sure, added it.

jack-transform:
> Nice! Let me know if you want to make this commit blocking.

Do you mean make it so that you can't merge unless it passes? I think that would be best- it only applies to the parsing folder so it shouldn't be a huge burden to deal with.

tlento:
Argh. I just realized it has to run on every PR otherwise nobody will be able to merge unless they happen to trigger the action. I wish GitHub didn't do that. I'll figure it out.

jack-transform:
> Argh. I just realized it has to run on every PR otherwise nobody will be able to merge unless they happen to trigger the action. I wish GitHub didn't do that. I'll figure it out.

That's a bizarre design choice lol. But if it's difficult it's super fine to drop. Not a huge deal, like in the worst case where the generated schema is out of date it won't break things.

PR #219, opened 2022-08-25
BUG: ChangeMeasureAggregationState wasn't copying the MeasureInstance fully

WilliamDee:
When using the transform function `ChangeMeasureAggregationState` if a measure instance has a spec with a non-null `NonAdditiveDimensionParameters` it ends up not copying it over along with it. This caused errors when filtering elements such as 
```python
MeasureSpec(element_name=x, non_additive_dimension=<not None>) not in
   [...,MeasureSpec(element_name=x, non_additive_dimension=None),...]
```

Also removed the previous workaround which was removing the non_additive_dimension property from MeasureSpec during measure aggregation, but that was incorrect.

### DIFF
--- a/metricflow/test/integration/test_cases/itest_semi_additive_measure.yaml
+++ b/metricflow/test/integration/test_cases/itest_semi_additive_measure.yaml
@@ -54,3 +54,55 @@ integration_test:
       a.ds = b.ds
     GROUP BY
       a.ds
+---
+integration_test:
+  name: semi_additive_measure_query_with_join_linkable_specs
+  description: |
+    Tests semi-additive measure with a path hitting join linkable specs. 
+    There are cases (like this query where we have a non-local linkable spec "user__home_state_latest")
+    we are directly copying a MeasureSpec via constructing it. Previously, this led to a code path not passing all the attributes,
+    then somewhere along the path it would end up throwing a MeasureSpec not in [List of MeasureSpec] due to an attribute
+    being missing. This test case hits that path and ensures that we are copying the correct.  
+  model: SIMPLE_MODEL
+  metrics: ["current_account_balance_by_user"]
+  group_bys: ["user"]
+  where_constraint: "user__home_state_latest = 'CA'"
+  check_query: |
+    SELECT
+      e.user AS user
+      , SUM(e.current_account_balance_by_user) AS current_account_balance_by_user
+    FROM (
+      SELECT
+        c.user AS user
+        , d.home_state_latest AS user__home_state_latest
+        , c.current_account_balance_by_user AS current_account_balance_by_user
+      FROM (
+        SELECT
+          a.user AS user
+          , a.current_account_balance_by_user AS current_account_balance_by_user
+        FROM (
+          SELECT
+            ds
+            , user_id AS user
+            , account_balance AS current_account_balance_by_user
+          FROM {{ source_schema }}.fct_accounts
+        ) a
+        INNER JOIN (
+          SELECT
+            user_id AS user
+            , MAX(ds) AS ds
+          FROM {{ source_schema }}.fct_accounts
+          GROUP BY
+            user_id
+        ) b
+        ON
+          a.ds = b.ds AND a.user = b.user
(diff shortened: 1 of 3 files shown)


tlento (commented):
Do you have any test cases that can illustrate this issue? Also, is this fully deterministic? I *think* we always want to preserve the non-additive dimension info in the spec, because it's always a needed piece of information for every dataflow plan operation we do. However, it's pretty hard to reason through this.

Note - I'm pretty sure this how we have to do things today, actually, but I'd like to see a test showing that we're handling this correctly, and ideally one which will break whenever we move to a world where specs are representative of dataflow plan state rather than their current role as "hash key references and also transient dataflow plan state but obviously not transient state because they're hash key references"

WilliamDee:
In order to test this case, I needed to create a query that requires joining linkable specs. In order to do so, I just added a where constraint that does that.

metricflow/test/integration/test_cases/itest_semi_additive_measure.yaml
```diff
@@ -53,4 +53,51 @@ integration_test:
     ON
       a.ds = b.ds
     GROUP BY
-      a.ds
\ No newline at end of file
+      a.ds
+---
+integration_test:
```
tlento, line 58: Is there a more minimal way to evaluate this condition, like maybe a test case in test_measure_additiveness.py? I think that'll be easier to understand (although having an integration test exercise this join path is a good idea!).

If you can come up with a more focused test, please add that as well and then you can update the docstring below to be a bit broader, more focused on ensuring correct results with a join condition.

tlento (approved):
Please look into the inliine suggestions before merge. At a minimum I think we want a documentation change, but something that tests the plan construction more directly would be ideal. Thanks!

metricflow/test/integration/test_cases/itest_semi_additive_measure.yaml
WilliamDee: I can't think of a more minimal way of evaluating this. In the bigger picture, this is generally a problem where in some instance we want to copy a `xSpec` (in this case `MeasureSpec`) and in the code it ended up just doing
```python
XSpec(element_name="...")
```
even though that spec has more attributes and should pass along the rest of the attributes. So this technically became a problem because we added a new attribute to `MeasureSpec`, which ended up unraveling this error. So whoever adds a new attribute (in this case, me) shouldve checked that all the constructors were passed correctly! Or we should think of a better way of copying these (maybe with `deepcopy`?) but these are immutable anyways so maybe we can just copy directly.

tlento: We should definitely find a better way to copy these things around, that's a known issue.

metricflow/test/integration/test_cases/itest_semi_additive_measure.yaml
```diff
@@ -53,4 +53,56 @@ integration_test:
...
+  name: semi_additive_measure_query_with_join_linkable_specs
+  description: |
+    Tests semi-additive measure with a path hitting join linkable specs. 
+    There are cases (like this query where we have a non-local linkable spec "user__home_state_latest")
+    we are directly copying a MeasureSpec via constructing it. Previously, this led to a code path not passing all the attributes,
+    then somewhere along the path it would end up throwing a MeasureSpec not in [List of MeasureSpec] due to an attribute
+    being missing. This test case hits that path and ensures that we are copying the correct.
```
tlento, line 65: Found some typos, but you may merge whenever you're ready!
```suggestion
    Tests semi-additive measure with a path hitting join linkable specs. 
    There are cases (like this query where we have a non-local linkable spec "user__home_state_latest") where
    we are directly copying a MeasureSpec via constructing it. Previously, this led to a code path not passing all the attributes,
    then somewhere along the path it would end up throwing a MeasureSpec not in [List of MeasureSpec] due to an attribute
    being missing. This test case hits that path and ensures that we are copying attributes correctly.  
```

tlento (approved):

PR #220, opened 2022-08-25
Mypy missing type errors debugging

tlento:
I have no idea why mypy is missing such a glaring type error.

### DIFF
--- a/metricflow/model/semantics/semantic_containers.py
+++ b/metricflow/model/semantics/semantic_containers.py
@@ -35,6 +35,7 @@ from metricflow.specs import (
     LinkableInstanceSpec,
     MeasureSpec,
     MetricSpec,
+    NonAdditiveDimensionSpec,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


metricflow/model/semantics/semantic_containers.py
```diff
@@ -306,6 +308,14 @@ def _add_data_source(
             self._entity_index[ident.entity].append(data_source)
             self._linkable_reference_index[ident.reference].append(data_source)
 
+    @staticmethod
+    def _make_measure_spec(measure: Measure) -> MeasureSpec:
+        """WTF mypy?"""
+        return MeasureSpec(
+            element_name=measure.name,
+            non_additive_dimension=measure.non_additive_dimension,
```
tlento, line 316: This error, along with the identical error above, is missed with and without the None default on the spec. It appears mypy just isn't checking the type of this dataclass element class correctly.

Note, however, that if I try to change either of these to a string value mypy fails as expected. I do not understand this.

Of course the unit tests blow up because this doesn't work at all.

tlento: Figured it out, mypy ignores a lot of stuff around Pydantic Model objects - up to and including treating them as if all of them are the same type regardless of subclass definitions.

tlento:
Filed #221 for improving our typechecking coverage

PR #223, opened 2022-08-26
Improve error message for missing aggregation time dimension

tlento:
If the user fails to include either a primary time dimension or a
measure-specific aggregation time dimension for a measure in their
model they will encounter an assertion check either during validation
or querying. The original error message was really intended for
developers, but it is incidentally raised in validation. This quick fix
makes it a bit more user-friendly.

### DIFF
--- a/metricflow/model/objects/elements/measure.py
+++ b/metricflow/model/objects/elements/measure.py
@@ -79,9 +79,11 @@ class Measure(HashableBaseModel, ModelWithMetadataParsing):
     @property
     def checked_agg_time_dimension(self) -> TimeDimensionReference:
         """Returns the aggregation time dimension, throwing an exception if it's not set."""
-        assert (
-            self.agg_time_dimension
-        ), f"Aggregation time dimension for {self.name} should have been set during model transformation"
+        assert self.agg_time_dimension, (
+            f"Aggregation time dimension for measure {self.name} is not set! This should either be set directly on "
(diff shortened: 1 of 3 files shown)


metricflow/model/validations/agg_time_dimension.py
```diff
@@ -56,8 +56,8 @@ def _validate_data_source(data_source: DataSource) -> List[ValidationIssueType]:
                     ValidationError(
                         context=measure_context,
                         message=f"In data source '{data_source.name}', measure '{measure.name}' has the aggregation "
-                        f"time dimension is set to '{agg_time_dimension_reference.element_name}', "
```
QMalcolm, line 59: Nice typo find!

metricflow/model/objects/elements/measure.py
```diff
@@ -79,9 +79,11 @@ class Measure(HashableBaseModel, ModelWithMetadataParsing):
     @property
     def checked_agg_time_dimension(self) -> TimeDimensionReference:
         """Returns the aggregation time dimension, throwing an exception if it's not set."""
-        assert (
-            self.agg_time_dimension
-        ), f"Aggregation time dimension for {self.name} should have been set during model transformation"
+        assert self.agg_time_dimension, (
```
QMalcolm, line 82: This is a lot friendlier for when people run into it :slightly_smiling_face:

QMalcolm (approved):
Thank you for this QoL improvement, this is gonna make things much clearer when people run into the issue.

Issue #214, opened 2022-08-22
Semantic conversions between objects and specs leads to circular import dependencies


tlento:
**Describe the bug**
While attempting to update #153 I encountered some circular dependencies that could not be expediently resolved. The time has come to clean this up.

We have a couple of situations where we need to move from a model object in `metricflow.model.objects` to a spec in `metricflow.specs` via a call to a semantic container in `metricflow.model.semantics.semantic_containers`.

Right now the existing logic (converting WhereClauseConstraint to SpecWhereClauseConstraint) lives in the query parser, which is a bit silly but neatly sidesteps the circular import dependency since the the query parser was a natural collector of these imports, and the DataFlowPlanBuilder didn't depend on anything in the query parser itself. However, #153 adds a conversion call into MetricSemantics.measures_for_metric(), which makes this problem unavoidable.

Splitting up MetricSemantics doesn't make a lot of sense - getting input measures for a metric is the kind of thing MetricSemantics is supposed to do. Naively forking the conversion logic into a separate module isn't going to work either, because that requires a dependency on a semantic container, which in turn requires a dependency on that converter.

**Proposal**

The Metricflow package essentially consists of:

- Primitive model/core query handling objects
- - Model objects (`metricflow.model.objects`)
- - Specs/instances/references (`metricflow.specs`, `metricflow.instances`, `metricflow.references`)
- Semantic containers for those data objects (`metricflow.model.semantics.semantic_containers`)
- Resolvers/converters related to semantic container operation
- - The WhereClauseConstraint -> SpecWhereClauseConstraint converter
- - LinkableInstanceSpecResolver
- Query Parser (`metricflow.query.query_parser`)
- Plan objects
- - DataflowPlanNode (`metricflow.dataflow.dataflow_plan`)
- - SqlQueryPlanNode (`metricflow.sql.sql_plan`)
- - etc.
- Plan builders (DataflowPlanBuilder, mainly, `metricflow.dataflow.builder.dataflow_plan_builder`)
- Various converter and resolver objects for handling transformations between object types
- - Visitors for plan conversion
- - InstanceSetConverters
- - NodeEvaluator
- - NodeProcessor
- - ColumnResolver

Some notes on how these fit together:

1. The primitive objects need to be loaded into the semantic containers, which are in turn passed around all over the place.
2. The semantic containers and specs/instances/references are needed for plan building and occasionally plan conversion.
3. The query parser requires both of the above

This implies:

[...]

PR #224, opened 2022-08-26
Add protocols for read-only semantic containers

tlento:
Many of our internal operations for building and converting dataflow
plans rely on our semantic containers (DataSourceSemantics and
MetricSemantics). These containers pull in a lot of dependencies and
also allow callers to modify them on the fly by adding data sources
or metrics to them.

This PR adds more limited scope accessor-only Protocols for
DataSourceSemantics and MetricSemantics, and converts all
relevant references to use it instead of the concrete implementations
found in semantic_containers.py

This gives us more protection against inadvertent use of methods
like `DataSourceSemantics.add_data_source` and limits the scope of
transitive imports running through our semantic_containers module.

PR #225, opened 2022-08-26
Clean up object and module organization

tlento:
We have a number of slightly wonky organizational artifacts in our
codebase which have been cropping up as circular imports of late.
Most recently, issues with conversion between WhereClauseConstraint
and SpecWhereClauseConstraint are blocking update and merge on
#153 , but that's not the only issue we've been encountering.

This PR does the following in order to tidy things up and bring us 
a bit closer to the layout described in #214:

- Move WhereClauseConstraint conversion out of the query parser
- Move AggregationType to a root module
- Split NonAdditiveDimensionParams into a model object/spec pair
- - Clean up naming and update plan files

### DIFF
--- a/metricflow/aggregation_properties.py
+++ b/metricflow/aggregation_properties.py
@@ -1,0 +1,39 @@
+from __future__ import annotations
+
+from metricflow.object_utils import ExtendedEnum
+
+
+class AggregationType(ExtendedEnum):
+    """Aggregation methods for measures"""
+
+    SUM = "sum"
+    MIN = "min"
+    MAX = "max"
+    COUNT_DISTINCT = "count_distinct"
+    # BOOLEAN is deprecated. Remove when customers have migrated.
+    BOOLEAN = "boolean"
+    SUM_BOOLEAN = "sum_boolean"
+    AVERAGE = "average"
+
+    # COUNT = "count" not yet implemented ... non-expansive as COUNT(COUNT(X), COUNT(Y)) != COUNT(COUNT(X ∪ Y))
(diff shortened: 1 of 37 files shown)


tlento:
* main
  * **PR #224** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/224" target="_blank">Graphite</a> 
    * **PR #225** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/225" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/transform-data/metricflow/225?utm_source=stack-comment).

metricflow/aggregation_properties.py
```diff
@@ -0,0 +1,39 @@
+from __future__ import annotations
```
WilliamDee, line 1: what if we had like a `metricflow/model/properties/` directory that stores these? I'm not aware of the best practices, but personally I like how we have most things organized in their respective directories rather than being in the root directory.

WilliamDee (approved):
this looks great! Thank you for cleaning everything!

SIDE NOTE: we will need to make some changes to our internal SourceNodeBuilder to accommodate the change from `NonAdditiveDimensionParameters` to `NonAdditiveDimensionSpec`

metricflow/aggregation_properties.py
tlento: I think that'd be fine, although I pulled these to root because I think of them as being on the level of specs and suchlike. To be honest I thought about moving those into a sub-directory as well, but decided not to get too carried away, so I'm open to some simple file move reorganization PRs.

PR #230, opened 2022-08-29
Fix the names of the top level objects in the generated schema.json

jack-transform:
In https://github.com/transform-data/metricflow/pull/211 (generating an explicit JSON schema), I made the mistake of using the schema names as the names of the properties at the top level of the generated Metricflow schema. 

Materialization objects are defined by a schema named `materialization_schema`, so right now the schema matches `materialization_schema` for materialization definitions as opposed to the correct `materialization`. This fixes the schema to match materializations properly. (I also fixed `derived_identifier`, but it appears this may not no longer be in use)

### DIFF
--- a/metricflow/model/parsing/explicit_schema.py
+++ b/metricflow/model/parsing/explicit_schema.py
@@ -7,10 +7,10 @@ from typing import Dict, List, Union
 from metricflow.model.parsing import schemas_internal
 
 TOP_LEVEL_SCHEMAS = {
-    "metric",
-    "data_source",
-    "derived_group_by_element_schema",
-    "materialization_schema",
+    "metric": "metric",
+    "data_source": "data_source",
+    "derived_group_by_element_schema": "derived_identifier",
+    "materialization_schema": "materialization",
 }
 
 BASE_SCHEMA = {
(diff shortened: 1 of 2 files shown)


metricflow/model/parsing/schemas/metricflow.json
```diff
@@ -713,10 +713,10 @@
         "data_source": {
             "$ref": "#/definitions/data_source"
         },
-        "derived_group_by_element_schema": {
+        "derived_identifier": {
```
QMalcolm, line 716: I think we can remove "derived_identifier" (and probably should as I don't believe it's currently supported)

metricflow/model/parsing/explicit_schema.py
```diff
@@ -7,10 +7,10 @@
 from metricflow.model.parsing import schemas_internal
 
 TOP_LEVEL_SCHEMAS = {
-    "metric",
-    "data_source",
-    "derived_group_by_element_schema",
-    "materialization_schema",
+    "metric": "metric",
+    "data_source": "data_source",
+    "derived_group_by_element_schema": "derived_identifier",
```
QMalcolm, line 12: Similarly removing it here

QMalcolm (approved):
We should probably remove "derived_identifier", otherwise looks good

metricflow/model/parsing/schemas/metricflow.json
jack-transform: I think I will leave it here for now until we strip it all out of the code.

PR #235, opened 2022-09-01
Run poetry update

tlento:
We're hitting yanked version dependency problems with grpcio in CI,
updating lock file with dependency version updates.

PR #236, opened 2022-09-01
Format core yaml config files

tlento:
The Metricflow model authoring extension has opinions about how YAML
should be formatted, so let's format everything here instead of
polluting unrelated changes with a bunch of whitespace and quote
replacements. Even non-configs get formatted because the extension
pulls in its underlying YAML extension library, so this covers things that
aren't strictly model files.

Note this does not include model templates because we're not likely to change
them, happy to run the formatter there if it seems worthwhile.

### DIFF
--- a/.github/workflows/cd-push-to-pypi.yaml
+++ b/.github/workflows/cd-push-to-pypi.yaml
@@ -2,8 +2,8 @@ name: Publish Metricflow Release
 on:
   workflow_dispatch:
   push:
-    tags:        
-      - '*'
+    tags:
+      - "*"
 
 jobs:
   poetry-publish:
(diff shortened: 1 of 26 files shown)


tlento:
* main
  * **PR #235** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/235" target="_blank">Graphite</a> 
    * **PR #153** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/153" target="_blank">Graphite</a> 
      * **PR #236** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/236" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/transform-data/metricflow/236?utm_source=stack-comment).

tlento:
Something wonky is happening but I think we should integrate with pre-commit and run a pass with that instead of the manual changes

PR #238, opened 2022-09-02
fix dataclass deserializer errors

tlento:
Our dataclass deserializer was crashing whenever we encountered a serialized value
at rest that did not have a value for a newly added dataclass field with a fallback
default. This makes it impossible to safely add new fields, even if they are strictly
optional and can be defaulted to None or something similar.

This PR fixes the default value propagation and does bits and piece of cleanup
along the way in an effort to make this a little less cumbersome to work with.


metricflow/dataclass_serialization.py
```diff
@@ -21,6 +19,8 @@
 
 # Any Pydantic object
 PydanticT = TypeVar("PydanticT", bound=Type[BaseModel])
+# Any value
+AnyValueType: TypeAlias = Any  # type: ignore[misc]
```
WilliamDee, line 23: nice

WilliamDee (approved):
this is great, thanks!

PR #243, opened 2022-09-08
Add metadata context info to dimensions and identifiers

QMalcolm:
Previously we hadn't been tracking the metadata/config file context for dimensions and identifiers. Doing so allows us to provide better context during validations (giving line numbers). Additionally it's useful for model investigating/debugging for projects using MetricFlow.

Additionally we this PR refactors the initialization of identifier objects to use pydantic's initializer and validators instead of a custom `__init__` method which clobbers pydantic initialization. This was necessary as part of this PR because it was required for metadata parsing to work for identifiers.

### DIFF
--- a/metricflow/model/objects/elements/identifier.py
+++ b/metricflow/model/objects/elements/identifier.py
@@ -3,7 +3,8 @@ from __future__ import annotations
 from pydantic import validator
 from typing import Any, Optional, List
 
-from metricflow.model.objects.base import HashableBaseModel
+from metricflow.model.objects.base import HashableBaseModel, ModelWithMetadataParsing
+from metricflow.model.objects.common import Metadata
 from metricflow.object_utils import ExtendedEnum
 from metricflow.references import IdentifierReference, CompositeSubIdentifierReference
 
@@ -32,7 +33,7 @@ class CompositeSubIdentifier(HashableBaseModel):
         return CompositeSubIdentifierReference(element_name=self.name)
 
 
-class Identifier(HashableBaseModel):
+class Identifier(HashableBaseModel, ModelWithMetadataParsing):
     """Describes a identifier"""
 
     name: str
@@ -42,6 +43,7 @@ class Identifier(HashableBaseModel):
     entity: Optional[str]
     identifiers: List[CompositeSubIdentifier] = []
     expr: Optional[str] = None
+    metadata: Optional[Metadata]
 
     @validator("entity", always=True)
     @classmethod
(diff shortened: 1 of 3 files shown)


metricflow/model/objects/elements/identifier.py
```diff
@@ -42,30 +43,25 @@ class Identifier(HashableBaseModel):
...
+
+        If an entity value is provided that is a string, that will become the value of
+        entity. If the provifed entity value is None, the entity value becomes the
+        element_name representation of the identifier's name.
+        """
+
+        if value is None:
+            if "name" not in values:
+                raise ValueError("Failed to default entity value because objects name value was not defined")
+            return IdentifierReference(element_name=values["name"]).element_name
```
tlento, line 61: Do we need to do this reference construction? We basically drop it on the floor after this statement, and there's no real validation benefit to the IdentifierReference initializer since Pydantic will validate the "name" field on its own.

metricflow/model/objects/elements/identifier.py
```diff
@@ -41,31 +43,27 @@ class Identifier(HashableBaseModel):
...
+
+        # guarantee value is string
+        if not isinstance(value, str):
+            raise ValueError(f"Entity value should be a string (str) type, but got {type(value)} with value: {value}")
```
tlento, line 65: I think we can simplify to this:
```suggestion
        if value is None:
            if "name" not in values:
                raise ValueError("Failed to default entity value because objects name value was not defined")
            value = values["name"]

        # guarantee value is string
        if not isinstance(value, str):
            raise ValueError(f"Entity value should be a string (str) type, but got {type(value)} with value: {value}")
```

If "name" is not a string then Pydantic's validation should catch that, although I admit I don't know if there's an order in which these things get applied, so maybe an error message update would be useful.

tlento (approved):
This looks great! Minor quibble inline, do whatever you feel is best.

Also the build failures should be resolved on main now.

metricflow/model/objects/elements/identifier.py
QMalcolm: I didn't know if there was any magic that meant sometimes doing `IndentifierReference(...).element_name` would ever not be the name. It looks like currently that is the case. We just need to be aware of if the initialization of `IndentifierReference` ever in the future begins using any fancy logic to modify the `element_name`

metricflow/model/objects/elements/identifier.py
QMalcolm: I agree that Pydantic's validations **should** catch when "name" isn't a string. But as you, I'm not sure if there is an order, and what it is, in which validations get applied, which is why I include the check. It should be unnecessary 🤷  but it also shouldn't slow things down dramatically.

PR #242, opened 2022-09-08
Improve clarity of model-level element naming validation operations

tlento:
Our model-level element naming validation is enforced primarily through two
validators today - the ElementConsistencyRule and the
DataSourceMeasuresUnique rule.

This PR clarifies their relationship via updates to documentation and test layout,
and restructures the behavior of the ElementConsistencyRule to be more
clear as to its purpose.

PR #239, opened 2022-09-03
Remove duplicate logic in aggregation node creation

tlento:

PR #246, opened 2022-09-09
Add tests for measure constraints with a single aliased measure

tlento:
In most cases with measure constraints there will be more than
one measure (ratio metrics, multi-measure expr metrics), and we can
rely on the existing join -> filter path to ensure proper use of
the alias. However, the single measure case runs through a separate
code path in the dataflow plan builder.

This works today because the alias is internal to the aggregated
measures node, and the expr is required to use the alias rather
than the name, so the aggregated measures node gets an alias applied
and the query proceeds as expected.

This case was not tested previously, but it's sufficiently important
and easily overlooked, so we add a case to cover it.


tlento:
* main
  * **PR #235** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/235" target="_blank">Graphite</a> 
    * **PR #153** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/153" target="_blank">Graphite</a> 
      * **PR #239** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/239" target="_blank">Graphite</a> 
        * **PR #246** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/246" target="_blank">Graphite</a>  👈
          * **PR #242** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/242" target="_blank">Graphite</a> 
      * **PR #236** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/236" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/transform-data/metricflow/246?utm_source=stack-comment).

WilliamDee (approved):

plypaul (approved):

Issue #240, opened 2022-09-06
Enable support for roles via the `entity` property in identifiers


tlento:
There are many cases where identifiers represent entities in the world, but those entities can play different roles. For example, you might have a marketplace where users can be either buyers or sellers, or a lodging service where users can be guests or hosts.

The config spec for supporting this is to have each identifier `name` represent the role, and an optional `entity` value represent the specific entity. If no `entity` value is provided, the `name` represents both role and identity. We have test configs with a [basic user account entity setup](https://github.com/transform-data/metricflow/blob/main/metricflow/test/fixtures/model_yamls/simple_model/data_sources/users_latest.yaml#L19-L22) and a [corresponding entity/role mapping in a separate data source](https://github.com/transform-data/metricflow/blob/main/metricflow/test/fixtures/model_yamls/simple_model/data_sources/bookings_source.yaml#L60-L67).

The linked specifications theoretically allow you to write a query like:

```
mf query --metrics bookings --dimensions guest__home_state_latest
```

That should resolve the role of `guest` (based on the identifier name) to the entity `user` (based on entity value). Today it throws an exception because the logic to process this resolution has not yet been implemented.

tlento:
Open question for discussion:

Should we enforce that no `role` can point to a different entity if an entity for that role already exists?

To put this another way, any identifier that is a `role` can never be an `entity` in its own right, although it can perhaps serve as a role-specific entity join key. Essentially, we would require the following in all models:

1. If you have a `guest` role tagged as a `user` entity, you can never add a `guest` entity - it must always point back to `user`, even if it's the primary key identifier in a hypothetical `guest_attributes` table. By extension, all dimensions in that table are assigned a dimension entity of `user`, albeit with a `guest` role.
2. If you have a `guest` entity without an explicit entity resolution link pointing it at something else (e.g., `user`), no `guest` role can ever point to another entity - it must always point to `guest`.

marshallkd:
>Should we enforce that no role can point to a different entity if an entity for that role already exists?
I think this is important if we want global unique namespacing. For example, if we have:

```
data-source-1
  identifiers: 
    - name: employee
      entity: user
      type: foreign
  measures:
    - name: X
       ...

data-source-2
  identifiers: 
    - name: employee
      entity: employee
      type: foreign
  measures:
    - name: Y
       ...
```

We would run into conflicts if we asked for `mf query --metrics X Y --dimensions ...`

tlento:
[...]

PR #249, opened 2022-09-14
Add accessor method for data sources by identifier to DataSourceSemantics

QMalcolm:
With a semantic model we want to be able to inspect how things are connected within the model. This commit adds a new accessor method `get_data_sources_for_identifier` which allows one to see what data sources can be directly linked given an identifier.


metricflow/protocols/semantics.py
```diff
@@ -107,6 +107,11 @@ def get_aggregation_time_dimensions_with_measures(
         """Return all aggregation time dimensions in the given data source with their associated measures"""
         raise NotImplementedError
 
+    @abstractmethod
+    def get_data_sources_for_identifier(self, identifier_reference: IdentifierReference) -> List[DataSource]:
+        """Return all data sources associated with and identifier reference"""
```
tlento, line 112: Typooooooooo.....

```suggestion
        """Return all data sources associated with an identifier reference"""
```

metricflow/test/model/test_data_source_container.py
```diff
@@ -187,3 +187,11 @@ def test_local_linked_elements_for_metric(new_metric_semantics: MetricSemantics)
...
         "listing__capacity_latest",
     }
+
+
+def test_get_data_sources_for_identifier(new_data_source_semantics: DataSourceSemantics) -> None:  # noqa: D
+    identifier_reference = IdentifierReference(element_name="user")
+    linked_data_sources = new_data_source_semantics.get_data_sources_for_identifier(
+        identifier_reference=identifier_reference
+    )
+    assert len(linked_data_sources) == 10
```
tlento, line 197: There are 9 distinct data sources with a user entity pointer - bookings has the `user` identity twice, and as such `bookings` will appear twice in the result set.

Note - as far as our current join resolution is concerned, `bookings` cannot be joined to `user`. #240 will fix this, though, so I think we should keep it as-is unless that discrepancy is going to be a significant problem for you.

metricflow/model/semantics/semantic_containers.py
```diff
@@ -328,3 +328,8 @@ def get_aggregation_time_dimensions_with_measures(
             data_source_reference in self._data_source_to_aggregation_time_dimensions
         ), f"Data Source {data_source_reference} is not known"
         return self._data_source_to_aggregation_time_dimensions[data_source_reference]
+
+    def get_data_sources_for_identifier(self, identifier_reference: IdentifierReference) -> List[DataSource]:
+        """Return all data sources associated with and identifier reference"""
```
tlento, line 333: We should either document that this might return multiple pointers to a data source, or else alter the signature in the Protocol to return a distinct set (either `Set[DataSource]` or else `Dict[str, DataSource]` where the str represents the data source name).

tlento (commented):
Inline subtleties need some consideration. If the duplication is intended that's fine, just let me know.

metricflow/test/model/test_data_source_container.py
QMalcolm: I don't believe it's a significant problem (at this time).

tlento (approved):
LGTM!

PR #250, opened 2022-09-15
Clean up measures_by_additiveness behavior

tlento:
An earlier commit added a property for getting measures by additiveness,
and returned a tuple of tuples, which made it a bit unclear at the
callsite that these inner tuples were grouped by a particular key.
We change it to a dict to make that relationship more obvious.

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -527,9 +527,7 @@ class DataflowPlanBuilder(Generic[SqlDataSetT]):
             measures_by_additiveness = grouped_measures_by_additiveness.measures_by_additiveness
 
             # Build output nodes for each distinct non-additive dimension spec
-            for measure_specs in measures_by_additiveness:
-                assert len(measure_specs) > 0, "received empy set of measure specs, this should not happen!"
-                non_additive_spec = measure_specs[0].non_additive_dimension_spec
+            for non_additive_spec, measure_specs in measures_by_additiveness.items():
                 logger.info(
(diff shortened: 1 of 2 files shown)


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -527,9 +527,7 @@ def build_aggregated_measures(
             measures_by_additiveness = grouped_measures_by_additiveness.measures_by_additiveness
 
             # Build output nodes for each distinct non-additive dimension spec
-            for measure_specs in measures_by_additiveness:
-                assert len(measure_specs) > 0, "received empy set of measure specs, this should not happen!"
-                non_additive_spec = measure_specs[0].non_additive_dimension_spec
+            for non_additive_spec, measure_specs in measures_by_additiveness.items():
```
plypaul, line 530: `non_additive_spec` can be `None` here?

tlento: That is correct, and if the typechecker is doing its job that should also be clear from the variable type itself (`Optional[str]`). 

The runtime behavior is identical to what was there before, since measure_spec.non_additive_dimension_spec is `Optional`

Also, weird fact about Python, the value `None` can be used as a key in a dict. I was surprised the first time I saw it, but it makes a certain amount of sense given the original goals of the language.

plypaul: Got it. Mind adding a comment about it? Also, the logging line below would seem a little weird with `None`, so a case to output just `f"Building aggregated measures for {data_source}"` or similar would be helpful.

plypaul (approved):

PR #253, opened 2022-09-19
Semi-additive (Support alternate granularity)

WilliamDee:
## Extension of https://github.com/transform-data/metricflow/pull/183
Previously, we supported semi-additive measures in a way that you would only be allowed to take the non-additive dimension and perform a MIN/MAX aggregation against the full time range and get values from that big "window". In this PR, we break this down such that if the `non_additive_dimension` is passed in the query as a group by, it would do a `GROUP BY` in the inner query which enables us to have `windows` of the granularity passed in the query.

These are the two scenarios 
- `query --metrics ... --dimensions non_additive_dim(__week/__month/__year)`
    - Uses the granularity of `non_additive_dim` as the window to perform min/max
- `query --metrics ... --dimensions dim1 (--where non_additive_dim between X AND Y)`
    - Does a min/max of the full window (either `FULL RANGE` or between `X and Y`)

## Changes
- defaulted `window_choice` to `MIN` instead of required
- Added the ability to group by a TimeDimension in the `SemiAdditiveJoinNode` inner query so that we can perform window filters on different granularity windows instead of just on the full time range.
- Refactored the way we build the `SemiAdditiveJoinNode`
  - Reverted the changes in `SourceNodeBuilder`
  - Build the `SemiAdditiveJoinNode` in the `DataflowPlanBuilder` so that we can alter the node depending on the query spec
  - This also fixes the bug in which constraints was not working properly for semi additive measures

## Examples
```
if we have a data set that includes "account_balances,user,date", we can build a
"latest_account_balance_by_user" data set using this node by filtering the data set such that for each user,
we get the MAX(date) and return the account_balance corresponding to that date.

Data transformation example,
| date       | account_balance | user |                             | date       | account_balance | user |
|:-----------|-----------------|-----:|     identifier_specs:       |:-----------|-----------------|-----:|
| 2019-12-31 |            1000 |    u1|       - user                | 2020-01-03 |            2000 |    u1|
| 2020-01-03 |            2000 |    u1| ->  time_dimension_spec: -> | 2020-01-12              1500 |    u2|
| 2020-01-09 |            3000 |    u2|       - date                | 2020-01-12 |            1000 |    u3|
| 2020-01-12 |            1500 |    u2|     agg_by_function:
| 2020-01-12 |            1000 |    u3|       - MAX


Similarly, if we don't provide any identifier_specs, it would end up performing the aggregation filter only on the
time_dimension_spec without any grouping by any identifiers.

Data transformation example,
| date       | account_balance | user |                             | date       | account_balance |
|:-----------|-----------------|-----:|     identifier_specs:       |:-----------|----------------:|
| 2019-12-31 |            1000 |    u1|                             | 2020-01-12 |            2500 |
| 2020-01-03 |            2000 |    u1| ->  time_dimension_spec: ->
| 2020-01-09 |            3000 |    u2|       - date
| 2020-01-12 |            1500 |    u2|     agg_by_function:
| 2020-01-12 |            1000 |    u3|       - MAX


Additionally, we can aggregate against 'windows' of a dataset. For example, if we group by a non_additive time dimension,
we perform the MIN/MAX filtering on the granularity window provided in that group by instead of the full time range.

Data transformation example,
| date       | account_balance | user |                                  | date       | account_balance |
|:-----------|-----------------|-----:|     identifier_specs:            |:-----------|----------------:|
| 2019-12-31 |            1500 |    u1|     time_dimension_spec:         | 2019-12-31 |            1500 |
| 2020-01-03 |            2000 |    u1| ->    - date                  -> | 2020-01-07 |            3000 |
| 2020-01-09 |            3000 |    u2|     agg_by_function:             | 2020-01-14 |            3250 |
| 2020-01-12 |            1500 |    u2|       - MIN
| 2020-01-14 |            1250 |    u3|     queried_time_dimension_spec:
| 2020-01-14 |            2000 |    u2|       - date__week
| 2020-01-15 |            4000 |    u1|
```

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -33,6 +33,7 @@ from metricflow.dataflow.dataflow_plan import (
     WhereConstraintNode,
     WriteToResultDataframeNode,
     WriteToResultTableNode,
+    SemiAdditiveJoinNode,
     SinkOutput,
 )
 from metricflow.dataflow.dataflow_plan_to_text import dataflow_dag_as_text
@@ -47,6 +48,7 @@ from metricflow.plan_conversion.column_resolver import DefaultColumnAssociationR
 from metricflow.plan_conversion.node_processor import PreDimensionJoinNodeProcessor
 from metricflow.plan_conversion.sql_dataset import SqlDataSet
 from metricflow.plan_conversion.time_spine import TimeSpineSource
+from metricflow.references import TimeDimensionReference
 from metricflow.specs import (
     MetricSpec,
     MetricInputMeasureSpec,
@@ -58,6 +60,7 @@ from metricflow.specs import (
     InstanceSpec,
     DimensionSpec,
     OrderBySpec,
+    NonAdditiveDimensionSpec,
     SpecWhereClauseConstraint,
     LinkableSpecSet,
     ColumnAssociationResolver,
@@ -83,6 +86,16 @@ class MeasureRecipe(Generic[SqlDataSetT]):
     measure_node: BaseOutput[SqlDataSetT]
     required_local_linkable_specs: Tuple[LinkableInstanceSpec, ...]
     join_linkable_instances_recipes: Tuple[JoinLinkableInstancesRecipe, ...]
+
+
+@dataclass(frozen=True)
+class MeasureSpecProperties:
+    """Input dataclass for grouping properties of a sequence of MeasureSpecs."""
+
+    measure_specs: Sequence[MeasureSpec]
+    data_source_name: str
+    agg_time_dimension: TimeDimensionReference
+    non_additive_dimension_spec: Optional[NonAdditiveDimensionSpec] = None
 
 
 class DataflowPlanBuilder(Generic[SqlDataSetT]):
@@ -296,7 +309,7 @@ class DataflowPlanBuilder(Generic[SqlDataSetT]):
 
         return sorted(nodes, key=sort_function)
 
-    def _select_source_nodes_with_meausres(
+    def _select_source_nodes_with_measures(
         self, measure_specs: Set[MeasureSpec], source_nodes: Sequence[BaseOutput[SqlDataSetT]]
     ) -> Sequence[BaseOutput[SqlDataSetT]]:
         nodes = []
@@ -309,9 +322,61 @@ class DataflowPlanBuilder(Generic[SqlDataSetT]):
                 nodes.append(source_node)
         return nodes
 
+    def _find_non_additive_dimension_in_linkable_specs(
+        self,
+        agg_time_dimension: TimeDimensionReference,
+        linkable_specs: Sequence[LinkableInstanceSpec],
+        non_additive_dimension_spec: NonAdditiveDimensionSpec,
+    ) -> Optional[TimeDimensionSpec]:
+        """Finds the TimeDimensionSpec matching the non_additive_dimension_spec, if any"""
+        queried_time_dimension_spec: Optional[LinkableInstanceSpec] = None
+        for linkable_spec in linkable_specs:
+            dimension_name_match = linkable_spec.element_name == non_additive_dimension_spec.name
+            metric_time_match = (
+                non_additive_dimension_spec.name == agg_time_dimension.element_name
+                and linkable_spec.element_name == self._metric_time_dimension_reference.element_name
+            )
+            if dimension_name_match or metric_time_match:
+                queried_time_dimension_spec = linkable_spec
+                break
+        assert queried_time_dimension_spec is None or isinstance(
+            queried_time_dimension_spec, TimeDimensionSpec
+        ), "Non-additive dimension can only be a time dimension, if specified."
+        return queried_time_dimension_spec
+
+    def _build_measure_spec_properties(self, measure_specs: Sequence[MeasureSpec]) -> MeasureSpecProperties:
+        """Ensures that the group of MeasureSpecs has the same non_additive_dimension_spec and agg_time_dimension"""
+        if len(measure_specs) == 0:
+            raise ValueError("Cannot build MeasureParametersForRecipe when given an empty sequence of measure_specs.")
+        data_sources = self._get_data_source_names_for_measures(measure_specs)
+        if len(data_sources) > 1:
+            raise ValueError(
+                f"Cannot find common properties for measures {measure_specs} coming from multiple "
+                f"data sources: {data_sources}. This suggests the measure_specs were not correctly filtered."
+            )
+
+        agg_time_dimension = agg_time_dimension = self._data_source_semantics.get_agg_time_dimension_for_measure(
+            measure_specs[0].as_reference
+        )
+        non_additive_dimension_spec = measure_specs[0].non_additive_dimension_spec
+        for measure_spec in measure_specs:
+            if non_additive_dimension_spec != measure_spec.non_additive_dimension_spec:
+                raise ValueError(f"measure_specs {measure_specs} do not have the same non_additive_dimension_spec.")
+            measure_agg_time_dimension = self._data_source_semantics.get_agg_time_dimension_for_measure(
+                measure_spec.as_reference
+            )
+            if measure_agg_time_dimension != agg_time_dimension:
+                raise ValueError(f"measure_specs {measure_specs} do not have the same agg_time_dimension.")
+        return MeasureSpecProperties(
+            measure_specs=measure_specs,
+            data_source_name=data_sources.pop(),
+            agg_time_dimension=agg_time_dimension,
+            non_additive_dimension_spec=non_additive_dimension_spec,
+        )
+
     def _find_measure_recipe(
         self,
-        measure_specs: Sequence[MeasureSpec],
+        measure_spec_properties: MeasureSpecProperties,
         linkable_specs: Sequence[LinkableInstanceSpec],
         time_range_constraint: Optional[TimeRangeConstraint] = None,
     ) -> Optional[MeasureRecipe]:
@@ -320,13 +385,7 @@ class DataflowPlanBuilder(Generic[SqlDataSetT]):
         Prior to calling this method we should always be checking that all input measure specs come from
         the same base data source, otherwise the internal conditions here will be impossible to satisfy
         """
-        data_sources = self._get_data_source_names_for_measures(measure_specs)
-        if len(data_sources) > 1:
-            raise ValueError(
-                f"Cannot find a single measure recipe for measures {measure_specs} coming from multiple "
-                f"data sources: {data_sources}. This suggests the measure_specs were not correctly filtered."
-            )
-
+        measure_specs = measure_spec_properties.measure_specs
         node_processor = PreDimensionJoinNodeProcessor(
             data_source_semantics=self._data_source_semantics,
             node_data_set_resolver=self._node_data_set_resolver,
@@ -335,7 +394,7 @@ class DataflowPlanBuilder(Generic[SqlDataSetT]):
         source_nodes: Sequence[BaseOutput[SqlDataSetT]] = self._source_nodes
 
         # We only care about nodes that have all required measures
-        potential_measure_nodes: Sequence[BaseOutput[SqlDataSetT]] = self._select_source_nodes_with_meausres(
+        potential_measure_nodes: Sequence[BaseOutput[SqlDataSetT]] = self._select_source_nodes_with_measures(
             measure_specs=set(measure_specs), source_nodes=source_nodes
         )
 
@@ -556,42 +615,6 @@ class DataflowPlanBuilder(Generic[SqlDataSetT]):
                 ),
             )
 
-    @staticmethod
-    def _add_where_and_aggregate(
-        metric_input_measure_specs: Tuple[MetricInputMeasureSpec, ...],
-        queried_linkable_specs: LinkableSpecSet,
-        unaggregated_node: BaseOutput[SqlDataSetT],
-        where_constraint: SpecWhereClauseConstraint,
-    ) -> BaseOutput[SqlDataSetT]:
-        """Applies the where constraint appropriately and produce aggregated measures.
-
-        Assumes that the input node has all measure specs, queried_linkable_specs, and also the specs required by the
-        where clause.
-        """
-        where_constrained_node = WhereConstraintNode(
-            parent_node=unaggregated_node,
-            where_constraint=where_constraint,
-        )
-
-        if where_constraint.linkable_spec_set.is_subset_of(queried_linkable_specs):
-            return AggregateMeasuresNode[SqlDataSetT](
-                parent_node=where_constrained_node, metric_input_measure_specs=metric_input_measure_specs
-            )
-
-        # At this point, it's the case that the linkable specs in the where clause are not a subset of the queried
-        # linkable specs. A filter is needed after a where clause so that the linkable specs in the where clause don't
-        # show up in the final result.
-        #
-        # e.g. for "bookings" by "ds" where "is_instant", "is_instant" should not be in the results.
-        filtered_node = FilterElementsNode(
-            parent_node=where_constrained_node,
-            include_specs=tuple(x.measure_spec for x in metric_input_measure_specs) + queried_linkable_specs.as_tuple,
-        )
-
-        return AggregateMeasuresNode[SqlDataSetT](
-            parent_node=filtered_node, metric_input_measure_specs=metric_input_measure_specs
-        )
-
     def _build_aggregated_measures_from_measure_source_node(
         self,
         metric_input_measure_specs: Tuple[MetricInputMeasureSpec, ...],
@@ -606,6 +629,9 @@ class DataflowPlanBuilder(Generic[SqlDataSetT]):
             linkable_spec.element_name for linkable_spec in queried_linkable_specs.as_tuple
         ]
         measure_specs = tuple(x.measure_spec for x in metric_input_measure_specs)
+        measure_properties = self._build_measure_spec_properties(measure_specs)
+        non_additive_dimension_spec = measure_properties.non_additive_dimension_spec
+
         cumulative_metric_adjusted_time_constraint: Optional[TimeRangeConstraint] = None
         if cumulative and time_range_constraint is not None:
             logger.info(f"Time range constraint before adjustment is {time_range_constraint}")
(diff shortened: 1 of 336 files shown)


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -450,8 +452,48 @@ def _find_measure_recipe(
...
+            # Since every build_aggregate_measures function call groups measure_specs by additiveness
+            # all of these measure_specs will have the same non_additive_dimension_spec
+            non_additive_dimension_spec: Optional[NonAdditiveDimensionSpec] = measure_specs[
+                0
+            ].non_additive_dimension_spec
+            if non_additive_dimension_spec is not None:
+                # Gets the dimension given in the query that matches the non-additive dimension
+                primary_time_dimension = self._data_source_semantics.find_primary_time_dimension(
+                    self._data_source_semantics.get_data_sources_for_measure(measure_specs[0].as_reference)[0]
+                )
```
tlento, line 465: These two accessors are making assumptions about how this method will be called, and they'll produce incorrect results if those assumptions ever change.

I'd prefer to make this part of the method signature, or find some other way of ensuring that this is correct. 

We could also bundle these up top, where the assertion about the data sources lives, and add invariant checks at that point.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -450,8 +452,48 @@ def _find_measure_recipe(
...
+                data_set = self._node_data_set_resolver.get_output_data_set(measure_node)
+                queried_time_dimension_spec: Optional[LinkableInstanceSpec] = None
+                for linkable_spec in linkable_specs:
+                    dimension_name_match = linkable_spec.element_name == non_additive_dimension_spec.name
+                    metric_time_match = (
+                        non_additive_dimension_spec.name == primary_time_dimension.element_name
```
tlento, line 471: Does this work with custom measure agg time dimensions? The `find_primary_time_dimension` checks at the data source level, and I don't see anything looking at the measure itself, but the measure might well have a different time dimension set. Does the source node builder step deal with this for us? If so it'd be good to document this behavior, and I recommend doing it via a helper method that returns the queried_time_dimension_spec and can then be documented independently with its assumptions.

It's actually kind of unclear to me how all of these things fit together, honestly, so more clarity in this block would be super helpful.

metricflow/model/semantics/semantic_containers.py
```diff
@@ -248,6 +248,13 @@ def get_identifier_references(self) -> List[IdentifierReference]:  # noqa: D
         return list(self._identifier_ref_to_entity.keys())
 
     # DSC interface
+    @staticmethod
+    def find_primary_time_dimension(data_source: DataSource) -> TimeDimensionReference:  # noqa: D
+        for dimension in data_source.dimensions:
+            if dimension.type == DimensionType.TIME and dimension.type_params and dimension.type_params.is_primary:
```
tlento, line 254: See this goes to the DataSource, not the source node set, which is why I was wondering if this accounts for metric_time usage on measures with a different agg_time_dimension.

It's possible for a data source to have no primary time dimension, I think, or it used to be. Regardless it's possible for data_source.primary_time to never be referenced anywhere.

tlento (commented):
Looks super promising! I have some questions and minor comments inline. I think it'd be good if @plypaul took a look as well, he might see some things we're missing.

metricflow/model/objects/elements/measure.py
```diff
@@ -14,7 +14,9 @@ class NonAdditiveDimensionParameters(HashableBaseModel):
     """
 
     name: str
-    window_choice: AggregationType
+
+    # Optional Fields
+    window_choice: AggregationType = AggregationType.MIN
```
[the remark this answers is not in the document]
WilliamDee: I talked with Nick here, we wanted to default to MIN simply because of how our granularity works and the preference of a minimal config. For example, if we did `ds__week` and say we get 3 rows of `2022-01-01`, `2022-01-08`, `2022-01-15` if we had a MIN default, the result would look like the following,
```
2022-01-01  | value at 2022-01-01
2022-01-08  | value at 2022-01-08
2022-01-15  | value at 2022-01-15
```
If we did max, the result would look like this 
```
2022-01-01  | value at 2022-01-07
2022-01-08  | value at 2022-01-14
2022-01-15  | value at 2022-01-21
```
so intuitively it makes more sense to default as MIN rather than MAX. Open for further discussion here though or we could just simply keep this as required until we decide which is better.


As for the `AggregationType`, I agree with it being weird, but I was going to add a validation on that to ensure it's `MIN/MAX`.  I was thinking that would be enough and makes sure it gets blocked on usage of other types.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1456,12 +1456,12 @@ def visit_semi_additive_join_node(self, node: SemiAdditiveJoinNode) -> SqlDataSe
...
                 ),
             ),
-            column_alias=time_dimension_column_name,
+            column_alias=f"{time_dimension_column_name}__{node.agg_by_function.value}",
         )
         column_equality_descriptions.append(
             ColumnEqualityDescription(
                 left_column_alias=time_dimension_column_name,
-                right_column_alias=time_dimension_column_name,
+                right_column_alias=time_dimension_select_column.column_alias,
```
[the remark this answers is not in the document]
WilliamDee: yea, we don't actually access this anywhere outside the subquery. I made this change since I actually had issues with BQ where the subquery was something like this 
```sql
SELECT 
  MIN(ds) AS ds
FROM data_set
GROUP BY ds
```
this threw a `cannot group by an aggregate` error. So by making this local alias, it allowed it to group by properly. So unless we might have a use case of `MIN(ds)` anywhere outside of the subquery I assumed it would be fine to do this.

metricflow/model/objects/elements/measure.py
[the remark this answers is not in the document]
tlento: The validation is fine for now, but a separate enum helps us out more, I think, because we can rely on the typechecker (and then the validation comes along with the JSON schema update).

The MIN thing makes sense, it's worth documenting that reasoning somewhere. I think the user-facing docs are the right place, actually, as long as it's not too in the weeds.

metricflow/model/semantics/semantic_containers.py
WilliamDee: now that I think about it, you're right. This doesn't cover the case of measures with different agg_time_dimension. With more thought on that, I actually realized we didn't need to do any of the finding primary time dimension stuff. Since the source node splits the nodes into measures with the same agg_time_dimension. I can just track the measure's agg_time_dimension and compare that to the non_additive_dimension_spec to check whether `metric_time` is a match or not.  So I ended up wiping all the code in favour of that. https://github.com/transform-data/metricflow/pull/253/commits/da5ce02dc2b0922e3a90f505e450c9abeea8f24b

metricflow/dataflow/builder/dataflow_plan_builder.py
WilliamDee: addressed here https://github.com/transform-data/metricflow/pull/253#discussion_r977898952

metricflow/dataflow/builder/dataflow_plan_builder.py
WilliamDee: addressed here https://github.com/transform-data/metricflow/pull/253#discussion_r977898952

metricflow/dataflow/dataflow_plan.py
```diff
@@ -454,11 +454,38 @@ class SemiAdditiveJoinNode(Generic[SourceDataSetT], BaseOutput[SourceDataSetT]):
...
+
+    Data transformation example,
+    | date       | account_balance | user |                                  | date       | account_balance |
+    |:-----------|-----------------|-----:|     identifier_specs:            |:-----------|----------------:|
+    | 2019-12-31 |            1500 |    u1|     time_dimension_spec:         | 2019-12-31 |            1500 |
+    | 2020-01-03 |            2000 |    u1| ->    - date                  -> | 2020-01-07 |            3000 |
+    | 2020-01-09 |            3000 |    u2|     agg_by_function:             | 2020-01-14 |            3250 |
+    | 2020-01-12 |            1500 |    u2|       - MIN
+    | 2020-01-14 |            1250 |    u3|     queried_time_dimension_spec:
+    | 2020-01-14 |            2000 |    u2|       - date__week
```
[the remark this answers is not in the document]
tlento: Never mind, I think I was getting confused by the identifier in the data itself.

The contract for semi-additive measures is that we can aggregate *within* values for the non-additive dimension, but not *across* them, and this is exactly what happens here. While it might be odd for somebody to ignore the `user` entry in this case, it's not wrong - in fact, if they really don't care about the `user` under normal circumstances (or if they're never allowed to have user-level aggregations in their reporting pipeline) this is exactly what they'd want.

metricflow/model/parsing/schemas.py
```diff
@@ -155,7 +155,7 @@
         },
     },
     "additionalProperties": False,
-    "required": ["name", "window_choice"],
+    "required": ["name"],
```
tlento, line 158: For validating aggregation type we could consider making it a restricted enum in the JSON schema here. It does seem cleaner to add a custom enum type for this, though - I suspect we'll need it for stuff like SCDs as well.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -309,6 +312,28 @@ def _select_source_nodes_with_meausres(
                 nodes.append(source_node)
         return nodes
 
+    def _find_queried_time_dimension(
+        self,
+        agg_time_dimension: TimeDimensionReference,
+        linkable_specs: Sequence[LinkableInstanceSpec],
+        non_additive_dimension_spec: NonAdditiveDimensionSpec,
+    ) -> Optional[TimeDimensionSpec]:
+        """Finds the equivalent TimeDimensionSpec in the sequence of linkable specs that matches the non_additive_dimension_spec."""
```
tlento, line 321: Nit - shorten this to avoid the 120 char overrun.

```
"""Finds the TimeDimensionSpec matching the non_additive_dimension_spec, if any"""
```

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -309,6 +312,28 @@ def _select_source_nodes_with_meausres(
                 nodes.append(source_node)
         return nodes
 
+    def _find_queried_time_dimension(
```
tlento, line 315: This is a very generic name. Do we want to make it more non-additive specific?

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -450,8 +475,36 @@ def _find_measure_recipe(
...
+            measure_node = node_with_lowest_cost
+            # Since every build_aggregate_measures function call groups measure_specs by additiveness
+            # all of these measure_specs will have the same non_additive_dimension_spec
+            non_additive_dimension_spec: Optional[NonAdditiveDimensionSpec] = measure_specs[
+                0
+            ].non_additive_dimension_spec
+            if non_additive_dimension_spec is not None:
+                agg_time_dimension = self._data_source_semantics.get_agg_time_dimension_for_measure(
+                    measure_specs[0].as_reference
+                )
```
tlento, line 487: These are still unchecked. If you look on line 349 we have an invariant check enforcing the internal contract for `_find_measure_recipe`. I think we need to extend that check, because we are now relying on this extended set of conditions (i.e., there is only one agg time dimension and only one non-additive spec) internally.

You can update the contract in one of two ways - either do it via a runtime assertion, or update the signature to accept a non-additive dimension spec and an agg time dimension. The former is simplest and probably cleanest, especially given that this is an internal helper method. The latter is more ergonomic as the signature itself tells developers what to expect.

The best thing might be to overhaul the signature to take in an object representing the measure specs, with properties for data source, agg time dimension, and non-additive dimension. That object can then maintain the invariants and encapsulate the work of doing the extraction of these values at the callsites. But it's more work.

tlento (commented):
I think this is good, but I'd value another set of eyes. Still not quite there, myself. Stupid COVID.

Anyway, please do add the invariant or helper object or whatever you deem most appropriate for ensuring `_find_measure_recipe` always gets appropriate input.

metricflow/plan_conversion/dataflow_to_sql.py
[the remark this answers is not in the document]
plypaul: One of the invariants is that all column aliases are generated through the column association resolver. Is this something you can handle there?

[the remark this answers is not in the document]
WilliamDee: as discussed offline, we are adding `Optional[AggregationState]` to `resolve_time_dimension_spec` and just in this semi_additive_join scenario, I’ll be passing in `AggregationState.COMPLETE` and we would render that column alias as `{time_dim}__complete` and that suffix `*__complete` may get translated to something else that’s simpler in the future.

metricflow/dataflow/builder/dataflow_plan_builder.py
WilliamDee: I added the invariant checks with a helper function that takes in a `Sequence[MeasureSpec]` and it asserts that all those measure_specs has the same `agg_time_dimension` and `non_additive_dimension_spec`

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -309,6 +320,48 @@ def _select_source_nodes_with_meausres(
...
+    ) -> Optional[TimeDimensionSpec]:
+        """Finds the TimeDimensionSpec matching the non_additive_dimension_spec, if any"""
+        queried_time_dimension_spec: Optional[LinkableInstanceSpec] = None
+        for linkable_spec in linkable_specs:
+            dimension_name_match = linkable_spec.element_name == non_additive_dimension_spec.name
+            metric_time_match = (
+                non_additive_dimension_spec.name == agg_time_dimension.element_name
+                and linkable_spec.element_name == self._metric_time_dimension_reference.element_name
+            )
+            if dimension_name_match or metric_time_match:
```
plypaul, line 337: Are there issues if the user queries for `metric_time__week` and `metric_time__month`? Or is this validated somewhere?

WilliamDee: They would be allowed to do that

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -450,8 +503,35 @@ def _find_measure_recipe(
                 for y in x.join_on_partition_time_dimensions
             )
 
+            measure_node = node_with_lowest_cost
+
+            # Retrieves the grouped non_additive_dimension_spec and agg_time_dimension
+            measure_properties = self._build_measure_spec_properties(measure_specs)
```
tlento, line 509: This is nice! I would love this to be part of the method signature, maybe containing the measure spec sequence as well. So this method would take in a MeasureSpecPropeties or MeasureSpecsByCommonProperties or whatever we end up calling it, and it's obvious to all of us that we want a collected set of measures. Maybe we could even put in the data source name and collect these at the very top, and then remove all of the validation checks inside of this method.

This is getting into separate PR territory, though, so I think it's fine as-is. Maybe add a TODO comment if you don't pull this into the method signature straightaway.

tlento (commented):
Ugh time granularities. Do we need another validation? Might be worth a quick chat separately just to make sure we're doing the right things here.

As long as our granularity handling is correct here this seems good to go to me. If we need another validation to prevent weirdness we can add it to the validation PR.

metricflow/dataflow/builder/dataflow_plan_builder.py
WilliamDee: Did this in this PR!

metricflow/dataflow/builder/dataflow_plan_builder.py
WilliamDee: We are validating that in here https://github.com/transform-data/metricflow/pull/259

metricflow/test/integration/test_cases/itest_semi_additive_measure.yaml
```diff
@@ -106,3 +109,119 @@ integration_test:
...
+        SELECT
+          ds
+        FROM {{ source_schema }}.fct_accounts
+      ) c
+      WHERE ds >= '2020-01-03'
+    ) d
+    ON
+      b.ds = d.ds__complete
+    GROUP BY
+      b.account_type
```
tlento, line 227: Newline! 😝

metricflow/test/model/validations/test_element_const.py
```diff
@@ -36,8 +36,8 @@ def test_cross_element_names(simple_model__pre_transforms: UserConfiguredModel)
 
     # We update the matching categorical dimension by reference for convenience
     ds_measure_x_dimension.get_dimension(dimension_reference).name = measure_reference.element_name
-    ds_measure_x_identifier.identifiers[1].name = measure_reference.element_name
-    ds_dimension_x_identifier.identifiers[1].name = dimension_reference.element_name
+    ds_measure_x_identifier.identifiers[0].name = measure_reference.element_name
+    ds_dimension_x_identifier.identifiers[0].name = dimension_reference.element_name
```
tlento, line 40: Thanks!

tlento (commented):
Seems reasonable! We need to refactor the dataflow plan builder, it's getting out of control, but I think that's beyond the scope of this PR.

I just realized we should probably run a whole lot of tests so I'll leave this in reviewer queues for a bit in case @plypaul wants to weigh in.

metricflow/specs.py
```diff
@@ -300,6 +302,15 @@ def bucket_hash(self) -> str:
...
         return hash_strings(values)
 
+    @property
+    def as_time_dimension_spec(self) -> TimeDimensionSpec:  # noqa: D
+        return TimeDimensionSpec.from_name(self.name)
+
+    def __eq__(self, other: Any) -> bool:  # type: ignore[misc] # noqa: D
+        if not isinstance(other, NonAdditiveDimensionSpec):
+            return False
+        return self.bucket_hash == other.bucket_hash
```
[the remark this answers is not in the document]
WilliamDee: I was actually not aware of that! This shouldn't be needed then!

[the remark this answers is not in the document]
WilliamDee: ooo wait, just tested it, it doesn't actually work. It converts the values of the as a tuple in the order of the fields so it has nothing to do with the value itself. ie., if we had this
```
NonAdditiveDimensionSpec(
   name="asd", window_choice="min",window_groupings=(1,2)
)
```
it would convert it to 
```
("asd", "min", (1,2))
```
which would not be equal to say 
```
("asd", "min", (2,1))
```
which it should be equal

tlento (approved):
336 files changed?! We might need to revisit this whole snapshot business... 🗄️

WilliamDee:
> 336 files changed?! We might need to revisit this whole snapshot business... 🗄️ 

Big change due to reverting the source node building logic which happens for every query!

PR #254, opened 2022-09-19
Enumerate Future SQL Engines

plypaul:
This PR adds enumerations for SQL engines that are not yet supported. There are some intermediate development uses cases where this is helpful.

### DIFF
--- a/metricflow/protocols/sql_client.py
+++ b/metricflow/protocols/sql_client.py
@@ -10,8 +10,8 @@ from metricflow.sql.sql_bind_parameters import SqlBindParameters
 from pandas import DataFrame
 
 
-class SupportedSqlEngine(Enum):
-    """Enumeration of DB engines currently supported by MetricFlow"""
+class SqlEngine(Enum):
+    """Enumeration of SQL engines, including ones that are not yet supported."""
 
     BIGQUERY = "BigQuery"
     DUCKDB = "DuckDB"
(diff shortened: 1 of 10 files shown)


metricflow/protocols/sql_client.py
```diff
@@ -10,15 +10,19 @@
 from pandas import DataFrame
 
 
-class SupportedSqlEngine(Enum):
-    """Enumeration of DB engines currently supported by MetricFlow"""
+class SqlEngine(Enum):
+    """Enumeration of SQL engines, including ones that are not yet supported."""
```
tlento, line 14: As context, I was strict on this originally because it was causing problems on the closed source side - it was very hard to figure out what was actually working and what was not - and I did not want to duplicate that here. This also made it clear which implementations were supportable on closed source and which were not, since we had things like MySQL clients that looked an awful lot like a SqlClient but weren't really (and weren't meant to be, it's just there are only so many query interfaces one might need).

The other issue here is we always have to switch through the unsupported ones if we want to use the enum as a complete set. I don't know if this is a practical concern, though, since we pretty much never operate on more than one of these at a time outside of tests.

tlento (commented):
Is this a short term hack around an annoyance? If so, do we really need to commit it? If not, I'm a bit confused by this, would you mind clarifying the use case?

I think this is probably ok in the short term if there's a reasonable justification, but I'd like to see if there's a cleaner approach.

tlento (approved):
Ok I just took a closer look at this thing and the right answer is, in fact, to remove this enum altogether as it serves no real purpose. When I get some time I'll rip this out along with the properties struct that contains it, but that'll be..... much later.

So whatever it is you're doing here seems fine in the short term as long as you're not adding rigid dependencies on this class.

plypaul:
Yeah, not adding new dependencies with this. We have an incomplete implementation for the MySQL `SqlClient` and this helps with existing code until that's complete.

PR #255, opened 2022-09-22
Added Validator for non_additive_dimension property

WilliamDee:
## Context
_This should be merged after this https://github.com/transform-data/metricflow/pull/253 goes in._ **actually this can be merged prior to this**

With the addition to semi-additive measures. We introduced a new `non_additive_dimension` property on measures. There are a few cases in which we need to validate such that the config is correct and ensures that the resultant query that is built is well-formed. There are 3 cases in that needs to be validated for a valid `non_additive_dimension`

1. `non_additive_dimension.name` must be a valid time-dimension in the data source
2. `non_additive_dimension.window_choice` must be either `min` or `max`
3. `non_additive_dimension_window_groupings` must be valid identifiers in the data source

### DIFF
--- a/metricflow/test/model/validations/test_measures.py
+++ b/metricflow/test/model/validations/test_measures.py
@@ -2,6 +2,8 @@ import copy
 import re
 
 import pytest
+from metricflow.aggregation_properties import AggregationType
+from metricflow.model.objects.elements.measure import Measure, NonAdditiveDimensionParameters
 from metricflow.model.objects.metric import Metric, MetricInputMeasure, MetricType, MetricTypeParams
 from metricflow.model.objects.user_configured_model import UserConfiguredModel
 from metricflow.model.model_validator import ModelValidator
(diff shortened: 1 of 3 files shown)


metricflow/test/model/validations/test_measures.py
```diff
@@ -221,3 +223,33 @@ def test_reused_measure_alias_within_metric(simple_model__pre_transforms: UserCo
...
+        f"that is not defined in data source '{model.data_sources[0].name}' as a time dimension."
+    )
+    expected_error_substring_2 = "has a non_additive_dimension with an invalid 'window_choice'"
+    expected_error_substring_3 = "has a non_additive_dimension with an invalid 'window_groupings'"
+    for expected_str in [expected_error_substring_1, expected_error_substring_2, expected_error_substring_3]:
+        exist = False
+        for actual_str in build.issues.errors:
+            if actual_str.as_readable_str().find(expected_str) != -1:
+                exist = True
+        assert exist, f"Expected error {expected_str} not found in error string! Instead got {actual_str}"
```
tlento, line 255: I think the error message on this last assertion only prints out the last actual_str value you happen to have seen.

I suggest something like this - I find it a little nicer to read and debug (well, once the formatting gets fixed up, anyway)

```suggestion
    missing_error_strings: Set[str] = {}
    for expected_str in [expected_error_substring_1, expected_error_substring_2, expected_error_substring_3]:
        if not any(actual_str.as_readable_str().find(expected_str) != -1 for actual_str in build.issues.errors):
            missing_error_strings.add(expected_str)

    assert len(missing_error_strings) == 0, f"Failed to match one or more expected errors: {missing_error_strings} in {set([x.as_readable_string() for x in build.issues.errors])}"
```

tlento (approved):
There are a handful of messaging items inline that need some attention before merge. Thanks!

QMalcolm (approved):
This new rule looks good!

PR #256, opened 2022-09-22
Enable the ability to not pass dimensions in interfaces

WilliamDee:
## Context
Currently, all the interfaces requires dimensions to be a non-empty list of inputs. This should not be the case as it is possible to not pass any dimension which outputs a metric without any group by dimension and returns an aggregate value of just the metric.

### Changes
- In `MetricFlowClient`, remove the requirement of a non-empty dimension input
- In the CLI, remove the requirement of a non-empty dimension input
- Restrict the ability to query for multiple metrics with no group bys due to a bug (Will re-enable once we implement cross join)

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -608,6 +608,12 @@ class DataflowPlanBuilder(Generic[SqlDataSetT]):
         if len(output_nodes) == 1:
             return output_nodes[0]
         else:
+            if len(queried_linkable_specs.as_tuple) == 0:
+                raise NotImplementedError(
+                    "Querying metrics that require a join with no dimensions is not yet supported."
+                    "This can happen with a metric that pulls measures from different data sources, "
+                    "or a metric that pulls measures with different constraints or non-additive dimension specifications."
+                )
             return FilterElementsNode(
                 parent_node=JoinAggregatedMeasuresByGroupByColumnsNode(parent_nodes=output_nodes),
                 include_specs=LinkableInstanceSpec.merge(
--- a/metricflow/test/integration/test_cases/itest_simple.yaml
+++ b/metricflow/test/integration/test_cases/itest_simple.yaml
@@ -14,6 +14,43 @@ integration_test:
       is_instant
 ---
 integration_test:
+  name: simple_query_with_no_group_bys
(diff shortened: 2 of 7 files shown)


tlento:
Is this open for review or still WIP?

WilliamDee:
> Is this open for review or still WIP?

For this PR itself, it is good, but there is a bug that exist for querying multiple metrics without dimensions so looking into that prior to getting this checked off! Thanks for checking in!

metricflow/test/integration/test_cases/itest_simple.yaml
```diff
@@ -13,6 +13,43 @@ integration_test:
     GROUP BY
       is_instant
 ---
+integration_test:
+  name: simple_query_with_no_group_bys
+  description: Tests selecting a metric with no group bys.
+  model: SIMPLE_MODEL
+  metrics: ["booking_value"]
```
tlento, line 20: Can you try this on a ratio metric requiring a join? Something like `bookings_per_listing` should do it. Last time I tried it generated invalid SQL, which means users will hit failure states with difficult to understand SQL syntax errors.

WilliamDee: oo yea this breaks on ratio metrics requiring a join.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -608,6 +608,10 @@ def build_aggregated_measures(
         if len(output_nodes) == 1:
             return output_nodes[0]
         else:
+            if len(queried_linkable_specs.as_tuple) == 0:
+                raise NotImplementedError(
+                    "Querying metrics that requires a join with no dimensions is not yet supported."
+                )
```
tlento, line 614: Let's give the user a little bit more context here. Something like this, perhaps, but with proper formatting.

```suggestion
                raise NotImplementedError(
                    "Querying metrics that require a join with no dimensions is not yet supported. This can happen with a metric that pulls measures from different data sources, or a metric that pulls measures with different constraints or non-additive dimension specifications."
                )
```

tlento (approved):
Ok, I think this will do for now. When we implement the join `on true` thing and whatever fix is appropriate for the multiple measures join case we should add some tests exercising those query paths.

PR #257, opened 2022-09-26
Add more helpful error message when file decoding fails

QMalcolm:
We've been running into an issue recently, specifically with Windows users wherein ASCII content is being copied to and saved in a UTF-8 file. During parsing this causes a cryptic error message

```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 2907500: character maps to `<undefined>`
```

to occur during reading of the model config files. Assuming / guessing the content of the file (and ignoring unknown characters) seems dangerous as this has potential to make parts of the config invalid during validation and thus becoming even harder to debug. Additionally, files could potentially contain mixed encoded content. Thus the safest thing to do seems to be explain what is happening and offer possible remediations.

### DIFF
--- a/metricflow/model/parsing/dir_to_model.py
+++ b/metricflow/model/parsing/dir_to_model.py
@@ -136,11 +136,24 @@ def parse_yaml_file_paths_to_model(
     template_mapping = template_mapping or {}
     yaml_config_files = []
     for file_path in file_paths:
-        with open(file_path) as f:
-            contents = Template(f.read()).substitute(template_mapping)
-            yaml_config_files.append(
-                YamlConfigFile(filepath=file_path, contents=contents),
-            )
+        try:
+            with open(file_path) as f:
+                contents = Template(f.read()).substitute(template_mapping)
+                yaml_config_files.append(
+                    YamlConfigFile(filepath=file_path, contents=contents),
+                )
+        except UnicodeDecodeError as e:
+            # We could alternatively return this as a validation issue, but this
+            # exception is hit *before* building the model. Currently the
+            # ModelBuildResult guarantees a UserConfiguredModel. We could make
+            # UserConfiguredModel optional on ModelBuildResult, but this has
+            # undesirable consequences.
+            raise Exception(
+                f"The content of file `{file_path}` doesn't match the encoding of the file."
+                " If you know the encoding the content is in, try resaving the file with that encoding explicitly."
+                " Alternatively this error generally arises due to copy and pasted content,"
+                " try manually typing up the problem file instead of copy and pasting"
+            ) from e
 
     build_result = parse_yaml_files_to_model(yaml_config_files)
     model = build_result.model


metricflow/model/parsing/dir_to_model.py
```diff
@@ -136,11 +136,24 @@ def parse_yaml_file_paths_to_model(
...
+                    YamlConfigFile(filepath=file_path, contents=contents),
+                )
+        except UnicodeDecodeError as e:
+            # We could alternatively return this as a validation issue, but this
+            # exception is hit *before* building the model. Currently the
+            # ModelBuildResult guarantees a UserConfiguredModel. We could make
+            # UserConfiguredModel optional on ModelBuildResult, but this has
+            # undesirable consequences.
+            raise Exception(
+                f"The content of file `{file_path}` doesn't match the encoding of the file."
```
tlento, line 152: Is it possible to provide the expected encoding in this error message? I think that'd be helpful for people if this throws on a file they've typed up or built programmatically or whatever.

tlento (approved):
I really hope this doesn't mean we don't support unicode emojis in descriptions. I mean, what if I NEED a snowman to properly express the meaning of some metric?

(I'm kidding. Mostly. ⛄ )

metricflow/model/parsing/dir_to_model.py
QMalcolm: `open()` allows one to specify the encoding. If an encoding is not specified I believe it tries to use the encoding specified in the file header, and if that isn't present uses the system default. I've been digging through python's source code and haven't found the implementation yet, but it's a matter of time. Python has to determine it somehow.

QMalcolm: Think I found it in [_pyio.py](https://github.com/python/cpython/blob/d79dd929accac13832bc13ed12be34466584ca81/Lib/_pyio.py#L77). Getting what encoding won't be a single line of code, but also isn't too complicated. I think it'll be something like
```
encoding = text_encoding()
if encoding == 'locale':
  try:
    import locale
  except ImportError:
    # Importing locale may fail if Python is being built
    return "utf-8"
  else:
    return locale.getencoding()
```

This duplicates the necessary logic without having to instantiate a `TextIOWrapper` object like [open(...)](https://github.com/python/cpython/blob/main/Lib/_pyio.py#L274) does

QMalcolm: Woof. It's actually not that straight forward. It looks like `text_encoding()` doesn't exist prior to python 3.10, and I think we're currently stuck on a versions less than 3.10. Additionally in Python 3.9.13 the following is used
```
                try:
                    import locale
                except ImportError:
                    # Importing locale may fail if Python is being built
                    encoding = "ascii"
                else:
                    encoding = locale.getpreferredencoding(False)
```
Instead of the code in the previous comment (starting with `try:`)

QMalcolm: I think we'd have to revisit our "what encoding was used for decoding" code anytime we change python versions until we think this bit of python source code has stabilized 😬

QMalcolm: Going to merge as is, and let surfacing which encoding was used for decoding to be a future improvement.

PR #259, opened 2022-09-27
Add more validations to NonAdditiveDimension

WilliamDee:
### Context
There is a issue where unexpected behaviour occurs when you try to have a different base time granularity for a measure's non_additive_dimension and it's agg_time_dimension. This validation is to ensure that both the base time granularities are the same.

### DIFF
--- a/metricflow/model/validations/measures.py
+++ b/metricflow/model/validations/measures.py
@@ -230,6 +230,32 @@ class MeasuresNonAdditiveDimensionRule(ModelValidationRule):
                 non_additive_dimension = measure.non_additive_dimension
                 if non_additive_dimension is None:
                     continue
+                agg_time_dimension = next(
+                    (
+                        dim
+                        for dim in data_source.dimensions
+                        if measure.checked_agg_time_dimension.element_name == dim.name
+                    ),
+                    None,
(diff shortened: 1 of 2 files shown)


metricflow/model/validations/measures.py
```diff
@@ -263,10 +273,54 @@ def validate_model(model: UserConfiguredModel) -> List[ValidationIssueType]:  #
...
+                            )
+                        )
+
+                    # Validates that the non_additive_dimension time_granularity is >= agg_time_dimension time_granularity
+                    if (
+                        matching_dimension.type_params
+                        and agg_time_dimension.type_params
+                        and (
+                            matching_dimension.type_params.time_granularity.name
+                            != agg_time_dimension.type_params.time_granularity.name
```
tlento, line 305: Is this meaningfully different from `matching_dimension.type_params.time_granularity != agg_time_dimension.type_params.time_granularity`? I think it amounts to the same thing, in which case I'd prefer the simpler direct comparison.

WilliamDee: ah yes, i added that there as a test due to a bug I encountered and forgot to take it out! thanks!

Issue #25, opened 2022-04-08
Add support for Databricks


tlento:
**Describe the Feature**
MetricFlow should be able to issue queries against Databricks. This requires:

1. Adding a Databricks SqlClient (with its associated attributes struct), 
2. (If necessary) updating the SQL rendering to handle any dialect-specific weirdness
3. Ensuring the test suite passes against a Databricks target warehouse

**Would you like to contribute?**
Are you interested in building this feature?

**Anything Else?**

tlento:
This is happening!

tlento:
Courtney did this!

PR #260, opened 2022-09-28
Databricks Support

courtneyholcomb:
Completes https://github.com/transform-data/metricflow/issues/25
Completes MF-496


.github/workflows/cd-sql-engine-tests.yaml
```diff
@@ -107,6 +107,39 @@ jobs:
...
+      - name: Install Poetry
+        run: pip install poetry==1.1.15 && poetry config virtualenvs.create false
+
+      - name: Install Deps
+        run: cd metricflow && poetry install
+
+      - name: Run MetricFlow unit tests with Databricks configs
+        run: pytest metricflow/test/
+        env:
+          MF_SQL_ENGINE_URL: ${{ secrets.MF_DATABRICKS_URL }}
```
courtneyholcomb, line 140: I put this in here but I still need to set up a service account and store these env vars!

tlento: Sweet! I can get the env vars stored once we have the service account credentials and appropriate URL ready. There's a special deployment secret we use to avoid leaking those credentials through forked repos.

metricflow/sql_clients/databricks.py
```diff
@@ -0,0 +1,177 @@
...
+                    cells = []
+                    for cell in row:
+                        if type(cell) in [str, pd.Timestamp]:
+                            # Wrap cell in quotes & escape existing single quotes
+                            escaped_cell = str(cell).replace("'", "\\'")
+                            cells.append(f"'{escaped_cell}'")
+                        else:
+                            cells.append(str(cell))
+                    values.append(f"({', '.join(cells)})")
+                cursor.execute(f"INSERT INTO {sql_table.sql} VALUES {', '.join(values)}")
```
tlento, line 162: I see what you mean about this being a bit hacky. I don't understand why they don't support creating tables from arrow or just dataframe -> sql. Might be worth opening an issue over there.

tlento (approved):
This looks great, thank you! Please remember to write a good commit message when you squash and merge - the default one is terrible, and this deserves a nice descriptor in the log.

Two minor inline nits worth looking at before merge:

1. Remove the databricks renderer class, we can just use the default one
2. Catch a concrete exception in that one spot, if possible

metricflow/sql/render/databricks.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class DatabricksSqlQueryPlanRenderer(DefaultSqlQueryPlanRenderer):
+    """Plan renderer for the Databricks engine."""
+
+    EXPR_RENDERER = DatabricksSQLExpressionRenderer()
+
+    @property
+    def expr_renderer(self) -> SqlExpressionRenderer:  # noqa :D
+        return self.EXPR_RENDERER
```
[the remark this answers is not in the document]
courtneyholcomb: Sounds good!

metricflow/sql_clients/databricks.py
```diff
@@ -102,31 +113,47 @@ def _engine_specific_dry_run_implementation(self, stmt: str, bind_params: SqlBin
...
+                for row in df.itertuples(index=False, name=None):
+                    cells = []
+                    for cell in row:
+                        if type(cell) in [str, pd.Timestamp]:
+                            # Wrap cell in quotes & escape existing single quotes
+                            cells.append(f"""'{str(cell).replace("'", '"')}'""")
+                        else:
+                            cells.append(str(cell))
+                    values.append(f"({', '.join(cells)})")
+                cursor.execute(f"INSERT INTO {sql_table.sql} VALUES {', '.join(values)}")
```
plypaul, line 161: @courtneyholcomb - We've had some issues in the past where without `chuck_size` support, we would get an error when loading larger data frames. Nit: mind breaking the values into separate lines? This could make it hard to read in the query logs and some chance it causes memory issues in Databricks if their parsing implementation is off.

metricflow/sql_clients/databricks.py
```diff
@@ -42,35 +52,36 @@ class DatabricksEngineAttributes(SqlEngineAttributes):
...
-            raise ValueError(f"Expected dialect '{dialect}' in {url}")
+            split_url = url.split(";")  # TODO: there might not only be http path in there
+            parsed_url = sqlalchemy.engine.url.make_url(split_url[0])
+            for piece in split_url:
+                if HTTP_PATH_KEY in piece.lower():
+                    __, http_path = piece.split("=")
+            dialect = SqlDialect.DATABRICKS.value
+            if not http_path or parsed_url.drivername != dialect or not parsed_url.host:
+                raise ValueError
+        except:  # noqa: E722
```
plypaul: Nit: Instead of swallowing the exception, can you do `raise ValueError(...) from e`? You would have to re-work throwing that value error though.

PR #261, opened 2022-09-28
Add tags to locked_metadata

QMalcolm:
**NOTE**
Transform's internal schema, `schemas_internal.py`, is in MetricFlow. We might one day change that, but for the time being it is where it is. Which means to modify `schemas_internal.py` we have to do so in MetricFlow.

# What's Changed
This PR adds a concept of tags to `locked_metadata` of `schemas_internal.py` for internal metric tagging functionality we're working on.


tlento (approved):
Ah, right, we don't need a model object update because we don't actually process this construct in MetricFlow. Thanks for the excellent PR summary!

PR #264, opened 2022-09-30
BUG: window_groupings should not need to be in query dimensions

WilliamDee:
### Context
Currently for a semi-additive measure, we can optionally add a set of identifiers to designate the grain to apply the non-additive filter. From a recent merge #253, we modified the implementation of where we did this filter join, which caused an edge case bug in which we have a window_grouping that is not passed as a dimension in the query which causes an `Invalid identifier '<one of the window_groupings that was not in query>'` Error.

### Solution
Similar to adding the non-additive dimension as an extraneous linkable spec, we add all the `window_groupings` identifiers to that extraneous_linkable_spec set so that we can reference the identifiers properly during the query.

### DIFF
--- a/metricflow/test/integration/test_cases/itest_semi_additive_measure.yaml
+++ b/metricflow/test/integration/test_cases/itest_semi_additive_measure.yaml
@@ -30,6 +30,40 @@ integration_test:
       a.user
 ---
 integration_test:
+  name: semi_additive_measure_query_with_identifier_grouping_no_group_by
+  description: |
+    Tests selecting a measure with semi additive properties with a window_grouping that are not in the query group by
+  model: SIMPLE_MODEL
+  metrics: ["current_account_balance_by_user"]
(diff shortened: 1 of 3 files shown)


metricflow/test/integration/test_cases/itest_semi_additive_measure.yaml
```diff
@@ -29,6 +29,40 @@ integration_test:
     GROUP BY
       a.user
 ---
+integration_test:
+  name: semi_additive_measure_query_with_identifier_grouping_no_group_by
+  description: |
+    Tests selecting a measure with semi additive properties with a window_grouping that are not in the query group by
+  model: SIMPLE_MODEL
+  metrics: ["current_account_balance_by_user"]
```
tlento, line 37: nit: We should probably rename this to something like `total_current_user_account_balances` or something.

tlento (approved):
LGTM!

PR #268, opened 2022-10-04
Add COUNT measure aggregation support

WilliamDee:
## Context
We currently work around a `COUNT` via a `SUM-CASE` in the config or as a SUM `expr: 1`. ie.,
```
measures:
    - name: referred_bookings
      expr: CASE WHEN referrer_id IS NOT NULL THEN 1 ELSE 0 END
      agg: sum
    - name: total_bookings
      expr: 1
      agg: sum
```

So as a cleanup we can just do the `SUM-CASE` conversion in the framework itself and have the config accept `COUNT` as an measure aggregation. From the above, the count equivalent would be,
```
measures:
    - name: referred_bookings
      expr: referrer_id
      agg: count
    - name: total_bookings
      expr: 1
      agg: count
```

## Additional Validation
It is awkward to have `expr` to be `Optional` as one may assume it would just do a `COUNT(*)` or `COUNT(1)`. However, in our framework when we don't pass `expr`, it is standardized to pass the measure name as the `expr`. This would lead to some confusion. This means if someone where to use a `COUNT` aggregation, they would be required to provide the `expr`.


metricflow/aggregation_properties.py
```diff
@@ -17,18 +17,29 @@ class AggregationType(ExtendedEnum):
...
     @property
     def is_expansive(self) -> bool:
         """Expansive ≝ Op( X ∪ Y ∪ ...) = Op( Op(X) ∪ Op(Y) ∪ ...)"""
-        return self in (AggregationType.SUM, AggregationType.MIN, AggregationType.MAX, AggregationType.BOOLEAN)
+        return self in (
+            AggregationType.SUM,
+            AggregationType.MIN,
+            AggregationType.MAX,
+            AggregationType.BOOLEAN,
+            AggregationType.COUNT,
```
tlento, line 31: Can you update the docstring to indicate that COUNT is expansive because it gets transformed into SUM prior to initial aggregation?

metricflow/test/fixtures/table_fixtures.py
```diff
@@ -39,18 +39,18 @@ def create_simple_model_tables(mf_test_session_state: MetricFlowTestSessionState
         # Columns for reference
         # ["guest_id", "host_id", "listing_id", "booking_value", "is_instant", "ds", "ds_partitioned",
         #  "booking_paid_at"]
```
tlento, line 41: Please add the referrer_id here as well.

metricflow/model/transformations/convert_count.py
```diff
@@ -0,0 +1,26 @@
...
+
+    @staticmethod
+    def transform_model(model: UserConfiguredModel) -> UserConfiguredModel:  # noqa: D
+        for data_source in model.data_sources:
+            for measure in data_source.measures:
+                if measure.agg == AggregationType.COUNT:
+                    if measure.expr is None:
+                        raise ModelTransformError(
+                            f"Measure '{measure.name}' is has a COUNT aggregation which requires an expr to be provided. "
+                            f"Provide 'expr: 1' if a count of all rows is desired."
```
tlento, line 20:
```suggestion
                            f"Measure '{measure.name}' uses a COUNT aggregation, which requires an expr to be provided. "
                            f"Provide 'expr: 1' if a count of all rows is desired."
```

metricflow/model/validations/measures.py
```diff
@@ -364,6 +364,37 @@ def validate_model(model: UserConfiguredModel) -> List[ValidationIssueType]:  #
...
+                        ValidationError(
+                            context=DataSourceElementContext(
+                                file_context=FileContext.from_metadata(metadata=data_source.metadata),
+                                data_source_element=DataSourceElementReference(
+                                    data_source_name=data_source.name, element_name=measure.name
+                                ),
+                                element_type=DataSourceElementType.MEASURE,
+                            ),
+                            message=(
+                                f"Measure '{measure.name}' is has a COUNT aggregation which requires an expr to be provided. "
```
tlento, line 390:
```suggestion
                                f"Measure '{measure.name}' uses a COUNT aggregation, which requires an expr to be provided. "
```

tlento, line 390: Do we need this in both places? @QMalcolm ?

metricflow/test/model/validations/test_measures.py
```diff
@@ -289,3 +289,27 @@ def test_invalid_non_additive_dimension_properties(simple_model__pre_transforms:
...
+    invalid_measure = Measure(
+        name="bad_measure",
+        agg=AggregationType.COUNT,
+        agg_time_dimension="ds",
+    )
+    data_source.measures.append(invalid_measure)  # type: ignore
+
+    build = ModelValidator().validate_model(model)
+    expected_error_substring = (
+        "Measure 'bad_measure' is has a COUNT aggregation which requires an expr to be provided. "
```
tlento, line 307:
```suggestion
        "Measure 'bad_measure' uses a COUNT aggregation, which requires an expr to be provided. "
```

tlento (commented):
Minor nits inline. I'll leave this up for test fixes and to see if @QMalcolm has a specific opinion on the transform/validation thing.

metricflow/model/validations/measures.py
QMalcolm: In both places meaning in the transformation and the validation? I think the answer is unfortunately yes. It shouldn't be that way, but currently the ordering of events is `Parsing -> Optionally Transforming -> Validating`. Because the transformation step is optional, we need the validation rule. But because if transforming happens, it happens prior to validation, it needs to be able to properly error out 😞

tlento: Ok, I thought it might be something like this, thanks for confirming!

metricflow/aggregation_properties.py
```diff
@@ -17,18 +17,29 @@ class AggregationType(ExtendedEnum):
     SUM_BOOLEAN = "sum_boolean"
     AVERAGE = "average"
 
-    # COUNT = "count" not yet implemented ... non-expansive as COUNT(COUNT(X), COUNT(Y)) != COUNT(COUNT(X ∪ Y))
+    COUNT = "count"  # is only expansive because it's transformed into a SUM agg during model transformation
```
tlento, line 20: Sorry, I meant in the docstring on is_expansive. It's fine here but the docstring is better since people might access the property definition/documentation in their IDEs or whatever.

metricflow/test/model/validations/test_measures.py
```diff
@@ -289,3 +289,27 @@ def test_invalid_non_additive_dimension_properties(simple_model__pre_transforms:
     assert (
         len(missing_error_strings) == 0
     ), f"Failed to match one or more expected errors: {missing_error_strings} in {set([x.as_readable_str() for x in build.issues.errors])}"
+
+
+def test_count_measure_missing_expr(simple_model__pre_transforms: UserConfiguredModel) -> None:
+    """Tests the all measures with COUNT agg should have expr provided."""
```
tlento, line 295: micro-nit
```suggestion
    """Tests that all measures with COUNT agg should have expr provided."""
```

tlento (approved):
LGTM, assuming tests pass!

tlento:
Databricks is failing. We'll need to figure out how to regenerate things over there.

courtneyholcomb (approved):

PR #270, opened 2022-10-05
Improve support for None types in Databricks create_table_from_dataframe

tlento:
In other client implementations of create_table_from_dataframe, we
rely on the pandas.io.sql.to_sql method to convert dataframe types
to SQLAlchemy types and then issue a query via SQLAlchemy. This
uses the infer_dtype method to change any object type columns
into something more refined, since object is assigned by default
to columns containing None type values. The Pandas library also
applies a mask to convert NA types to None, which SQLAlchemy can
then insert as the appropriate null value type.

Our logic only partially covered these cases, with the result that
any None-type input in a numeric or timestamp column would fail to
upload, since Databricks naturally can't understand the `numpy.nan`
or `pandas.NaT` values that result. Worse, without the inference
in place, any boolean type column containing a None value would
be converted to a STRING type in the CREATE TABLE statement logic.

This change addresses both of these problems, and extends our
integration test for create_table_as_dataframe to include None-typed
values as inputs on all of the column types we check.

### DIFF
--- a/metricflow/sql_clients/databricks.py
+++ b/metricflow/sql_clients/databricks.py
@@ -1,10 +1,13 @@
 from __future__ import annotations
-from typing import Optional, List, ClassVar, Dict
-import pandas as pd
+
 import logging
 import time
+from typing import Optional, List, ClassVar, Dict
+
+import pandas as pd
 import sqlalchemy
 from databricks import sql
+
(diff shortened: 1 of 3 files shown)


courtneyholcomb (approved):
Looks great!! Thanks Tom!

PR #279, opened 2022-10-12
Add ability to parse dbt projects

QMalcolm:
> **NOTE**: This is being merged into `feature-branch--issue-266-dbt-integration` _not_ `main`

This PR is parses dbt configs to python objects the dbt way. That is we instantiate the dbt profile/target and then build then have dbt build the dbt manifest. This has some pros and cons:

### Pros

- Our parsing will always be up to date with how dbt does parsing
  - We get `refs` and what not built in
- it makes the code on our end really simple

### Cons

- We have to install all `dbt-*` adapters for warehouses we support
  - This is because they are required for instantiating the manifest

### Other things to note
This implementation requires we `model_path` points to the root of the dbt project root

## Testing

With this version of metricflow install to your virtual env
```
$ python
>>> from metricflow.model.parsing import dbt_dir_to_model
>>> dbt_dir_to_model.parse_dbt_project_to_model("/path/to/dbt/project/root")
NotImplementedError(...)

>>> dbt_dir_to_model.get_dbt_project_manifest("/path/to/dbt/project/root")
Manifest(...)
```

## Stacked PR
* feature-branch--issue-266-dbt-integration
  * #279 👈 You are here
    * #280


metricflow/model/parsing/dbt_dir_to_model.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def get_dbt_project_manifest(directory: str) -> DbtManifest:
+    """Builds the dbt Manifest object from the dbt project"""
+
+    dbt_config = get_dbt_config(project_dir=directory)
+    # If we don't disable tracking, we have to setup a full
+    # dbt User object to build the manifest
+    tracking.disable_tracking()
+    return DbtManifestLoader.get_full_manifest(config=dbt_config)
```
tlento, line 15: Nice, I think this is better than selective custom parsing.

tlento (approved):
LGTM!

PR #281, opened 2022-10-13
Derived Metrics

WilliamDee:
## Feature
Similar to EXPR metrics, this enables the ability to define a derived metric. A derived metric is essentially a metric that is defined as an expression of other metrics. 

## Config
Introducing a new metric type, `derived`. The following is a derived metric defined in the config,
```yaml
metric:
  name: net_sales_per_user
  owners:
    - nick@company.com
  type: derived
  type_params:
    expr: gross_sales - cogs / active_users
    metrics:
      - name: gross_sales # these are all metrics (can be a derived metric, meaning building a derived metric with derived metrics)
      - name: cogs
      - name: users
        constraint: is_active # Optional additional constraint
        alias: active_users # Optional alias to use in the expr
```

## Changes
- Updated `DataflowToSqlQueryPlanConverter._build_metrics_output_node` to be able to handle building nested metrics node (which is essentially a reducing repeated logic)
- Updated `MetricSpec` to have `constraint` and `alias` field
- Added new transform rule that links all the corresponding measures of each referenced metric to the parent derived metric


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -146,48 +151,72 @@ def build_plan(
...
-                    aggregated_measures_node=aggregated_measures_node,
+                compute_metrics_nodes.append(
+                    ComputeMetricsNode[SqlDataSetT](
+                        parent_node=compute_metric_node,
+                        metric_specs=[metric_spec],
+                    )
                 )
-            )
+            else:
+                metric_input_measure_specs = self._metric_semantics.measures_for_metric(blank_metric_spec)
```
WilliamDee, line 186: This is the same logic as before but placed into this else block

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -146,48 +151,72 @@ def build_plan(
...
-                    metric_spec=metric_spec,
-                    aggregated_measures_node=aggregated_measures_node,
+                compute_metrics_nodes.append(
+                    ComputeMetricsNode[SqlDataSetT](
+                        parent_node=compute_metric_node,
+                        metric_specs=[metric_spec],
+                    )
                 )
-            )
+            else:
```
WilliamDee, line 185: same logic as the prior build compute metric node but placed into this `else` block

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -146,48 +151,72 @@ def build_plan(
...
+        for metric_spec in metric_specs:
             logger.info(f"Generating compute metrics node for {metric_spec}")
-            metric = self._metric_semantics.get_metric(metric_spec)
-            metric_input_measure_specs = self._metric_semantics.measures_for_metric(metric_spec)
-
-            logger.info(
-                f"For {metric_spec}, needed measures are:\n"
-                f"{pformat_big_objects(metric_input_measure_specs=metric_input_measure_specs)}"
-            )
+            blank_metric_spec = MetricSpec.from_element_name(metric_spec.element_name)
```
WilliamDee, line 164: This is a bit of a special case where in `MetricSemantics` we use `MetricSpec` as a reference key to obtain things in the container, but due to adding the new fields it ends up not being able to find the matching key. 

ie., `MetricSpec(name="m1", alias="test")` won't match as in the `MetricSemantic` it is only defined as `MetricSpec(name="m1")`.

I originally wanted to refactor it to use `MeasureReferences` instead of `MetricSpec` as reference keys, but it ended up being more work that is outside of this PR's scope. To get around this, I ended up creating this "blank" MetricSpec to use as the key to access the `MetricSemantics`.

plypaul: > I originally wanted to refactor it to use MeasureReferences instead of MetricSpec as reference keys

I could help with that if needed.

WilliamDee: it's not blocking this PR so can be done at anytime, i could take this on as well if you don't bandwidth.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -146,48 +151,73 @@ def build_plan(
...
+        for metric_spec in metric_specs:
             logger.info(f"Generating compute metrics node for {metric_spec}")
-            metric = self._metric_semantics.get_metric(metric_spec)
-            metric_input_measure_specs = self._metric_semantics.measures_for_metric(metric_spec)
-
-            logger.info(
-                f"For {metric_spec}, needed measures are:\n"
-                f"{pformat_big_objects(metric_input_measure_specs=metric_input_measure_specs)}"
-            )
+            # TODOmetricsem: refactor MetricSemantics to use MetricReference instead of MetricSpec as an accessor key
```
plypaul, line 164: Nit: typo

plypaul (commented):
Nice, looks good overall! That `blank_metric` does indicate that we should adjust the semantics classes. Also, could you provide context on why the `AddInputMetricMeasuresRule` is needed as opposed to resolving it in the builder?

WilliamDee:
> Also, could you provide context on why the AddInputMetricMeasuresRule is needed as opposed to resolving it in the builder?

re: @plypaul 
This is because there were places where the logic was that we would retrieve the metric from the user_configured_model, then it would get the measures via `Metric.input_measures` or `Metric.measure_references` then do some calculation with that. But because derived metrics doesn't have any `input_measures` they end up returning an empty list and those calculations will break. I originally had a logic where in each of those instances I would use a conditional to check whether or not the metric is `derived` and would populate the measures at runtime time. But this meant I had to do this `metric_type conditional + go through each sub metric in the derived metric` in every case that had that. On the other hand, just doing it in a transform rule allowed me to just make that rule and all the other functions work as expected without any changes. An example of this callsite is `get_linkable_elements_for_metrics` in `linkable_spec_resolver.py`

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -146,48 +151,73 @@ def build_plan(
...
+                logger.info(
+                    f"For derived metric: {metric_spec}, needed metrics are:\n"
+                    f"{pformat_big_objects(metric_input_specs=metric_input_specs)}"
+                )
 
-            combined_where = query_spec.where_constraint
-            if metric.constraint:
-                metric_constraint = WhereConstraintConverter.convert_to_spec_where_constraint(
-                    self._data_source_semantics, metric.constraint
+                compute_metric_node = self._build_metrics_output_node(
```
tlento, line 174: Ok, so this includes all metrics in the dataflow plan, but when we render the SQL we strip them out of the result sets for each metric node (i.e., every visit to the metric node returns only the results of that metric node, not any of its inputs).

Could we get some plan file test cases for this logic just so we can see how it all works?

tlento (commented):
Nice! I think some plan files showing the derived metric behavior would be great.

tlento:
I think the transform rule gives us some additional flexibility for optimization as well - it might behoove us to convert derived input MEASURE_PROXY metrics into constrained measures. Or it might not. If it doesn't we might want to re-think how we manage measure inputs in general. As long as there aren't other major issues I'm missing, I like the minimalist approach even if it feels a little janky at the moment.

We should sketch out design options for these optimizations once we get the feature working correctly.

metricflow/dataflow/builder/dataflow_plan_builder.py
WilliamDee: pushed the refactor and made the appropriate changes on this PR

metricflow/dataflow/builder/dataflow_plan_builder.py
WilliamDee: added plan files!

metricflow/test/snapshots/test_dataflow_to_sql_plan.py/SqlQueryPlan/DuckDbSqlClient/test_derived_metric__plan0_optimized.sql
```diff
@@ -0,0 +1,57 @@
...
+        , CASE WHEN referrer_id IS NOT NULL THEN 1 ELSE 0 END AS referred_bookings
+      FROM (
+        -- User Defined SQL Query
+        SELECT * FROM ***************************.fct_bookings
+      ) bookings_source_src_10001
+    ) subq_13
+    GROUP BY
+      metric_time
+  ) subq_19
+  FULL OUTER JOIN (
```
tlento, line 33: Oh, huh.... should this be a FULL OUTER join?

With multi-measure ratio and expr metrics we do an inner join, because those don't make sense when NULL inputs get involved - you end up with a bunch of uncontrolled failure states. For example:

https://github.com/transform-data/metricflow/blob/main/metricflow/test/snapshots/test_dataflow_to_sql_plan.py/SqlQueryPlan/DuckDbSqlClient/test_compute_metrics_node_ratio_from_multiple_data_sources__plan0_optimized.sql#L40

tlento (commented):
This isn't consistent with how we do ratio/expr computations today, and I think that'll make the next migration step hard.

I think we can do this in one of four ways:

1. Adjust the join type here pre-merge
2. Merge this and apply a follow-up PR to adjust the join type
3. Merge this and move ratio/expr to FULL OUTER JOIN

I'm not inclined towards option 3 - we used INNER JOIN on those aggregations for a reason - but we should probably think about it a bit more just to make sure the derived metrics case doesn't open up some other considerations.

tlento (approved):
Per separate discussion, @WilliamDee will follow up on the join semantics and apply any necessary changes before we announce this feature.

PR #282, opened 2022-10-13
Add SqlClient.list_tables

plypaul:
This PR adds `list_tables` to the `SqlClient` protocol instead of `BaseSqlClientImplementation` so that it's exposed.


tlento (approved):
Looks good to me! This is technically a breaking change on the API since custom SqlClient implementations out in the wild will fail on the package update, but since it's an easy fix I think we just roll it out and announce.

PR #283, opened 2022-10-13
Add Async Query / Cancel Methods to SqlClient

plypaul:
This PR adds an async interface to the query execution methods in `SqlClient` along with the associated query cancellation methods. The cancellation of queries is handled by returning a request ID in the async method, which is later used to call the cancel method. The queries are tagged using serialized metadata that is added as a comment to the query text, which allows for easy inspection of running queries in the DW engine. It also enables cancelling queries from separate `SqlEngine` objects.

### DIFF
--- a/metricflow/protocols/async_sql_client.py
+++ b/metricflow/protocols/async_sql_client.py
@@ -1,0 +1,52 @@
+from __future__ import annotations
+
+from abc import abstractmethod
+from typing import Protocol, Sequence
+
+from metricflow.protocols.sql_client import SqlClient
+from metricflow.protocols.sql_request import SqlRequestId, SqlRequestResult, SqlRequestTagSet
+from metricflow.sql.sql_bind_parameters import SqlBindParameters
+
+
+class AsyncSqlClient(SqlClient, Protocol):
+    """Defines methods for executing SQL statements asynchronously."""
+
+    def async_query(
+        self,
+        statement: str,
+        bind_parameters: SqlBindParameters = SqlBindParameters(),
+        tags: SqlRequestTagSet = SqlRequestTagSet(),
+    ) -> SqlRequestId:
+        """Execute a query asynchronously."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def async_request_result(self, request_id: SqlRequestId) -> SqlRequestResult:
+        """Wait until a async query has finished, and then return the result"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def async_execute(
+        self,
+        statement: str,
+        bind_parameters: SqlBindParameters = SqlBindParameters(),
+        tags: SqlRequestTagSet = SqlRequestTagSet(),
+    ) -> SqlRequestId:
+        """Execute a statement that does not return values asynchronously."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def cancel_request(self, pattern_tag_set: SqlRequestTagSet) -> int:
+        """Make a best-effort at canceling requests that have a superset of the given tags.
+
+        Returns the number of cancellation commands sent.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def active_requests(self) -> Sequence[SqlRequestId]:
+        """Return requests that are still in progress.
+
+        If the results for a request have not yet been fetched with async_request_result(), it's considered in progress.
+        """
+        raise NotImplementedError
--- a/metricflow/protocols/sql_client.py
+++ b/metricflow/protocols/sql_client.py
@@ -186,6 +186,7 @@ class SqlEngineAttributes(Protocol):
     # SQL Dialect replacement strings
     double_data_type_name: ClassVar[str]
     timestamp_type_name: ClassVar[Optional[str]]
+    random_function_name: ClassVar[str]
 
     # MetricFlow attributes
     sql_query_plan_renderer: ClassVar[SqlQueryPlanRenderer]
--- a/metricflow/protocols/sql_request.py
+++ b/metricflow/protocols/sql_request.py
@@ -1,0 +1,104 @@
+from __future__ import annotations
+
+import typing
+from collections import OrderedDict
+from dataclasses import dataclass
+from enum import Enum
+from operator import itemgetter
+from typing import Optional, Sequence, Dict
+
+import pandas as pd
+from pydantic import Field
+
+from metricflow.model.objects.base import FrozenBaseModel
+from metricflow.object_utils import assert_exactly_one_arg_set
+
+
+@dataclass(frozen=True)
+class SqlRequestId:
+    """Identifies a request (i.e. a call to SqlClient.query() or SqlClient.execute()) to the SQL engine."""
+
+    id_str: str
+
+    def __repr__(self) -> str:  # noqa: D
+        return self.id_str
+
+
+@dataclass(frozen=True)
+class SqlRequestResult:
+    """The result from a request to the SQL engine."""
+
+    df: Optional[pd.DataFrame] = None
+    exception: Optional[Exception] = None
+
+    def __post_init__(self) -> None:  # noqa: D
+        assert_exactly_one_arg_set(df=self.df, exception=self.exception)
+
+
+class SqlRequestTagSet(FrozenBaseModel):
+    """Set of tags as a Pydantic model for easy serialization."""
+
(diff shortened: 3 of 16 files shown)


metricflow/sql_clients/big_query.py
```diff
@@ -13,7 +13,7 @@
 from metricflow.sql_clients.sqlalchemy_dialect import SqlAlchemySqlClient
 
 
-class BigQueryEngineAttributes:
+class BigQueryEngineAttributes(SqlEngineAttributes):
```
tlento, line 18: As a general rule we should not inherit from protocols, and we should NEVER inherit from protocols that define bare properties (such as `SqlEngineAttributes`) as this defeats the purpose of using a protocol as a type definition.

With this change, if you left this class empty it'd pass the typechecker and then blow up in the runtime.

metricflow/sql_clients/async_request.py
```diff
@@ -0,0 +1,120 @@
...
+            logger.error(
+                f"Got multiple tag sets from parsing comments:\n"
+                f"{pformat_big_objects(tag_sets)}\n"
+                f"Using the first one."
+            )
+
+        return tag_sets[0] if len(tag_sets) > 0 else None
+
+
+class SqlRequestExecutorThread(threading.Thread):
```
tlento, line 62: Oof. Do we really have to roll our own thread class implementation for this? I know we can't use asyncio Futures or anything because of Snowflake, but it just feels like we should be able to use something that's a bit more out of the box. I guess the serialization/deserialization on something like ThreadPoolExecutor gets onerous?

metricflow/sql_clients/big_query.py
plypaul: In that case, how can we improve the developer experience? E.g. I want to add an attribute to the protocol, and I need to find all classes that need updates.

metricflow/sql_clients/async_request.py
plypaul: One of the things that turned out to be an issue is that with process pools, the name of the worker thread is opaque, so when we're debugging and trying to see what the threads are doing, we would see a bunch of threads named `xyz-1-worker` (or similar), and it would be hard to figure out what's going on.

metricflow/sql_clients/big_query.py
tlento: I generally add the attribute to the protocol and then run mypy (or find new mypy errors in the IDE) if I don't already know what I'm looking for. This can lead to a big fanout explosion of errors, but it's easy enough to tame those.

metricflow/sql_clients/async_request.py
tlento: le sigh, python.....

metricflow/protocols/sql_request.py
```diff
@@ -0,0 +1,104 @@
...
+
+@dataclass(frozen=True)
+class SqlRequestResult:
+    """The result from a request to the SQL engine."""
+
+    df: Optional[pd.DataFrame] = None
+    exception: Optional[Exception] = None
+
+    def __post_init__(self) -> None:  # noqa: D
+        assert_exactly_one_arg_set(df=self.df is not None, exception=self.exception)
```
tlento, line 35: Is this boolean conversion because an empty dataframe is falsey? 

Also, that's an unexpected bit of news for me about assert_exactly_one_arg_set - I assumed it was doing `is None` checks. So does its docstring, so I just put up #290

metricflow/sql_clients/base_sql_client_implementation.py
```diff
@@ -203,3 +217,61 @@ def close(self) -> None:  # noqa: D
...
+        tags: SqlRequestTagSet = SqlRequestTagSet(),
+    ) -> SqlRequestId:
+        with self._state_lock:
+            request_id = SqlRequestId(f"mf_rid__{random_id()}")
+            thread = SqlRequestExecutorThread(
+                sql_client=self,
+                request_id=request_id,
+                statement=statement,
+                bind_parameters=bind_parameters,
+                tags=tags.add_request_id(request_id),
```
tlento, line 234: You're doing this twice. Is that intentional? It's a bunch of objects to create and throw away.

I think I like it better if request_id is special and internal - always added, etc. as it is in the SqlRequestExecutorThread - and this `tags` parameter is really for arbitrary tags that people want to pass through.

metricflow/test/sql_clients/test_async.py
```diff
@@ -0,0 +1,69 @@
...
+            CROSS JOIN {table_with_1000_rows.sql} b
+            CROSS JOIN {table_with_1000_rows.sql} c
+            -- CROSS JOIN {table_with_100_rows.sql} d
+            """
+        )
+    )
+    # Need to wait a little bit as some clients like BQ doesn't show the query as running right away.
+    time.sleep(2)
+    num_cancelled = sql_client.cancel_request(SqlRequestTagSet.create_from_request_id(request_id))
+    sql_client.async_request_result(request_id)
```
tlento, line 68: Is this robust? I worry that the sleep timeout is going to make this test brittle.

Also it's a little odd to request result in here. Maybe that gets updated later?

tlento (commented):
Ok I have reviewed this in full. We shouldn't inherit that protocol - use mypy to find things that need updating instead.

Regarding the SqlClient protocol structure, I'm inclined to think a separate protocol is the way to go - at least for now - since it gives us flexibility and some time to think about what final API we wish to land on without having to impose too many breaking changes on outside developers. I know of at least one playing around with a custom SqlClient class, so this is no longer a purely theoretical concern.

That said, I don't feel too strongly about this, and if you're clear on where you want this interface to end up we should go for it in a more direct manner while the developer base is still small.

tlento (changes requested):
As discussed, we'll try splitting the protocol out. I'm also trying out this request changes thing, let's see what happens!

metricflow/protocols/sql_client.py
```diff
@@ -158,6 +159,46 @@ def render_execution_param_key(self, execution_param_key: str) -> str:
         """Wrap execution parameter key with syntax accepted by engine."""
         raise NotImplementedError
 
+    def async_query(
```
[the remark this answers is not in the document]
plypaul: Updated with `AsyncSqlClient`.

metricflow/protocols/sql_request.py
plypaul: Updated, taking #290 into account.

metricflow/sql_clients/base_sql_client_implementation.py
plypaul: Moved to `SqlRequestExecutorThread`.

metricflow/test/sql_clients/test_async.py
plypaul: It's tricky to make this robust, but moved to an alternative that seems better.

metricflow/protocols/async_sql_client.py
```diff
@@ -0,0 +1,57 @@
...
+from metricflow.sql.sql_bind_parameters import SqlBindParameters
+
+
+class AsyncSqlClient(SqlClient, Protocol):
+    """Defines methods for executing SQL statements asynchronously."""
+
+    @abstractmethod
+    def render_execution_param_key(self, execution_param_key: str) -> str:
+        """Wrap execution parameter key with syntax accepted by engine."""
+        raise NotImplementedError
```
tlento, line 17: nit - this is duplicated, can remove it from here if you like

metricflow/test/sql_clients/test_async.py
```diff
@@ -68,12 +69,18 @@ def test_cancel_request(  # noqa: D
...
-    time.sleep(2)
-    num_cancelled = async_sql_client.cancel_request(SqlRequestTagSet.create_from_request_id(request_id))
-    async_sql_client.async_request_result(request_id)
+    start_time = time.time()
+    num_cancelled: Optional[int] = None
+    while time.time() - start_time < 30:
+        time.sleep(1)
+        num_cancelled = async_sql_client.cancel_request(SqlRequestTagSet.create_from_request_id(request_id))
+        if num_cancelled > 0:
+            break
```
tlento, line 83: Nice, this seems more robust. We'll see how it goes in CI!

tlento (approved):
Thanks for the fixup, this looks great!

PR #284, opened 2022-10-13
Implement Cancel Methods for Sql Engines

plypaul:
This PR implements the query cancellation methods for the SQL engines (minus Databricks).

### DIFF
--- a/metricflow/sql_clients/big_query.py
+++ b/metricflow/sql_clients/big_query.py
@@ -1,18 +1,28 @@
 from __future__ import annotations
 
 import json
-from typing import ClassVar, Optional, Sequence
+import logging
+from typing import ClassVar, Optional, Dict
+from typing import Sequence
 
+import google.oauth2.service_account
 import sqlalchemy
+from google.cloud import bigquery
+from google.cloud.bigquery import QueryJob
 
 from metricflow.protocols.sql_client import SqlEngine
-from metricflow.protocols.sql_client import SqlEngineAttributes
+from metricflow.protocols.sql_client import (
+    SqlEngineAttributes,
+)
 from metricflow.protocols.sql_request import SqlRequestTagSet
 from metricflow.sql.render.big_query import BigQuerySqlQueryPlanRenderer
 from metricflow.sql.render.sql_plan_renderer import SqlQueryPlanRenderer
 from metricflow.sql.sql_bind_parameters import SqlBindParameters
+from metricflow.sql_clients.async_request import SqlStatementCommentMetadata
 from metricflow.sql_clients.common_client import SqlDialect
 from metricflow.sql_clients.sqlalchemy_dialect import SqlAlchemySqlClient
+
+logger = logging.getLogger(__name__)
 
 
 class BigQueryEngineAttributes:
@@ -31,7 +41,7 @@ class BigQueryEngineAttributes:
     timestamp_type_supported: ClassVar[bool] = True
     timestamp_to_string_comparison_supported: ClassVar[bool] = False
     # Cancelling should be possible, but not yet implemented.
-    cancel_submitted_queries_supported: ClassVar[bool] = False
+    cancel_submitted_queries_supported: ClassVar[bool] = True
 
     # SQL Dialect replacement strings
     double_data_type_name: ClassVar[str] = "FLOAT64"
@@ -46,7 +56,7 @@ class BigQuerySqlClient(SqlAlchemySqlClient):
     """BigQuery implementation of SQL client"""
 
     @staticmethod
-    def from_connection_details(url: str, password: Optional[str]) -> BigQuerySqlClient:  # noqa: D
+    def from_connection_details(url: str, password: Optional[str] = None) -> BigQuerySqlClient:  # noqa: D
         parsed_url = sqlalchemy.engine.url.make_url(url)
         dialect = SqlDialect.BIGQUERY.value
         if parsed_url.drivername != dialect:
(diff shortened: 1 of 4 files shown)


metricflow/sql_clients/snowflake.py
```diff
@@ -244,4 +245,25 @@ def cancel_submitted_queries(self) -> None:  # noqa: D
                     conn.execute(f"SELECT SYSTEM$cancel_all_queries({session_id})")
 
     def cancel_request(self, pattern_tag_set: SqlRequestTagSet) -> int:  # noqa: D
-        raise NotImplementedError
+        # Running queries have an end_time set to the epoch time:
```
tlento, line 248: What? So many interesting choices in the world....

metricflow/sql_clients/big_query.py
```diff
@@ -57,29 +67,39 @@ def from_connection_details(url: str, password: Optional[str]) -> BigQuerySqlCli
...
-        super().__init__(
-            engine=BigQuerySqlClient._create_bq_engine(
-                query_string=query_string, project_id=project_id, password=password
-            )
+        password_json = json.loads(password) if password else {}
+        self._project_id = project_id or password_json.get("project_id")
+        bq_engine = BigQuerySqlClient._create_bq_engine(project_id=project_id, password=password)
+        self._bq_client = bigquery.Client(
+            project=self._project_id,
+            credentials=google.oauth2.service_account.Credentials.from_service_account_info(password_json),
```
tlento, line 80: Why are all of these changes in this commit?

Also, does this work if we aren't using service accounts?

metricflow/sql_clients/big_query.py
```diff
@@ -90,10 +110,15 @@ def _create_bq_engine(
            - Google's decision tree (https://google.aip.dev/auth/4110)
 
         Args:
-            query_string: Additional query params to pass to engine construction.
+            query_field_values: The field/value pairs to use in the query string.
```
tlento, line 113: This I get. That was super weird with the bare string.

metricflow/sql_clients/big_query.py
```diff
@@ -57,29 +67,39 @@ def from_connection_details(url: str, password: Optional[str]) -> BigQuerySqlCli
...
+            query_field_values={"dry_run": "true"}, project_id=self._project_id, password=self._password
         )
         with _engine.connect() as conn:
             conn.execute(sqlalchemy.text(stmt), bind_params.param_dict)
 
+    @staticmethod
+    def _create_bq_url(project_id: str = "", query_field_values: Optional[Dict[str, str]] = None) -> str:
+        if query_field_values is None:
+            query_field_values = {}
+        query_items = tuple(f"{field}:{value}" for field, value in query_field_values.items())
```
tlento, line 96: Wait, `{field}:{value}` not `{field}={value}`?

metricflow/sql_clients/big_query.py
```diff
@@ -90,10 +110,15 @@ def _create_bq_engine(
...
 
         Args:
-            query_string: Additional query params to pass to engine construction.
+            query_field_values: The field/value pairs to use in the query string.
             project_id: Project ID listed on the GCP platform.
             password: String representation of keyfile.json.
         """
+        if query_field_values is None:
+            query_field_values = {}
+        query_items = tuple(f"{field}:{value}" for field, value in query_field_values.items())
```
tlento, line 119: Again, why `:` instead of `=` between param key and value?

tlento (commented):
The BigQuery change is weirdly expansive and not all of it makes sense to me. Mind providing some more context, or splitting out the other bits if appropriate?

tlento (dismissed):
Hey can you look over the inlines? I don't fully understand all of the changes in here, and I'm not 100% sure they're all correct, but if you've verified and this was what had to happen just some explanation is enough.

metricflow/sql_clients/big_query.py
plypaul: Yeah, some of these changes could have been a separate PR, but saw some things that could use fixing.

Not entirely sure about initialization without credentials, but this is the way that the docs create a client: https://github.com/googleapis/python-bigquery/blob/HEAD/samples/client_query.py

metricflow/sql_clients/big_query.py
plypaul: That's a mistake - I was looking at the syntax for labels in the URL. Fixed.

metricflow/sql_clients/big_query.py
plypaul: Same as previous comment.

metricflow/sql_clients/big_query.py
```diff
@@ -57,29 +67,33 @@ def from_connection_details(url: str, password: Optional[str]) -> BigQuerySqlCli
...
-        super().__init__(
-            engine=BigQuerySqlClient._create_bq_engine(
-                query_string=query_string, project_id=project_id, password=password
-            )
+        password_json = json.loads(password) if password else {}
+        self._project_id = project_id or password_json.get("project_id")
+        bq_engine = BigQuerySqlClient._create_bq_engine(project_id=project_id, password=password)
+        self._bq_client = bigquery.Client(
+            project=self._project_id,
+            credentials=google.oauth2.service_account.Credentials.from_service_account_info(password_json)
```
tlento, line 80: Still a little worried about this one, and I'd prefer it be separate just so we don't get confused with the canceling vs the connection change, but I guess it's small enough we can figure it out.

tlento (approved):
Ok!

PR #286, opened 2022-10-14
Make dbt deps optional

QMalcolm:
Making the dbt-* dependencies optional means they aren't installed when one does `poetry add metricflow` or `pip install metricflow`. Instead one would do `poetry add metricflow -E dbt-snowflake` or `pip install metricflow[dbt-snowflake]`. In this case we support the following extras `dbt-snowflake`, `dbt-redshift`, `dbt-postgres` and `dbt-bigquery`.

This is important because installing dbt-* deps adds to MetricFlow package size fairly dramatically, thus it should only be installed if people are going to be using the dbt-MetricFlow functionality.

### DIFF
--- a/.github/workflows/ci-metricflow-unit-tests.yaml
+++ b/.github/workflows/ci-metricflow-unit-tests.yaml
@@ -32,7 +32,7 @@ jobs:
         run: pip install poetry==1.1.15 && poetry config virtualenvs.create false
 
       - name: Install Metricflow
-        run: poetry install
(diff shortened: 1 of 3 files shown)


tlento (approved):
I'm a little bit concerned about what dbt-snowflake et al pull in, since some of the engine libraries are not compatible with each other due to version conflicts on stuff like pyarrow.

For now this should be just fine, especially with dbt-* being optional, but it's possible we'll have some build patterns that work with dbt and some that do not.

PR #288, opened 2022-10-14
When building the MetricFlowEngine from config, check if building with dbt

QMalcolm:
Currently the CLI doesn't use the `SemanticModel` or the `UserConfiguredModel` provided by the `CLIContext`, instead it rebuilds the `SemanticModel` each time. Specifically most CLI commands use `CLIContext.mf` which directly builds the `MetricFlowEngine` from the config. Thus to build the model from dbt configs using most CLI commands, we have to update the `MetricFlowEngine.from_config` method to check how the underlying `SemanticModel`/`UserConfiguredModel` should be built, and building it accordingly.

## Stacked PR
* feature-branch--issue-266-dbt-integration
  * #279 
    * #280
      * #288 👈 You are here

### DIFF
--- a/metricflow/engine/metricflow_engine.py
+++ b/metricflow/engine/metricflow_engine.py
@@ -9,7 +9,7 @@ from typing import Optional, List, Sequence
 
 import pandas as pd
 
-from metricflow.configuration.constants import CONFIG_DWH_SCHEMA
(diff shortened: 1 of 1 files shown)


tlento (approved):
LGTM!

PR #289, opened 2022-10-14
use MetricReference as the key for MetricSemantics

WilliamDee:
## Context
We currently use `MetricSpec` as a key for everything in `MetricSemantics`. This restricts us from adding properties to `MetricSpec` as we are using `MetricSpec(element_name="..")` as the key and if we were to add any properties it would end up not matching. For example,
```
MetricSpec(element_name="metric", alias="test") 
     would not match 
MetricSpec(element_name="metric")
```
So in order to resolve this, replace all the usage of `MetricSpec` where we treat it as a key and replace it with `MetricReference`.

### DIFF
--- a/metricflow/model/semantics/semantic_containers.py
+++ b/metricflow/model/semantics/semantic_containers.py
@@ -31,6 +31,7 @@ from metricflow.references import (
     IdentifierReference,
     LinkableElementReference,
     MeasureReference,
+    MetricReference,
     TimeDimensionReference,
 )
 from metricflow.specs import (
@@ -51,11 +52,11 @@ class MetricSemantics:  # noqa: D
         self, user_configured_model: UserConfiguredModel, data_source_semantics: DataSourceSemantics
     ) -> None:
         self._user_configured_model = user_configured_model
-        self._metrics: Dict[MetricSpec, Metric] = {}
+        self._metrics: Dict[MetricReference, Metric] = {}
         self._data_source_semantics = data_source_semantics
 
         # Dict from the name of the metric to the hash.
-        self._metric_hashes: Dict[MetricSpec, str] = {}
+        self._metric_hashes: Dict[MetricReference, str] = {}
 
         for metric in self._user_configured_model.metrics:
             self.add_metric(metric)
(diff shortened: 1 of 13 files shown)


tlento:
It's 9 PM and I need to stop working. I'll review this Monday!

metricflow/model/semantics/semantic_containers.py
```diff
@@ -67,59 +68,65 @@ def __init__(  # noqa: D
...
-            raise MetricNotFoundError(f"Unable to find metric `{metric_name}`. Perhaps it has not been registered")
-        return self._metrics[metric_name]
+    def get_metric(self, metric_reference: MetricReference) -> Metric:  # noqa:D
+        if metric_reference not in self._metrics:
+            raise MetricNotFoundError(f"Unable to find metric `{metric_reference}`. Perhaps it has not been registered")
+        return self._metrics[metric_reference]
+
+    def get_metric_spec(self, metric_reference: MetricReference) -> MetricSpec:  # noqa: D
+        metric = self.get_metric(metric_reference)
+        return MetricSpec(element_name=metric.name)
```
tlento, line 107: Will we be able to eliminate this in the future as well? Seems like an API we'd rather not have in the long run.

metricflow/query/query_parser.py
```diff
@@ -217,7 +217,10 @@ def _parse_and_validate_query(
         else:
             parsed_where_constraint = where_constraint
 
-        metric_specs = self._parse_metric_names(metric_names)
+        metric_references = self._parse_metric_names(metric_names)
+        metric_specs = tuple(
```
tlento, line 221: Maybe we make the metric spec from the reference inline here instead of doing it in MetricSemantics? Then we can remove that API.

Might be good to have MetricSpec have a `.from_reference` factory method if it doesn't already.

metricflow/query/query_parser.py
```diff
@@ -371,11 +374,16 @@ def _parse_and_validate_query(
...
-        metric_specs: Sequence[MetricSpec],
+        metric_references: Sequence[MetricReference],
         linkable_specs: LinkableSpecSet,
     ) -> None:
         """Checks that the order by specs references an item in the query."""
 
+        # TODO: this is a workaround
+        # Need to figure out whether we should clean up OrderBySpec or if we have to actually pass a fully resolved MetricSpec here
+        metric_specs = [
+            MetricSpec(element_name=metric_reference.element_name) for metric_reference in metric_references
```
tlento, line 385: If you make a factory method you can use it here, too.

tlento (approved):
Nice! See inlines for a small suggestion to clean out that extra bit of metric spec access from MetricSemantics. I think we'd rather the reference -> spec conversion happen directly instead of mediating it through the MetricSemantics container.

Issue #266, opened 2022-10-03
MetricFlow should know that dbt knows about metrics


nhandel:
**Describe the Feature**

[dbt metrics](https://docs.getdbt.com/docs/building-a-dbt-project/metrics) users should be able to use MetricFlow to build metrics queries. To do this, MetricFlow should be able to parse dbt metrics configs and build a Metricflow model. That model would then provide all the same interfaces that it does today (list metrics, get dimension values, query, materializations, etc.)

**Context** 

dbt is a popular data transformation framework with support for metric definitions. dbt metrics are new, but there is growing interest in the space, and MetricFlow should be able to support users of dbt natively. 

This feature would enable anyone using dbt metrics to use Metricflow as an interface to their metrics and transform as a self-serve BI tool without additional configuration.

**Requirements**

The happy product flow would look like this:

1. Connect to a dbt repo with dbt metrics
    1. The first version is just a way to query metrics from a dbt model using MetricFlow to keep things simple.
    2. MetricFlow parses dbt metrics into a MetricFlow model object, and then MetricFlow does everything as it would on a MetricFlow model.
2. Users can run all of the same MetricFlow commands just like they would for any other metric defined using MetricFlow’s config spec.

This flow would require the following new features in MetricFlow.

1. [P0] A new dbt metrics Parser [See details below]
2. [P1] Model updates to support count metric type(rather than sum, metrics as an input to expr metrics, a new meta parameter, and making owners optional
3. [P1] Validations would need some updates to run on a dbt metrics constructed model.
4. [P2] We should probably add a testing suite at some point that runs dbt metrics and Metricflow on the same dbt metrics spec and validates that we are keeping MetricFlow in sync to any dbt metrics changes.

**Parser Proposal**

There are some logic mapping decisions here, but it’s mostly pretty straightforward. I’m open to feedback!

1. Read in dbt metrics from YAML to a simple dbt metric object.
2. Build Data Sources - Use dbt metrics objects to generate data source MetricFlow objects, but do not write out new YAML. This is not conversion. It’s just reading from dbt metrics.
    1. Make data sources of type `sql_table`.
    2. Add a reference to the `dbt_model` param so that Transform/users building on MetricFlow could connect to dbt Cloud metadata and lineage. 
    3. Unfortunately, there are no identifiers, so there would be no joins.
    4. Collect all dimensions that are in the set of metrics built from a dbt model and generate dimensions from them
[...]

PR #292, opened 2022-10-17
Enable MetricFlow to be used with dbt projects with metrics

QMalcolm:
# The What
This is the initial implementation of the MetricFlow dbt metrics integration (#266). Allowing MetricFlow to work on top of a dbt project that has metrics 🎉 

# The How
Whats happening under the hood is actually fairly straightforward. First we parse the dbt project using `dbt` as library, into a dbt `Manfiest` object. This manifest gives us access to the ref resolver, all the metrics, and much more. With these we simply map specs of dbt metrics to a MetricFlow `UserConfiguredModel`. The `Metric.model` becomes the data source, the `Metric.dimensions` becomes the data source's dimensions, the metric itself becomes a measure and a measure proxy metric, and etc. Once we have the `UserConfiguredModel` it all becomes smooth sailing ⛵ 

# Using a dbt project with MetricFlow

## Reinstall MetricFlow with appropriate dbt-* extras.

MetricFlow's dbt integration currently supports `snowflake`, `redshift`, `postgres`, and `bigquery`. We plan on supporting this integration for `duckdb` and `databricks` shortly. For `duckdb` we are waiting on https://github.com/jwills/dbt-duckdb/issues/21 to be resolved. For `databricks` we need to resolve some dependency conflicts involving `databricks-sql-connector`.

### Reinstalling with Poetry
`poetry install metricflow -E dbt-<data-warehouse>`

### Reinstalling with Pip
`poetry install metricflow[dbt-<data-warehouse>]`

## Updating your `.metricflow/config.yaml`

To begin using this new functionality two changes must take place in ones `.metricflow/config.yml`.
1) update `model_path` to point at your dbt project root
   * e.g. `model_path: /path/to/your/dbt_project`
2) add `dbt_repo: true`
   * this is what tells MetricFlow it should expect dbt configs

After this is done, all other MetricFlow commands, `$ mf <command> <options>`, should work as normal but now from your dbt metrics! 🎉


tlento (approved):
Exciting!




Per discussion, original branch was reviewed piece by piece and this large merge is now ready with the one fixup commit appended!

Let's allow the engine tests to run before merging just to be sure there aren't any wonky dependency conflicts, although I don't think anything there will be different in any meaningful way.

PR #293, opened 2022-10-18
Re-establish static checking for exhaustive switch statements 

tlento:
A type change to our `assert_values_exhausted` helper resulted in
mypy accepting Enum-typed input for that method, which opened the
door to runtime failures that mypy would have caught had the
typing remained as NoReturn.

This restores the original behavior, adds some documentation, and
fixes the broken callsites.

This PR includes some minor poetry lock file updates in a separate commit.

### DIFF
--- a/metricflow/model/validations/validator_helpers.py
+++ b/metricflow/model/validations/validator_helpers.py
@@ -239,11 +239,11 @@ class ModelValidationResults(FrozenBaseModel):
         errors: List[ValidationError] = []
 
         for issue in issues:
-            if issue.level == ValidationIssueLevel.WARNING:
+            if issue.level is ValidationIssueLevel.WARNING:
                 warnings.append(issue)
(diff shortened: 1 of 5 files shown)


metricflow/model/validations/validator_helpers.py
```diff
@@ -239,11 +239,11 @@ def from_issues_sequence(cls, issues: Sequence[ValidationIssueType]) -> ModelVal
         errors: List[ValidationError] = []
 
         for issue in issues:
-            if issue.level == ValidationIssueLevel.WARNING:
```
QMalcolm, line 242: 👍

QMalcolm (approved):
Nice catch, thank you for figuring out what was up with `assert_values_exhausted` not working like expected

WilliamDee (approved):

PR #294, opened 2022-10-18
Fix dbt imports

QMalcolm:
The `dbt-*` packages are optional. This means that typically they won't be installed, and they shouldn't be installed unless you are actively developing `MetricFlow`/`dbt` functionality or using `MetricFlow`/`dbt` functionality. If `dbt-*` deps aren't installed, imports which eventually lead to dbt imports cause fatal exceptions. To get around this, we've moved the import of methods/classes which lead to dbt imports into the places they get used. Thus the exceptions will now only occur if you try to use dbt functionality without having installed an appropriate `dbt-*` dependency.

Additionally we've reverted changes to some of the github workflows which began installing the `dbt-snowflake` optional dependency.

-----

An example of the errors that were happening which no longer happen:
```
$ make test

ImportError while loading conftest '/Users/tqm/Developer/transform-data/metricflow/metricflow/test/conftest.py'.
metricflow/__init__.py:1: in <module>
    from metricflow.api.metricflow_client import MetricFlowClient  # noqa: D
metricflow/api/metricflow_client.py:11: in <module>
    from metricflow.engine.metricflow_engine import (
metricflow/engine/metricflow_engine.py:22: in <module>
    from metricflow.engine.utils import build_user_configured_model_from_config, build_user_configured_model_from_dbt_config
metricflow/engine/utils.py:10: in <module>
    from metricflow.model.parsing.dbt_dir_to_model import parse_dbt_project_to_model
metricflow/model/parsing/dbt_dir_to_model.py:4: in <module>
    from metricflow.model.transformations.dbt_to_metricflow import DbtManifestTransformer
metricflow/model/transformations/dbt_to_metricflow.py:42: in <module>
    class DbtManifestTransformer:
metricflow/model/transformations/dbt_to_metricflow.py:50: in DbtManifestTransformer
    def __init__(self, manifest: DbtManifest) -> None:
E   NameError: name 'DbtManifest' is not defined
make: *** [test] Error 4
```

### DIFF
--- a/.github/workflows/ci-metricflow-unit-tests.yaml
+++ b/.github/workflows/ci-metricflow-unit-tests.yaml
@@ -32,7 +32,7 @@ jobs:
         run: pip install poetry==1.1.15 && poetry config virtualenvs.create false
 
       - name: Install Metricflow
-        run: poetry install -E dbt-snowflake
+        run: poetry install
 
       - name: Run MetricFlow Unit tests suites
         run: pytest metricflow/test
--- a/.github/workflows/ci-schema-consistency.yaml
+++ b/.github/workflows/ci-schema-consistency.yaml
@@ -36,7 +36,7 @@ jobs:
         run: pip install poetry==1.1.15 && poetry config virtualenvs.create false
 
       - name: Install Metricflow
-        run: poetry install -E dbt-snowflake
+        run: poetry install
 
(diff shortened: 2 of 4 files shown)


marcodamore (approved):
Looks good to me, only request is that we add a comment explaining why we need to move the import down before merge

PR #296, opened 2022-10-18
Add validation to block DISTINCT keyword on COUNT aggregations

tlento:
The DISTINCT keyword is a problem for aggregation expressions,
because it effectively overrides the additivity of the aggregation
type. Since we rely on the aggregation type to tell us whether or
not a given measure is additive, we run the risk of doing an
inappropriate partial aggregation -> re-aggregation of these
measures.

This would produce incorrect results for a subset of queries,
which makes the framework behavior unpredictable.

Rather than allowing for this, we block the DISTINCT keyword from
use in COUNT expressions. Follow-ups should enable this safeguard for
SUM and AVG types, but COUNT(DISTINCT x) is generally more common
than SUM(DISTINCT x) so we start here.

### DIFF
--- a/metricflow/model/validations/measures.py
+++ b/metricflow/model/validations/measures.py
@@ -376,19 +376,39 @@ class CountAggregationExprRule(ModelValidationRule):
 
         for data_source in model.data_sources:
             for measure in data_source.measures:
+                context = DataSourceElementContext(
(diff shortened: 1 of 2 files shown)


WilliamDee (approved):
ah right, thank you!

QMalcolm (approved):
This looks great! Nice that there was an easy fix 🙂 Thanks for doing this

PR #300, opened 2022-10-19
added validator for derived metrics

WilliamDee:
## Context
With the addition of derived metrics, we need to validate that they are correctly configured.

## Cases
1. Derived metrics allows you to attach an alias to an input metric. The alias must be unique from the other aliases and other input metric names.
2. Validates that the input metric provided exists in the model

### DIFF
--- a/metricflow/model/validations/metrics.py
+++ b/metricflow/model/validations/metrics.py
@@ -5,6 +5,7 @@ from metricflow.errors.errors import ParsingException
 from metricflow.instances import MetricModelReference
 from metricflow.model.objects.metric import Metric, MetricType, CumulativeMetricWindow
 from metricflow.model.objects.user_configured_model import UserConfiguredModel
+from metricflow.model.validations.unique_valid_name import UniqueAndValidNameRule
 from metricflow.model.validations.validator_helpers import (
     FileContext,
     MetricContext,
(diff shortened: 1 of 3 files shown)


metricflow/model/validations/metrics.py
```diff
@@ -61,3 +62,64 @@ def validate_model(model: UserConfiguredModel) -> List[ValidationIssueType]:  #
...
+        if metric.type == MetricType.DERIVED:
+            metric_context = MetricContext(
+                file_context=FileContext.from_metadata(metadata=metric.metadata),
+                metric=MetricModelReference(metric_name=metric.name),
+            )
+            used_names = {input_metric.name for input_metric in metric.input_metrics}
+            for input_metric in metric.input_metrics:
+                if input_metric.alias:
+                    issues += UniqueAndValidNameRule.check_valid_name(input_metric.alias, metric_context)
+                    if input_metric.alias in used_names:
```
tlento, line 84: Do these need to be global?

What I mean is, the alias gets used at some point. Will that ever bubble up in the SQL to a cross-metric join? I think I made measure aliases be globally unique among all measures in the model just in case we end up with some optimization path that puts an alias and a measure with the same name into the same query.

I'm thinking of something like this:

```
metrics:
  - name: input_1
    type: measure_proxy
  - name: duplicate_name
    type: measure_proxy
  - name: ruh_roh
    type: derived
    expr: duplicate_name / input_1
    input_metrics:
      - name: input_1
        constraint: some_dimension > 5
        alias: duplicate_name 
      - name: input_1
 ```

Comment (WilliamDee) [reply metricflow/model/validations/metrics.py:84]: mmm is `ruh_roh` supposed to be configured incorrectly here? Cause from the current implementation, that would actually be correctly configured. Since it does't matter if there is another metric called `duplicate_name` as it'll only be affected by metrics within the `input_metrics`. For example, a query (`query --metric ruh_roh --dimensions metric_time`) for this would actually work as it will render an equivalent to
```sql
SELECT 
  metric_time
  duplicate_name / input_1 AS ruh_roh
FROM (
  SELECT
     input_1 AS duplicate_name, metric_time
  FROM ds
  WHERE some_dimension > 5
) a
FULL OUTER JOIN (
  SELECT input_1, metric_time FROM ds
) b
ON  a.metric_time = b.metric_time
```
not complete sql, excluded some group bys. But, the metric `duplicate_name` would be irrelevant to this metric.

tlento: Yeah that's what I'm wondering about - do we want to restrict `duplicate_name` in these aliases.

The issue we might encounter is one where there's some optimization where we re-use derived metric inputs, but I guess we can adapt the aliases somehow if that happens.

Now that I've written this out it's more of an issue with measures because the aliasing could happen inside a data source at some point, but metrics always live outside of data sources so we should be able to manage this.

tlento (approved):
I'm convinced we don't need to be stricter about our validation!

PR #301, opened 2022-10-19
Databricks: Enable secondary connection for table renames

courtneyholcomb:
When using databricks-sql-connector, you can connect using a cluster or a SQL warehouse. Some users prefer SQL warehouse because it can be more affordable. If your Databricks connection uses S3, however, you will get an error when you try to rename tables. We fix this issue for cluster connections by setting this config: `spark.databricks.delta.alterTable.rename.enabledOnAWS=true`. That config is not supported for SQL warehouse connections, so users are unable to rename tables when using a SQL warehouse connection with S3.
To work around this this for users who want to use S3 with a SQL warehouse, we will require them to specify a cluster connection to use exclusively for table renames. We can specify which connection to use by changing the HTTP path used when connecting to Databricks.

### DIFF
--- a/metricflow/sql_clients/databricks.py
+++ b/metricflow/sql_clients/databricks.py
@@ -18,6 +18,9 @@ from metricflow.sql_clients.common_client import SqlDialect
 logger = logging.getLogger(__name__)
 
 HTTP_PATH_KEY = "httppath="
+HTTP_PATH_RENAME_KEY = "httppathrename="
+SQL_RENAME = " RENAME TO "
+SQL_ALTER_TABLE = "ALTER TABLE "
 
 # This is a non-exhaustive list of pandas dtypes, but in theory it will cover the ones we need to support
 # for data frames generated and run through type inference.
@@ -58,10 +61,13 @@ class DatabricksEngineAttributes(SqlEngineAttributes):
 class DatabricksSqlClient(BaseSqlClientImplementation):
     """Client used to connect to Databricks engine."""
 
-    def __init__(self, host: str, http_path: str, access_token: str) -> None:  # noqa: D
+    def __init__(  # noqa: D
+        self, host: str, http_path: str, access_token: str, http_path_for_table_renames: Optional[str] = None
+    ) -> None:
         self.host = host
         self.http_path = http_path
         self.access_token = access_token
+        self.http_path_for_table_renames = http_path_for_table_renames
 
     @staticmethod
     def from_connection_details(url: str, password: Optional[str]) -> DatabricksSqlClient:  # noqa: D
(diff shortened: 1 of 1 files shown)


metricflow/sql_clients/databricks.py
```diff
@@ -18,6 +18,9 @@
 logger = logging.getLogger(__name__)
 
 HTTP_PATH_KEY = "httppath="
+HTTP_PATH_RENAME_KEY = "httppathrename="
```
tlento, line 21: Does this only trigger on renames, or other alter table commands? If not, we should name this key something more generic, and maybe broaden out our match pattern below.

If we don't know let's leave it this way - since this isn't part of the public interface for SqlClient we should be able to change it easily later.

metricflow/sql_clients/databricks.py
```diff
@@ -87,17 +101,27 @@ def from_connection_details(url: str, password: Optional[str]) -> DatabricksSqlC
...
-        return DatabricksSqlClient(host=parsed_url.host, http_path=http_path, access_token=password)
+        return DatabricksSqlClient(
+            host=parsed_url.host,
+            http_path=http_path,
+            access_token=password,
+            http_path_for_table_renames=http_path_for_table_renames,
+        )
 
-    def get_connection(self) -> sql.client.Connection:
+    def get_connection(self, table_rename: bool = False) -> sql.client.Connection:
```
tlento, line 118: micro-nit: `is_table_rename` or `is_for_table_rename`

metricflow/sql_clients/databricks.py
```diff
@@ -87,17 +101,27 @@ def from_connection_details(url: str, password: Optional[str]) -> DatabricksSqlC
...
         """Get connection to Databricks cluster/warehouse."""
-        return sql.connect(server_hostname=self.host, http_path=self.http_path, access_token=self.access_token)
+        return sql.connect(
+            server_hostname=self.host,
+            http_path=self.http_path_for_table_renames if table_rename else self.http_path,
```
tlento, line 122: Do we want to do this here? The pro is anybody using this as a standalone client can actually do so even if Databricks throws that error in their face, but that's not really a thing we support.

tlento (commented):
As a higher level note - we don't really do alter table commands like this in open source as far as I know. In the contexts where we do, I guess we need at least parts of this change so that we can collect the alternate path information, is that right? Or could we somehow get that information into the Transform client implementation and just apply the overrides to the relevant methods over there?

metricflow/sql_clients/databricks.py
courtneyholcomb: Only renames!

tlento (approved):
As far as I know we don't currently need this in the open source project, but if we ever do have rename operations this will be useful.

Based on this and our other offline discussion I think it makes sense to do this here unless it's super trivial to support the rename scenario on the Transform side.

PR #304, opened 2022-10-20
Enable pre-commit typechecking for dbt packages

tlento:
Metricflow has an override in its .pre-commit-config to ignore all
missing imports, which means third party packages simply don't get
typechecked in any case where the mypy invocation is mediated by
pre-commit (which is roughly all of them, for us).

This means the dbt integration was done with a relative minimum of
typechecker support.

Here we enable dbt typchecking in the .pre-commit-config (the remaining
changes were an unfortunate side effect of auto-formatting) by adding
dbt-core to the dependency list, and clean up the typing errors this
exposed as a result.

Note these fixes are just wallpapering over the underlying issues, so
more robust changes may be warranted.

### DIFF
--- a/metricflow/model/transformations/dbt_to_metricflow.py
+++ b/metricflow/model/transformations/dbt_to_metricflow.py
@@ -65,7 +65,9 @@ class DbtManifestTransformer:
         dimension should be primary for
         """
         if not self._time_dimension_stats:
-            self._time_dimension_stats = self.collect_time_dimension_stats(dbt_metrics=self.manifest.metrics.values())
+            self._time_dimension_stats = self.collect_time_dimension_stats(
+                dbt_metrics=list(self.manifest.metrics.values())
+            )
 
         return self._time_dimension_stats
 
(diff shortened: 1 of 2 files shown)


metricflow/model/transformations/dbt_to_metricflow.py
```diff
@@ -339,7 +342,7 @@ def collect_time_dimension_stats(cls, dbt_metrics: List[DbtMetric]) -> Dict[str,
         # dimenson for the data source that is built for the `DbtMetric.model`
         time_stats_for_metric_models: Dict[str, Dict[str, int]] = {}
         for dbt_metric in dbt_metrics:
-            if dbt_metric.calculation_method != "derived":
+            if dbt_metric.calculation_method != "derived" and dbt_metric.model is not None:
```
QMalcolm, line 345: I believe all dbt metrics that aren't `derived` must have a `model`. However, I'm not sure if dbt enforces this when parsing the dbt project / building the manifest. Thus although we likely don't need the extra check, it doesn't hurt.

QMalcolm (approved):

PR #310, opened 2022-10-20
Actions cleanup: use more recent Poetry and Python versions

tlento:
Our actions were pinned to Poetry 1.1.15 due to an outstanding issue 
with virtual env creation configuration. For historical reasons, we were
also testing exclusively in Python 3.8.

This PR moves us to a model where we're using the latest Poetry version
for build management, and running all actions against Python 3.9 (our latest
supported Python version) EXCEPT our core unit tests, which are run against
all supported Python versions.

This shift to matrixed Python versions presents a separate issue, resolved here
as well, where our merge-blocking unit test job is effectively forked into N
separate jobs, one for each version. Since branch protection is bound to job
name, and the job names are generated in the matrix configuration, we'd need
to keep updating the branch protection against every python version we test.

The solution used here is to have the required job depend on success for all
upstream job via the `needs` keyword, the `always()` condition for running,
and a test against input job status. This will always fail if any of its input jobs
fails, and always succeed if all of its input jobs succeed.

### DIFF
--- a/.github/workflows/cd-push-to-pypi.yaml
+++ b/.github/workflows/cd-push-to-pypi.yaml
@@ -12,10 +12,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Set up Python 3.8
-        uses: actions/setup-python@v2
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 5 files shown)


marcodamore (approved):
Great use of the matrix 💪

PR #312, opened 2022-10-21
Use INNER JOIN for derived metrics when combining metrics

WilliamDee:
## Context
With the merge of Derived Metrics, we supported the ability to reference metrics and build an expr out of it. The way we aggregate all the metrics data is through the `CombineMetricsNode` which utilizes a `FULL OUTER JOIN` to combine the metric data. This would mean that there are cases where some metrics would return rows of `NaN` or `null` values because it doesn't have data associated to the metric time that other metrics have. This would result in the final result having `NaN` or `null` value. 

## Changes
With further discussions on https://github.com/transform-data/metricflow/pull/281#discussion_r998590378, we decided to use an `INNER JOIN` when combining metrics for derived metrics so that we don't have null values in the results.

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -66,6 +66,7 @@ from metricflow.specs import (
     ColumnAssociationResolver,
     LinklessIdentifierSpec,
 )
+from metricflow.sql.sql_plan import SqlJoinType
 from metricflow.time.time_granularity import TimeGranularity
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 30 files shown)


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -157,6 +158,7 @@ def _build_metrics_output_node(
         queried_linkable_specs: LinkableSpecSet,
         where_constraint: Optional[SpecWhereClauseConstraint] = None,
         time_range_constraint: Optional[TimeRangeConstraint] = None,
+        combine_metrics_with_inner_join: bool = False,
```
tlento, line 161: Can we add a docstring to explain the usage of this boolean?

Alternatively, we could remove the boolean in favor of a MetricType, something like `output_metric_type: Optional[MetricType] = None` and then set the join type based on a check for `output_metric_type is MetricType.DERIVED`.

tlento (approved):

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -157,7 +158,17 @@ def _build_metrics_output_node(
         queried_linkable_specs: LinkableSpecSet,
         where_constraint: Optional[SpecWhereClauseConstraint] = None,
         time_range_constraint: Optional[TimeRangeConstraint] = None,
+        combine_metrics_join_type: SqlJoinType = SqlJoinType.FULL_OUTER,
```
tlento, line 161: Nice!

PR #313, opened 2022-10-21
Databricks: parse dry run errors properly

courtneyholcomb:
SQL warehouse and cluster connections return differently structured results for `EXPLAIN` queries. This PR adds support for properly parsing the errors raised by SQL warehouse connections.


metricflow/sql_clients/databricks.py
```diff
@@ -165,10 +167,15 @@ def _engine_specific_dry_run_implementation(self, stmt: str, bind_params: SqlBin
...
-                    error = result.split("== Physical Plan ==")[1].split(";")[0]
+                # If the plan contains errors, they won't be raised. Parse results to find & raise errors.
+                result = cursor.fetchall_arrow()["plan"]
+                str_result = str(result)
+                if SQL_WAREHOUSE_ERROR_KEY in str_result:
+                    error = "".join([str(row) for row in result])
+                    raise sql.exc.ServerOperationError(error)
+
+                if CLUSTER_ERROR_KEY in str_result:
+                    error = str(result[0]).split("== Physical Plan ==")[1].split(";")[0]
```
tlento, line 178: Le sigh....

tlento (approved):
Maybe we can document this along with the general S3 issue and open an issue on their core project. I don't think this is the Python DBAPI, really - seems to be an issue on the engine side - so I'm not sure where we'd raise this. I'll look into it.

Thanks for the fix!

PR #314, opened 2022-10-21
Fix errors in SQL client implementations

tlento:
- Update dependencies to explicitly include google-auth
- Clean up wonky imports in BigQuery sql client implementation
- Fix initialization gap in Databricks client


courtneyholcomb (approved):
Thank you!

tlento:
I think the BigQuery test might be deterministically broken, but I'm going to merge this to unbreak the Databricks client and debug that test failure separately since I have to run it in CI to repro.

plypaul:
TY!

PR #315, opened 2022-10-21
Provide exception state info in test_cancel_request

tlento:
BigQuery is throwing exceptions in CI in test_cancel_request.
The CI logs don't indicate anything other than an assertion error,
which is unhelpful for debugging.


tlento:
Never mind, I can't read.

PR #317, opened 2022-10-21
Databricks: only use secondary connection if it exists

courtneyholcomb:
Currently, if the user has not set up a secondary connection and tries to execute a table rename, this fails. This PR fixes that!


tlento (approved):
This S3 table rename business is super confusing for me.

PR #318, opened 2022-10-21
Make derived metrics tests use double types

tlento:
The derived metrics tests were all implicitly checking BIGINT types,
since that's what SUM returns. This changes them to use DOUBLE
for proper ratio computation, which will bring the answers in line
with expectation based on manual calculation against the table fixtures.

Note we cannot do explicit casting in the metric definition expr because
DOUBLE types are not consistently named across engines, so we rely on
the implicit casting shorthand of multiplying by a double literal.

### DIFF
--- a/metricflow/test/compare_df.py
+++ b/metricflow/test/compare_df.py
@@ -1,7 +1,7 @@
 import logging
 import math
 import pandas as pd
-
+from typing import SupportsFloat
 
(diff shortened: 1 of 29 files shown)


plypaul (approved):

tlento:
Flaky BigQuery cancel test strikes again!

PR #322, opened 2022-10-26
FIX: DW measure validation was using wrong source node

WilliamDee:
## Context
In the SourceNodeBuilder, we split the measures into smaller source nodes by the `agg_time_dimension`  property. In the DW validator, we get all the source nodes for the data source and just use the first one via `source_nodes[0]`. This means if we have a data source with 2 measures with a different `agg_time_dimension`, we would correctly validate 1 measure while the other measure would throw an error that looks like (`MeasureSpec(measure1) not in specset ...`). Again, this is due to simply using 1 source node for validating all the measures when we should really be selecting the correct source node that contains the measure we are trying to query.

## Changes
- Select the correct source node to pass to the `FilterElementNode` when validating a measure

### DIFF
--- a/metricflow/model/data_warehouse_model_validator.py
+++ b/metricflow/model/data_warehouse_model_validator.py
@@ -1,12 +1,14 @@
 from __future__ import annotations
 from copy import deepcopy
 
+import collections
 from dataclasses import dataclass, field
 from functools import partial
 from math import floor
 from time import perf_counter
 import traceback
-from typing import Callable, List, Optional, Tuple
+from typing import Callable, DefaultDict, Dict, List, Optional, Sequence, Tuple
+from metricflow.dataflow.builder.node_data_set import DataflowPlanNodeOutputDataSetResolver
 from metricflow.dataflow.builder.source_node import SourceNodeBuilder
 from metricflow.dataflow.dataflow_plan import BaseOutput, FilterElementsNode
 from metricflow.dataset.convert_data_source import DataSourceToDataSetConverter
@@ -36,7 +38,7 @@ from metricflow.plan_conversion.column_resolver import DefaultColumnAssociationR
 from metricflow.plan_conversion.dataflow_to_sql import DataflowToSqlQueryPlanConverter
 from metricflow.plan_conversion.time_spine import TimeSpineSource
 from metricflow.protocols.sql_client import SqlClient
-from metricflow.specs import DimensionSpec, LinkableInstanceSpec
+from metricflow.specs import DimensionSpec, LinkableInstanceSpec, MeasureSpec
 from metricflow.sql.sql_bind_parameters import SqlBindParameters
 
 
@@ -66,6 +68,11 @@ class QueryRenderingTools:
             semantic_model=self.semantic_model,
             time_spine_source=self.time_spine_source,
         )
+        self.node_resolver = DataflowPlanNodeOutputDataSetResolver[DataSourceDataSet](
+            column_association_resolver=DefaultColumnAssociationResolver(self.semantic_model),
+            semantic_model=self.semantic_model,
+            time_spine_source=self.time_spine_source,
+        )
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


metricflow/model/data_warehouse_model_validator.py
```diff
@@ -340,25 +358,6 @@ def gen_measure_tasks(
                         error_message=f"Unable to query measure `{spec.element_name}` on data source `{data_source.name}` in data warehouse",
                     )
                 )
-
```
WilliamDee, line 343: @QMalcolm I think we don't actually need this step, as we are already querying for all the measures above one by one, this step seems to simply query them altogether which seems redundant. Unless there's another reason why we need it, then feel free to let me know and I'll add it back.

To iterate on that, what I mean is we already query 1 by 1 as followed in the above code,
```sql
SELECT measure1 FROM data_source; SELECT measure2 FROM data_source; SELECT measure3 FROM data_source;
```
then this next part we just do
```sql
SELECT measure1, measure2, measure2 FROM data_source
```

metricflow/model/data_warehouse_model_validator.py
```diff
@@ -341,25 +359,6 @@ def gen_measure_tasks(
...
-            filter_elements_node = FilterElementsNode(parent_node=source_node, include_specs=data_source_specs)
-            tasks.append(
```
tlento, line 346: The test failures are caused, at least in part, by removing this task. It does serve a purpose but @QMalcolm is a lot more familiar with it than I am so he'd know best how to get it fixed up.

metricflow/model/data_warehouse_model_validator.py
```diff
@@ -341,25 +359,6 @@ def gen_measure_tasks(
                         error_message=f"Unable to query measure `{spec.element_name}` on data source `{data_source.name}` in data warehouse",
                     )
                 )
-
-            filter_elements_node = FilterElementsNode(parent_node=source_node, include_specs=data_source_specs)
```
tlento, line 345: I think we need to make the filter specs be the union of all measure specs here and keep the other task in but I admit I'm guessing a bit.

tlento (commented):
The test failures seem legit to me - I think we are no longer validating all of the measures, and that's why the bad one isn't failing - it just so happens to be getting filtered out now. But I'm only like 60% sure if that's true.

metricflow/model/data_warehouse_model_validator.py
WilliamDee: ok i realized what the differences here are. This union of all measure spec is the first task, then if it fails we call the individual querys from above

WilliamDee: there seems to be an issue tho with the validation

metricflow/model/data_warehouse_model_validator.py
WilliamDee: NVM!

QMalcolm: Some extra context on this. Apologies for future people looking pack at this. Also, sorry that happened through a separate line of communication. I wasn't at my computer during this, and thus the details were communicated through slack at the time.

The big query is a short cut query. The queries for individual measures are run only if the short cut query fails. This is because we want to reduce the number of total queries run, thus if the short cut query passes, that means everything is okay and we don't need to run the individual queries. But if the short cut query fails, that means one or more of the items queried caused an issue. The key part is *one or more*. The error we get back from the data ware house will only be for the *first* queried item that failed. But, we want to create issues *for all* bad query items. Thus to find out how many and which of the queried items are bad we have to do the individual queries. This makes the happy path REALLY FAST. It does mean that in the worst case, wherein each data source has one one or more failed dimensions / measures / identifiers, *a lot* of queries get run. But this is to ensure we give complete information on all everything that is in a bad state.

QMalcolm (approved):
Thanks for putting so much time in on this @WilliamDee. Sorry that I couldn't pair directly. This looks great 💯 Thanks again!

PR #323, opened 2022-10-26
Remove `update_cron` Default From Mutability

plypaul:
This PR removes the `update_cron` default as it is not used in the open source project so it will be confusing.


tlento (approved):
Seems reasonable!

PR #325, opened 2022-10-31
Enable querying with no dimension using CROSS JOIN

WilliamDee:
## Context
Follow up on https://github.com/transform-data/metricflow/pull/256. We blocked the ability to query multiple metrics or a metric that joins across multiple data sources with no dimension. This was because it would break the SQL generation as the current assumption is that we would have linkable elements to join on. Which means if no dimensions were passed in, we would be left with an incomplete JOIN clause. ie.,
```sql
SELECT ... FROM ds1
INNER JOIN ds2
ON
```
It was also not possible to just remove the `ON` clause or do an `ON TRUE` for it to act as a `CROSS JOIN` because redshift did not allow that syntax. This meant that CROSS JOIN would need to be integrated in as a join type.

## Changes
- Added `CROSS JOIN` as a `SqlJoinType`
- Made `SqlJoinDescription.on_condition` optional as `CROSS JOIN` does not use an `ON` clause
- Use a cross join when we combine metrics and join measures across data sources if no linkable element is provided.
- Alleviate "not supported" assertions


metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -915,8 +916,10 @@ def visit_join_aggregated_measures_by_groupby_columns_node(  # noqa: D
                         column_aliases=join_aliases,
                         left_source_alias=from_data_set_alias,
                         right_source_alias=right_data_set_alias,
-                    ),
-                    join_type=SqlJoinType.INNER,
+                    )
+                    if not use_cross_join
+                    else None,
```
tlento, line 921: Oof, this is hard to read, but don't worry about it - I'm refactoring all of these join helpers separately so I'll deal with it when I get to this node type.

metricflow/query/query_parser.py
```diff
@@ -158,6 +159,18 @@ def parse_and_validate_query(
...
+    def _validate_no_dimension_query_for_metric(self, metric_references: Sequence[MetricReference]) -> None:
+        """Validate if all requested metrics are queryable without any dimensions."""
+        for metric_reference in metric_references:
+            metric = self._metric_semantics.get_metric(metric_reference)
+            if metric.type == MetricType.CUMULATIVE:
+                # Cumulative metrics configured with a window/grain_to_date cannot be queried without a dimension.
+                if metric.type_params.window or metric.type_params.grain_to_date:
+                    raise UnableToSatisfyQueryError(
+                        f"Metric {metric.name} is a cumulative metric specified with a window/grain_to_date "
+                        f"which cannot be queried without a dimension.",
```
tlento, line 171: Why not? Is this because they require a time dimension? If there's a specific dimension requirement (e.g., must include `metric_time`) it'd be good to update the error message accordingly.

metricflow/test/integration/test_cases/itest_metrics_no_dim.yaml
```diff
@@ -0,0 +1,77 @@
+---
```
tlento, line 1: Can we rename this test file to either `itest_metric_queries_no_dimensions` or similar?

metricflow/test/plan_conversion/test_dataflow_to_sql_plan.py
```diff
@@ -1449,3 +1449,50 @@ def test_multi_hop_to_scd_dimension(
...
+    convert_and_check(
+        request=request,
+        mf_test_session_state=mf_test_session_state,
+        dataflow_to_sql_converter=dataflow_to_sql_converter,
+        sql_client=sql_client,
+        node=dataflow_plan.sink_output_nodes[0].parent_node,
+    )
+
+
+def test_metric_with_measures_from_multiple_sources(  # noqa: D
```
tlento, line 1479: Can we put the fact that there are no groupbys in this test name somewhere? You'll have to move all the plan files around.

tlento (commented):
Looks great! I left a couple of nit-picky comments but nothing really blocking.

Please update with the remaining plan files and run the full suite of engine tests - I think we should run the CROSS JOIN in CI on all engines before merging just in case something weird happens.

tlento (approved):
Looks great!

PR #326, opened 2022-11-01
Factor join building logic out of DataflowToSqlQueryPlanBuilder

tlento:
The DataflowToSqlQueryPlanConverter is enormously long and difficult to read
due to the sheer number of node-specific operations it manages. In particular,
join building logic includes a number of ancillary, heavily overlapping helper
functions plus a fair bit of metadata juggling in service of constructing
SqlJoinDescription objects to describe how the join should be rendered.

This PR moves most of the SqlJoinDescription building into a dedicated helper
class, and refactors the individual methods to make them more tightly focused
and hopefully easier to understand.

The lone exception is the join described in the visitor method for combining metrics,
as this logic is already delegated to a special InstanceSetTransform, and moving it
did not appear to be a meaningful improvement.

### DIFF
--- a/metricflow/plan_conversion/sql_join_builder.py
+++ b/metricflow/plan_conversion/sql_join_builder.py
@@ -25,23 +25,46 @@ def _make_cumulative_metric_join_condition(
     time_spine_data_set_alias: str,
     time_spine_data_set_time_col: str,
     window: Optional[CumulativeMetricWindow],
+    grain_to_date: Optional[TimeGranularity],
 ) -> SqlExpressionNode:
-    """Creates SqlLogicalExpression representing a cumulative metric self-join condition"""
-    if window is None:  # accumulate over all time
-        return SqlComparisonExpression(
-            left_expr=SqlColumnReferenceExpression(
-                SqlColumnReference(
-                    table_alias=metric_data_set_alias,
-                    column_name=metric_data_set_metric_time_col,
-                )
-            ),
-            comparison=SqlComparison.LESS_THAN_OR_EQUALS,
-            right_expr=SqlColumnReferenceExpression(
-                SqlColumnReference(
-                    table_alias=time_spine_data_set_alias,
-                    column_name=time_spine_data_set_time_col,
-                )
-            ),
+    """Creates SqlExpression representing a cumulative metric self-join conditions"""
+
+    # Comparison expression against the endpoint of the cumulative time range,
+    # meaning the metric time must always be BEFORE the time spine time
+    end_of_range_comparison_expression = SqlComparisonExpression(
+        left_expr=SqlColumnReferenceExpression(
+            SqlColumnReference(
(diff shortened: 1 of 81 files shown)


tlento:
* main
  * **PR #326** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/326" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/transform-data/metricflow/326?utm_source=stack-comment).

metricflow/plan_conversion/sql_join_builder.py
```diff
@@ -14,18 +15,14 @@
...
 def _make_cumulative_metric_join_condition(
-    metric_data_set_alias: str,
-    metric_data_set_metric_time_col: str,
-    time_spine_data_set_alias: str,
-    time_spine_data_set_time_col: str,
-    window: Optional[CumulativeMetricWindow],
-    grain_to_date: Optional[TimeGranularity],
+    metric_data_set: AnnotatedSqlDataSet,
+    time_spine_data_set: AnnotatedSqlDataSet,
+    node: JoinOverTimeRangeNode[SqlDataSetT],
```
WilliamDee: is there a reason we put a node here instead of using the window, grain_to_date arguments? Are we just restricting this to only be used by places that utilizes the JoinOverTimeRangeNode? I was wondering cause for conversion metrics, I did want to make use of this but it's not really joining to a time_spine, but rather joining against some time dimension between the 2 datasets. But other than that the entire windowing logic would be the same.

WilliamDee (approved):
wow... this looks really good!! thank you for consolidating and cleaning everything up. Went through each commit and everything looked good to me. Though had a question related to join logic in conversion metrics that I'm working on now. There's alot of moving things around and simplifying code here, but I didn't see anything jarring there. I approved, but would be great if @plypaul took a pass on this as well!

metricflow/plan_conversion/sql_join_builder.py
tlento: I made the helper very explicitly about cumulative metrics because I found the JoinOverTimeRangeNode to really be a CumulativeMetricTimeSpineJoinNode - it's not readily generalizable because of the way it couples to the visitor method that converts it to SQL. We can - and probably should - refactor this along with the other 3-4 time window join use cases we're working on.

If you'd like to modify the name and signature of this helper and broaden its utility that's perfectly fine. Personally, I was thinking of getting everything in there and then figuring out how to make the design more sensible. We need a holistic approach to these time window joins (and likely to range joins more generally), because specifying which point in the range to favor is something users will need to be able to do. I could be very wrong about this, but the ad hoc "window" and "grain to date" properties in the cumulative metrics node don't feel like the right foundation to me.

plypaul (approved):
LGTM - thanks @tlento !

PR #329, opened 2022-11-04
Add TimeRangeConstraint Representation Invariant

plypaul:
As per title.


metricflow/constraints/time_constraint.py
```diff
@@ -17,6 +17,20 @@ class TimeRangeConstraint(SerializableDataclass):
     start_time: datetime.datetime
     end_time: datetime.datetime
 
+    def __post_init__(self) -> None:  # noqa: D
+        if self.start_time > self.end_time:
+            raise RuntimeError(
+                f"start_time must not be > end_time. start_time={self.start_time} end_time={self.end_time}"
+            )
+
+        if self.start_time < TimeRangeConstraint.ALL_TIME_BEGIN():
```
tlento, line 26: Shall we switch this to epoch start instead of the year 2000? Seems more in line with other computer date systems.

metricflow/constraints/time_constraint.py
```diff
@@ -17,6 +17,20 @@ class TimeRangeConstraint(SerializableDataclass):
...
+            raise RuntimeError(
+                f"start_time={self.start_time} exceeds the limits of {TimeRangeConstraint.ALL_TIME_BEGIN()}"
+            )
+
+        if self.end_time > TimeRangeConstraint.ALL_TIME_END():
+            raise RuntimeError(f"end_time={self.end_time} exceeds the limits of {TimeRangeConstraint.ALL_TIME_END()}")
```
tlento, line 32: 2040 isn't too far away but I guess that's alright.

Out of historical instance, why is this here? And why is it a fixed constant? Shouldn't we make it slide forward based on current date?

tlento (commented):
Conceptually great, needs a fix for that test failure, and I think the right thing to do is update our range boundaries.

metricflow/constraints/time_constraint.py
plypaul: So the limits of the `TimeRangeConstraint` are used for the time spine table. We could increase this, but that would cause the size of the time spine table to be larger, which adds a bit of overhead when creating it.

Generally prefer not to make it sliding to get better consistency + easier debugging.

tlento: > the limits of the TimeRangeConstraint are used for the time spine table

Ah, makes sense!

metricflow/test/time/test_time_granularity_solver.py
```diff
@@ -223,10 +223,7 @@ def test_adjusted_time_constraint_at_boundaries(
     """Tests the time range adjustment behavior at the min / max boundaries"""
     assert (
         time_granularity_solver.adjust_time_range_to_granularity(
-            time_range_constraint=TimeRangeConstraint(
-                start_time=TimeRangeConstraint.all_time().start_time + datetime.timedelta(days=-30),
-                end_time=TimeRangeConstraint.all_time().end_time + datetime.timedelta(days=30),
```
tlento, line 228: LOL. The reason this failed is because the timedelta signs were backwards.

tlento (approved):
LGTM!

PR #331, opened 2022-11-05
Support multiple types for SqlBindParameters

courtneyholcomb:
Currently we only support strings in `SqlBindParameters`. This change allows us to support the primary column types used in our supported DB engines. Note that when a pydantic param is a string, anything passed into it will be converted to string.
This also updates `hash_strings` to be `hash_items` (not used here, but will be used in the paid product).
Completes MF-551

### DIFF
--- a/metricflow/object_utils.py
+++ b/metricflow/object_utils.py
@@ -215,4 +215,4 @@ class ExtendedEnum(Enum):
 
 
 # Supported SQL column types (not comprehensive).
-SqlColumnType = Union[str, int, float, datetime.datetime, datetime.date, bool, None]
+SqlColumnType = Union[str, int, float, datetime.datetime, datetime.date, bool]
--- a/metricflow/object_utils.py
+++ b/metricflow/object_utils.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 import itertools
 import logging
 import pprint
@@ -8,9 +9,9 @@ from collections import OrderedDict
 from collections.abc import Mapping
 from dataclasses import is_dataclass, fields
 from enum import Enum
+import datetime
 from hashlib import sha1
-from typing import Sequence, TypeVar, Tuple, NoReturn, Type, Any, List
-from metricflow.sql_clients.sql_utils import SQLColumnType
+from typing import Sequence, TypeVar, Tuple, NoReturn, Type, Any, List, Union
 
 from metricflow.model.objects.base import HashableBaseModel
 
@@ -211,3 +212,7 @@ class ExtendedEnum(Enum):
     def list_names(cls) -> List[str]:
         """List valid names within this enum class"""
         return list(cls.__members__.keys())
+
+
+# Supported SQL column types (not comprehensive).
+SQLColumnType = Union[str, int, float, datetime.datetime, datetime.date, bool]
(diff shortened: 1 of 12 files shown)


metricflow/sql_clients/databricks.py
```diff
@@ -137,7 +138,7 @@ def sql_engine_attributes(self) -> SqlEngineAttributes:
         return DatabricksEngineAttributes()
 
     @staticmethod
-    def params_or_none(bind_params: SqlBindParameters) -> Optional[Dict[str, str]]:
+    def params_or_none(bind_params: SqlBindParameters) -> Optional[Dict[str, SQLColumnType]]:
```
tlento, line 141: This currently works by accident, I think:

https://github.com/databricks/databricks-sql-python/blob/main/src/databricks/sql/client.py#L301

MyPy isn't catching it because we aren't typechecking external libraries correctly, but as soon as we fix that issue this will break.

For most things it should work in their runtime as long as the parameter spec and value type match (e.g., the spec is `%(colname)d` for an `int` parameter value), because their parameter escaper takes in an `Any` and returns the escaped value true to type, so when they do `query_string % escaped_parameters` everything will be fine.

However, they don't support booleans:

https://github.com/databricks/databricks-sql-python/blob/main/src/databricks/sql/utils.py#L160-L174

This seems like a bummer. :(

metricflow/sql/sql_bind_parameters.py
```diff
@@ -12,7 +13,7 @@ class SqlBindParameters(HashableBaseModel):
     These can be used as per https://docs.sqlalchemy.org/en/14/core/tutorial.html#using-textual-sql
     """
 
-    param_dict: Dict[str, str] = OrderedDict()
+    param_dict: Dict[str, SQLColumnType] = OrderedDict()
```
tlento, line 16: Can we add test cases for non-string param types? Since we don't directly use the parameters in Metricflow itself there's no integration test coverage for this - we'll only know it breaks if an end user passes a parameterized query down.

tlento (commented):
I like this because it allows us to properly handle typed values, but I'm sad because I suspect the downstream services might do silly things - the only one I checked (Databricks) is already causing problems. :(

I'm not sure how best to proceed at this point.

metricflow/object_utils.py
```diff
@@ -210,3 +212,7 @@ def for_name(cls: Type[T], name: str) -> T:
     def list_names(cls) -> List[str]:
         """List valid names within this enum class"""
         return list(cls.__members__.keys())
+
+
+# Supported SQL column types (not comprehensive).
+SQLColumnType = Union[str, int, float, datetime.datetime, datetime.date, bool]
```
plypaul, line 218: Nit: We've been using `Sql` instead of `SQL` for names.

plypaul (approved):

metricflow/sql_clients/databricks.py
courtneyholcomb: As discussed:
1. I'll put up some integration tests to make sure this works for all engines.
2. For any that don't support other types, I'll add engine-specific logic to convert to string before passing to the SQL connector.

tlento (commented):
Something weird is happening with those tests...

tlento:
@courtneyholcomb looks like this is just out of date - @plypaul added a new test case, and the environment deploy is basing this change on current main, and that's where the failure is coming from: https://github.com/transform-data/metricflow/blob/main/metricflow/test/sql_clients/test_sql_client.py#L220-L224

tlento:
And now every engine is failing for a different reason. 😭

PR #332, opened 2022-11-08
Fix DuckDB Concurrency Issues

plypaul:
* Using a Lock instead of RLock as DuckDB can throw odd errors with concurrent queries, even in the same thread.
* Protecting `list_tables` with a lock as it also accesses the engine.

### DIFF
--- a/metricflow/sql_clients/duckdb.py
+++ b/metricflow/sql_clients/duckdb.py
@@ -62,7 +62,7 @@ class DuckDbSqlClient(SqlAlchemySqlClient):
     def __init__(self, file_path: Optional[str] = None) -> None:  # noqa: D
         # DuckDB is not designed with concurrency, but in can work in multi-threaded settings with
         # check_same_thread=False, StaticPool, and serializing of queries via a lock.
-        self._concurrency_lock = threading.RLock()
+        self._concurrency_lock = threading.Lock()
         super().__init__(
             sqlalchemy.create_engine(
                 f"duckdb:///{file_path if file_path else ':memory:'}",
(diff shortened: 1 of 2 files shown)


tlento (approved):

metricflow/sql_clients/duckdb.py
```diff
@@ -114,3 +114,7 @@ def create_table_from_dataframe(  # noqa: D
 
     def cancel_request(self, pattern_tag_set: SqlRequestTagSet) -> int:  # noqa: D
         raise NotImplementedError
+
+    def list_tables(self, schema_name: str) -> Sequence[str]:  # noqa: D
+        with self._concurrency_lock:
+            return self._engine.table_names(schema=schema_name)
```
tlento, line 120: I'm assuming this has been tested.

table_names is deprecated so at some point we should add better integration test coverage for this method and move on to the new approach.

tlento (approved):

metricflow/sql_clients/duckdb.py
plypaul: It was tested as a part of an integration test, and I assumed that a test for `list_tables` was in `test_sql_client.py`. Surprisingly, it was not, so I added one there.

plypaul: I did look into using the `information_schema` for `list_tables`, but there was one issue with Snowflake that I couldn't resolve, so that's been on the back burner.

tlento: Yeah I tried to update a while back and hit issues as well, probably the same deal. Having the explicit test will hopefully make it easier to futz with it - I'll open an issue.

Issue #158, opened 2022-07-14
Allow aggregating measures by a custom time dimension 


tlento:
**Describe the Feature**

Prior to this all data sources had to have a primary time dimension with the same name, but there are lots of use cases for:

1. Allowing data sources to have different time dimension names (like maybe one has a primary time of "day" and another has a primary time of "hour")
2. Allowing measures within a data source to have different primary times (e.g., an `expected_revenue` measure might be based on `billing_date` and `revenue` might be based on `payment_received_date`, and people might want those in the same data source)

tlento:
This has been merged to master, but as of today it has not been formally released.

Note this is potentially a breaking change on the query side and in materialization schema output, so we'll leave the issue open as a reminder to announce the nuances.

PR #334, opened 2022-11-08
If a data source's measures all have an `agg_time_dimension`, a primary time dimension is not requried

QMalcolm:
We added the concept of measure `agg_time_dimension` in #158. Which means that it's no longer required for a data source with measures to have a primary time dimension, so long as every measure has an `agg_time_dimension` set. This PR catches validations up to that reality.

### DIFF
--- a/metricflow/dataset/convert_data_source.py
+++ b/metricflow/dataset/convert_data_source.py
@@ -177,7 +177,6 @@ class DataSourceToDataSetConverter:
         self,
         data_source_name: str,
         measures: Sequence[Measure],
-        measure_time_dimension_spec: TimeDimensionSpec,
         table_alias: str,
     ) -> Tuple[Sequence[MeasureInstance], Sequence[SqlSelectColumn]]:
         # Convert all elements to instances
@@ -387,11 +386,9 @@ class DataSourceToDataSetConverter:
 
         # Handle measures
         if len(data_source.measures) > 0:
-            primary_time_dimension = self._find_primary_time_dimension(data_source)
             measure_instances, select_columns = self._convert_measures(
                 data_source_name=data_source.name,
                 measures=data_source.measures,
-                measure_time_dimension_spec=primary_time_dimension,
                 table_alias=from_source_alias,
             )
             all_measure_instances.extend(measure_instances)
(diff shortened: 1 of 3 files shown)


metricflow/dataset/convert_data_source.py
```diff
@@ -353,26 +352,6 @@ def _create_identifier_instances(
                 )
         return identifier_instances, select_columns
 
-    @staticmethod
-    def _find_primary_time_dimension(data_source: DataSource) -> TimeDimensionSpec:
```
WilliamDee, line 357: maybe it's outside of the changes in this PR, but we would still need to check that a data source has a primary time dimension if not all measures have agg_time_dimension set right?

WilliamDee (commented):
checking in on one thing

metricflow/dataset/convert_data_source.py
QMalcolm: Good question! I'm unsure. Specifically I'm unsure if we want to do that in assertions during `convert_data_source` stuff. The validations will catch this, so if they put changes to their model through validations, it won't be a problem.

What we currently do with the a "primary time dimension" is set it to the `agg_time_dimension` for measures that don't have an `agg_time_dimension` set in this [transformation](https://github.com/transform-data/metricflow/blob/33f18ac5ad47b1df08c694316849c82e3ea2cf50/metricflow/model/transformations/agg_time_dimension.py#L24). Then if the model goes through validations, it'll get caught by the above rule. If it doesn't go through validations it will fail somewhere. Specifically if something tries to get the `agg_time_dimension` directly and do something with it, a error involving something about `None` will be raised. However we do have [checked_agg_time_dimension](https://github.com/transform-data/metricflow/blob/33f18ac5ad47b1df08c694316849c82e3ea2cf50/metricflow/model/objects/elements/measure.py#L39) which will raise an assertion error (which is what the the `convert_data_source` used to do).

QMalcolm: Basically if the expectation is that an error will get raised at some point for measure that doesn't have an `agg_time_dimension`, I'm confident that is the case with these changes.

WilliamDee (approved):

PR #340, opened 2022-11-10
Add Tag Support for Snowflake Queries

plypaul:
This PR updates the Snowflake SQL client so that the request tags that are passed in get converted to a JSON string and used as the `QUERY_TAG` session parameter when executing a SQL query.


courtneyholcomb (approved):
Looks good! Thank you!

PR #344, opened 2022-11-14
Remove python-Levenshtein Package

plypaul:
`python-Levenshtein` is an optional package for `fuzzywuzzy` that improves matching performance. However, it has system dependencies, and since the matching used in MF is only used in error cases with a small data set, the dependencies will be easier to manage if we remove it.


marcodamore (approved):

tlento:
That whole pre-commit thing was very odd..... oh well. Thanks for the cleanup!

PR #348, opened 2022-11-18
Show underlying error from query or execution

courtneyholcomb:
The current way of raising the exception is burying the actual error, so we get errors like this:
image

Instead of this, which is much more helpful:
image

Changing to the latter here!


tlento (approved):
Thank you!

PR #350, opened 2022-11-18
Databricks: Move execute logic into helper function

courtneyholcomb:
We're working to add more logging for Databricks in Transform. Here I did a minor refactor to move execution logic into a helper function that we can easily override in Transform (to add a Datadog trace) without having to duplicate any additional logic.

### DIFF
--- a/metricflow/sql_clients/databricks.py
+++ b/metricflow/sql_clients/databricks.py
@@ -145,6 +145,7 @@ class DatabricksSqlClient(BaseSqlClientImplementation):
     def _execute_stmt(
         self, cursor: sql.client.Cursor, stmt: str, bind_params: SqlBindParameters = SqlBindParameters()
     ) -> None:
+        """Execute SQL statement. Abstracted into a function that can be easily overridden for logging purposes."""
         logger.info(f"Executing SQL statement: {stmt}")
         cursor.execute(operation=stmt, parameters=self.params_or_none(bind_params))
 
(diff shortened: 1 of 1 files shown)


metricflow/sql_clients/databricks.py
```diff
@@ -142,6 +142,12 @@ def params_or_none(bind_params: SqlBindParameters) -> Optional[Dict[str, SqlColu
         """If there are no parameters, use None to prevent collision with `%` wildcard."""
         return None if bind_params == SqlBindParameters() else bind_params.param_dict
 
+    def _execute_stmt(
```
tlento, line 145: The other option is to have a _log_before_execute or similar and then call that inside execute, but of course that only works if we just need to make a logging call before we run the query.

Might be worth adding a docstring to explain why this is set up this way in a general manner (so people can add custom logging hooks before and after the call to execute, or whatever description is appropriate).

tlento (approved):
Ok, this works! I left a suggestion inline for a more self-documenting alternative but I don't know if it'll meet your needs.

In the long run we might want to find a way to do this more consistently across all of these clients but I don't have any bright ideas at the moment. I'm vaguely thinking of something like a MonitoringSqlClient class that has simple override hooks for custom telemetry around the core client operations, but I have no idea how to make that work yet. If you've got any ideas I'm all ears - you've got a lot more experience working with these clients than I do.

metricflow/sql_clients/databricks.py
courtneyholcomb: We're looking to add a Datadog trace, for which the syntax is
```
with tracer.trace:
    <SQL execution here>
```
So I think we do unfortunately have to override the whole thing! But then I can just call `super()._execute_stmt()` so hopefully not too much redundancy here.
For all other SQL clients, we use SQLAlchemy which has a nice integration with Datadog automatically, so this trace is much more easily created. (Something to keep in mind in the future when we're deciding whether or not to use SQLAlchemy for other DB engines.)

I will add a docstring!

PR #351, opened 2022-11-21
Allow Tags for SQL Requests  in JSON Format

plypaul:
This PR updates the tag format used for SQL requests to be of the JSON format so that they are easier to consume.


courtneyholcomb (approved):
Looks good!! Thank you!

Issue #252, opened 2022-09-19
[SL-1290] Conversion Metrics


nhandel:
**Describe the Feature**
A new conversion metric type would be useful. A conversion is when a base event happens and then a conversion event happens for a specific entity within some time range. For example, we might want to measure how often a user(entity) completes a visit(base event) and then purchase(conversion event) with 7 days(time window). This requires a time range and an entity to join on. Today the closest alternative is a ratio metric but that does not include the entity in the pre-aggregated join.

This could be defined using the following spec:

```yaml
metric:
  name: visit_to_purchase_7d
  owners:
    - nick@company.com
  type: conversion
  type_params:
    entity: user
    base_data_source: visits # required
    base_params: # all optional
      time_dimension: created_at # default: primary time
    conversion_data_source: transactions # required
    conversion_params: # all optional
      time_dimension: created_at # default: primary time
      constraint: status = 'completed' # default: none
      role: buyer # default: none. This is needed for disambiguation when there are two roles for the same entity.
    window: 7d # default: inf. window to join the two events on. 
    multiple_conversions: false # default: false. Is this true/false or a count of how many conversions happen? default: false
```

would yield SQL that doesn't look like this but accomplishes the same thing:

```sql
SELECT
  DATE_TRUNC(visits.created_at, 'day') AS created_at
  COUNT(distinct transaction_id) / COUNT(visit_id) AS conversion_rate
FROM
  visits
LEFT JOIN
  transactions
ON
  visits.created_at - transactions.created_at BETWEEN 0 AND 7 days
  AND visits.user_id = transactions.user_id
WHERE
  transactions.status = 'completed'
```

A few notes about query parameters:
* [NEEDS DISCUSSION] For time, you would use the base_events time as the primary time for this metric because that is when the conversion started. You can still request either time dimension in the API. Asking for both dimensions would allow you to create cohort metrics.
* [NEEDS DISCUSSION] For mutability, because of the note above, a conversion metric cannot be treated as mutable until the end of it's window has passed. In the case of an infinite window, the metric is permanently mutable and should be expected to change.
* For time over time, this would behave like any other metric. Metricflow would calculate those two conversion rates for the specified granularity and then take a ratio of the two time period.
* For granularity, this would behave like any other metric. You would take the values across the entire window of time and then aggregated to that granularity. 
* For filters, this would behave like any other metric with multiple data sources (such as ratio metrics).
[...]

PR #352, opened 2022-11-21
Conversion Metrics

WilliamDee:
## Issue
Context in #252 
#### Added the ability to define Conversion Metrics
Config skeleton:
```
metric:
  name: _string_ # required
  description: _string_ # optional
  type: conversion # required
  type_params:
    conversion_type_params:
      entity: _entity_
      calculation: _calculation_type_ # default: conversion_rate. options: conversions(buys) or conversion_rate (buys/visits) + more to come
      base_measure: _measure_ # required
      conversion_measure: _measure_ # required
      window: _time_window_ # optional. default: inf. window to join the two events on. Follows similar format as time windows elsewhere (i.e. 7 days)
```
#### Added `AddGenerateUuidNode`
There would be times where a data source does not have a set of primary keys to uniquely identify rows. For conversion metrics, we need the ability to uniquely identify rows in order to deduplicate events properly. This node simply adds a randomly generated uuid column to enable that.

#### Added `JoinConversionEventsNode`
Created a node to handle retrieving all successful conversions. This node takes the conversion and base data set and joins them against an entity and a valid time range to get successful conversions. It then deduplicates opportunities via the window function `first_value` to take the closest opportunity event to the corresponding conversion event. Then it returns a data set with each row representing a successful conversion. Because an opportunity event can lead to multiple conversions, there would be duplication in the final dataset from this node.

Example of this node,
```
VISIT DATA SOURCE
| visit_ds            | user     | referrer   
|:--------------------|:---------|:---------------
| 2020-01-01 00:00:00 | u0003141 | homepage_1 
| 2020-01-01 00:00:00 | u0003452 | fb_ad_2        
| 2020-01-01 00:00:00 | u0004114 | fb_ad_1           
| 2020-01-02 00:00:00 | u0003452 | fb_ad_1         
| 2020-01-04 00:00:00 | u0004114 | fb_ad_5
| 2020-01-07 00:00:00 | u0004114 | fb_ad_2     


BUY DATA SOURCE
| buy_ds              | user     | referrer   
|:--------------------|:---------|:---------------     
| 2020-01-02 00:00:00 | u0004114 | fb_ad_1   
| 2020-01-04 00:00:00 | u0003452 | fb_ad_1             
| 2020-01-07 00:00:00 | u0004114 | fb_ad_2 
| 2020-01-07 00:00:00 | u0003141 | homepage_1
| 2020-01-25 00:00:00 | u0004114 | google_ad_1


Result from 'JoinConversionEventsNode' on a 7 day window
| visit_ds            | buy_ds              | user     | referrer   
|:--------------------|:--------------------|:---------|:---------------
| 2020-01-01 00:00:00 | 2020-01-07 00:00:00 | u0003141 | homepage_1       
| 2020-01-01 00:00:00 | 2020-01-02 00:00:00 | u0004114 | fb_ad_1        
| 2020-01-02 00:00:00 | 2020-01-04 00:00:00 | u0003452 | fb_ad_1      
| 2020-01-07 00:00:00 | 2020-01-07 00:00:00 | u0004114 | fb_ad_2 
```


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -203,9 +303,63 @@ def _build_metrics_output_node(
...
+                ), "Conversion measure needs to be defined in conversion metrics. Something likely went wrong in validation."
+
+                combined_where = where_constraint
+                if metric.constraint:
+                    metric_constraint = WhereConstraintConverter.convert_to_spec_where_constraint(
+                        self._data_source_semantics, metric.constraint
+                    )
+                    combined_where = combined_where.combine(metric_constraint) if combined_where else metric_constraint
+
+                if metric_spec.constraint:
```
tlento, line 343: Wait, both of these are a thing? Would they ever be different?

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -441,7 +603,7 @@ def _find_measure_recipe(
         measure_spec_properties: MeasureSpecProperties,
         linkable_specs: Sequence[LinkableInstanceSpec],
         time_range_constraint: Optional[TimeRangeConstraint] = None,
-    ) -> Optional[MeasureRecipe]:
+    ) -> MeasureRecipe:
```
tlento, line 606: Nice

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -164,6 +170,100 @@ def build_plan(
...
+            where_constraint=where_constraint,
+            time_range_constraint=time_range_constraint,
+            cumulative=False,
+            cumulative_window=None,
+            cumulative_grain_to_date=None,
+            measure_recipe=conversion_measure_recipe,
+        )
+
+        # Combine the aggregated opportunities and conversion values
+        return JoinAggregatedMeasuresByGroupByColumnsNode(
```
tlento, line 263: Oh interesting. And that just.... works? That's pretty nice.

metricflow/test/integration/test_cases/itest_conversion_metrics.yaml
```diff
@@ -0,0 +1,44 @@
+---
+integration_test:
+  name: conversion_rate_metric
+  description: Query a conversion metric that calculates the conversion rate
+  model: SIMPLE_MODEL
+  metrics: ["visit_buy_conversion"]
+  group_bys: ["metric_time"]
```
tlento, line 7: We also want a test where we group by referrer_id. If that's a follow-up PR it's fine, just making sure we don't forget it.

metricflow/sql/optimizer/rewriting_sub_query_reducer.py
```diff
@@ -522,9 +527,11 @@ def _rewrite_node_with_join(node: SqlSelectStatementNode) -> SqlSelectStatementN
             order_bys=tuple(clauses_to_rewrite.order_bys),
             where=clauses_to_rewrite.combine_wheres(additional_where_clauses),
             limit=node.limit,
+            distinct=node.distinct,
         )
 
     def visit_select_statement_node(self, node: SqlSelectStatementNode) -> SqlQueryPlanNode:  # noqa: D
+        # print(node.description)
```
tlento, line 534: rm

tlento (commented):
This seems reasonable, although it's hard to follow the abstract object construction we're doing here. The SQL makes sense.

I'm not sure if this covers all of the fanout scenarios that could happen, so it might be a good idea to add tests for things like joins to linkable dimension datasets, aggregations against multiple dimensions, etc.

metricflow/dataflow/builder/dataflow_plan_builder.py
WilliamDee: this is how our current logic works, I didn't actually pay much attention here. Looking at this now I'm not entirely sure why we have both of these. But our current querying logic doesn't actually put the constraint into each MetricSpec during query parsing. So I'm assuming this doesn't get used at all unless somewhere is calling this and explicitly passing in MetricSpecs with constraints defined (which I'm fairly certain no one is doing that). Here's what I'll do, I'm going to remove the `if metric.constraint` block and have that be done in the query parser instead. That way we don't have this messy duplicate here.

tlento:
@WilliamDee is this ready for re-review?

tlento (changes requested):
Sadly this is going to need an overhaul once we come back to it. Sorry we couldn't shepherd this stack through before the integration.

I'm requesting changes here just because I'm getting confused about which things need review and which things do not.

WilliamDee:
reopened here #916

PR #355, opened 2022-11-22
Added `SqlWindowFunctionExpression` and `MetadataSpec/Instance`

WilliamDee:
## Context
In aid of Conversion Metrics #352, It was best to break it out into multiple PRs. This PR is used to add the sql rendering node for window functions ~~and the dataflow plan node for appending a row number column. The reason this was broken out into a separate PR was due to the nature of adding a row number column into a data set.~~ Meaning there was no where to define this column in the current framework as this was never defined from anywhere (ie., in the config). In this PR, we introduce a new spec/instance type called `MetadataSpec` and `MetadataInstance` respectively. This is used for places that requires an additional spec or instance. 

## Changes
- `SqlFunctionExpression` is now a parent object that is inherited by all sql function objects
- `SqlFunctionExpression` -> `SqlAggregationFunctionExpressionNode`
- Added `SqlWindowFunctionExpressionNode`
- Added `MetadataSpec` and `MetadataInstance` to SpecSets
- ~~Added `AppendRowNumberColumnNode`~~

## Note
After further discussion, `AppendRowNumberColumnNode` is unnecessary, but everything else could still be useful so keeping those in.


metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1264,3 +1271,58 @@ def visit_semi_additive_join_node(self, node: SemiAdditiveJoinNode) -> SqlDataSe
                 order_bys=(),
             ),
         )
+
+    def visit_append_row_number_column_node(self, node: AppendRowNumberColumnNode) -> SqlDataSet:
+        """Appends a row number identifier column to the data set."""
+        input_data_set: SqlDataSet = node.parent_node.accept(self)
+        input_data_set_alias = self._next_unique_table_alias()
+
+        row_number_spec = ExtraSpec.from_name("mf_row_number")
```
WilliamDee, line 1280: how should I name this?

metricflow/sql/sql_exprs.py
```diff
@@ -1094,7 +1100,7 @@ class SqlRatioComputationExpression(SqlExpressionNode):
     """Node for expressing Ratio metrics to allow for appropriate casting to float/double in each engine
 
     In future we might wish to break this up into a set of nodes, e.g., SqlCastExpression and SqlMathExpression
-    or even add CAST to SqlFunctionExpression. However, at this time the only mathematical operation we encode
+    or even add CAST to SqlAggregateFunctionExpression. However, at this time the only mathematical operation we encode
```
tlento, line 1250: Revert this one, CAST is not an aggregation function

tlento (commented):
I didn't have time to read this thoroughly. Directionally it seems fine to me. A couple of notes:

1. ExtraSpec does need a new name, especially in the InstanceSpec interface. Maybe `internal_specs` or `added_metadata_spec` (as I believe @plypaul suggested) - something a bit more obvious that it's a framework internal thing.
2. You mentioned a new identifier type (`internal_key` or `row_number` or something). It's hard for me to see how that will shake out - on the one hand it eliminates the need for a special container for extra specs. On the other hand, it mixes internal elements with model elements in the instance spec interface, and requires runtime checking to see if one of these things exists anywhere in there. My instinct is to stick with what you have here, but an internal identifier type is an idea worth exploring.

plypaul (commented):
Remind me how the row number is used again?

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1264,3 +1271,58 @@ def visit_semi_additive_join_node(self, node: SemiAdditiveJoinNode) -> SqlDataSe
...
+        row_number_spec_column_name = self._column_association_resolver.resolve_extra_spec(row_number_spec)
+
+        input_sql_column_references = input_data_set.instance_set.transform(
+            CreateSqlColumnReferencesForInstances(input_data_set_alias, self._column_association_resolver)
+        )
+        # Build enumerated column using ROW_NUMBER function
+        row_number_select_column = SqlSelectColumn(
+            expr=SqlWindowFunctionExpression(
+                sql_function=SqlWindowFunction.ROW_NUMBER,
+                order_by_args=[SqlWindowOrderByArg(expr=x) for x in input_sql_column_references],
```
plypaul, line 1290: There will have to be a `FilterElementsNode` before this node so we don't have a long list of columns here.

metricflow/sql/sql_exprs.py
```diff
@@ -776,6 +781,148 @@ def matches(self, other: SqlExpressionNode) -> bool:  # noqa: D
...
+    @property
+    def lineage(self) -> SqlExpressionTreeLineage:  # noqa: D
+        return SqlExpressionTreeLineage.combine(
+            tuple(x.lineage for x in self.parent_nodes) + (SqlExpressionTreeLineage(function_exprs=(self,)),)
+        )
+
+    def matches(self, other: SqlExpressionNode) -> bool:  # noqa: D
+        if not isinstance(other, SqlWindowFunctionExpression):
+            return False
+        order_by_matches = all(x == y for x, y in itertools.zip_longest(self.order_by_args, other.order_by_args))
```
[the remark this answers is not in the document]
WilliamDee: updated the check to do `len(x) == len(y) and set(x) == set(y)`.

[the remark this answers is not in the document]
WilliamDee: nvm, we should actually just do an equality check since ordering matters for this. So I just made it a simple `self.order_by_args == other.order_by_args`

metricflow/specs.py
```diff
@@ -107,6 +111,24 @@ def qualified_name(self) -> str:
         raise NotImplementedError()
 
 
+@dataclass(frozen=True)
+class MetadataSpec(InstanceSpec):
```
plypaul, line 115: This looks fine for now, but there will likely be different types of metadata later on so there might need to be subclasses of this. I'm also thinking whether to make this `Visitable` (also for later).

plypaul (approved):
LGTM

PR #359, opened 2022-11-24
Update cancel_request to filter by function

plypaul:
This PR updates `AsyncSqlClient.cancel_request` so that a filtering / matching function can be used to select which queries to cancel.


jpreillymb (approved):
👍

PR #361, opened 2022-11-28
Add Dataflow Plan Optimizer to Reduce Source Scans

plypaul:
This PR adds a dataflow plan optimizer that transforms a dataflow plan to reduce the number of scans of a data source needed to resolve multi-metric queries when the metrics are computed from a common data source. For example, in a test model, the metrics `bookings, booking_value by metric_time` can be both computed from the `booking_source` data source. It's is currently rendered as:

```
-- Combine Metrics
SELECT
  COALESCE(subq_18.metric_time, subq_19.metric_time) AS metric_time
  , subq_18.bookings AS bookings
  , subq_19.booking_value AS booking_value
FROM (
  -- Aggregate Measures
  SELECT
    metric_time
    , SUM(bookings) AS bookings
  FROM (
    -- Read Elements From Data Source 'bookings_source'
    -- Metric Time Dimension 'ds'
    -- Pass Only Elements:
    --   ['bookings', 'metric_time']
    SELECT
      ds AS metric_time
      , 1 AS bookings
    FROM (
      -- User Defined SQL Query
      SELECT * FROM ***************************.fct_bookings
    ) bookings_source_src_10001
  ) subq_12
  GROUP BY
    metric_time
) subq_18
FULL OUTER JOIN (
  -- Read Elements From Data Source 'bookings_source'
  -- Metric Time Dimension 'ds'
  -- Pass Only Elements:
  --   ['booking_value', 'metric_time']
  -- Aggregate Measures
  SELECT
    ds AS metric_time
    , SUM(booking_value) AS booking_value
  FROM (
    -- User Defined SQL Query
    SELECT * FROM ***************************.fct_bookings
  ) bookings_source_src_10001
  GROUP BY
    ds
) subq_19
ON
  subq_18.metric_time = subq_19.metric_time
```

With the optimizer transforming the current dataflow plan:

```
-- Aggregate Measures
SELECT
  metric_time
  , SUM(bookings) AS bookings
  , SUM(booking_value) AS booking_value
FROM (
  -- Read Elements From Data Source 'bookings_source'
  -- Metric Time Dimension 'ds'
  -- Pass Only Elements:
  --   ['bookings', 'booking_value', 'metric_time']
  SELECT
    ds AS metric_time
    , 1 AS bookings
    , booking_value
  FROM (
    -- User Defined SQL Query
    SELECT * FROM ***************************.fct_bookings
  ) bookings_source_src_10001
) subq_6
GROUP BY
  metric_time
```


metricflow/dataflow/optimizer/source_scan/cm_branch_combiner.py
```diff
@@ -0,0 +1,437 @@
...
+    def visit_compute_metrics_node(  # noqa: D
+        self, node: ComputeMetricsNode[SourceDataSetT]
+    ) -> ComputeMetricsBranchCombinerResult:
+        current_right_node = node
+        self._log_visit_node_type(current_right_node)
+        combined_parent_nodes = self._combine_parent_branches(current_right_node)
+        if combined_parent_nodes is None:
+            return ComputeMetricsBranchCombinerResult()
+
+        if not isinstance(self._current_left_node, current_right_node.__class__):
```
WilliamDee, line 283: we could probably have a helper for this, it seems like we would do this alot here. There's already 3 instances of the exact same code here.

metricflow/dataflow/optimizer/source_scan/source_scan_optimizer.py
```diff
@@ -0,0 +1,368 @@
...
+    def visit_pass_elements_filter_node(  # noqa: D
+        self, node: FilterElementsNode[SourceDataSetT]
+    ) -> OptimizeBranchResult[SourceDataSetT]:
+        self._log_visit_node_type(node)
+        return self._default_base_output_handler(node)
+
+    @staticmethod
+    def _combine_branches(
+        left_branches: Sequence[BaseOutput[SourceDataSetT]], right_branch: BaseOutput[SourceDataSetT]
+    ) -> Sequence[BranchCombinationResult]:
```
WilliamDee, line 240: from our chat earlier, comment here to explain the assumption of how this is optimal would be nice!

WilliamDee (commented):
This is amazing! 👏👏

metricflow/dataflow/optimizer/source_scan/cm_branch_combiner.py
```diff
@@ -0,0 +1,437 @@
...
+            # Avoid combining branches if the AggregateMeasuresNode specifies a metric with an alias to avoid
+            # collisions e.g. two metrics use the same alias for two different measures. This is not always the case,
+            # so this could be improved later.
+            if spec.alias is not None:
+                self._log_combine_failure(
+                    left_node=self._current_left_node,
+                    right_node=current_right_node,
+                    combine_failure_reason=f"Metric input measure spec {spec} has an alias",
+                )
+                return ComputeMetricsBranchCombinerResult()
```
tlento, line 261: This is all very difficult to reason about. In theory this shouldn't be an issue if we take the distinct set of combined input measure specs, but I guess if we combine parents in a way that assembles two different measures (or metrics, when we do this for derived metrics) that happen to have the same alias that'll cause problems.

We could make this less aggressive by updating the conditional to be something like:

```
aliases = [spec.alias for spec in combined_metric_input_measure_specs]
distinct_aliases = set(aliases)
if len(aliases) > len(distinct_aliases):
    return ...
```

metricflow/test/dataflow/optimizer/source_scan/test_source_scan_optimizer.py
```diff
@@ -0,0 +1,337 @@
...
+    def visit_semi_additive_join_node(self, node: SemiAdditiveJoinNode[SourceDataSetT]) -> int:  # noqa: D
+        return self._sum_parents(node)
+
+    def visit_metric_time_dimension_transform_node(  # noqa: D
+        self, node: MetricTimeDimensionTransformNode[SourceDataSetT]
+    ) -> int:
+        return self._sum_parents(node)
+
+    def count_source_nodes(self, dataflow_plan: DataflowPlan[SourceDataSetT]) -> int:  # noqa: D
+        return dataflow_plan.sink_output_node.accept(self)
```
tlento, line 104: It really feels like this entire class should just be `len(dataflow_plan.node_list)` or similar, but I guess that interface is really just "you have a sink node" at the moment.

tlento (commented):
This moves us into the following dataflow plan building workflow:

1. Read model and query params
2. Construct dataflow plan 1 metric at a time
3. Run recursive visitor optimizer over naively constructed dataflow plan to merge branches and eliminate duplicate reads generated by our approach in step 2

If step 3 was really easy to reason about and clearly and deterministically doing the correct thing independent of other input considerations, I'd say it's totally fine. Probably not perfectly optimal, since we do lose some context in the translation from model + query params to dataflow plan, but a fine trade for the relative simplicity.

As it stands now, I'm struggling with this PR - I'm finding these optimizers to be very difficult to reason about, much more difficult than I expect it to be to build the plan in a more query-aware way, and I'm not convinced that the optimizations themselves are robust. I am reasonably sure they produce correct results today, but ultimately I am concerned that their correctness is implicitly dependent on model level restrictions which may or may not stand the test of time. In particular, I think there will be cases in the future where we allow for the potential of resolvable ambiguous join paths, and in those cases the dataflow plan nodes themselves may not be distinguishable in the way these optimizers check such things. I'm also concerned that almost immediately on the heels of this will come demand for partial aggregation optimization, which will be even more complicated to manage via an external optimizer.

We're on a timeline, and rewriting the dataflow plan builder will take a considerable amount of time and effort, and as such I think this is probably the best approach for now, but I'd like us to reconsider this the moment the immediate time-sensitive demand is past.

tlento:
Your test failure looks like a change on the main branch that affects the new plans you generated, so that should be an easy cleanup.

plypaul (commented):
Overall, in order to produce more optimal queries, we have to do a grouping based on the type of measure and what's available in the sources, whether in the dataflow plan builder or in a separate optimizer.  There's going to be complexity somewhere, and the question is whether to put that in the dataflow plan builder or in a separate optimizer. It could be the case that putting the grouping logic in the dataflow plan builder is simpler or easier, but based on past experience, it was more challenging.

The argument for doing as an optimizer step now is that:

* It's easier to add new features in the dataflow plan because we can retain the "build one metric at a time" approach.
* The optimizer can be made so that it only handles a specific set of cases and we can have higher confidence in producing correct results at the expense of optimal results.
* We can always turn off / remove the optimizer and move the logic in the dataflow plan builder once we figure out the details of that. This makes it more of a type 2 decision.
* In case there's an exception in the optimizer, we could fall back to the naive plan.

Arguments against the optimizer approach are
* It can be more brittle and relies on invariants on structure / behavior of the dataflow plan that are handled in a separate class.
* There are no code mechanisms that enforce these invariants.
* Transformations of graphs and the recursive functions required to do them can be complex as well.

Context on the way I was reasoning about the correctness of this optimizer:
* With the behavior of the `CombinedMetricsNode`, as long as the branch combinations are correct, the optimized plan will have the same behavior as the original plan.
* Two branches can be combined if they have the same structure (i.e. connections / node types) and for each node type, we can reason that if the parents of the nodes are a combination, we can produce a node that properly handles combined parents.

PR #364, opened 2022-12-02
Add SqlGenerateUuidExpression

WilliamDee:
## Changes
- Added `SqlGenerateUuidExpression` node
- Added renderer for different dialects
- (Yet to do) create a python UDF for redshift users under the function name `MF_FN_UUID`


metricflow/sql/render/redshift.py
```diff
@@ -10,6 +16,13 @@ def double_data_type(self) -> str:
         """Custom double data type for the Redshift engine"""
         return "DOUBLE PRECISION"
 
+    def visit_generate_uuid_expr(self, node: SqlGenerateUuidExpression) -> SqlExpressionRenderResult:  # noqa: D
+        """Utilizes a python UDF that needs to get added in."""
+        return SqlExpressionRenderResult(
+            sql="MF_FN_UUID()",
```
plypaul, line 22: How does this get registered?

metricflow/sql/render/redshift.py
```diff
@@ -10,6 +16,19 @@ def double_data_type(self) -> str:
...
+    def visit_generate_uuid_expr(self, node: SqlGenerateUuidExpression) -> SqlExpressionRenderResult:  # noqa: D
+        """Generates a "good enough" random key to simulate a UUID.
+
+        NOTE: This is a temporary hacky solution as redshift does not have any UUID generation function.
+
+        Proposed solutions that requires more thinking:
+            - create a python UDF (Could we insert this without needing additional permissions?)
+        """
+        return SqlExpressionRenderResult(
+            sql="CONCAT(CAST(RANDOM()*100000 AS INT),CAST(RANDOM()*100000 AS INT),CAST(RANDOM()*100000 AS INT))",
```
plypaul, line 28: Seems like you could get ~16 significant digits (52 bits) out of RAND(). Thinking that since the UUIDs are 128-bit, we go with a similar number of bits?

plypaul (approved):
LGTM but see comment

metricflow/sql/render/redshift.py
tlento: How does RANDOM work? I'd be worried about some weirdness in how they seed it by default.

Oftentimes what people do is concat the current timestamp in milliseconds to have a value as part of the sequence that is guaranteed to repeat only for a very small number of records.

I don't know if that's better or worse than three RANDOM() calls, so I don't think you need to change anything.

metricflow/sql/render/snowflake.py
```diff
@@ -0,0 +1,28 @@
...
+from metricflow.sql.sql_bind_parameters import SqlBindParameters
+from metricflow.sql.sql_exprs import SqlGenerateUuidExpression
+
+
+class SnowflakeSqlExpressionRenderer(DefaultSqlExpressionRenderer):
+    """Expression renderer for the Snowflake engine."""
+
+    def visit_generate_uuid_expr(self, node: SqlGenerateUuidExpression) -> SqlExpressionRenderResult:  # noqa: D
+        return SqlExpressionRenderResult(
+            sql="UUID_STRING()",
```
tlento, line 16: Let us all pause for a moment to be irritated about how inconsistent this function naming is across engines...

tlento (approved):
Nice!

metricflow/model/validations/unique_valid_name.py
```diff
@@ -44,6 +45,8 @@ def get_reserved_reason(keyword: MetricFlowReservedKeywords) -> str:
                 "Used as the query input for creating time series metrics from measures with "
                 "different time dimension names."
             )
+        if keyword is MetricFlowReservedKeywords.MF_INTERNAL_UUID:
```
tlento, line 48: I'm pretty sure this has to be an `elif`

metricflow/model/validations/unique_valid_name.py
```diff
@@ -35,6 +35,7 @@ class MetricFlowReservedKeywords(enum.Enum):
     """Enumeration of reserved keywords with helper for accessing the reason they are reserved"""
 
     METRIC_TIME = "metric_time"
+    MF_INTERNAL_UUID = "mf_internal_uuid"
```
tlento, line 38: Oh I see, this is because we're adding this under the covers, right?

Shall we make the prefix `metricflow` instead of `mf`? I'm ok either way, `mf_internal_` is already pretty distinctive.

metricflow/model/validations/unique_valid_name.py
WilliamDee: doesn't need to, but it'll be structurally more correct I think. So changing it

PR #365, opened 2022-12-05
Get joinable data sources

courtneyholcomb:
Adds a function to get all joinable data sources for a given data source, as well as the way to join them.


metricflow/model/semantics/join_validator.py
```diff
@@ -58,34 +75,140 @@ class DataSourceJoinValidator:
...
+                            right_data_source_reference=secondary_joinable_data_source_reference,
+                            identifier_joins=[
+                                DataSourceIdentifierJoin(
+                                    identifier_reference=identifier_reference, join_type=join_type
+                                ),
+                                DataSourceIdentifierJoin(
+                                    identifier_reference=secondary_identifier_reference, join_type=secondary_join_type
+                                ),
+                            ],
+                        )
```
tlento, line 138: This nested logic is quite confusing for me. This might be a case where operating on batches will be easier to understand, at least at the top level of this function. Something like:

```
# we can wrap this in a function and call it per data source, passing in an optional memoize list to cut out repeated work 
data_sources = {
    identifier.reference: self._data_source_semantics.get_data_sources_for_identifier(identifier.reference)
    for identifier in data_source.identifiers
}
valid_targets = {
    idref: filter(candidate_sources, lambda x: self.is_valid_data_source_join(...)) 
    for idref: candidate_sources in data_sources
}
# This produces a data_source_reference -> List[DataSourceIdentifierJoin]
return convert_to_data_source_to_join_path(valid_targets)
```

We'd have to do some extra stuff for making sure the multi-hop calls include all leading identifier references.

Let me know what you think! Note I like these kind of pipeline processing flows, so I might be missing an easier readability improvement.

metricflow/model/semantics/join_validator.py
```diff
@@ -58,34 +75,140 @@ class DataSourceJoinValidator:
...
+                            == left_data_source_reference.data_source_name
+                        ):
+                            continue
+                        data_source_joins[secondary_joinable_data_source_reference.data_source_name] = DataSourceJoin(
+                            left_data_source_reference=left_data_source_reference,
+                            right_data_source_reference=secondary_joinable_data_source_reference,
+                            identifier_joins=[
+                                DataSourceIdentifierJoin(
+                                    identifier_reference=identifier_reference, join_type=join_type
+                                ),
```
tlento, line 133: Do we need to keep the intermediate data source reference for this?

metricflow/model/semantics/join_validator.py
```diff
@@ -18,6 +18,23 @@ class DataSourceIdentifierJoinType:
...
+    identifier_reference: IdentifierReference
+    join_type: DataSourceIdentifierJoinType
+
+
+@dataclass(frozen=True)
+class DataSourceJoin:
+    """Links between two data sources and the valid way to join them."""
+
+    left_data_source_reference: DataSourceReference
+    right_data_source_reference: DataSourceReference
```
tlento, line 34: I noticed this below, but do we need to have a different structure for multi-hop joins here, or are we just trying to give people the `identifier1__identifier2__target_data_source` marker?

In the future we'll definitely need something that's more explicit about the dependency sequencing, because we'll want to use the logic here to find data source join paths based on a source and target where the target is referred by dimension entity. Conceptually it'd look like `{entity: user, dimension: country}` and we'd be grabbing the path to that separately.

metricflow/model/semantics/join_validator.py
```diff
@@ -58,34 +75,140 @@ class DataSourceJoinValidator:
...
+                continue
+
+            right_data_source_reference = DataSourceReference(data_source_name=identifier_data_source.name)
+            valid_join_type = self.get_valid_data_source_identifier_join_type(
+                left_data_source_reference=left_data_source_reference,
+                right_data_source_reference=right_data_source_reference,
+                on_identifier_reference=identifier_reference,
+            )
+            if valid_join_type is not None:
+                new_joinable_data_sources.append((right_data_source_reference, valid_join_type))
```
tlento, line 166: If we don't need to pass this join type info from down here out to the return value the whole function becomes a pair of filters - memoization and validity.

tlento (commented):
Sorry for the delay! I've been struggling with the vagaries of SQL engines and lost track of my review queue.

The main question for me here is about what we need to return - depending on how we're using this there might be some easier implementations we can consider.

metricflow/model/semantics/join_validator.py
```diff
@@ -18,6 +18,23 @@ class DataSourceIdentifierJoinType:
...
+    join_type: DataSourceIdentifierJoinType
+
+
+@dataclass(frozen=True)
+class DataSourceJoin:
+    """Links between two data sources and the valid way to join them."""
+
+    left_data_source_reference: DataSourceReference
+    right_data_source_reference: DataSourceReference
+    identifier_joins: List[DataSourceIdentifierJoin]
```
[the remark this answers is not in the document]
courtneyholcomb: For the model page, we want to be able to show which identifiers need to join together to connect two data sources, and what type of join needs to be executed to join them. So that's why I put together this data structure.
But also, how are you imagining the logic would be simplified if that were removed?

metricflow/model/semantics/join_validator.py
courtneyholcomb: Yeah I think I will need to refactor this to make it recursive anyway since we want to support more than 2 hops! So l'll see if this structure will work for that while I refactor.

metricflow/model/semantics/join_validator.py
courtneyholcomb: Hmmm good question.

metricflow/model/semantics/join_validator.py
courtneyholcomb: We do need the join type for the model page!

metricflow/model/semantics/join_validator.py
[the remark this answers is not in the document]
tlento: I don't recall, but I suspect when you switch to an arbitrary-depth recursive approach whatever I was thinking will likely be obvious (or obviously wrong) so I think we should just move on. :)

metricflow/model/semantics/join_validator.py
```diff
@@ -1,14 +1,16 @@
...
-from typing import List
+from typing import List, Dict, Optional
 
 from metricflow.instances import DataSourceReference, DataSourceElementReference, IdentifierInstance, InstanceSet
 from metricflow.model.objects.elements.identifier import IdentifierType
 from metricflow.object_utils import pformat_big_objects
 from metricflow.protocols.semantics import DataSourceSemanticsAccessor
 from metricflow.references import IdentifierReference
 
+MAX_JOIN_HOPS = 2
```
courtneyholcomb, line 12: Moved this constant to resolve circular import

metricflow/model/semantics/join_validator.py
```diff
@@ -58,34 +75,140 @@ class DataSourceJoinValidator:
     def __init__(self, data_source_semantics: DataSourceSemanticsAccessor) -> None:  # noqa: D
         self._data_source_semantics = data_source_semantics
 
-    def is_valid_data_source_join(
+    def get_joinable_data_sources(
```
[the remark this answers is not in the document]
courtneyholcomb: If you have any ideas for a better place to put it, LMK!

metricflow/test/model/semantics/test_join_validator.py
```diff
@@ -269,6 +274,83 @@ def test_foreign_target_instance_set_join_validation(
...
+                right_data_source_reference=DataSourceReference(data_source_name="bridge_table"),
+                identifier_reference=IdentifierReference(element_name="account_id"),
+                join_type=DataSourceIdentifierJoinType(
+                    left_identifier_type=IdentifierType.PRIMARY, right_identifier_type=IdentifierType.PRIMARY
+                ),
+            )
+        ],
+    )
+
+    # 2-hop
```
tlento, line 297: nit: can we split the 1 hop and 2 hop tests into two test cases? That way we can see which ones are failing in the test runner.

tlento (approved):
Yay! I think this will be very useful for when we need to resolve joins by dimension entity instead of multi-hop path.

I left a few nits in line but they can all be done without further review. Thanks!

PR #367, opened 2022-12-07
Add time granularity checks to where constraint parsing

kyleli626:
Previously, where constrains with too narrow time granularities would only error in the data plan with a recipe not found error. This pr adds a check for time dimension specs using the Time Granularity Solver to make sure this error would get caught in parsing. Where constraints like "metric_time__day >= '2022-01-01'" on a metric with month-level granularity now would throw a RequestTimeGranularityException. 

This pr also refactors the query_parser tests to use a model created from in-line yaml in the test itself, rather than a simple model created from the model fixtures. This is so the model used is simpler and more easily viewed.

Changes
- add a yaml -> MetricFlowQueryParser function in model fixtures
- transfer and simplify test models yaml into tests themselves from the simple model
- add a call to validate_time_granularity in the _parse_and_validate_query method for where constraint time dimensions
- add a test for the query parser to verify bad constraints would throw a RequestTimeGranularityException


metricflow/test/fixtures/model_fixtures.py
```diff
@@ -42,6 +49,23 @@ def _data_set_to_source_nodes(
     return source_node_builder.create_from_data_sets(list(data_sets.values()))
 
 
+def query_parser_from_yaml(
+    yaml_contents: List[YamlConfigFile], time_spine_source: TimeSpineSource
+) -> MetricFlowQueryParser:
+    """Given yaml files, return a query parser using default source nodes, resolvers and time_sp"""
```
tlento, line 55: nit: `time_sp` -> `time spine source` (or whatever)

metricflow/test/fixtures/model_fixtures.py
```diff
@@ -42,6 +49,23 @@ def _data_set_to_source_nodes(
...
+    semantic_model = SemanticModel(parse_yaml_files_to_validation_ready_model(yaml_contents).model)
```
tlento, line 56: You might want to run the semantic validators on this, that way people get better error messages when they write a test model that's invalid.

You can either refactor dir_to_model a tiny bit more or add a call to `ModelValidator().checked_validations()` here.

metricflow/test/query/test_query_parser.py
```diff
@@ -16,11 +18,81 @@
...
 
-def test_query_parser(query_parser: MetricFlowQueryParser) -> None:  # noqa: D
+BOOKINGS_YAML = textwrap.dedent(
+    """\
+    data_source:
+      name: bookings_source
+
+      sql_query: |
+        -- User Defined SQL Query
+        SELECT * FROM $bookings_source_table
```
tlento, line 34: If you like you can put in a string literal name for the table, there's no underlying data involved here. Doesn't matter, either way it'll likely fail if anything tries to run a query.

metricflow/test/query/test_query_parser.py
```diff
@@ -240,6 +241,19 @@ def test_parse_and_validate_where_constraint_dims(time_spine_source: TimeSpineSo
     assert DimensionSpec(element_name="is_instant", identifier_links=()) not in query_spec.dimension_specs
 
 
+def test_parse_and_validate_where_constraint_metric_time(time_spine_source: TimeSpineSource) -> None:
+    """Test that granularity of metric_time reference in where constraint is at least that of the ds dimension."""
+    revenue_yaml_file = YamlConfigFile(filepath="inline_for_test_2", contents=REVENUE_YAML)
+
+    query_parser = query_parser_from_yaml([revenue_yaml_file], time_spine_source)
+    with pytest.raises(RequestTimeGranularityException):
```
tlento, line 249: You might want to add the `match` param here to assert that the correct error message is being thrown, but given that this is a granularity-specific exception I don' think it's necessary.

tlento (approved):
Looks great, thanks! I left some small inline comments which you can address before merge as you see fit. No need to send this back to me unless you want me to take a look at something you change.

You will need to update this onto the most recent commit to main to get the CI tests to run.

PR #368, opened 2022-12-07
Pin CI actions to ubuntu 20.04

tlento:
The recent update to ubunutu 22.04 seems to have broken a bunch of
stuff with python installs. In all likelihood the fix is going to
involve hacky environment variable overrides or continuous
reinstallation of python-distutils.

Rather than deal with testing all of those things at this time, we
simply pin the ubuntu version back to 20.04 in CI and hope for the
best.


marcodamore (approved):

WilliamDee (approved):
thanks!

PR #371, opened 2022-12-10
Add percentile renderers

kyleli626:
Context: Adding percentile aggregations for measures.

Changes:
- Create new SqlPercentileExpression class that subclasses SqlFunctionExpression.
- Add engine-specific and default renderers visitor functions for new percentile expression.
- Add snapshot tests for new rendering.


metricflow/sql/render/expr_renderer.py
```diff
@@ -124,6 +126,21 @@ def visit_function_expr(self, node: SqlAggregateFunctionExpression) -> SqlExpres
             execution_parameters=combined_params,
         )
 
+    def visit_percentile_expr(self, node: SqlPercentileExpression) -> SqlExpressionRenderResult:
```
conglei, line 129: Should we make this as the default implementation so it won't be duplicated among several sql engines?

metricflow/sql/sql_exprs.py
```diff
@@ -786,6 +791,86 @@ def matches(self, other: SqlExpressionNode) -> bool:  # noqa: D
...
+    """Type of percentile function used."""
+
+    DISCRETE = "discrete"
+    CONTINUOUS = "continuous"
+
+
+class SqlPercentileExpression(SqlFunctionExpression):
+    """An percentile aggregation expression."""
+
+    def __init__(self, arg: SqlExpressionNode, percentile: float, function_type: SqlPercentileFunctionType) -> None:
```
conglei, line 804: Instead of passing two variables, we can creat a data class PercentileArgs and it has two fields.

metricflow/sql/render/expr_renderer.py
kyleli626: yes - this is currently in the DefaultSqlExpressionRenderer, so it would be the default for snowflake, duckdb, postgres, and databricks.

conglei: Oops, I was reading the code wrong 😂

metricflow/sql/sql_exprs.py
```diff
@@ -786,6 +791,89 @@ def matches(self, other: SqlExpressionNode) -> bool:  # noqa: D
...
+
+    def __init__(self, arg: SqlExpressionNode, percentile_args: SqlPercentileExpressionArgument) -> None:
+        """Constructor.
+
+        Args:
+            arg: The expression that should go into the function. e.g. for "percentile_cont(col, 0.1)", the arg
+            expressions should be "col".
+            percentile_args: Auxillary information including percentile value and type.
+        """
+        self.arg = arg
```
WilliamDee, line 825: lets keep it consistent with the other and keep the parameters as a protected attribute in the class and expose it as a property for retrieval. I believe it also helps for type checking.

metricflow/sql/sql_exprs.py
```diff
@@ -786,6 +791,89 @@ def matches(self, other: SqlExpressionNode) -> bool:  # noqa: D
...
+
+
+class SqlPercentileExpression(SqlFunctionExpression):
+    """An percentile aggregation expression."""
+
+    def __init__(self, arg: SqlExpressionNode, percentile_args: SqlPercentileExpressionArgument) -> None:
+        """Constructor.
+
+        Args:
+            arg: The expression that should go into the function. e.g. for "percentile_cont(col, 0.1)", the arg
```
WilliamDee, line 821: `arg` seems kind of ambiguous. It goes into the `ORDER BY` clause so maybe rename to `order_by_arg`

WilliamDee (commented):
this looks good!

metricflow/sql/sql_exprs.py
```diff
@@ -786,6 +791,98 @@ def matches(self, other: SqlExpressionNode) -> bool:  # noqa: D
...
+
+@dataclass(frozen=True)
+class SqlPercentileExpressionArgument:
+    """Dataclass for holding percentile arguments"""
+
+    percentile: float
+    function_type: SqlPercentileFunctionType
+
+    @property
+    def function_name(self) -> str:
```
conglei, line 809: 👍

conglei (approved):
It looks good to me! But want to get a second opinion from metricflow team :)

metricflow/sql/sql_exprs.py
```diff
@@ -786,6 +791,98 @@ def matches(self, other: SqlExpressionNode) -> bool:  # noqa: D
...
+        """Helper to build function name to use in rendering"""
+        return "PERCENTILE_CONT" if self.function_type == SqlPercentileFunctionType.CONTINUOUS else "PERCENTILE_DISC"
+
+
+class SqlPercentileExpression(SqlFunctionExpression):
+    """An percentile aggregation expression."""
```
tlento, line 815: micro-nit: 
```suggestion
    """A percentile aggregation expression."""
```

tlento (approved):
Nice! I think you'll need to base this on latest main, there have been some updates that might affect this under the hood even if there aren't direct merge conflicts.

metricflow/sql/sql_exprs.py
kyleli626: For sure! Will merge after the query_parser pr goes through. Think I neglected to rebase before merging today, my bad!

PR #374, opened 2022-12-12
add back query_parser

WilliamDee:
## Context
tests failed due to 2 PRs being merge concurrently. Both PR modified the test file and both CI succeeded individually but once merged together it fails.


metricflow/test/fixtures/model_fixtures.py
```diff
@@ -68,6 +68,23 @@ def query_parser_from_yaml(
...
+    return MetricFlowQueryParser(
+        model=simple_semantic_model,
+        source_nodes=consistent_id_object_repository.simple_model_source_nodes,
+        node_output_resolver=DataflowPlanNodeOutputDataSetResolver(
+            column_association_resolver=DefaultColumnAssociationResolver(simple_semantic_model),
+            semantic_model=simple_semantic_model,
+            time_spine_source=time_spine_source,
+        ),
+    )
+
```
tlento, line 86: The reason this was removed was because it's doing a bunch of stuff against stored models and that makes these tests very rigid. The whole file was actually cut over to focus pretty much exclusively on parsing and processing a simple local model directly illustrating the behavior we need to test.

Can we update the new test case added in #370 to bring it in line with the rest of the tests in the test_query_parser module?

tlento (approved):
Approving to unblock since we need these tests passing, but let's fix up the new query parser test and delete this fixture again, since it's not really necessary.

metricflow/test/fixtures/model_fixtures.py
WilliamDee: gotcha, i updated it to not use the stored models

PR #376, opened 2022-12-13
Derived Metric Offset

courtneyholcomb:
Preliminary code for Derived Metric Offset. Check out [this doc](https://www.notion.so/transformdata/Derived-Metric-Offset-f235a0aa4aab4100b42b7aa4afa87a1e) for an overview of what that feature should look like.
Basic approach:
1. Pass the offset params through the query via `MetricSpec`
2. In `CombineMetricsNode`, check if offset is required. If so, add time spine as the FROM dataset (to ensure we don't lose any rows we shouldn't), and update the join condition to offset as needed.

**Main question:**
Does this approach work (using the `CombineMetricsNode` to offset the join) or do we need to add a new node to join the time spine?
An edge case that's not currently handled is if a user tries to apply an offset to a derived metric with only one metric input, since that won't go through the `CombineMetricsNode` at all. But do we even want to allow that case? (Product decision to be made.) If we do allow it, we could add this same logic to the `ComputeMetricsNode` - but at that point it might be cleaner just to add a new node rather than duplicate logic. LMK your thoughts!

Other notes:
- Tom's join PR will impact the same section of the code so I'll need to do a refactor after his merges.
- What does it look like to apply granularity to these metrics? Will need to test and see if we need to prevent that.
- Need to communicate to optimizer that it shouldn't combine metrics into one query if they're from the same data source if their offsets don't match. This and one other known issue are causing unit test failures.


nhandel:
Just so excited 🧑🏼‍🎤

metricflow/plan_conversion/spec_transforms.py
```diff
@@ -88,34 +90,49 @@ class CreateOnConditionForCombiningMetrics(InstanceSpecSetTransform[SqlExpressio
...
+        left_expr: SqlExpressionNode = _make_coalesced_expr(self._table_aliases_in_coalesce, column_alias)
+        right_expr: SqlExpressionNode = SqlColumnReferenceExpression(
             col_ref=SqlColumnReference(
                 table_alias=self._table_alias_on_right_equality,
                 column_name=column_alias,
             )
         )
-        # if either has offset window, build date_add expr
-        # if either has offset grain to date, build date_trunc expr
+        if offset.left_offset:
```
tlento: Sorry, this whole class no longer exists........ check out the SqlQueryPlanJoinBuilder

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -923,8 +924,31 @@ def visit_combine_metrics_node(self, node: CombineMetricsNode[SqlDataSetT]) -> S
...
+        input_metric_specs: Dict[str, Sequence[MetricSpec]] = defaultdict(list)
+        parent_aliases_to_data_sets: OrderedDict[str, SqlDataSet] = OrderedDict()
+        table_alias_to_metric_specs: OrderedDict[str, Sequence[MetricSpec]] = OrderedDict()
+        for i, parent_node in enumerate(node.parent_nodes):
+            parent_data_set = parent_data_sets[i]
+            table_alias = self._next_unique_table_alias()
+            parent_data_sets.append(parent_data_set)
+
+            # If metric is derived, get input metrics.
+            if isinstance(parent_node, ComputeMetricsNode):
```
tlento, line 938: You'll want to rely on the derived-ness directly, rather than node type checks, since derived metrics can be derived from other derived metrics, which might themselves be represented here as a CombineMetricsNode. Like a ratio of the sums of multiple other metrics.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -923,8 +924,31 @@ def visit_combine_metrics_node(self, node: CombineMetricsNode[SqlDataSetT]) -> S
...
+                if not uses_time_offset:
+                    uses_time_offset = any(
+                        (metric_spec.offset_window or metric_spec.offset_to_grain_to_date)
+                        for metric_spec in parent_node.metric_specs
+                    )
```
tlento, line 943: Now that we've agreed to support single-measure offsets, is it the case that you need to use a time offset on any metric computation that has that value set? I believe it is, in which case making a separate node for the time spine join and applying the window/offset parameters in that node is likely to be a cleaner approach to covering all of the cases.

Like there's nothing preventing a derived metric from having an offset as an input, or from having two metrics with two different offsets that feed into a derived metric with its own offset. Like you might have some ad revenue conversion thing where you have a conversion event, a billing event, a revenue event, and a finalized revenue event, and those are all running on different offsets to keep them aligned to the conversion date (or to the finalized revenue date, depending).

In that case you'd have a separate TimeSpineJoinNode (or whatever) for each of those offsets, and it'd show up in the DataflowPlan as a formal step. You'd basically be rewriting the metric_time column with the time spine, which is something we need to support anyway, so we can come up with a lightweight extended design that parameterizes the node in the way we think we'll need.

tlento (commented):
My main high level thought is this will be easier to reason about if we make the Time Spine join its own node, and apply it to the metric based on its offset as kind of a way to rewrite metric time. We should chat about this because the approach I was planning to take for the generic time spine metric time might or might not work well with this scenario, so it'd be good to figure out if we can come up with something consistent, if not something that allows us to share most of the logic.

Mechanically speaking, this PR would be a faster review for me if the commits were more consolidated into unified chunks, and that'll be even more true if we add a new node since it involves modifying the dreaded DataflowPlanBuilder.

You may split things out into whatever form makes sense (and is easiest to get to from here, although I think `git add --patch` will work pretty well for whatever you choose), but for reference we typically split out our commits into something like:

1. Model/parser updates
2. Model validation (sometimes its own PR, just because)
3. New node with method stubs and integration into DataflowPlanBuilder
4. New SQL expressions (if needed)
5. Visitor implementation for SQL rendering

Here you might also consider making the class rename to MetricTimeWindow as its own PR, that's a great change and it can go in even while we think through design considerations.

PR #377, opened 2022-12-13
BUG: Aggregate functions gets mixed with other exprs when optimized

WilliamDee:
## Context
There are certain cases where the sub query reducer collapses cross joins, it potentially leads to a final select node containing a mix of aggregate functions and other exprs (ie., column references). This ends up throwing an invalid GROUP BY error.

### Before optimizer
```sql
-- query
SELECT
  src2.bookings AS bookings
  , src3.listings AS listings
FROM (
  -- src2
  SELECT
    SUM(src1.bookings) AS bookings
  FROM (
    -- src1
    SELECT
      1 AS bookings
    FROM demo.fct_bookings src0
  ) src1
) src2
CROSS JOIN (
  -- src4
  SELECT
    SUM(src4.listings) AS listings
  FROM demo.fct_listings src4
) src3
```
### Current optmizer
```sql
-- query
SELECT
  src2.bookings AS bookings
  , SUM(src4.listings) AS listings
FROM (
  -- src2
  SELECT
    SUM(1) AS bookings
  FROM demo.fct_bookings src1
) src2
CROSS JOIN
  demo.fct_listings src4
```
This would throw an error similar to this
```
SQL compilation error: [src2.bookings] is not a valid group by expression
```

### Optimizer with this PR
```sql
-- query
SELECT
  src2.bookings AS bookings
  , src3.listings AS listings
FROM (
  -- src1
  -- src2
  SELECT
    SUM(1) AS bookings
  FROM demo.fct_bookings src0
) src2
CROSS JOIN (
  -- src4
  SELECT
    SUM(src4.listings) AS listings
  FROM demo.fct_listings src4
) src3
```

### DIFF
--- a/metricflow/sql/optimizer/rewriting_sub_query_reducer.py
+++ b/metricflow/sql/optimizer/rewriting_sub_query_reducer.py
@@ -145,9 +145,10 @@ class SqlRewritingSubQueryReducerVisitor(SqlQueryPlanNodeVisitor[SqlQueryPlanNod
     def _is_simple_source(node: SqlSelectStatementNode) -> bool:
         """Returns true if the node is simple.
 
-        Simple is defined as having no JOINs, WHERE, GROUP BYs, ORDER BYs, LIMIT and there are no strings in the column
+        Simple is defined as having no JOINs, WHERE, GROUP BYs, ORDER BYs, LIMIT, AGG functions, and there are no strings in the column
         select. Strings are avoided so that the child node doesn't use the string expression in a group by or cause
-        aliasing issues when used in the child query.
+        aliasing issues when used in the child query. Aggregate functions are avoided due to the nature of applying on grouped rows which
+        is essentially the effect as group bys and should be treated in here as such.
 
         e.g.
 
@@ -180,6 +181,8 @@ class SqlRewritingSubQueryReducerVisitor(SqlQueryPlanNodeVisitor[SqlQueryPlanNod
             if select_column.expr.lineage.contains_string_exprs:
                 return False
             if select_column.expr.lineage.contains_column_alias_exprs:
+                return False
+            if select_column.expr.lineage.contains_aggregate_exprs:
                 return False
         return (
             len(node.parent_nodes) <= 1
--- a/metricflow/sql/sql_exprs.py
+++ b/metricflow/sql/sql_exprs.py
@@ -146,6 +146,10 @@ class SqlExpressionTreeLineage:
     def contains_ambiguous_exprs(self) -> bool:  # noqa: D
         return self.contains_string_exprs or self.contains_column_alias_exprs
 
+    @property
+    def contains_aggregate_exprs(self) -> bool:  # noqa: D
+        return any(x.is_aggregate_function for x in self.function_exprs)
+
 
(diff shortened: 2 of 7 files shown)


tlento (approved):
LGTM!

plypaul (changes requested):
Hold on, why can't there be group bys in a cross join?

tlento:
Ah, good question @plypaul 

There can be no group bys in our cross joins today, because the only cases where we use them are for combining aggregated nodes based on their group by columns. If there are group bys, we cannot use a cross join, because that would produce wildly incorrect results.

If we ever change this assumption, and allow for other CROSS JOIN contexts, this optimizer will then become somewhat less effective. As such, we should probably handle this case explicitly, or at least document the choice more explicitly. 

I think we have three optoins:

1. Keep the logic as-is and update the documentation to indicate that we aren't going to consolidate CROSS JOIN nodes with group by elements and an aggregation function because that could lead to confusing behavior post-optimization. CROSS JOIN is hard to reason about, so this seems reasonable to me.
2. Keep the core logic as is, but add an assertion that there are no group bys in the cross join context, and throw an error if that isn't true.
3. Update the core logic to include a check for an empty group by, and only do the bypass if there are no group bys AND there is an aggregation function present

I have a slight preference for option 2. We really shouldn't be running CROSS JOIN on group by attributes off an aggregation (although CROSS JOIN UNNEST is an interesting case), that's a recipe for bad data. If we don't have an assertion in the non-optimized path we probably should, although it's a bit tricky. Adding one here is a reasonable safeguard in any case, since it's useful to have the optimizer not be considered applicable for use cases where CROSS JOIN is needed with a group by.

metricflow/sql/optimizer/rewriting_sub_query_reducer.py
```diff
@@ -266,6 +275,12 @@ def _current_node_can_be_reduced(self, node: SqlSelectStatementNode) -> bool:  #
         ) > 0 and not SqlRewritingSubQueryReducerVisitor._select_columns_are_column_references(node.select_columns):
             return False
 
+        # It is best to avoid when a parent node contains an aggregate expression for simplicity
+        if SqlRewritingSubQueryReducerVisitor._select_columns_contain_aggregate_expressions(
```
WilliamDee, line 279: This causes us to lose a few optimization lines

metricflow/sql/sql_exprs.py
```diff
@@ -145,6 +145,10 @@ def contains_column_alias_exprs(self) -> bool:  # noqa: D
     def contains_ambiguous_exprs(self) -> bool:  # noqa: D
         return self.contains_string_exprs or self.contains_column_alias_exprs
 
+    @property
+    def contains_aggregate_exprs(self) -> bool:  # noqa: D
+        return any(isinstance(x, SqlAggregateFunctionExpression) for x in self.function_exprs)
```
plypaul, line 150: Would like to avoid the use of `isinstance`. Maybe add a boolean property like `is_aggregate_function` or similar to `SqlFunctionExpression`.

WilliamDee: fixed

plypaul (approved):

PR #378, opened 2022-12-14
Add percentile parsing

kyleli626:
Context: Percentile engine specific rendering merged previously - https://github.com/transform-data/metricflow/pull/371
This PR incorporates the SqlPercentileExpression into the dataplan and creates the parsing for this new aggregation type in the model yaml. 

Changes:
- Adds percentile and median aggregation type. 
- Replace SqlAggregateFunctionExpression.from_aggregation_type with method build_expression_from_aggregation_type, which can return a SqlPercentileExpression as well and takes in a MeasureAggregationParameter object.
- Add schemas for measures, including adding agg_params, a sub-schema which specifies percentile (indicating which percentile) and disc (whether to use discrete versions of percentiles). 
- Validation rule that checks percentile exists for percentile aggregation and aggregation measures only. Tests for validation rule.
- Transformation rule that converts the median aggregation type to a percentile one, and sets percentile to be 0.5
- Integration tests for percentile aggregation using new percentile measures defined in simple_model fixture. 
- Test for CreateSelectColumnsWithMeasuresAggregated, which actually creates the SqlPercentileExpression.


metricflow/model/transformations/convert_median.py
```diff
@@ -0,0 +1,21 @@
...
+
+    @staticmethod
+    def transform_model(model: UserConfiguredModel) -> UserConfiguredModel:  # noqa: D
+        for data_source in model.data_sources:
+            for measure in data_source.measures:
+                if measure.agg == AggregationType.MEDIAN:
+                    measure.agg = AggregationType.PERCENTILE
+                    if not measure.agg_params:
+                        measure.agg_params = MeasureAggregationParameters()
+                    measure.agg_params.percentile = MEDIAN_PERCENTILE
```
conglei, line 20: We should also force `disc` to be false.

metricflow/model/validations/measures.py
```diff
@@ -415,6 +415,62 @@ def validate_model(model: UserConfiguredModel) -> List[ValidationIssueType]:  #
...
+                                ),
+                            )
+                        )
+                    elif measure.agg_params.percentile <= 0 or measure.agg_params.percentile >= 1:
+                        issues.append(
+                            ValidationError(
+                                context=context,
+                                message=(
+                                    f"Percentile aggregation parameter for measure '{measure.name}' is '{measure.agg_params.percentile}', but"
+                                    "must be between 0 and 1 (non-inclusive). For example, to indicate the 65th percentile value, set 'percentile: 0.65'. "
```
conglei, line 453: We can add. `for 0 , please use MIN, for 1, please use MAX`

metricflow/test/model/validations/test_measures.py
```diff
@@ -332,3 +336,88 @@ def test_count_measure_with_distinct_expr(simple_model__pre_transforms: UserConf
     assert (
         actual_error.find(expected_error_substring) != -1
     ), f"Expected error {expected_error_substring} not found in error string! Instead got {actual_error}"
+
+
+def test_percentile_measure_missing_agg_params(simple_model__pre_transforms: UserConfiguredModel) -> None:
```
conglei, line 341: We should also add some tests for `MEDIAN` agg

kyleli626: I can't really think of any test cases that go wrong based on this validation case specifically. The median agg doesn't have any specific validations attached to it, and, just like the sum agg bad_measure_3, would raise an error if any agg_params are provided, which i think is what we want?

conglei: Yep, that is exactly what I thought: it should raise an error if `agg_params` is provided.

metricflow/test/model/validations/test_measures.py
```diff
@@ -332,3 +336,90 @@ def test_count_measure_with_distinct_expr(simple_model__pre_transforms: UserConf
...
+    """Tests that all measures with COUNT agg should have expr provided."""
```
conglei, line 342: the comment seems to be unrelated to this test.

metricflow/model/objects/elements/measure.py
```diff
@@ -20,6 +20,13 @@ class NonAdditiveDimensionParameters(HashableBaseModel):
     window_groupings: List[str] = []
 
 
+class MeasureAggregationParameters(HashableBaseModel):
+    """Describes parameters for aggregations"""
+
+    percentile: Optional[float] = None
+    disc: bool = False
```
conglei, line 27: This parameter should be optional as well.

conglei (approved):
some nit, very overall it looks good to me! Thanks for adding this!

metricflow/model/validations/measures.py
```diff
@@ -415,6 +415,63 @@ def validate_model(model: UserConfiguredModel) -> List[ValidationIssueType]:  #
...
+    """Checks that only PERCENTILE measures have agg_params and valid percentile value provided."""
+
+    @staticmethod
+    @validate_safely(
+        whats_being_done="running model validation ensuring the agg_params.percentile value exist for measures with percentile aggregation"
+    )
+    def validate_model(model: UserConfiguredModel) -> List[ValidationIssueType]:  # noqa: D
+        issues: List[ValidationIssueType] = []
+
+        for data_source in model.data_sources:
```
tlento, line 428: We generally prefer to have a separate function for the data source level checks with the `@validate_safely` decorator, as it helps us understand exactly what's happening in the validation stack.

metricflow/model/validations/measures.py
```diff
@@ -415,6 +415,63 @@ def validate_model(model: UserConfiguredModel) -> List[ValidationIssueType]:  #
...
+                            )
+                        )
+                    elif measure.agg_params.percentile <= 0 or measure.agg_params.percentile >= 1:
+                        issues.append(
+                            ValidationError(
+                                context=context,
+                                message=(
+                                    f"Percentile aggregation parameter for measure '{measure.name}' is '{measure.agg_params.percentile}', but"
+                                    "must be between 0 and 1 (non-inclusive). For example, to indicate the 65th percentile value, set 'percentile: 0.65'. "
+                                    "For percentile values of 0, please use MIN, for percentile values of 1, please use MAX."
```
tlento, line 454: Nice!

metricflow/model/validations/measures.py
```diff
@@ -415,6 +415,63 @@ def validate_model(model: UserConfiguredModel) -> List[ValidationIssueType]:  #
...
+                    measure.agg != AggregationType.PERCENTILE
+                    and measure.agg_params
+                    and (measure.agg_params.percentile or measure.agg_params.disc)
+                ):
+                    issues.append(
+                        ValidationError(
+                            context=context,
+                            message=(
+                                f"Measure '{measure.name}' with aggregation '{measure.agg.value}' uses agg_params "
+                                "only relevant to Percentile measures."
```
tlento, line 468: Might be nice to include the agg params in this message.

metricflow/test/model/validations/test_measures.py
tlento: The post-validation transform rules are all applied pre-validation. Yes, I know. But that's how it works.

We should assert that the model input doesn't have wacky values, but that has to happen in the transform rule.

metricflow/model/transformations/convert_median.py
```diff
@@ -0,0 +1,22 @@
...
+                    measure.agg_params.disc = False
```
tlento, line 21: Let's assert that these are not set to different values. Like if someone says they want a median with percentile=0.25 they should get an error telling them that this is conflicting and they should remove the percentile parameter.

We can allow them to code the correct values in the model, that's fine.

metricflow/test/model/validations/test_measures.py
```diff
@@ -332,3 +336,90 @@ def test_count_measure_with_distinct_expr(simple_model__pre_transforms: UserConf
...
+        agg_params=MeasureAggregationParameters(),
+    )
+
+    invalid_measure_3 = Measure(
+        name="bad_measure_3",
+        agg=AggregationType.SUM,
+        agg_time_dimension="ds",
+        agg_params=MeasureAggregationParameters(percentile=0.3),
+    )
+    data_source.measures.extend([invalid_measure_1, invalid_measure_2, invalid_measure_3])  # type: ignore
```
tlento, line 363: I'd like to move away from this pattern and on to one where we make a simple data source that has the measures we need, just for validation. See https://github.com/transform-data/metricflow/blob/main/metricflow/test/model/validations/test_validity_param_definitions.py#L11-L39 for an example.

tlento (commented):
Woohoo! Looks great. Left a few minor comments for cleanup and clarity inline.

I do think we should move away from `disc` - it's an abbreviation of a term of art, and we typically favor self-descriptive model properties, so I'd like to see that happen before merge.

metricflow/model/objects/elements/measure.py
kyleli626: or by default its false, we don't want it ever to be None, correct? so it should be set to false.

metricflow/sql/sql_exprs.py
```diff
@@ -708,6 +719,22 @@ class SqlFunctionExpression(SqlExpressionNode):
     pass
 
 
+def build_expression_from_aggregation_type(
```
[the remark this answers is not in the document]
kyleli626: one thing conglei mentioned is that perhaps superclasses shouldn't necessarily have knowledge of their subclasses as a general convention. don't think it's too big of a deal right now, and can't think of any bad consequences right now, so can make this change

metricflow/model/validations/measures.py
kyleli626: every other measure validation (except MeasureConstraintAliasesRule) seems to do data source level checks in the main validate_model function without having a separate function, would you like me to separate those?

And would the main validate_model just look like: 

`def validate_model(model: UserConfiguredModel) -> List[ValidationIssueType]:  # noqa: D
        issues: List[ValidationIssueType] = []

        for data_source in model.data_sources:
            for measure in data_source.measures:
                context = DataSourceElementContext(
                    file_context=FileContext.from_metadata(metadata=data_source.metadata),
                    data_source_element=DataSourceElementReference(
                        data_source_name=data_source.name, element_name=measure.name
                    ),
                    element_type=DataSourceElementType.MEASURE,
                )
                issues.append(PercentileAggregationRule._validate_percentile_measure(measure, context))

        return issues`

metricflow/sql/render/redshift.py
```diff
@@ -21,11 +21,15 @@ def visit_percentile_expr(self, node: SqlPercentileExpression) -> SqlExpressionR
         arg_rendered = self.render_sql_expr(node.order_by_arg)
         params = arg_rendered.execution_parameters
 
+        approximate_str = (
+            "APPROXIMATE " if node.percentile_args.function_type is SqlPercentileFunctionType.DISCRETE else ""
```
tlento, line 25: Ooof, this doesn't work? https://docs.aws.amazon.com/redshift/latest/dg/r_WF_PERCENTILE_DISC.html

If that's the case, I'd rather validate and throw an error if we're requesting the approx for discrete and the true value otherwise. approx_percentile can be divergent from percentile by a fair bit, depending on the algorithm and underlying distribution of the data, so giving users approx in one case and "true" in the other without making it clear that this is what's happening is best avoided.

I don't think we can do this as a semantic validation on the model, since it's an engine capability thing, but we could throw an informative error here instructing the user to disable the discrete percentile flag in the model. If that proves overly burdensome we could downgrade it to a warning.

tlento (commented):
Thanks for refactoring that test file, looks great!

I don't think we should be silently falling back to approximate percentile in Redshift. I'd rather throw an error if the model uses a discrete percentile on a Redshift engine.

metricflow/test/integration/test_configured_cases.py
```diff
@@ -102,6 +105,22 @@ def render_date_trunc(self, expr: str, granularity: TimeGranularity) -> str:
             renderable_expr
         ).sql
 
+    def render_percentile_expr(self, expr: str, percentile: float, use_discrete_percentile: bool) -> str:
+        """Return the percentile call that can be used for ."""
```
tlento, line 109: nit - incomplete docstring

metricflow/test/sql/test_engine_specific_rendering.py
```diff
@@ -106,39 +106,39 @@ def test_percentile_expr(
...
 ) -> None:
     """Tests rendering of the percentile expression in a query."""
-
-    select_columns = [
-        SqlSelectColumn(
-            expr=SqlPercentileExpression(
-                order_by_arg=SqlColumnReferenceExpression(SqlColumnReference("a", "col0")),
-                percentile_args=SqlPercentileExpressionArgument(
-                    percentile=0.5, function_type=SqlPercentileFunctionType.CONTINUOUS
+    if sql_client.sql_engine_attributes.percentile_aggregation_supported:
```
tlento, line 109: Let's mark this to skip instead. So it'd be something like:

```
if not sql_client.sql_engine_attributes.percentile_aggregation_supported:
    pytest.skip("Warehouse does not support percentile expressions")

select_columns = [
# .... rest of test here
```

Then we get a skip notice in the test runner rather than a spurious pass.

tlento (approved):
Looks good, thanks! Let's hold merges until the new year, but after that you may merge this at your leisure!

PR #379, opened 2022-12-15
Fix incomplete JOIN condition for combining derived metrics nodes

tlento:
The CombineMetricsNode was originally designed to join together multiple
fully computed metrics nodes for rendering unified results to an
end user. As such, it relied on a FULL OUTER JOIN, and doing so in
a way that failed to account for NULL values on the selected dimensions
would result in potential duplicate sets of dimension values.

In addressing this annoyance we noticed that derived metrics were using
this node to join input metrics for subsequent expression computation,
and using an INNER JOIN for consistency with multiple-measure
computations, which themselves use INNER JOIN because we have no
current mechanism for users to consistently handle NULL inputs to
measure/metric expressions.

This means the on condition MUST account for NULL inputs correctly. The
original join, which was essentially `ON a.dimension = b.dimension`, would
exclude the joined row altogether if both `a.dimension` and `b.dimension` were
NULL. The technique we were forced to adopt for the FULL OUTER JOIN case
would not work here, because that does a re-aggregation pass, but here we
need to prevent the rows from being thrown away.

Therefore, this PR effectively converts the ON condition for any join type
other than FULL OUTER JOIN to a compound test of the form:

`ON a.dimension = b.dimension OR (a.dimension IS NULL AND b.dimension IS NULL)`

Doing this required a certain amount of additional testing and restructuring of code,
which is documented in the commit structure of this PR.

### DIFF
--- a/metricflow/plan_conversion/dataflow_to_sql.py
+++ b/metricflow/plan_conversion/dataflow_to_sql.py
@@ -951,23 +951,46 @@ class DataflowToSqlQueryPlanConverter(Generic[SqlDataSetT], DataflowPlanNodeVisi
         linkable_spec_set = from_data_set.data_set.instance_set.spec_set.transform(SelectOnlyLinkableSpecs())
         join_type = SqlJoinType.CROSS_JOIN if len(linkable_spec_set.all_specs) == 0 else node.join_type
 
-        joins_descriptions = []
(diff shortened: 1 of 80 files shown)


WilliamDee (approved):
this is awesome, thanks for cleaning up everything along the way!

PR #383, opened 2022-12-21
Fix `MetricFlow SQL Engine Tests` that got broken due to #324

QMalcolm:
After my merge of #324, the `MetricFlow Sql Engine Tests` began failing. This is because I missed the fact that `MetricFlow SQL Engine Tests` run the entire MetricFlow test suite. This raised two issues
1. `MetricFlow Sql Engine Tests` shouldn't run the dbt cloud support tests
2. dbt cloud support tests need to be easy to isolate

This PR isolates the dbt cloud support tests (and related materials like fixtures) into one dir making them easy to skip: `poetry run pytest metricflow/test --ignore=metricflow/test/model/dbt_cloud_parsing`. Additionally this PR begins skipping the dbt cloud support tests in the `MetricFlow Sql Engine Tests`.

### DIFF
--- a/.github/workflows/cd-sql-engine-tests.yaml
+++ b/.github/workflows/cd-sql-engine-tests.yaml
@@ -36,7 +36,7 @@ jobs:
         run: cd metricflow && poetry install
 
       - name: Run MetricFlow unit tests with Snowflake configs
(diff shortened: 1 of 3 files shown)


tlento (approved):
Thank you!

PR #384, opened 2022-12-23
Write JoinToTimeSpineNode for dataflow plan

courtneyholcomb:
Add a new DataflowPlanNode to join a dataset to the time spine, potentially with a time offset.


metricflow/dataflow/dataflow_plan.py
```diff
@@ -707,6 +712,95 @@ class ComputedMetricsOutput(Generic[SourceDataSetT], BaseOutput[SourceDataSetT],
...
 
+class JoinToTimeSpineNode(Generic[SourceDataSetT], BaseOutput[SourceDataSetT], ABC):
+    """Join parent dataset to time spine dataset."""
+
+    def __init__(
+        self,
+        parent_node: BaseOutput[SourceDataSetT],
+        time_range_constraint: Optional[TimeRangeConstraint] = None,
+        offset_window: Optional[MetricTimeWindow] = None,
+        offset_to_grain_to_date: Optional[TimeGranularity] = None,
```
tlento, line 723: Nice, I can use this as a source node as well if I need to.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1297,3 +1301,64 @@ def visit_semi_additive_join_node(self, node: SemiAdditiveJoinNode) -> SqlDataSe
...
+    def visit_join_to_time_spine_node(self, node: JoinToTimeSpineNode[SourceDataSetT]) -> SqlDataSet:  # noqa: D
+        parent_data_set = node.parent_node.accept(self)
+        parent_alias = self._next_unique_table_alias()
+
+        # Build time spine dataset
+        metric_time_dimension_instance: Optional[TimeDimensionInstance] = None
+        for instance in parent_data_set.metric_time_dimension_instances:
+            if len(instance.spec.identifier_links) == 0:
+                metric_time_dimension_instance = instance
+                break
```
tlento, line 1314: This returns whatever granularity happens to be first. I think this is usually DAY but we don't have any guarantees around it.

We probably want to do something similar to visit_constrain_time_range_node, because I think we want the time spine comparison to be against the smallest granularity metric time dimension:

https://github.com/transform-data/metricflow/blob/main/metricflow/plan_conversion/dataflow_to_sql.py#L1007-L1030

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1297,3 +1301,64 @@ def visit_semi_additive_join_node(self, node: SemiAdditiveJoinNode) -> SqlDataSe
...
+            right_source_alias=parent_alias,
+            on_condition=SqlComparisonExpression(
+                left_expr=left_expr,
+                comparison=SqlComparison.EQUALS,
+                right_expr=SqlColumnReferenceExpression(
+                    col_ref=SqlColumnReference(table_alias=parent_alias, column_name=metric_time_dimension_column_name)
+                ),
+            ),
+            join_type=SqlJoinType.LEFT_OUTER,  # TODO: test other join types
+        )
```
tlento, line 1351: Longer term we'll want to move this to the SqlQueryJoinBuilder class. It's going to get more complicated when I add the other scenarios here. 

It's up to you if you want to move this over there now or leave it here while we work on this.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1297,3 +1301,64 @@ def visit_semi_additive_join_node(self, node: SemiAdditiveJoinNode) -> SqlDataSe
...
+        left_expr: SqlExpressionNode = SqlColumnReferenceExpression(
+            col_ref=SqlColumnReference(table_alias=time_spine_alias, column_name=metric_time_dimension_column_name)
+        )
+        if node.offset_window:
+            left_expr = SqlTimeDeltaExpression(
+                arg=left_expr, count=node.offset_window.count, granularity=node.offset_window.granularity
+            )
+        elif node.offset_to_grain_to_date:
+            left_expr = SqlDateTruncExpression(time_granularity=node.offset_to_grain_to_date, arg=left_expr)
+        join_description = SqlJoinDescription(
```
tlento, line 1340: nit: line break after the else block please. :)

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1297,3 +1301,64 @@ def visit_semi_additive_join_node(self, node: SemiAdditiveJoinNode) -> SqlDataSe
...
+        )
+
+        # Build join expression
+        left_expr: SqlExpressionNode = SqlColumnReferenceExpression(
+            col_ref=SqlColumnReference(table_alias=time_spine_alias, column_name=metric_time_dimension_column_name)
+        )
+        if node.offset_window:
+            left_expr = SqlTimeDeltaExpression(
+                arg=left_expr, count=node.offset_window.count, granularity=node.offset_window.granularity
+            )
```
tlento, line 1337: Ah, ok, I see. The offset here is always exact to granularity, not capturing a range, right? I mean if it's 10 days, we don't look for all dates between 10 days ago, we match on things that were exactly 10 days ago.

Note this is applied to the time spine, not the metric data set. So we're offsetting the spine and filtering out any rows from 10 days after the start of the time spine.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1297,3 +1301,64 @@ def visit_semi_additive_join_node(self, node: SemiAdditiveJoinNode) -> SqlDataSe
...
+        elif node.offset_to_grain_to_date:
+            left_expr = SqlDateTruncExpression(time_granularity=node.offset_to_grain_to_date, arg=left_expr)
```
tlento, line 1339: See comment on the SQL - this does the date trunc on the time spine value, not the metric time value. So the EQUALS below filters all of the intervening values. I believe we need the date_trunc on the other side of the join comparison as well, so we probably need to build up a right_expr here.

metricflow/test/snapshots/test_dataflow_to_sql_plan.py/SqlQueryPlan/DuckDbSqlClient/test_join_to_time_spine_node_with_offset_window__plan0_optimized.sql
```diff
@@ -0,0 +1,43 @@
...
+      SELECT * FROM ***************************.fct_bookings
+    ) bookings_source_src_10001
+    GROUP BY
+      ds
+      , ds
+      , listing_id
+  ) subq_9
+) subq_10
+ON
+  subq_11.metric_time - INTERVAL 10 day = subq_10.metric_time
```
tlento, line 43: Two things:

1. The aliases on the outer select columns don't look right, they should be coming from subq_10 or subq_11, not subq_9. I think I see the issue in the visitor, will comment there.
2. I don't think this is offsetting the values. What we want to do with the offset is re-assign the metric value to a different date, which I believe means the metric_time has to come from the time_spine with offset applied, not the original metric dataset. Is that right?

So it'd be like:

```
SELECT spine.metric_time, metric.metric_value
FROM (
    SELECT metric_time
    FROM time_spine
    WHERE <constraint>
) spine
LEFT OUTER JOIN (
  SELECT metric_time, metric_value
  FROM metric_data_set
) metric
ON spine.metric_time - INTERVAL 10 day = metric.metric_time
```

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1297,3 +1301,64 @@ def visit_semi_additive_join_node(self, node: SemiAdditiveJoinNode) -> SqlDataSe
...
+                ),
+            ),
+            join_type=SqlJoinType.LEFT_OUTER,  # TODO: test other join types
+        )
+
+        return SqlDataSet(
+            instance_set=InstanceSet.merge([time_spine_dataset.instance_set, parent_data_set.instance_set]),
+            sql_select_node=SqlSelectStatementNode(
+                description=node.description,
+                select_columns=parent_data_set.sql_select_node.select_columns,
```
tlento, line 1357: See commentary on SQL below, but I think we need two changes here:

1. This has to be built out using `create_select_columns_for_instance_sets` like in the `visit_join_to_base_output_node` method. We first need to map the time_spine_alias and parent_alias values to the columns we want to include in the select and then let this update those aliases accordingly.
2. I'm less certain of this, but I think for the offsetting to work we'll want to make sure the output metric_time column is the one from the time_spine, not the metric dataset. You can do that as part of this alias mapping process or in a custom transform on the instance set. I might be wrong, though, it's really hard to be sure without seeing the input/output data.

tlento (commented):
Wheee! Getting close! The query conversion needs a couple of fixes - my comments about the join condition itself might be off base, but I'm pretty sure we're filtering when we want to be re-mapping.

I mentioned we have a setup for adding new nodes so we can have the full test suite in place in the same PR while maintaining readability, but I just went looking and I couldn't find ready single-PR examples with new nodes. I may try to dig some up for our contributor guide, I think it'd help new people get moving on tasks like this one. If I find one would love your input on whether or not it would've helped here.

Anyway, as we discussed, splitting into separate PRs as you're doing with the model updates/validations/integration tests works just fine, we've done that in the past as well.

metricflow/test/snapshots/test_dataflow_to_sql_plan.py/SqlQueryPlan/DuckDbSqlClient/test_join_to_time_spine_node_with_offset_to_grain_to_date__plan0_optimized.sql
```diff
@@ -0,0 +1,43 @@
...
+      SELECT * FROM ***************************.fct_bookings
+    ) bookings_source_src_10001
+    GROUP BY
+      ds
+      , ds
+      , listing_id
+  ) subq_9
+) subq_10
+ON
+  DATE_TRUNC('month', subq_11.metric_time) = subq_10.metric_time
```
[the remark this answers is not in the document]
courtneyholcomb: We actually are just looking for the value on the first day of the granularity period. Not very useful for a derived metric with only one metric input, but it is useful when compared with another. An example use case is something like this: `bookings - bookings_start_of_month` gives you the metric `booking_since_start_of_month`, or `(bookings - bookings_start_of_month) / bookings_start_of_month` gives you `bookings_growth_this month`.

courtneyholcomb:
> I mentioned we have a setup for adding new nodes so we can have the full test suite in place in the same PR while maintaining readability, but I just went looking and I couldn't find ready single-PR examples with new nodes. I may try to dig some up for our contributor guide, I think it'd help new people get moving on tasks like this one. If I find one would love your input on whether or not it would've helped here.
It was actually not so bad to implement a new node! (Assuming I didn't miss anything...) The linter very helpfully shows you all the abstract classes where you need to implement a new function.

metricflow/plan_conversion/sql_join_builder.py
```diff
@@ -60,7 +60,7 @@ class SqlQueryPlanJoinBuilder:
     """Helper class for constructing various join components in a SqlQueryPlan"""
 
     @staticmethod
-    def make_sql_join_description(
+    def make_column_equality_sql_join_description(
```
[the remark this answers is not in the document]
courtneyholcomb: Updated the name because I started trying to use this function to build my join descriptions, and quickly realized it wouldn't support that!

metricflow/plan_conversion/dataflow_to_sql.py
courtneyholcomb: Correct on both counts!

> So we're offsetting the spine and filtering out any rows from 10 days after the start of the time spine.
Hmmm. Do we worry that users will want to include those missing 10 days that get removed from the time spine? When querying an offset metric, we could adjust the requested time range constraint when there is an offset window. But that won't fix the issue if they're querying from the start of the time range available in the time spine.

metricflow/test/snapshots/test_dataflow_to_sql_plan.py/SqlQueryPlan/DuckDbSqlClient/test_join_to_time_spine_node_with_offset_window__plan0_optimized.sql
courtneyholcomb: Ah you're right, should be selecting metric_time from `subq_11` instead of `subq_9`! Will update.

metricflow/test/snapshots/test_dataflow_to_sql_plan.py/SqlQueryPlan/DuckDbSqlClient/test_join_to_time_spine_node_with_offset_to_grain_to_date__plan0_optimized.sql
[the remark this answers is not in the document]
tlento: Oh I see! That's not what grain_to_date means to me, I think of it as being like year_to_date or month_to_date, not "value at start of period". 

Maybe we should give this a different name?

metricflow/plan_conversion/dataflow_to_sql.py
tlento: I *think* this gets resolved when we grab the time_spine metric_time. The only missing values in that case would be values that are missing in the data.

It'd be good to have our integration test cases in the next PR check that boundary so we can see where the NULLs show up, but I'm pretty sure your next update will address this!

metricflow/test/snapshots/test_dataflow_to_sql_plan.py/SqlQueryPlan/DuckDbSqlClient/test_join_to_time_spine_node_with_offset_to_grain_to_date__plan0_optimized.sql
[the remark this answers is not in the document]
courtneyholcomb: Renamed as discussed in slack!

metricflow/plan_conversion/dataflow_to_sql.py
courtneyholcomb: We're good on this, as discussed in slack!

metricflow/test/snapshots/test_dataflow_to_sql_plan.py/SqlQueryPlan/DuckDbSqlClient/test_join_to_time_spine_node_with_offset_window__plan0_optimized.sql
courtneyholcomb: Updated if you want to check out the new snapshots.

metricflow/plan_conversion/dataflow_to_sql.py
courtneyholcomb: Updated to fix both!

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1297,3 +1301,84 @@ def visit_semi_additive_join_node(self, node: SemiAdditiveJoinNode) -> SqlDataSe
...
+        ), "Can't query offset metric without a time dimension. Validations should have prevented this."
+        metric_time_dimension_column_name = self.column_association_resolver.resolve_time_dimension_spec(
+            metric_time_dimension_instance.spec
+        ).column_name
+        time_spine_alias = self._next_unique_table_alias()
+        time_spine_dataset = _make_time_spine_data_set(
+            metric_time_dimension_instance=metric_time_dimension_instance,
+            metric_time_dimension_column_name=metric_time_dimension_column_name,
+            time_spine_source=self._time_spine_source,
+            time_spine_table_alias=time_spine_alias,
```
tlento, line 1326: Ugh, there's a bug in this API and it's causing a repeat table alias in your rendered SQL. The only current callsite does this:

```suggestion
            time_spine_table_alias=self._next_unique_table_alias(),
```

Feel free to copy that for now, since removing that parameter is a little bit annoying.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1297,3 +1301,84 @@ def visit_semi_additive_join_node(self, node: SemiAdditiveJoinNode) -> SqlDataSe
...
+
+        # Build join expression
+        left_expr: SqlExpressionNode = SqlColumnReferenceExpression(
+            col_ref=SqlColumnReference(table_alias=time_spine_alias, column_name=metric_time_dimension_column_name)
+        )
+        if node.offset_window:
+            left_expr = SqlTimeDeltaExpression(
+                arg=left_expr, count=node.offset_window.count, granularity=node.offset_window.granularity
+            )
+        elif node.offset_to_grain:
```
tlento, line 1338: Do we want any validation on the configuration of this node? Might be nice so we know illegal states are always excluded.

I think that's work for a separate PR, though, since we'll be messing about with this a bit anyway.

metricflow/test/snapshots/test_dataflow_to_sql_plan.py/SqlQueryPlan/BigQuerySqlClient/test_join_to_time_spine_node_with_offset_to_grain__plan0_optimized.sql
```diff
@@ -0,0 +1,39 @@
+-- Join to Time Spine Dataset
+SELECT
+  subq_11.metric_time AS metric_time
```
tlento, line 3: Was this supposed to be truncated to month as well?

metricflow/test/snapshots/test_dataflow_to_sql_plan.py/SqlQueryPlan/BigQuerySqlClient/test_join_to_time_spine_node_with_offset_to_grain__plan0_optimized.sql
```diff
@@ -0,0 +1,39 @@
...
+  , subq_10.listing AS listing
+  , subq_10.booking_fees AS booking_fees
+FROM (
+  -- Date Spine
+  SELECT
+    ds AS metric_time
+  FROM ***************************.mf_time_spine subq_11
```
tlento, line 10: subq_11 is repeated. Left a comment for the fix in the dataflow_to_sql visitor

metricflow/plan_conversion/dataflow_to_sql.py
tlento: We should update this before merge as well.

tlento (approved):
Small things! I think they can all be dealt with before merge without further review, but I'm happy to take another look if you'd like. Details inline, but the relevant issues are:

1. Repeated subquery alias
2. Unordered metric_time instances (we might get wrong granularity)
3. Do we need to do date_trunc in the select on the offset_to_grain? If so we should probably do that now. If not, no need to do anything (although a comment might be helpful)!
4.

metricflow/plan_conversion/dataflow_to_sql.py
courtneyholcomb: I will have validation for the derived metric to confirm that you don't set both a grain and a window. That will be in the next PR! Let me know if you had additional validation in mind.

metricflow/test/snapshots/test_dataflow_to_sql_plan.py/SqlQueryPlan/BigQuerySqlClient/test_join_to_time_spine_node_with_offset_to_grain__plan0_optimized.sql
courtneyholcomb: Nope! We want to keep that at daily so each day shows the value at the start of the grain. (Again, not very useful for an individual metric, but useful when comparing to another metric.)

metricflow/plan_conversion/dataflow_to_sql.py
courtneyholcomb: Oops, missed this before!

courtneyholcomb: Just updated to loop through all of them and use the one with the lowest granularity.

PR #385, opened 2022-12-23
Rename `CumulativeMetricWindow` to `MetricTimeWindow`

courtneyholcomb:
This class will soon be used to specify offset windows for derived metrics. Renaming to something that makes sense for both use cases!


tlento (approved):
I love mechanical cleanup PRs. Thank you!

PR #386, opened 2022-12-23
Derived Metric Offset

courtneyholcomb:
Enable users to set a time offset for input metrics within a derived metric, specifying either an `offset_window` or an `offset_to_grain`. Examples for what these metrics should look like are available [here](https://www.notion.so/transformdata/Derived-Metric-Offset-f235a0aa4aab4100b42b7aa4afa87a1e).


metricflow/query/query_parser.py
```diff
@@ -274,9 +282,6 @@ def _parse_and_validate_query(
             # If the time constraint is all time, just ignore and not render
             time_constraint = None
 
-        if len(group_by_names) == 0:
```
courtneyholcomb, line 277: Previously, we were only checking if there were no dimensions at all for metrics that require a time dimension. This wouldn't catch the scenario when there is a dimension in the query but it isn't a time dimension (in which case we should error).

metricflow/query/query_parser.py
```diff
@@ -511,7 +518,7 @@ def _parse_metric_names(self, metric_names: Sequence[str]) -> Tuple[MetricRefere
             metric = self._metric_semantics.get_metric(metric_reference)
             if metric.type == MetricType.DERIVED:
                 input_metrics = self._parse_metric_names([metric.name for metric in metric.input_metrics])
-                metric_references.extend(list(input_metrics))
+                metric_references.extend([metric_reference] + list(input_metrics))
```
courtneyholcomb, line 521: Here we were verifying the metric inputs, but not the derived metric itself.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -205,16 +205,6 @@ def _build_metrics_output_node(
                     ),
                     metric_specs=[metric_spec],
                 )
-                if metric_spec.offset_window or metric_spec.offset_to_grain:
```
courtneyholcomb, line 208: Updated this logic to handle an offset in any metric, not just a derived metric. This is because this function will run for the input metric, not the derived metric. (Config restrictions will still prevent non-derived metrics from having an offset.)

Issue #387, opened 2022-12-29
Test runs fail due to missing dbt dependencies


tlento:
**Describe the bug**
MetricFlow has a set of dependencies on dbt core and dbt cloud APIs, which provide support for an optional integration with dbt. These dependencies are all flagged as optional, since they're only necessary for end users who wish to use the MetricFlow <-> dbt integration.

Unfortunately, by default pytest's test runner does a full walk of all conftest.py files, including the ones in the dbt integration directories. This means the `make test` command (and any pytest invocation that does not explicitly skip the dbt tests) will fail due to missing import dependencies unless the developer has already installed the required dbt packages in their local environment.

Ideally, the package structure would be updated such that the test runner naturally runs required tests, and anybody working in the dbt packages can run those tests in a natural way. Whether we do this by updating our overall package organization (e.g., by splitting the dbt integrations into a set of packages that depend on the core MetricFlow libraries), by restructuring our test package (e.g., by putting the dbt tests into a different tests directory), or by some other approach is undecided.

**Steps To Reproduce**

1. Create a clean virtual environment with poetry installed and NO dbt installations available via the venv PATH or PYTHONPATH
2. Run `make install`
3. Run `make test`

**Expected behavior**
Tests should run rather than failing with an import dependency exception.

PR #388, opened 2022-12-29
Update makefile and developer docs for dbt-cloud

tlento:
Our integration with dbt-cloud adds optional dependencies
which are currently required to run the complete test suite.

This doesn't resolve the broader issue raised in #387, but it
does bring our developer documentation up to date with the current
state of our main branch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,15 @@ install:
 # Testing and linting
 .PHONY: test
 test:
-	poetry run pytest metricflow/test/
+	poetry run pytest metricflow/test/ --ignore metricflow/test/model/dbt_cloud_parsing
+
+.PHONY: test-dbt
+test-dbt:
+	poetry run pytest metricflow/test/model/dbt_cloud_parsing
+
+.PHONY: test-all
+test-all:
+	poetry run pytest metricflow/test
 
 .PHONY: test-postgresql
 test-postgresql:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -21,7 +21,15 @@ install:
...
 test:
-	poetry run pytest metricflow/test/
+	poetry run pytest metricflow/test/ --ignore metricflow/test/model/dbt_cloud_parsing
+
+.PHONY: test-dbt
+test-dbt:
+	poetry run pytest metricflow/test/model/dbt_cloud_parsing
+
+.PHONY: test-all
+test-all:
```
QMalcolm, line 31: Unfortunate that we have to have a `test` and a `test-all`, but also for the time being I don't know what the alternative would be 😞

QMalcolm (approved):
Thanks for doing this work! I should have included it in my work introducing this functionality, and I'll be more cognizant of that going forward.

Makefile
tlento: I didn't even know this was needed until my `make test` blew up. I'm not sure how you would've even noticed given that you had the dbt packages installed for your dev work.

I think the longer term solution will involve re-packaging somehow, either making the dbt bits required as part of metricflow core library, or else having the dbt bits packaged as an extension that imports the metricflow core library. But that's quite a lot of work so I figured we can use this as a stopgap.

One smaller scale solution would be to have the dbt tests in their own test package, and then we could have test and test-dbt.

PR #381, opened 2022-12-20
Restrict SQLAlchemy package version to <1.4.42

tlento:
Due to an incompatibility between Snowflake's SQLAlchemy connector
and SQLAlchemy version 1.4.42, we must either update our
Snowflake dependency to 1.4.3 or restrict SQLAlchemy updates to
versions prior to 1.4.42.

Rather than try to rush an update from snowflake-sqlalchemy 1.2.3
to 1.4.3, this commit temporarily limits our SQLAlchemy version
dependency to versions >=1.4.27 and <1.4.42

See https://github.com/snowflakedb/snowflake-sqlalchemy/issues/350
for more information on the underlying issue.

PR #389, opened 2023-01-04
Add all group by columns to select in semi-additive join node

tlento:
Under certain conditions the semi-additive join node rendering
produces slightly strange SQL, where columns in the `GROUP BY`
statement are not included in the `SELECT`. For example, the inner
query for the filter join might look like this:

```
SELECT MIN(ds) AS ds__complete
FROM input_table
GROUP BY DATE_TRUNC(ds, isoweek)
```

This is confusing, as the DATE_TRUNC expression in the group by
allows for multiple rows of output. In this case, since the `ds`
columns are dates, there will always be exactly 1 row per `isoweek`,
however it's not obvious that this is the intent of the query.

This also causes problems with certain rendering approaches, as
we typically render either a raw expression OR an alias in `GROUP BY`
statements. In cases where we render the latter, this query will
break because the `DATE_TRUNC()` *must* be rendered as a full expression.
Depending on inputs and optimizer layouts this could, for certain engines,
result in broken SQL, a problem which is easily avoided if the expression
is included in the `SELECT` statement as a properly aliased column expression.

This change adds the full set of `GROUP BY` columns to the `SELECT` expression,
which results in the following updated SQL (for expr-style group by references):

```
SELECT DATE_TRUNC(ds, isoweek) AS ds__week, MIN(ds) AS ds__complete
FROM input_table
GROUP BY DATE_TRUNC(ds, isoweek)
```

### DIFF
--- a/metricflow/plan_conversion/dataflow_to_sql.py
+++ b/metricflow/plan_conversion/dataflow_to_sql.py
@@ -1265,7 +1265,7 @@ class DataflowToSqlQueryPlanConverter(Generic[SqlDataSetT], DataflowPlanNodeVisi
         # Construct SelectNode for Row filtering
         row_filter_sql_select_node = SqlSelectStatementNode(
             description=f"Filter row on {node.agg_by_function.name}({time_dimension_column_name})",
-            select_columns=tuple(identifier_select_columns) + (time_dimension_select_column,),
+            select_columns=row_filter_group_bys + (time_dimension_select_column,),
             from_source=from_data_set.sql_select_node,
             from_source_alias=inner_join_data_set_alias,
             joins_descs=(),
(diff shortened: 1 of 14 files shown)


tlento:
* main
  * **PR #381** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/381" target="_blank">Graphite</a> 
    * **PR #389** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/389" target="_blank">Graphite</a>  👈
      * **PR #390** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/390" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/transform-data/metricflow/389?utm_source=stack-comment).

WilliamDee (approved):
lgtm!

PR #390, opened 2023-01-04
Render aliases in group by expressions for BigQuery

tlento:
Metricflow's default renderer produces `GROUP BY` expressions
that exactly mirror the `SELECT` expression in a query. This would
look something like:

```
  SELECT COALESCE(x, y) AS x_or_y, SUM(1)
  FROM source_table
  GROUP BY COALESCE(x, y)
```

The use of the `COALESCE` expression in the `GROUP BY` causes BigQuery
to throw an exception.

Previously, all queries involving a `GROUP BY` expression have not
caused issues, because they were either column name references
(e.g., `SELECT x, SUM(1) FROM source_table GROUP BY x`), or else
they were run through our optimizer stack. The optimizer stack
includes a subquery reducer, which will convert consolidated
select expressions into appropriately aliased `GROUP BY` references
as needed.

With the recent updates to how we join metric data sets together,
we have uncovered some edge cases where that optimizer will not
trigger. It is, therefore, finally time for us to render BigQuery
`GROUP BY` expressions correctly. This commit changes BigQuery's
rendering to:

```
  SELECT COALESCE(x, y) AS x_or_y, SUM(1)
  FROM source_table
  GROUP BY x_or_y
```

### DIFF
--- a/metricflow/sql/render/big_query.py
+++ b/metricflow/sql/render/big_query.py
@@ -12,6 +12,7 @@ from metricflow.sql.sql_exprs import (
     SqlPercentileExpression,
     SqlTimeDeltaExpression,
 )
+from metricflow.sql.sql_plan import SqlSelectColumn
 from metricflow.time.time_granularity import TimeGranularity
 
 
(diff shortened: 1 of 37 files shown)


tlento:
* main
  * **PR #381** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/381" target="_blank">Graphite</a> 
    * **PR #389** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/389" target="_blank">Graphite</a> 
      * **PR #390** <a href="https://app.graphite.dev/github/pr/transform-data/metricflow/390" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/transform-data/metricflow/390?utm_source=stack-comment).

tlento:
> Was this just an oversight previously? Or were we actively expecting it in some circumstances?

@jzhu13 this was an oversight. In cases where the expression includes a subquery identifier I think it doesn't matter, but when there's an expression in the SELECT you must reference the alias in the GROUP BY. This is specific to BigQuery.

PR #392, opened 2023-01-05
Importing fuzzywuzzy causes warning due to non-installed levenshstein package. Suppress this by catching the warning during the import in query_parser because this package is not required.

jzhu13:
After digging around to find where it was triggered I discovered that the fuzzywuzzy.fuzz import in query_parser.py is what produces the levenshtein warning. Suppressed it using this warning capture but not sure if there are preferred other ways of doing so.

After making this change the original error (from running 'poetry run pytest  metricflow/test/model/validations/') no longer shows up.

Original error:

../../miniconda3/envs/mf3.9/lib/python3.9/site-packages/fuzzywuzzy/fuzz.py:11
/Users/jimmyzhu/miniconda3/envs/mf3.9/lib/python3.9/site-packages/fuzzywuzzy/fuzz.py:11: UserWarning: Using slow pure-python SequenceMatcher. Install python-Levenshtein to remove this warning
warnings.warn('Using slow pure-python SequenceMatcher. Install python-Levenshtein to remove this warning')

### DIFF
--- a/metricflow/query/query_parser.py
+++ b/metricflow/query/query_parser.py
@@ -6,7 +6,6 @@ import time
 from dataclasses import dataclass
 from typing import List, Optional, Tuple, Dict, Sequence
 
-import fuzzywuzzy.fuzz
 import fuzzywuzzy.process
 
 from metricflow.constraints.time_constraint import TimeRangeConstraint
@@ -42,6 +41,11 @@ from metricflow.time.time_granularity_solver import (
     PartialTimeDimensionSpec,
     RequestTimeGranularityException,
 )
+
+logging.captureWarnings(True)
+import fuzzywuzzy.fuzz  # noqa: E402
+
+logging.captureWarnings(False)
 
 logger = logging.getLogger(__name__)
 


metricflow/query/query_parser.py
```diff
@@ -6,7 +6,11 @@
 from dataclasses import dataclass
 from typing import List, Optional, Tuple, Dict, Sequence
 
-import fuzzywuzzy.fuzz
+import warnings
+
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore", category=UserWarning)
+    import fuzzywuzzy.fuzz
```
tlento, line 13: > Note: this can only be guaranteed in a single-threaded application. If two or more threads use the [catch_warnings](https://docs.python.org/3/library/warnings.html#warnings.catch_warnings) context manager at the same time, the behavior is undefined.

From https://docs.python.org/3/library/warnings.html

MetricFlow is not single-threaded, although in the CLI the query_parser is effectively used in exactly one place. I think this is still ok given what we need here, but I don't know if there are more robust alternatives available.

There's this:

https://docs.python.org/3/library/logging.html#logging.captureWarnings

Seems like maybe we could enable capture, suppress that one warning during the import, and then disable capture.

@plypaul did you have any other ideas in mind?

metricflow/query/query_parser.py
```diff
@@ -6,7 +6,11 @@
 from dataclasses import dataclass
 from typing import List, Optional, Tuple, Dict, Sequence
 
-import fuzzywuzzy.fuzz
+import warnings
+
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore", category=UserWarning)
```
tlento, line 12: Is it possible to filter this specific warning? If not that's fine, it's a nice-to-have. We don't maintain fuzzywuzzy.

metricflow/query/query_parser.py
jzhu13: > If two or more threads use the [catch_warnings](https://docs.python.org/3/library/warnings.html#warnings.catch_warnings) context manager at the same time, the behavior is undefined.

Yeah but this sounds exactly like a classic Python Logger thing to do.

metricflow/query/query_parser.py
jzhu13: I saw a warning capture that let you filter on the error message but it was only available as a method decorator (and I also now can no longer find an example of the thing I saw previously).

It seems like warnings.filterwarnings can only filter on the class of the warning, rather than the content of it.

jzhu13:
Updated to use the thread safe (seemingly? It doesn't say it's NOT thread safe) per Tom's suggestion above.

Warnings output for the levenshtein warning produced by fuzzy wuzzy is still suppressed in test runs.

tlento (approved):
Let's try this and see how it goes!

tlento:
> Let's try this and see how it goes!

You know, I should read check the actions output first. I think Flake8 is angry because of my suggestion. Sigh.

metricflow/query/query_parser.py
```diff
@@ -43,6 +42,11 @@
     RequestTimeGranularityException,
 )
 
+logging.captureWarnings(True)
+import fuzzywuzzy.fuzz  # noqa: D
```
tlento, line 46: I don't think this should be `noqa: D` as that's typically used for missing docstring errors (and it seems the D is interpreted as "all errors" which is very blah).  

I suspect this should be `# noqa: E402`

PR #393, opened 2023-01-05
Consolidate metric constraints in DataflowPlanBuilder

WilliamDee:
## Context
https://github.com/transform-data/metricflow/pull/352#discussion_r1049226384

## Changes
Essentially in the DataflowPlanBuilder, given a MetricSpec
1. we get the Metric object corresponding to that MetricSpec and combine the where constraint defined in the Metric
2. combine on the where constraint defined in the MetricSpec as well

Currently, it doesn't seem like there's a use case of doing both of these other than to cause slight confusion. At runtime when building the query, we just pass a list of MetricSpec to the builder that was initialized by the metric names only. So to remove the redundancy here, in the query parser we should attach the constraint to the MetricSpec that gets passed to the builder so that we don't have to retrieve the metric constraint from the Metric object directly downstream.

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -43,7 +43,6 @@ from metricflow.dataset.dataset import DataSet
 from metricflow.errors.errors import UnableToSatisfyQueryError
 from metricflow.model.objects.metric import MetricType, MetricTimeWindow
 from metricflow.model.semantic_model import SemanticModel
-from metricflow.model.spec_converters import WhereConstraintConverter
 from metricflow.object_utils import pformat_big_objects, assert_exactly_one_arg_set
 from metricflow.plan_conversion.column_resolver import DefaultColumnAssociationResolver
 from metricflow.plan_conversion.node_processor import PreDimensionJoinNodeProcessor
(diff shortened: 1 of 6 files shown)


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -243,6 +235,9 @@ def _build_metrics_output_node(
                         aggregated_measures_node=aggregated_measures_node,
                     )
                 )
+
+        assert len(compute_metrics_nodes) > 0, "ComputeMetricsNode was not properly constructed"
```
jzhu13, line 239: Should this assertion have existed before or is this now needed due to how the constraints are being bundled?

WilliamDee: just a sanity check that shouldve existed before, something wouldve went wrong if `len == 0`.

metricflow/query/query_parser.py
```diff
@@ -376,8 +377,26 @@ def _parse_and_validate_query(
                 metric_references=metric_references, time_dimension_specs=where_time_specs
             )
 
+        metric_specs: List[MetricSpec] = []
+        for metric_reference in base_metric_references:
```
tlento, line 381: Hmmm. It was a little hard to wrap my head around this.

The reason we do base_metric_references is because we only evaluate constraints on metrics at this point. The input_metrics to a DERIVED metric have the constraints pushed into the spec inside MetricSemantics.

It might be worth it to consolidate this further - right now we do top level spec construction in the query parser and upstream dependent spec construction inside MetricSemantics. But I also think that's likely opening up a can of worms, since MetricSemantics is in need of some restructuring itself.

So, I've convinced myself we should leave this here. But let's add a comment somewhere in this vicinity for why the base_metric_references are the things we care about here.

Also, I suggest initializing the local variable right above this block, it's not as important as putting it on line 233 makes it seem. We might even want to factor this all out into a helper method, then use the docstring to explain everything that's going on. I think I like that best.

tlento (approved):
Thanks, this looks less weird! I do think some added clarity around how we populate the spec constraint would be good to have before merge.

PR #395, opened 2023-01-06
Add approx percentiles

kyleli626:
Add the approximate flag in agg_params to allow for the usage of approximate percentile measures. 

- New percentile function type
- New renderers for every sql engine
- Sql engine flags for testing + engine capabilities - only some engines support approximate continuous + discrete percentiles
- integration tests for new approximate percentiles

<p>Percentile Availability Table</p>

  | Cont. | Disc. | Approx. Cont | Approx. Disc
-- | -- | -- | -- | --
Snowflake | [Y](https://docs.snowflake.com/en/sql-reference/functions/percentile_cont.html) | [Y](https://docs.snowflake.com/en/sql-reference/functions/percentile_disc.html) | [Y](https://docs.snowflake.com/en/sql-reference/functions/approx_percentile.html) (t-digest) | N
Bigquery | N (window) | N (window) | [Y](https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#approx_quantiles) | N
Databricks | [Y](https://docs.databricks.com/sql/language-manual/functions/percentile_cont.html) | [N](https://docs.databricks.com/sql/language-manual/functions/percentile_disc.html) | N | [Y](https://docs.databricks.com/sql/language-manual/functions/approx_percentile.html)
Redshift | [Y](https://docs.aws.amazon.com/redshift/latest/dg/r_PERCENTILE_CONT.html) | N (window) | N | [Y](https://docs.aws.amazon.com/redshift/latest/dg/r_APPROXIMATE_PERCENTILE_DISC.html)
[Postgres](https://www.postgresql.org/docs/9.4/functions-aggregate.html) | Y | Y | N | N
[DuckDB](https://duckdb.org/docs/sql/aggregates.html) | Y | Y | Y (t-digest) | N


metricflow/sql/render/databricks.py
```diff
@@ -0,0 +1,49 @@
...
+)
+from metricflow.sql.render.sql_plan_renderer import DefaultSqlQueryPlanRenderer
+from metricflow.sql.sql_exprs import SqlPercentileExpression, SqlPercentileFunctionType
+
+
+class DatabricksSqlExpressionRenderer(DefaultSqlExpressionRenderer):
+    """Expression renderer for the Databricks engine."""
+
+    def visit_percentile_expr(self, node: SqlPercentileExpression) -> SqlExpressionRenderResult:
+        """Render a percentile expression for Databricks. Add additional over() syntax for window."""
```
conglei, line 15: what do you mean by adding `over()` here? it seems it is not added ?

metricflow/sql/sql_exprs.py
```diff
@@ -849,18 +851,22 @@ def from_aggregation_parameters(agg_params: MeasureAggregationParameters) -> Sql
         """Given the measure parameters, returns a SqlPercentileExpressionArgument with the corresponding percentile args."""
         if not agg_params.percentile:
             raise RuntimeError("Percentile value is none - this should have been caught during model parsing.")
+        flags_to_function_type = {
```
conglei, line 854: This one is not easy to understand without seeing how it is used later. Can we use some type annotation to make it easier to understand?

conglei: I was thing something like this:

```python
UseDiscretePercentile = bool
UseApproximatePercentile = bool
flags_to_function_type: Map[Tuple[UseDiscretePercentile, UseApproximatePercentile], bool]
```

metricflow/model/transformations/convert_median.py
```diff
@@ -28,4 +28,5 @@ def transform_model(model: UserConfiguredModel) -> UserConfiguredModel:  # noqa:
                             )
                     measure.agg_params.percentile = MEDIAN_PERCENTILE
                     measure.agg_params.use_discrete_percentile = False
+                    # let's not set use_approximate_percentile to be false due to valid performance reasons
```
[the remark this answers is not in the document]
kyleli626: I don't think so - there's a valid use case where user wants fast median or exact median and they can specify. Setting this flag would also make it impossible to use median aggregation in some engines where it isn't supported.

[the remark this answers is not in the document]
conglei: ohh, that makes sense: we still allow users to specify the parameters.

metricflow/protocols/sql_client.py
```diff
@@ -197,7 +197,10 @@ class SqlEngineAttributes(Protocol):
     timestamp_type_supported: ClassVar[bool]
     timestamp_to_string_comparison_supported: ClassVar[bool]
     cancel_submitted_queries_supported: ClassVar[bool]
-    percentile_aggregation_supported: ClassVar[bool]
+    continuous_percentile_aggregation_supported: ClassVar[bool]
+    discrete_percentile_aggregation_supported: ClassVar[bool]
+    approximate_continuous_percentile_aggregation_supported: ClassVar[bool]
+    approximate_discrete_percentile_aggregation_supported: ClassVar[bool]
```
tlento, line 203: Ooof. I have a low key goal to eliminate these things altogether, but this extends the set immensely. For now I think it's ok, there isn't really a better way to do this with the way things are structured at the moment, but we need to rethink our approach to this before the divergence grows any further.

metricflow/model/transformations/convert_median.py
tlento, line 31: Do we need to assert that approx is set to something?

Also, we missed the validation on discrete - people shouldn't be asking for discrete medians, right?

metricflow/sql/render/databricks.py
```diff
@@ -0,0 +1,49 @@
...
+        params = arg_rendered.execution_parameters
+        percentile = node.percentile_args.percentile
+
+        if node.percentile_args.function_type is SqlPercentileFunctionType.CONTINUOUS:
+            function_str = "PERCENTILE_CONT"
+        elif node.percentile_args.function_type is SqlPercentileFunctionType.DISCRETE:
+            function_str = "PERCENTILE_DISC"
+        elif node.percentile_args.function_type is SqlPercentileFunctionType.APPROXIMATE_CONTINUOUS:
+            raise RuntimeError(
+                "Approximate continuous percentile aggregate not supported for Snowflake. Set use_discrete_percentile to true and/or use_approximate_percentile to false in all measures."
```
tlento, line 26: nit: copy/paste error Snowflake -> Databricks. Also same thing as last one.
```suggestion
                "Approximate continuous percentile aggregate not supported for Databricks. Set "
                "use_discrete_percentile to true and/or use_approximate_percentile to false in all percentile measures."
```

metricflow/sql/render/big_query.py
```diff
@@ -42,8 +44,26 @@ def render_group_by_expr(self, group_by_column: SqlSelectColumn) -> SqlExpressio
...
+            percentile = node.percentile_args.percentile
+
+            fraction = Fraction(percentile).limit_denominator()
+
+            return SqlExpressionRenderResult(
+                sql=f"APPROX_QUANTILES({arg_rendered.sql}, {fraction.denominator})[OFFSET({fraction.numerator})]",
+                execution_parameters=params,
+            )
+        raise RuntimeError(
+            "Only approximate continous percentile aggregations are supported for BigQuery. Set use_approximate_percentile and disable use_discrete_percentile in all measures"
```
tlento, line 65: Maybe this instead? It's a little more clear:

```suggestion
            "Only approximate continous percentile aggregations are supported for BigQuery. Set use_approximate_percentile and disable use_discrete_percentile in all percentile measures"
```

tlento: micro-nit: line length as well:
```suggestion
            "Only approximate continous percentile aggregations are supported for BigQuery. Set "
            "use_approximate_percentile and disable use_discrete_percentile in all percentile measures"
```

metricflow/sql/render/postgres.py
```diff
@@ -44,6 +47,32 @@ def visit_generate_uuid_expr(self, node: SqlGenerateUuidExpression) -> SqlExpres
...
+        elif node.percentile_args.function_type is SqlPercentileFunctionType.DISCRETE:
+            function_str = "PERCENTILE_DISC"
+        elif node.percentile_args.function_type is SqlPercentileFunctionType.APPROXIMATE_CONTINUOUS:
+            raise RuntimeError(
+                "Approximate continuous percentile aggregate not supported for Postgres. Set use_approximate_percentile to false in all measures."
+            )
+        elif node.percentile_args.function_type is SqlPercentileFunctionType.APPROXIMATE_DISCRETE:
+            raise RuntimeError(
+                "Approximate discrete percentile aggregate not supported for Postgres. Set use_approximate_percentile to false in all measures."
+            )
```
tlento, line 67: Same thing on this and all other error messages re: line length and increased clarity.

metricflow/sql/render/snowflake.py
```diff
@@ -17,6 +18,33 @@ def visit_generate_uuid_expr(self, node: SqlGenerateUuidExpression) -> SqlExpres
             execution_parameters=SqlBindParameters(),
         )
 
+    def visit_percentile_expr(self, node: SqlPercentileExpression) -> SqlExpressionRenderResult:
+        """Render a percentile expression for Snowflake. Add additional over() syntax for window."""
```
tlento, line 22: This `over` syntax comment looks like the same leftover from earlier work as in the Databricks renderer. Shall we remove it? Or should this be `WITHIN GROUP` instead of `over()`?

tlento (commented):
This implementation looks good to me, although I have a lightweight design question. Do we want users to be adding a flag, or is approx_percentile its own aggregation type? If the latter the boolean switching logic is simpler, we validate on agg type availability and it uses the same parameters as the regular percentile. 

Similarly they can do either MEDIAN or APPROXIMATE_MEDIAN and we'll transform it accordingly.

I think we also have some larger issues around that engine specific support struct but that's way beyond the scope of your work here. I'll have to think about it a bit more. Open to suggestions!

metricflow/sql/render/expr_renderer.py
```diff
@@ -136,15 +136,8 @@ def visit_function_expr(self, node: SqlAggregateFunctionExpression) -> SqlExpres
...
-        arg_rendered = self.render_sql_expr(node.order_by_arg)
-        params = arg_rendered.execution_parameters
-
-        function_str = node.percentile_args.function_name
-        percentile = node.percentile_args.percentile
-
-        return SqlExpressionRenderResult(
-            sql=f"{function_str}({percentile}) WITHIN GROUP (ORDER BY ({arg_rendered.sql}))",
-            execution_parameters=params,
+        raise RuntimeError(
```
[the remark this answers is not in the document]
kyleli626: unfortunately, a god number of tests rely on the default expression renderer, which wouldn't be able to exist if abstract. Is there an alternative way? Not sure if using the duckdb renderer or anything makes sense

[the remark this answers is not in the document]
tlento: Ah, bummer, let's leave this alone then. I suspect everything using the default renderer should in fact be using the engine-specific variants, but given all of the test instances using it I think that update is beyond the scope of this PR.

kyleli626:
@tlento 

Can't figure out to respond to the parent level comment. regarding making approximate percentile a different aggregation type, if we were to do that, I think discrete could also be a different aggregation type, so 4 different aggs. I don't think it simplifies too much in my opinion (like it shouldn't change how validation works i think), but either way works for me!

tlento:
> Can't figure out to respond to the parent level comment. regarding making approximate percentile a different aggregation type, if we were to do that, I think discrete could also be a different aggregation type, so 4 different aggs. I don't think it simplifies too much in my opinion (like it shouldn't change how validation works i think), but either way works for me!

Let's just leave it as is. In thinking about this a bit more I believe `percentile` as the measure aggregation with options for what kind of percentile computation makes the most sense. I don't really understand why engines have some window function for percentiles that can't be used as a regular aggregation and an approx_percentile function that can't be used in window functions, it's weird. We shouldn't reflect that forward to our users.

metricflow/model/validations/measures.py
```diff
@@ -455,14 +455,29 @@ def validate_model(model: UserConfiguredModel) -> List[ValidationIssueType]:  #
...
                         )
-                elif (
-                    measure.agg != AggregationType.PERCENTILE
-                    and measure.agg_params
-                    and (
-                        measure.agg_params.percentile
-                        or measure.agg_params.use_discrete_percentile
-                        or measure.agg_params.use_approximate_percentile
-                    )
+                elif measure.agg == AggregationType.MEDIAN:
```
tlento, line 458: With our current setup this will never happen but I think it's good to have this here as well. It might be useful for models pulled in from dbt configs, for example.

tlento (approved):
I think this is good to go, assuming all of the engine tests are passing. Thanks for doing this, and thanks for the thorough error messaging!

PR #398, opened 2023-01-11
Replace usage of moz_parser with mo_parser since it's deprecated and throwing warnings

jzhu13:
Title explains things, moz_parser is deprecated because the person who made it now works on: https://github.com/klahnakoski/mo-sql-parsing

Looks like the usage is identical based on documentation. 

All validation tests still pass.


metricflow/model/objects/constraints/where.py
```diff
@@ -4,7 +4,7 @@
 import re
 from typing import List, Optional, Dict, Any
 
-from moz_sql_parser import parse as moz_parse
+from mo_sql_parsing import parse as mo_parse
```
tlento, line 7: You need to update the pyproject.toml to reflect the dependency change and then run poetry update to reflect those changes in the poetry.lock file.

https://python-poetry.org/docs/basic-usage/#specifying-dependencies
https://python-poetry.org/docs/dependency-specification/
https://python-poetry.org/docs/cli/#update

tlento (approved):

PR #401, opened 2023-01-14
Make python logger swallow Levenshtein warnings

jzhu13:


metricflow/query/query_parser.py
```diff
@@ -44,7 +42,7 @@
 
 logging.captureWarnings(True)
 import fuzzywuzzy.fuzz  # noqa: E402
-
+import fuzzywuzzy.process  # noqa: E402
```
tlento, line 45: LOL. Is that why it didn't work?

tlento (approved):
You'll need to fix the format lint locally and update.

metricflow/query/query_parser.py
jzhu13: Yes apparently not putting both fuzzywuzzy imports into the capture caused it to still complain.

Why the capture is *actually* suppressing things...I'm not sure. Maybe Python Logger output is already suppressed during test runs? Like print statements.

PR #402, opened 2023-01-14
Add Databricks SQL Warehouse runtime to sql engine tests

tlento:
The existing SQL engine tests only exercised our Databricks
engine implementations as SQL queries against a Databricks
cluster configuration. However, it's more common for analytics
query workloads to target a Databricks SQL warehouse configuration.

Since we support both, we should test both.


courtneyholcomb (approved):
Perfect thank youuu!!

PR #405, opened 2023-01-17
Bug fix: error for restricted granularity

courtneyholcomb:
Previously, a bug was allowing some queries to succeed for multiple metrics with no overlapping granularities. This fixes that.

Example:
`bookings_cumulative` only supports `day` granularity
`bookings_monthly` only supports `month` granularity or larger
Querying for `bookings_cumulative` and `bookings_monthly` together with a granularity of `month` would succeed, though we would expect it to fail since querying `bookings_cumulative` alone with `month` granularity would fail.

### DIFF
--- a/metricflow/time/time_granularity_solver.py
+++ b/metricflow/time/time_granularity_solver.py
@@ -107,7 +107,7 @@ class TimeGranularitySolver:
         named 'ds' with granularity DAY.
 
         The 'monthly_bookings' measure is in defined in the 'fct_bookings_monthly' data source. 'fct_bookings_monthly'
-        has a local time dimension named 'ds' with granularity DAY.
+        has a local time dimension named 'ds' with granularity MONTH.
 
         Then this would return [DAY, MONTH].
         """
--- a/metricflow/time/time_granularity_solver.py
+++ b/metricflow/time/time_granularity_solver.py
@@ -156,14 +156,16 @@ class TimeGranularitySolver:
                 # If there is a cumulative metric, granularity changes aren't supported.
                 for metric_reference in metric_references:
                     metric = self._semantic_model.metric_semantics.get_metric(metric_reference)
-                    if (
-                        metric.type == MetricType.CUMULATIVE
-                        and time_dimension_spec.time_granularity != min_granularity_for_querying
-                    ):
-                        raise RequestTimeGranularityException(
-                            f"For querying cumulative metric '{metric_reference.element_name}', the granularity of "
-                            f"'{time_dimension_spec.qualified_name}' must be {min_granularity_for_querying.name}"
+                    if metric.type == MetricType.CUMULATIVE:
+                        _, only_queryable_granularity = self.local_dimension_granularity_range(
+                            metric_references=[metric_reference],
+                            local_time_dimension_reference=time_dimension_spec.reference,
                         )
+                        if time_dimension_spec.time_granularity != only_queryable_granularity:
+                            raise RequestTimeGranularityException(
+                                f"For querying cumulative metric '{metric_reference.element_name}', the granularity of "
+                                f"'{time_dimension_spec.qualified_name}' must be {only_queryable_granularity.name}"
+                            )
 
         # TODO: Validate non-local time dimension granularities here instead of during plan building.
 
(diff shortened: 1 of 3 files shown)


metricflow/time/time_granularity_solver.py
```diff
@@ -156,14 +156,16 @@ def validate_time_granularity(
...
-                        metric.type == MetricType.CUMULATIVE
-                        and time_dimension_spec.time_granularity != min_granularity_for_querying
-                    ):
-                        raise RequestTimeGranularityException(
-                            f"For querying cumulative metric '{metric_reference.element_name}', the granularity of "
-                            f"'{time_dimension_spec.qualified_name}' must be {min_granularity_for_querying.name}"
+                    if metric.type == MetricType.CUMULATIVE:
+                        _, only_queryable_granularity = self.local_dimension_granularity_range(
+                            metric_references=[metric_reference],
+                            local_time_dimension_reference=time_dimension_spec.reference,
```
tlento, line 162: Ooof, all of our granularity resolution stuff is impossibly complicated. :(

I don't know if this is works correctly in all cases. The time_dimension_spec could be pulled in from a join, in which case it's not clear to me what happens. To be honest, it's not clear to me what happens if we somehow have multiple "local" time_dimension_specs, either, and of course that's been sitting in this code path since we open sourced.

I think all of this logic was implicitly tied to the old model where everybody queried by a single primary time dimension which was consistent model-wide (essentially, `metric_time`). But I'd expect that to work correctly already - like if `metric_time` was in the wrong granularity we should throw an error here.

tlento (commented):
I think I need to try this out to see what's really going on. This is a very odd block of old logic, and I don't think I ever fully understood how it works....

dog astronaut has no idea

metricflow/time/time_granularity_solver.py
courtneyholcomb: There is some weirdness about how we get here for sure - for example, if I query in MetricFlow  `mf --metrics wau_rolling_7,gusto_survey_response_rate --dimensions metric_time__month`, I get the expected error `For querying cumulative metric 'wau_rolling_7', the granularity of 'metric_time__month' must be DAY`.
But if I run the pseudo-equivalent query in Transform `mql query --metrics wau_rolling_7,gusto_survey_response_rate --dimensions metric_time__month` I unexpectedly get a successful result. Working on figuring out why that difference occurs. (But this change fixes it on the Transform side.)

But the logic itself here seems flawed regardless - `local_dimension_granularity_range`  would return `[DAY, MONTH]` for the 2 metrics in question. We then treat `MONTH` as the minimum queryable granularity for BOTH metrics, even though `wau_rolling_7` only supports `DAY`. For cumulative metrics, we need to check their supported granularity individually.

tlento: Ok, I get it now. The solver is confusing because it doesn't really understand metric types. In the long run we'll need to address this, but for now I think this is the right fix. Let's do one of the following just to make it more obvious why this is here:

1. Add a comment that we have to do this again for cumulative metrics because the range method might return the wrong value
2. Factor the granularity range extraction for a single metric into its own method (instead of having it all inlined in `local_dimension_granularity_range`) and then assert against that here.

tlento (approved):
I figured out why this is so weird. Maybe. The problem here is this:

We do this granularity check against the union of all measures, meaning the minimum queryable granularity is the smallest common granularity value across all measures in all metrics. So if you have two metrics, one that is DAY and one that is MONTH, the minimum queryable granularity is MONTH, because that's the time granularity that works for both metrics.

However, if the DAY granularity metric is cumulative, it is incompatible with a MONTH setting for granularity, because cumulative metrics have a fixed valid granularity due to the window specification. So we encounter weird behavior in the specific case where the cumulative metric has a smaller granularity than one of the other metrics in the query.

The bug is that `local_dimension_granularity_range()` is oblivious to the nature of the individual metrics, and so it implicitly adjusts granularity for cumulative metrics if another metric is included in the input set from the query.

Whew. Thanks for finding and fixing this!

courtneyholcomb:
> because the range method might return the wrong value

Technically, I think the range method is actually returning the correct value. The range method is just designed to tell you the min and max granularities that are _specified in the configs_ for these metrics, and that's what it's doing. We just needed the additional logic to recognize that cumulative metrics don't care about that range.

I'll add that comment before merging!

plypaul (commented):
Thanks for the fix! Mind adding a test case in `test_time_granularity_solver.py`?

PR #406, opened 2023-01-18
Add workflow to measure validation test duration on PRs and main

jzhu13:
Chatted with Marco offline and he suggested a workflow be added for test runs to avoid local caching/pytest artifacts interfering with getting a stable read on what the proper baseline is for tests (after I was playing around with a local branch that changed our SQL parser and then completely changed the test duration baseline).

### DIFF
--- a/.github/workflows/ci-validation-duration.yaml
+++ b/.github/workflows/ci-validation-duration.yaml
@@ -1,0 +1,24 @@
+name: Measure Validation Test Duration
+
+on:
+  workflow_dispatch:
+
+jobs:
+  validation-tests:
+    name: Run validation tests and measure duration
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.9"
+
+      - name: Install Poetry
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-validation-duration.yaml
```diff
@@ -0,0 +1,34 @@
...
+  # run these jobs against the PR + master.  To run tests against just the PR, use "push" instead of "pull_request"
+  pull_request:
+    # run these jobs when a PR is opened, edited, reopened, or updated (synchronize)
+    # edited = title, body, or the base branch of the PR is modified
+    # synchronize = commit(s) pushed to the pull request
+    types:
+      - opened
+      - edited
+      - reopened
+      - synchronize
```
tlento, line 14: I suggest making this a workflow_dispatch only. Then you can run it once against your merge base commit (usually main) and once against your test commit and compare the output.

Another option is to run this on label, and add a custom label that you can use to trigger the action on any PR. It's only marginally easier than a manual trigger through the actions UI, so I don't think it's worth the overhead to get that set up.

```suggestion
```

QMalcolm (approved):
Thanks for putting this together. It will be nice to have a repeatable way to check durations to track how they change

PR #408, opened 2023-01-18
Derived Metric Offset: Use `INNER JOIN` + better testing

courtneyholcomb:
Updates `JoinToTimeSpineNode` to use `INNER JOIN` instead of `OUTER JOIN`.
This ensures the resulting dataset only ends up with rows where there is actually data for the metric being joined. Most derived metrics will use multiple metrics, in which case the `INNER JOIN` between those metrics will remove most of the time spine dataset. But when you only have one metric in a derived metric with offset, the `OUTER JOIN` causes the query to return the entire time spine dataset unnecessarily.

Additional testing changes:
- Simplifies test case expressions to minimize null results
- Adds more data to the `bookings` table fixture


metricflow/test/fixtures/model_yamls/simple_model/metrics.yaml
```diff
@@ -627,7 +627,7 @@ metric:
     - support@transformdata.io
   type: derived
   type_params:
-    expr: (bookings - bookings_2_weeks_ago) / bookings_2_weeks_ago
```
courtneyholcomb, line 630: Simplified these expressions so we would end up with a more complete set of results

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1349,7 +1349,7 @@ def visit_join_to_time_spine_node(self, node: JoinToTimeSpineNode[SourceDataSetT
                     col_ref=SqlColumnReference(table_alias=parent_alias, column_name=metric_time_dimension_column_name)
                 ),
             ),
-            join_type=SqlJoinType.LEFT_OUTER,  # TODO: test other join types
+            join_type=SqlJoinType.INNER,
```
WilliamDee, line 1352: shall we move the # Build join expression block to sql_join_builder.py? Sorry I missed this before, but Tom made some efforts to clean this stuff up by splitting out building join logics to that module.

courtneyholcomb: Sure!

WilliamDee (approved):

PR #409, opened 2023-01-19
New granularity syntax + {{ }} escaping

kyleli626:
Context: Replace use of dunders and start using "function" syntax for things like granularity

**Granularity**
old: mf query —metrics revenue —dimensions metric_time__week
new : mf query —metrics revenue —dimensions metric_time.granularity(week)

- add new function style parser class
- new from_name method for dimensions+constraints/linkable specs only for granularity
- tests

**New {{ }} syntax**
Context: Since our new granularity syntax can be invoked in arbitrary where constraint strings, we need some way to differentiate our syntax from the surrounding sql. For example, if we try 
` mf query --metrics transactions --dimensions customer__billing_country --where "metric_time.granularity(week) >= '2021-10-01'"`

Since we use the moz sql parser to parse the arbitrary where string and extract all the dimension and identifier references, we need a mechanism to separate these dimension/identifier references from the surrounding sql. Otherwise, it's not valid sql and doesn't parse, and the linkable name string, in this case 'metric_time.granularity(week)', isn't extractable. Thus, the new syntax would be 
` mf query --metrics transactions --dimensions customer__billing_country --where "{{metric_time.granularity(week)}} >= '2021-10-01'"`

Internally, when we parse a string segment surrounded by the double curly braces, we parse that into a StructuredLinkableName and use the qualified_name to replace it. This is because the where string is directly appended to the end of the sql query, and it needs to match the generated column name in the select statement. Thus, the where statement becomes valid. Inside the linkable_names, which dictate what dimensions/identifiers we might need or join with, we use the original string inside the braces and exclude the new qualified_name one. 

Thus "{{metric_time.granularity(week)}} >= '2021-10-01'"` is parsed to 
WhereClauseConstraint(where = "metric_time__week >= '2021-10-01'", linkable_names = [metric_time.granularity(week)]

### DIFF
--- a/metricflow/naming/linkable_spec_name.py
+++ b/metricflow/naming/linkable_spec_name.py
@@ -1,6 +1,9 @@
 from __future__ import annotations
+from collections import OrderedDict
 
 import logging
+import re
+import typing
 from dataclasses import dataclass
 from typing import Optional, Tuple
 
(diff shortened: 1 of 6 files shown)


metricflow/naming/linkable_spec_name.py
```diff
@@ -51,6 +99,21 @@ def from_name(qualified_name: str) -> StructuredLinkableSpecName:
         else:
             return StructuredLinkableSpecName(tuple(name_parts[:-1]), name_parts[-1])
 
+    @staticmethod
+    def from_name_v2(modified_name: str) -> StructuredLinkableSpecName:
+        """Construct from a name with newer syntax. e.g. listing__ds.granularity(month)."""
+
+        structured_modified_spec = StructuredModifiedSpecName.from_name(modified_name)
+        linked_name = structured_modified_spec.qualified_name.split(DUNDER)
```
conglei, line 107: Why do we still need to `split` ?

metricflow/test/query/test_query_parser.py
```diff
@@ -367,3 +378,35 @@ def test_derived_metric_with_offset_parsing(time_spine_source: TimeSpineSource)
...
+
+def test_query_parser_granularity_v2(time_spine_source: TimeSpineSource) -> None:  # noqa: D
+    bookings_yaml_file = YamlConfigFile(filepath="inline_for_test_1", contents=BOOKINGS_YAML)
+    query_parser = query_parser_from_yaml([bookings_yaml_file], time_spine_source)
+
+    with pytest.raises(RuntimeError):  # incorrect syntax
+        query_parser.parse_and_validate_query(
+            metric_names=["bookings"],
+            group_by_names=["is_instant", MTD + ".granularity(day"],
+            order=[MTD + ".granularity(day)"],
```
conglei, line 391: nit： 
```suggestion
            order=[f"{MTD}.granularity(day)"],
```

metricflow/naming/linkable_spec_name.py
kyleli626: there are other identifier links that still use the dunder, which we don't change, like user__country__created_on.granularity(week) for example.

conglei: gotcha! thanks for the clarification!

callum-mcdata:
@plypaul should this PR be closed out with the new WhereFilter implementation, right?

plypaul:
Yeah, closing this out as we recently put in templates in filters.

PR #411, opened 2023-01-20
Clean Up Logging indents and Bind Parameter Logging

plypaul:
This PR cleans up how logging indents are generated and also formats the bind parameters so they are more legible when there are many parameters.


tlento (approved):
Seems reasonable!

PR #412, opened 2023-01-25
Improve Terminology and Implementation of `TransformModel`

QMalcolm:
There are no longer any `pre` validation transformations or `post` validation transformations. There are however transformation rules which require other transformation rules to be run first. Additionally, we can imagine
there being more ordered levels or transformation rules than just two. This PR adds a new `transform` method to the `ModelTransformer` which takes ina tuple of rule sequences. These rule sequences are operated over in order, i.e.
the first sequence is run before the second sequence, and etc. This makes transformations more flexible going forward, resolves the naming with the actual state of things, and simlifies the majority of use cases involving `ModelTransformer`. Additionally, we have deprecated the `pre` and `post`functions, which will eventually be removed.

# Summary
* Adds new `ModelTransformer.transform` method
* Deprecates `ModelTransfomer` `.pre_validation_transform_rules` and `.post_validation_transform_rules` methods
* Switches use of `.pre_validation_transform_rules` and `.post_validation_transform_rules` throughout the code base to `.transform` accordingly
* BREAKING: Alters the call signatures of the following in `dir_to_model.py`
  - `parse_directory_of_yaml_files_to_model`
  - `parse_yaml_file_paths_to_model`
  - `parse_yaml_files_to_validation_ready_model`

### DIFF
--- a/metricflow/model/model_transformer.py
+++ b/metricflow/model/model_transformer.py
@@ -1,7 +1,7 @@
 import copy
 import logging
 
-from typing import Sequence
+from typing import Sequence, Tuple
 
 from metricflow.model.objects.user_configured_model import UserConfiguredModel
 from metricflow.model.transformations.add_input_metric_measures import AddInputMetricMeasuresRule
@@ -23,12 +23,12 @@ class ModelTransformer:
     Generally used to make it more convenient for the user to develop their model.
     """
 
-    DEFAULT_PRE_VALIDATION_RULES: Sequence[ModelTransformRule] = (
+    PRIMARY_RULES: Sequence[ModelTransformRule] = (
         LowerCaseNamesRule(),
         SetMeasureAggregationTimeDimensionRule(),
     )
 
-    DEFAULT_POST_VALIDATION_RULES: Sequence[ModelTransformRule] = (
+    SECONDARY_RULES: Sequence[ModelTransformRule] = (
         CreateProxyMeasureRule(),
         BooleanMeasureAggregationRule(),
         CompositeIdentifierExpressionRule(),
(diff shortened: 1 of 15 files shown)


QMalcolm:
The linting failure
```
metricflow/model/transformations/dbt_to_metricflow.py:112: error: Argument "target_model_name" to "resolve_ref" of "Manifest" has incompatible type "Optional[Any]"; expected "str"  [arg-type]
```
appears unrelated to this PR. Locally I had had `pre-commit 2.x` installed and not a version of `dbt-*` adapter. Upon upgrading my local `pre-commit` to `3.x` the lint error began raising locally. Then after installing `dbt-snowflake` locally, the linting error ceased. I believe the issue is that we need to install a `dbt-*` adapter during the linting workflow, but I attempted that in #413 which did not fix the problem. Note also that #413 is branched off of `main` and does not include the changes of this PR, which solidifies that this PR is not causing the linting issue.

QMalcolm:
https://github.com/transform-data/metricflow/pull/413 now resolves the issue.

metricflow/model/model_transformer.py
```diff
@@ -37,25 +37,49 @@ class ModelTransformer:
...
+        PRIMARY_RULES,
+        SECONDARY_RULES,
+    )
+
     @staticmethod
-    def pre_validation_transform_model(
-        model: UserConfiguredModel, rules: Sequence[ModelTransformRule] = DEFAULT_PRE_VALIDATION_RULES
+    def transform(
+        model: UserConfiguredModel,
+        ordered_rule_sequences: Tuple[Sequence[ModelTransformRule], ...] = DEFAULT_RULES,
```
WilliamDee, line 48: what's the benefit of having a tuple of sequences of rules rather than just a sequence of rules?

WilliamDee (approved):
nice!

metricflow/model/model_transformer.py
QMalcolm: There are a few benefits. one is of explicitness, one is of ease of ordering.

## Explicitness
Our transformation rules _require_ that some happen before or after one another. Specifically currently we have two rules which _must_ be run before any of our other current rules. There are two ways to do this. One way is a one dimension explicitly ordered list. I.e. a list wherein the first rule must be run before the second rule and so on. The other way is the way I've implemented in this PR, ranked/ordered groups of rules (primary rules, secondary rules, tertiary rules, and etc). I.e. primary rules have to be run before secondary rules, secondary rules before tertiary rules, and etc.

There is a difference in explicitness in the two solutions. With a flat ordered list you know the relation of one rule to the next. But how does a rule related to all of the other rules that come before and after it where is the dividing line? With an ordered sequence of non strictly ordered rule sequences it becomes very apparent. This plays into the second part, easy of updating

## Ease of updating
You create a new transform rule, where does it go? In the the flat ordered list you need to know where the dividing lines are of how rules relate. In the ordered sequence of non strictly ordered rule sequences it, the dividing lines are made apparent. Thus this implementation should make it easier for both future contributors and if people using MF as a package to add new/custom rules.

PR #414, opened 2023-01-26
Query interface change

kyleli626:
Context: As syntax for modifying identifiers and dimensions gets more complicated - adding modifiers like period over period, trim, granularity, etc. it becomes more cumbersome on the client side to render all these different attributes into a string. It would be more convenient for the query parser interface to use object input directly. 

This PR makes the query parser take in object input rather than string input. Creates GroupByInputSpec, MetricInputSpec, OrderByInputSpec classes which inherit from LinkableSpecName, and the relevant query parser functions that take in objects instead of strings. This pretty much means removing the .from_name methods everywhere.

### DIFF
--- a/metricflow/naming/linkable_spec_name.py
+++ b/metricflow/naming/linkable_spec_name.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import logging
 from dataclasses import dataclass
 from typing import Optional, Tuple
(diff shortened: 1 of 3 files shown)


tlento:
We've refactored all the things and it's not clear if this is the query interface we'll be using anymore, so I'll just close this out to keep our PR queue clean.

Hope all is well, Kyle!

PR #418, opened 2023-01-26
Stop returning the model from ModelValidator

QMalcolm:
Model validation rules do not alter the model. By extension the ModelValidator does not alter the model. The fact that the ModelValidator was returning the model was a carry over from a now long bygone era when model validators and model transformations were much more tightly coupled.  We've, for the better, gotten away from that. This now vestigial return value's existence can muddle the mental model of how the ModelValidator works for people learning about it. Thus this change cleans this no longer needed functionality. Indeed it's desirable to stop returning the model from the ModelValidator because we want to cement the idea that the ModelValidator does not and should not modify the model.


tlento (commented):
I'm back!

Please request review if this is still relevant. Thanks!

QMalcolm:
Gonna rebase off main to resolve conflicts

tlento (approved):
YAY! That was always so confusing for me...

PR #419, opened 2023-01-26
Ensure consistent sort order for order-dependent integration test cases

tlento:
Recent changes added some values to our dummy test data which
produced multiple correct sort orders for output rows in the
itest_order_limit test cases. Although this has not yet come up in
our CI we have seen intermittent output order failures on these tests
in other runtime contexts, so adding explicit ordering
directives are necessary to ensure test robustness.

This commit makes a similar update to the composite_key_order integration
test to prevent a future recurrence of this issue.

This does not affect other tests as they rely on
comparisons of unordered output.


tlento:
Note - the failing check is addressed by #421

jzhu13 (approved):

PR #420, opened 2023-01-26
Replace uses of `ValidationIssueType` with `ValidationIssue`

QMalcolm:
Our different validation issue classes (`ValidationWarning`, `ValidationFutureError`, `ValidationError`) used to be entirely separate classes (i.e. didn't inherit from a parent class). However, they now all inherit from `ValidationIssue`. Before this we needed the `ValidationIssueType` type for saying something could contain any of those separate classed. However, since we now have `ValidationIssue` as the parent class, we should be able to replace `ValidationIssueType` with it. This will make typing easier.


tlento (commented):
I'm back!

Please request review if this is still relevant. Thanks!

QMalcolm:
Gonna rebase off main to resolve conflicts

tlento (approved):

PR #423, opened 2023-01-30
Clean up object_utils.py because it does a bunch of unrelated things (bag of crap)

jzhu13:


metricflow/sql/sql_column_type.py
```diff
@@ -0,0 +1,5 @@
+import datetime
+from typing import Union
+
+
+SqlColumnType = Union[str, int, float, datetime.datetime, datetime.date, bool]
```
tlento, line 5: We might consider putting this into the sql_bind_parameters file

metricflow/object_utils.py
```diff
@@ -1,15 +1,14 @@
...
 import textwrap
 from collections.abc import Mapping
 from dataclasses import is_dataclass, fields
-import datetime
 from hashlib import sha1
-from typing import Sequence, TypeVar, Tuple, Union
+from typing import Sequence
 
 from metricflow.model.objects.base import HashableBaseModel
+from metricflow.sql.sql_column_type import SqlColumnType
```
tlento, line 11: I don't think we want this here, right? But of course then I don't know where hash_items should live. Hmmm.

That will get easier after #444 is merged, so maybe we just leave this here for now.

tlento (approved):
You know me, I love this kind of stuff. No more bags of crap!

I have a bunch of quibbles inline but this is mostly good to merge for now pending the commit level cleanups and maybe some added movement on the hash - @plypaul is removing the definition_hash method from the Metric object in #444 so that makes things easier.

If you'd like to keep going, I'd say do it in a follow up PR and get this one merged, since it would be helpful for some of the work we're starting on dbt integration. 

Meanwhile, @QMalcolm can weigh in and help you shepherd this through as one last gift to us. :)

I'm approving here assuming updating the PR as re-quested and re-running tests doesn't break anything, but @QMalcolm is your reviewer if you need him to look in on something first.

metricflow/sql/sql_column_type.py
plypaul: Tried this, but it causes a circular import.

plypaul:
Fixed the merge conflicts and made some rebase fixes, but otherwise going to leave @jzhu13's commits alone and merge to unblock the other PRs.

plypaul:
@jzhu13 also had signed the CLA

Issue #416, opened 2023-01-26
Support for dbt v1.4.0


callum-mcdata:
**Describe the Feature**
dbt-core v1.4.0 released this week and made some changes under the hood around classes/constructs that are referenced in MetricFlow's dbt integration. 

### What changed that is relevant to MetricFlow
**Consolidating Classes:**
Previously, dbt had the concept of Parsed & Compiled nodes that had significiant overlap in the codebase. We cleaned this up under the hood and combined them into a single construct just called Nodes. Why is this relevant? In the `dbt_to_metricflow` [file](https://github.com/transform-data/metricflow/blob/d5dd7e916adf7d1865cf820d9357c8440721b287/metricflow/model/transformations/dbt_to_metricflow.py#L6) on line 6 you import `ParsedMetric` and `ParsedModelNode` - these have been renamed to `Metric` and `Model` respectively and now come from [this file.](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/contracts/graph/nodes.py)

**Metrics & Timestamps**
We made the `timestamp` property in a dbt metric an optional input. This [line](https://github.com/transform-data/metricflow/blob/d5dd7e916adf7d1865cf820d9357c8440721b287/metricflow/model/transformations/dbt_to_metricflow.py#L158) appears to define the timestamp so it would just add an empty string. Pretty sure that doesn't break anything (including the filter logic a[ few lines down](https://github.com/transform-data/metricflow/blob/d5dd7e916adf7d1865cf820d9357c8440721b287/metricflow/model/transformations/dbt_to_metricflow.py#L167)) but wanted to flag.

**Anything Else?**
I think those are the only ones that we'd want to call out!

callum-mcdata:
@QMalcolm not sure if this is a new issue or part of this one but I noticed something when testing out the new `derived` functionality. There was an issue where one of the component metrics was returning NULL (issue with the base dataset that I've since resolved) and that was causing the derived metric to return NULL instead of the other values. In the below example, `net_profit` is equal to `revenue - (costs + taxes)`

<img width="500" alt="image" src="https://user-images.githubusercontent.com/101437052/215144678-8eb6ba6c-3e59-4c5b-bcc7-7115531178e8.png">

[...]

PR #424, opened 2023-01-31
dbt 1.4.0 support

QMalcolm:
This PR is to enable `dbt 1.4.0` support (issue #416). Notably as part of the upgrade to `dbt 1.4.0`, we had to upgrade our `google-cloud-bigquery` dependency to `3.x`. Ideally the upgrade of `google-cloud-bigquery` would be part of a separate PR, so depending on what people think, we may delay merging this until that has happened (and then we'll rebase this appropriately). Mostly the changes of this PR are some import changes from dbt as some classes got renamed/moved. Additionally the `timestamp` property of a dbt metric is now optional in `dbt 1.4.0`, so we had to put some protections around this. These protections do mean that it is possible that a data source may be built without a timestamp, which is a problem. The plan though is to shortly refactor `dbt_to_metricflow` to use the `DbtConverter` which will resolve that issue.


tlento (approved):
Thanks!

PR #425, opened 2023-01-31
Enable different granularities for metrics with time offset

courtneyholcomb:
With the current placement of the `JoinToTimeSpineNode` in the dataflow plan, time granularity aggregation happens before joining to time spine, which means queries with any granularity besides `DAY` don't return expected results. Here, we move the time spine join to happen before granularity aggregation.
For cumulative metrics, we also move the `FilterElementsNode` to after the `JoinOverTimeRange` node to ensure we keep all columns needed when joining to time spine.


metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -197,34 +197,15 @@ def _make_time_spine_data_set(
...
-                    select_columns=(
-                        SqlSelectColumn(
-                            expr=SqlColumnReferenceExpression(
-                                SqlColumnReference(
-                                    table_alias=time_spine_table_alias,
-                                    column_name=time_spine_source.time_column_name,
-                                ),
-                            ),
-                            column_alias=metric_time_dimension_column_name,
+            select_columns = (
```
courtneyholcomb, line 200: Removed some redundant code here

PR #426, opened 2023-02-03
Enable different granularities for metrics with time offset

courtneyholcomb:
With the current placement of the `JoinToTimeSpineNode` in the dataflow plan, we filter down to time columns with the requested granularity BEFORE joining to time spine. Since joining to time spine expects an unaggregated time column, this means queries with any granularity besides `DAY` don't return expected results. Here, we move the time spine join to happen before time column aggregation.
For cumulative metrics, we also needed to change the order of the dataflow plan, from `FilterElementsNode -> JoinOverTimeRangeNode` (both required) to `JoinOverTimeRangeNode (optional, only used if a time dimension is requested) -> JoinToTimeSpineNode (optional, only used for offset metrics) -> FilterElementsNode`. This ensures we'll have the columns needed when joining to time spine.
Unrelated to time offset, this change also removes the unnecessary `JoinOverTimeRangeNode` step for a cumulative metric queried without a time dimension, which 1) does nothing in SQL and 2) might result in inaccurate costing for that query.


metricflow/test/fixtures/model_yamls/simple_model/metrics.yaml
```diff
@@ -209,7 +209,7 @@ metric:
       - txn_revenue
 ---
 metric:
-  name: "every_two_days_bookers" # because the bookings test data only spans 3 days
```
courtneyholcomb, line 212: No longer true

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -197,34 +197,15 @@ def _make_time_spine_data_set(
 
         # If the requested granularity is the same as the granularity of the spine, do a direct select.
         if metric_time_dimension_instance.spec.time_granularity == time_spine_source.time_column_granularity:
-            return SqlDataSet(
```
courtneyholcomb, line 200: Removed some redundant code here

plypaul (changes requested):
As per discussion, there are some uncertainties on how the feature should work (from a product perspective), and we haven't prioritized a resolution. Since I'm using github to manage my review queue, marking this is `Request Changes` so that it's not on my list.

courtneyholcomb:
PR outdated; new one here https://github.com/dbt-labs/metricflow/pull/726

PR #435, opened 2023-03-09
Add ability for the data warehouse model validator to check metric query return values

QMalcolm:
If we actually want to do this, we might want to take this a step further by sub-classing the available data warehouse model validator task types, and calling different task handling functions in `run_tasks` such that those functions do the necessary work and task type specific issue generation handling.


tlento (changes requested):
I like the concept! Let's hold off on this until we've figured out exactly how the warehouse validations will run post-integration.

tlento:
I'm going to close this out just to tidy up our open PR list. I think we'll be looking to remove our pandas dependencies over the course of time, so whatever implementation we go with will almost certainly be protocol-based with a corresponding external API of some kind.

PR #441, opened 2023-03-24
Upgrade cryptography library

WilliamDee:
## Context
Updating cryptography library to patch vulnerability: https://security.snyk.io/vuln/SNYK-PYTHON-CRYPTOGRAPHY-3316038

## Other library that needed to be upgraded
- `python-snowflake-connector`: `2.7.12 -> 3.0.2`
- `dbt-snowflake`: `1.4.0 -> 1.4.2`
- `snowflake-sqlalchemy`: `1.4.4 -> 1.4.7`


WilliamDee:
also had to bump `pyopenssl`: `22.0.0 -> 23.1.0`

pyproject.toml
```diff
@@ -52,6 +52,8 @@ dbt-redshift = {version="^1.4.0", optional=true}
 dbt-postgres = {version="^1.4.0", optional=true}
 dbt-bigquery = {version="^1.4.0", optional=true}
 dbt-metadata-client = {version="^0.1.0", optional=true}
+cryptography = "^39.0.1"
+pyopenssl = "^23.0.0"
```
tlento, line 56: OK, this buys us time to update the things that depend on these things, so let's do it. Could you add a comment to indicate that we don't need these, they're just included to ensure we stay ahead of vulnerabilities in transitive package installations?

I think the reason this works is because the cryptography version we were installing was pinned to an optional dependency of some package.

tlento (approved):
🚢 🔒 

Thanks @WilliamDee !

PR #444, opened 2023-04-06
Clean up DataSourceSemantics and MetricSemantics.

plypaul:
This PR cleans up `DataSourceSemantics` and `MetricSemantics` to

* Remove functionality that is no longer used. 
* Move from using strings to look up data sources to using `DataSourceReference`.
* Change signatures to use `Sequence` instead of `List` or `Tuple` to follow principle of using the most general interfaces in the signatures. `Iterable` or something else may be more apt, however.

Cleanup of some of the internal data structures will be handled in another PR.

### DIFF
--- a/metricflow/model/semantics/data_source_semantics.py
+++ b/metricflow/model/semantics/data_source_semantics.py
@@ -135,39 +135,23 @@ class DataSourceSemantics(DataSourceSemanticsAccessor):
     def get_by_reference(self, data_source_reference: DataSourceReference) -> Optional[DataSource]:  # noqa: D
         return self._data_source_reference_to_data_source.get(data_source_reference)
 
-    def _add_data_source(
-        self,
-        data_source: DataSource,
-        fail_on_error: bool = True,
-        logging_context: str = "",
-    ) -> None:
+    def _add_data_source(self, data_source: DataSource) -> None:
         """Add data source semantic information, validating consistency with existing data sources."""
         errors = []
 
-        if data_source.name in self._data_source_names:
-            errors.append(f"name {data_source.name} already registered - please ensure data source names are unique")
+        if data_source.reference in self._data_source_reference_to_data_source:
+            errors.append(f"Data source {data_source.reference} already added.")
 
         for measure in data_source.measures:
             if measure.reference in self._measure_aggs and self._measure_aggs[measure.reference] != measure.agg:
                 errors.append(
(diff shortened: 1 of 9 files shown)


metricflow/model/semantics/data_source_semantics.py
```diff
@@ -26,7 +26,7 @@
 logger = logging.getLogger(__name__)
 
 
-class DataSourceSemantics:
+class DataSourceSemantics(DataSourceSemanticsAccessor):
```
tlento, line 29: It's best not to do this, because this makes DataSourceSemantics inherit all properties from the protocol, which means anything that isn't tagged as an `@abstractmethod` in the protocol is automatically "correct" as far as the typechecker is concerned.

Today this makes no practical difference with regard to typechecking behavior, but if anybody adds a bare property without a default value to the protocol the typechecker will pass and anything accessing that property may fail in the runtime.

metricflow/model/semantics/metric_semantics.py
```diff
@@ -1,35 +1,31 @@
...
 from metricflow.references import MetricReference
 from metricflow.specs import MetricSpec, LinkableInstanceSpec, MetricInputMeasureSpec, MeasureSpec
-from metricflow.model.semantics.data_source_join_evaluator import MAX_JOIN_HOPS
-
 
 logger = logging.getLogger(__name__)
 
 
-class MetricSemantics:  # noqa: D
+class MetricSemantics(MetricSemanticsAccessor):  # noqa: D
```
tlento, line 19: Same thing about protocol inheritance from within a concrete implementation - seems it's best avoided

tlento (commented):
The first commit is really hard to read. Is it too much to ask that this be split into a few more? Like:

1. Delete unused metric_semantics.valid_hashes method
2. Remove configured data source container
3. Convert data source names to data source references

That'd be really fast and easy to review. Might be kind of a hassle to split now, sadly.

metricflow/model/semantics/data_source_semantics.py
plypaul: Changed this to be an ABC. Without it having something inherited, it was hard to tell in the IDE which functions needed to be updated in `DataSourceSemantics` when  `DataSourceSemanticsAccessor` changed.

tlento (approved):
Thanks so much for splitting things out, this was a much easier review!

I am frustrated at the lack of tooling support for protocols. It's super irritating.

PR #445, opened 2023-04-07
Removing Materializations

callum-mcdata:
### Materializations
This PR removes materializations. Originally it was done in three phases:
- Removing materialization CLI commands
- Removing the tests that failed with those CLI commands gone (probably should have been grouped with the first one)
- Removing the materialization object and all its downstream uses

Unfortunately I ran into some issue with graphite because I was working on a forked version of metricflow. When trying to get it back onto this repo I ran into some real issues and just decided to fall back onto the old PR. In the future though, I've got graphite all set up and should be good to go for review and stacked commits!


QMalcolm (approved):
Thank you for being so thorough! This looks great 🙂 after reading through the changes I pull down the branch and double checked for any references to materializations, and the only thing I found was a mention in the change log (and that seems appropriate to stay). Can't wait to get this merged as I think removing materializations moves the cutover to `dbt-semantic-interfaces` forward dramatically (a bunch of the tangled code was in materializations).

metricflow/test/model/validations/test_configurable_rules.py
```diff
@@ -1,43 +0,0 @@
...
-            )
-        ],
-    )
-
-    # confirm that with the default configuration, an issue is raised
-    issues = ModelValidator().validate_model(model).issues
-    assert len(issues.all_issues) == 1, f"ModelValidator with default rules had unexpected number of issues {issues}"
-
-    # confirm that a custom configuration excluding ValidMaterializationRule, no issue is raised
-    rules = [rule for rule in ModelValidator.DEFAULT_RULES if rule.__class__ is not ValidMaterializationRule]
```
QMalcolm, line 30: I think this can be changed to the following
```
    model = copy.deepcopy(simple_model__with_primary_transforms)
    model.metrics.append(metric_with_guaranteed_meta(
        name='metric_doesnt_exist_squared',
        type=MetricType.DERIVED,
        type_params=MetricTypeParams(
            expr='metric_doesnt_exist * metric_doesnt_exist',
            metrics=[MetricInput(name='metric_doesnt_exist')]
        ),
    ))

    # confirm that with the default configuration, an issue is raised
    issues = ModelValidator().validate_model(model).issues
    assert len(issues.all_issues) == 1, f"ModelValidator with default rules had unexpected number of issues {issues}"

    # confirm that a custom configuration excluding ValidMaterializationRule, no issue is raised
    rules = [rule for rule in ModelValidator.DEFAULT_RULES if rule.__class__ is not DerivedMetricRule]
```

QMalcolm (changes requested):
I  was a dumbdumb and by the time I reviewed  `test_configurable_rules.py` my brain was on autopilot 🙃

metricflow/test/model/validations/test_configurable_rules.py
callum-mcdata: Added back!

metricflow/test/model/validations/test_configurable_rules.py
```diff
@@ -1,43 +0,0 @@
-import pytest
```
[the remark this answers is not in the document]
callum-mcdata: Made changes based on your suggestions, let me know if that resolves the issue!

metricflow/test/model/validations/test_configurable_rules.py
QMalcolm: Thank you! ❤️

PR #451, opened 2023-04-19
BUG: Input Metric's original constraint is not being propogated

WilliamDee:
## Context
In derived metric, there are multiple layers of constraints.
1. The constraint in the metric used as an input metric
2. The input_metric constraint defined in the derived metric definition

`(1)` is not being respected so if one were to define a derived metric as so,
```yaml
metric:
  name: "instant_bookings"
  description: "bookings metric"
  type: measure_proxy
  type_params:
    measures:
      - bookings
  constraint: is_instant
---
metric:
  name: "double_instant_bookings"
  description: "2x bookings metric"
  type: derived
  type_params:
    expr: instant_bookings * 2
    metrics:
      - name: instant_bookings
```
The `constraint: is_instant` clause would be missing in the metric `double_instant_bookings`. This PR fixes this bug.

### DIFF
--- a/metricflow/model/semantics/metric_semantics.py
+++ b/metricflow/model/semantics/metric_semantics.py
@@ -128,7 +128,10 @@ class MetricSemantics(MetricSemanticsAccessor):  # noqa: D
         input_metric_specs: List[MetricSpec] = []
 
         for input_metric in metric.input_metrics:
-            spec_constraint = (
+            original_metric_obj = self.get_metric(input_metric.as_reference)
+
+            # This is the constraint parameter added to the input metric in the derived metric definition
+            combined_spec_constraint = (
                 WhereConstraintConverter.convert_to_spec_where_constraint(
                     data_source_semantics=self._data_source_semantics,
                     where_constraint=input_metric.constraint,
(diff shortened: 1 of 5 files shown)


metricflow/test/integration/test_cases/itest_metrics.yaml
```diff
@@ -803,4 +803,70 @@ integration_test:
...
+      INNER JOIN (
+        SELECT
+          SUM(booking_value) AS booking_value
+          , ds AS metric_time
+        FROM {{ source_schema }}.fct_bookings
+        GROUP BY
+          ds
+      ) b
+      ON a.metric_time = b.metric_time
+    )
```
tlento, line 872: This needs an alias

tlento (commented):
Seems reasonable! Test needs a fix first.

Would you mind squashing this into a single commit either here or on merge? I don't think we need to add the tests separately (and it'll get rid of the empty commit you added for clabot).

tlento (commented):
One more thought, sorry, off my game. I wish there was a better way to make it more obvious why we need both of these things and why they're different, but I can't come up with it right now so comments will have to do.

metricflow/model/semantics/metric_semantics.py
```diff
@@ -128,17 +128,33 @@ def metric_input_specs_for_metric(self, metric_reference: MetricReference) -> Se
         input_metric_specs: List[MetricSpec] = []
 
         for input_metric in metric.input_metrics:
-            spec_constraint = (
+            input_metric_obj = self.get_metric(input_metric.as_reference)
```
tlento, line 131: nit: maybe we call this `original_metric_obj` to match the `original_metric_constraint` naming below? I guess I should've noticed this the first time....

tlento (approved):

metricflow/model/semantics/metric_semantics.py
WilliamDee: done!

Issue #455, opened 2023-04-19
Convert aggregation_properties to dbt_semantic_interfaces


callum-mcdata:
**Describe the Feature**
This issue entails moving all the aggregation properties into their new home.

PR #460, opened 2023-04-24
Moving AggregationType enum into dbt-semantic-interfaces

QMalcolm:
Resolves: #455

### DIFF
--- a/dbt_semantic_interfaces/dbt_semantic_interfaces/objects/aggregation_type.py
+++ b/dbt_semantic_interfaces/dbt_semantic_interfaces/objects/aggregation_type.py
@@ -1,7 +1,7 @@
-from metricflow.object_utils import ExtendedEnum
+from enum import Enum
 
 
-class AggregationType(ExtendedEnum):
+class AggregationType(Enum):
     """Aggregation methods for measures"""
 
     SUM = "sum"
--- a/dbt_semantic_interfaces/dbt_semantic_interfaces/objects/aggregation_type.py
+++ b/dbt_semantic_interfaces/dbt_semantic_interfaces/objects/aggregation_type.py
@@ -1,0 +1,17 @@
+from metricflow.object_utils import ExtendedEnum
+
+
+class AggregationType(ExtendedEnum):
+    """Aggregation methods for measures"""
+
+    SUM = "sum"
+    MIN = "min"
+    MAX = "max"
+    COUNT_DISTINCT = "count_distinct"
+    # BOOLEAN is deprecated. Remove when people have migrated.
+    BOOLEAN = "boolean"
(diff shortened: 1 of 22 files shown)


QMalcolm:
* main
  * **PR #460** <a href="https://github.com/dbt-labs/metricflow/pull/460" target="_blank">Graphite</a>  👈
    * **PR #461** <a href="https://github.com/dbt-labs/metricflow/pull/461" target="_blank">Graphite</a>

dbt_semantic_interfaces/dbt_semantic_interfaces/objects/aggregation_type.py
```diff
@@ -0,0 +1,17 @@
+from metricflow.object_utils import ExtendedEnum
+
+
+class AggregationType(ExtendedEnum):
```
tlento, line 4: I think we can make this extend Python's builtin Enum. The ExtendedEnum doesn't do that much for us.

If we do need it, we can put it in the interfaces package, but I suspect we will not.

metricflow/aggregation_properties.py
```diff
@@ -2,77 +2,60 @@
...
-        Currently, this means:
-        1. The dimensions we care about most are the ones with the highest numeric values
-        2. We can calculate the "other" column in the postprocessor (meaning the metric is expansive)
-        """
-        return self in (AggregationType.SUM, AggregationType.SUM_BOOLEAN, AggregationType.COUNT)
+from dbt_semantic_interfaces.objects.aggregation_type import AggregationType
+from metricflow.object_utils import assert_values_exhausted
+
+
+def is_expansive(agg_type: AggregationType) -> bool:
```
tlento, line 9: I have a slight preference for keeping these inside of a class, but I guess it doesn't matter because Python gonna Python.

metricflow/aggregation_properties.py
```diff
@@ -2,77 +2,60 @@
...
-    @property
-    def can_limit_dimension_values(self) -> bool:
-        """Indicates if we can limit dimension values in charts.
-
-        Currently, this means:
-        1. The dimensions we care about most are the ones with the highest numeric values
-        2. We can calculate the "other" column in the postprocessor (meaning the metric is expansive)
-        """
-        return self in (AggregationType.SUM, AggregationType.SUM_BOOLEAN, AggregationType.COUNT)
+from dbt_semantic_interfaces.objects.aggregation_type import AggregationType
```
tlento, line 5: Oh wow, none of this is used anywhere. These methods were used in the postprocessor in Transform, but we don't need to support that at this time. I guess we can leave this here for now - if we don't need it any time soon we can delete it separately.

tlento (approved):
LGTM! :shipit:

PR #467, opened 2023-04-26
Add Support for Testing Table Snapshots

plypaul:
This PR adds support for a new way of defining tables that are used in tests that interact with the SQL engine. The new approach uses configuration files (instead of code) so that they can be more easily defined and manipulated. Only table snapshots for the simple model are migrated in this PR - snapshots for other models will be migrated in subsequent PRs after this is merged.


metricflow/test/fixtures/source_table_snapshots/simple_model/dim_listings.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - name: country
+      type: STRING
+    - name: capacity
+      type: INT
+    - name: is_lux
+      type: BOOLEAN
+    - name: user_id
+      type: STRING
+  rows:
+    - ["2020-01-01", "2020-01-02", "l3141592", "us", "2", "True", "u0004114"]
```
tlento, line 20: How did you move all of these? Manually?

.changes/unreleased/Under the Hood-20230425-180020.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Under the Hood
+body: Added Support for Testing Table Snapshots
+time: 2023-04-25T18:00:20.57818-07:00
+custom:
+  Author: plypaul
+  Issue: N/A
```
tlento, line 6: Let's just open issues and link them, it's what the tool demands. Also, it's a better place for discussion.

The other option is to add the label to skip changelog - for something like this that might be better, since it doesn't matter much to community members. 

Or you can skip everything in the stack until the last changelog entry.

metricflow/test/table_snapshot/table_snapshots.py
```diff
@@ -0,0 +1,240 @@
...
+    """A hash that can be used to compare whether a SQL table snapshot has changed."""
+
+    str_value: str
+
+    @staticmethod
+    def create_from_hashes(hashes: Sequence[SqlTableSnapshotHash]) -> SqlTableSnapshotHash:  # noqa: D
+        return SqlTableSnapshotHash(hash_items(tuple(one_hash.str_value for one_hash in hashes)))
+
+
+class SqlTableColumnType(Enum):  # noqa: D
```
tlento, line 35: How many sql type enums do we have now? Three?

I don't know if we should accumulate more calls for the others, so maybe this is ok, but it does pop up quite often.

tlento (commented):
I've been thinking about this a bit - while we're moving all of this stuff from code to config, do we want to make these a dbt project? I need to think more about what that looks like in CI, but all of these integration tests execute queries against models targeted at given warehouses. If our test package built the dbt + metricflow setup and we invoked the package API, we could have an end to end test suite using the semantic manifest from the dbt project, targeted at whatever warehouse we were pointing to.

I don't know if we can run the same kinds of assertions from within that test framework, but it seems like it's worth considering.

metricflow/test/fixtures/source_table_snapshots/simple_model/dim_listings.yaml
plypaul: I wrote a script that takes care of the bulk of the work in migrating them.

.changes/unreleased/Under the Hood-20230425-180020.yaml
plypaul: Yeah, skipping sounds good for this set.

plypaul (commented):
> I've been thinking about this a bit - while we're moving all of this stuff from code to config, do we want to make these a dbt project? I need to think more about what that looks like in CI, but all of these integration tests execute queries against models targeted at given warehouses. If our test package built the dbt + metricflow setup and we invoked the package API, we could have an end to end test suite using the semantic manifest from the dbt project, targeted at whatever warehouse we were pointing to.

Yeah, it would be beneficial to consolidate the structure w/ dbt. Once we migrate to configuration files though, it's easy to dump to another format, so I'm thinking that we would wait until things are more settled to do that.

metricflow/test/table_snapshot/table_snapshots.py
plypaul: What was the other one?

tlento: There's one in inference and another that has something to do with bind parameters.

plypaul: They're not quite equivalent use cases, so need to have a think about how to consolidate those.

tlento:
Per discussion, let's proceed with this for now and then we can cut over to using dbt at a later time if that turns out to be the most reasonable path forward.

Your test failures appear to be legit - something's different about the SQL as a result of this change.

plypaul:
@tlento - fix the test failure. It was due to an error in one of the snapshots.

metricflow/test/table_snapshot/table_snapshots.py
tlento: Yeah, it's mostly annoying. No big deal.

metricflow/test/table_snapshot/table_snapshots.py
```diff
@@ -0,0 +1,233 @@
...
+                elif column_type is SqlTableColumnType.STRING:
+                    type_converted_row.append(column_value)
+                elif column_type is SqlTableColumnType.TIME:
+                    type_converted_row.append(dateutil.parser.parse(column_value))
+                elif column_type is SqlTableColumnType.INT:
+                    type_converted_row.append(int(column_value))
+                elif column_type is SqlTableColumnType.FLOAT:
+                    type_converted_row.append(float(column_value))
+                elif column_type is SqlTableColumnType.BOOLEAN:
+                    type_converted_row.append(SqlTableSnapshot._parse_bool_str(column_value))
```
tlento, line 113: Urgh. I feel like we should be able to use pydantic union type collections for all of this and just have typed rows, but I guess that doesn't actually work. Sigh.

This is fine, I'm just complaining.

metricflow/test/table_snapshot/table_snapshots.py
```diff
@@ -0,0 +1,233 @@
...
+    def __init__(self, sql_client: SqlClient, schema_name: str) -> None:  # noqa: D
+        self._sql_client = sql_client
+        self._schema_name = schema_name
+
+    def restore(self, table_snapshot: SqlTableSnapshot, overwrite: bool = False) -> None:  # noqa: D
+        sql_table = SqlTable(schema_name=self._schema_name, table_name=table_snapshot.table_name)
+        if overwrite:
+            self._sql_client.drop_table(sql_table)
+        elif self._sql_client.table_exists(sql_table):
+            logger.info(f"Table {sql_table.sql} already exists, so not restoring")
```
tlento, line 136: Maybe `already exists, no need to restore` here? Makes it more obvious that this for the case where you want to load new tables but not clobber the ones that are there.

metricflow/test/table_snapshot/table_snapshots.py
```diff
@@ -0,0 +1,233 @@
...
+                        f"Conflicting paths: {table_snapshot.file_path} "
+                        f"{self._table_snapshots[table_snapshot.table_name].file_path}"
+                    )
+                self._table_snapshots[table_snapshot.table_name] = table_snapshot
+
+    @staticmethod
+    def _parse_config_yaml(file_path: Path) -> Sequence[SqlTableSnapshot]:
+        """Parse the YAML file at the given path into table snapshots."""
+        results = []
+        document_key = "table_snapshot"
```
tlento, line 177: Should this be a constant someplace?

metricflow/test/table_snapshot/table_snapshots.py
```diff
@@ -0,0 +1,233 @@
...
+    def _find_all_yaml_file_paths(directory: Path) -> Sequence[Path]:  # noqa: D
+        """Recursively search through the given directory for YAML files."""
+        yaml_file_paths = []
+
+        for root, dirs, files in os.walk(directory):
+            for file in files:
+                if file.endswith(".yaml"):
+                    yaml_file_paths.append(Path(root, file))
+
+        return sorted(yaml_file_paths)
```
tlento, line 220: I feel like we have a bunch of these, but we'll get rid of some of them soon so I guess it's ok.

tlento (approved):
Seems reasonable! I left a couple of nits inline along with my usual run of complaining about stuff.

We can investigate moving this to dbt later.

metricflow/test/table_snapshot/table_snapshots.py
plypaul: Updated.

metricflow/test/table_snapshot/table_snapshots.py
plypaul: Updated.

metricflow/test/table_snapshot/table_snapshots.py
plypaul: Yeah, was thinking of a loader class with generic typing, but put that on the back burner.

PR #468, opened 2023-04-26
Remove MYSQL from `SqlEngine` & `SqlDialect` enums

courtneyholcomb:
This was an artifact of Transform and is no longer needed.


plypaul (approved):
LGTM aside from the change log issue.

Issue #470, opened 2023-04-26
Remove Fuzzy Wuzzy from MetricFlow


callum-mcdata:
**Describe the Feature**
Fuzzy Wuzzy has a GPL license, which is not compatible with BSL. As such, we need to remove our dependency on it.

**Would you like to contribute?**
Potentially!

Issue #465, opened 2023-04-25
Update MetricFlow License to BSL


tlento:
MetricFlow will change its license to BSL to match the dbt labs licensing configuration for the semantic layer.

PR #474, opened 2023-04-26
Changing to RapidFuzz from FuzzyWuzzy

### LINKED ISSUE #471: Remove Python-Levenshtein from MetricFlow
**Describe the Feature**
Python-Levenshtein is a GPL licensed library, which is not compatible with MetricFlow's new licensing. We should  remove it.

**Would you like to contribute?**
Potentially!

callum-mcdata:
Resolves #471 and #470 (as 471 is a dependency within 470)

## This PR
Given #465 , we need to remove our dependency on FuzzyWuzzy. I found an alternative library called RapidFuzz that has the same APIs but with an MIT license and supposedly increased performance.

@tlento and his eagle eyes also noted that we have to hard-pin GitPython in order to resolve test failures with the migration.

**I have not done security review and we should absolutely do so before we add this repo**


tlento:
I know what's causing your failures. You can work around this for now by pegging GitPython's version to `">=3.1.27,<3.1.31"` - there's some test shim that's broken and passing a bad value into the GitPython command, but fixing it here doesn't seem worthwhile.

tlento (approved):
Thanks for the quick update!

For merging, I recommend squashing instead of merging with individual commits, and pulling your PR description into the commit message. That way we get rid of the breakpoint stuff and nothing is broken in between commits. The change will be focused enough in the source tree for people to follow.

Issue #475, opened 2023-04-27
Pin dbt-core to 1.4, as 1.5 causes mypy issues


callum-mcdata:
**Describe the bug**
With the release of dbt-core 1.5 today, our linting CI job is now raising errors. dbt-core 1.5 expects some information about the target model name that is not accounted for in our dbt functionality. Given that we're going to get rid of that functionality in the near future, I propose that we temporarily pin dbt-core to 1.4 to unblock.

PR #476, opened 2023-04-27
Hard pinning dbt-core to 1.4

callum-mcdata:
This PR addresses issue #475

### DIFF
--- a/.changes/unreleased/Under the Hood-20230427-102540.yaml
+++ b/.changes/unreleased/Under the Hood-20230427-102540.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Pinning dbt-core to 1.4
+time: 2023-04-27T10:25:40.769074-05:00
+custom:
+  Author: callum-mcdata
+  Issue: "475"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,5 +31,5 @@ repos:
             pytest,
             python-dateutil,
             types-python-dateutil,
-            dbt-core,
+            dbt-core==1.4.0,
           ]
--- a/metricflow/model/transformations/dbt_to_metricflow.py
+++ b/metricflow/model/transformations/dbt_to_metricflow.py
@@ -5,7 +5,6 @@ import traceback
 from typing import DefaultDict, Dict, List, Optional, Set
 from dbt.contracts.graph.nodes import Metric as DbtMetric, ModelNode as DbtModelNode
 from dbt.contracts.graph.unparsed import MetricFilter as DbtMetricFilter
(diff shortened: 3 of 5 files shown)


callum-mcdata:
@QMalcolm would you have any idea why the below errors would be triggered if I've hard pinned our dev testing to 1.4?

> metricflow/model/transformations/dbt_to_metricflow.py:8: error: Module "dbt.exceptions" has no attribute "ref_invalid_args"  [attr-defined]
metricflow/model/transformations/dbt_to_metricflow.py:113: error: Missing positional argument "target_model_version" in call to "resolve_ref" of "Manifest"  [call-arg]
Found 2 errors in 1 file (checked 355 source files)

I am able to recreate it locally but I am very confused because `target_model_version` would only exist in 1.5.

QMalcolm:
> @QMalcolm would you have any idea why the below errors would be triggered if I've hard pinned our dev testing to 1.4?
> 
> > metricflow/model/transformations/dbt_to_metricflow.py:8: error: Module "dbt.exceptions" has no attribute "ref_invalid_args"  [attr-defined]
> > metricflow/model/transformations/dbt_to_metricflow.py:113: error: Missing positional argument "target_model_version" in call to "resolve_ref" of "Manifest"  [call-arg]
> > Found 2 errors in 1 file (checked 355 source files)
> 
> I am able to recreate it locally but I am very confused because `target_model_version` would only exist in 1.5.

I think agree, `target_model_version` doesn't exist in 1.4. Locally it looks correct with 1.4 installed. I can't see in the failing job what verion of dbt is being installed 🤔 My guess is that the job is still getting 1.5 for whatever reason

QMalcolm:
> > ...
> 
> I think agree, `target_model_version` doesn't exist in 1.4. Locally it looks correct with 1.4 installed. I can't see in the failing job what verion of dbt is being installed 🤔 My guess is that the job is still getting 1.5 for whatever reason

So on [line 15 of Run Linter](https://github.com/dbt-labs/metricflow/actions/runs/4822118216/jobs/8588922358?pr=476#step:6:16) it looks like it gets the types for dbt-core via a mypi-mirror (https://github.com/pre-commit/mirrors-mypy:dbt-core) but it doesn't specify a version. My guess is that gets the latest?

QMalcolm:
@callum-mcdata I think I found the issue. We also need to correctly constrain the dbt-core version referenced in [.pre-commit-config.yaml](https://github.com/dbt-labs/metricflow/blob/8800ed5e36e2f7795d458fd46564aeb9f991a995/.pre-commit-config.yaml#L34)

QMalcolm (approved):
Looks great to me! 🚀

PR #477, opened 2023-04-27
Removing YamlLint

### LINKED ISSUE #472: Remove YamlLint from MetricFlow
**Describe the Feature**
YamlLint is GPL licensed, which is not compatible with MetricFlow's new licensning structure. We should remove the dependency on YamlLint 

**Would you like to contribute?**
Potentially!

callum-mcdata:
Resolves #472 

## This PR
This PR contains changes to remove YamlLint as it does not work with the new BSL License. Additionally we remove the ConfigLinter functionality as it was based on YamlLint.


dbt_semantic_interfaces/pyproject.toml
```diff
@@ -20,7 +20,6 @@ pandas = "^1.3.0"
 PyYAML = "^6.0"
 mo-sql-parsing = "^9.328.23003"
 more-itertools = "8.10.0"
-yamllint = "^1.26.3"
```
tlento, line 23: You'll need to run poetry update in both package directories, otherwise we'll technically still ship with this in the package.

metricflow/cli/main.py
```diff
@@ -658,21 +657,7 @@ def validate_configs(
...
-        lint_spinner.start()
-
-        lint_results = ConfigLinter().lint_dir(path_to_models(handler=cfg.config))
-        if not lint_results.has_blocking_issues:
-            lint_spinner.succeed(f"🎉 Successfully linted config YAML files ({lint_results.summary()})")
-        else:
-            lint_spinner.fail(f"Breaking issues found in config YAML files ({lint_results.summary()})")
-            _print_issues(lint_results, show_non_blocking=show_all, verbose=verbose_issues)
-            return
+    lint_results = ModelValidationResults()
```
tlento, line 660: Can we remove this altogether? Looks like it's only used in the removed block, plus as an input into a multi-result merge, so we should be able to delete it from that array as well.

metricflow/cli/main.py
```diff
@@ -51,7 +51,7 @@
 from metricflow.model.model_validator import ModelValidator
 from dbt_semantic_interfaces.objects.user_configured_model import UserConfiguredModel
 from metricflow.protocols.sql_client import SqlEngine
-from metricflow.engine.utils import model_build_result_from_config, path_to_models
```
tlento, line 54: I think this is what broke the tests. If path_to_models isn't used anymore we should update the patch calls. There are a bunch in the test_cli.py module, for example:

https://github.com/dbt-labs/metricflow/blob/main/metricflow/test/cli/test_cli.py#L64

Also, I'm 99% sure those patches are the thing that caused the GitPython issues, so deleting those patch calls might incidentally resolve that problem for us.

tlento (commented):
Nice! Left some notes inline on resolving test failures and updating the build.

dbt_semantic_interfaces/pyproject.toml
callum-mcdata: TIL

metricflow/cli/main.py
callum-mcdata: That fixed it!

metricflow/cli/main.py
callum-mcdata: Removed it!

tlento (approved):
I love red diffs! Thanks so much!

PR #484, opened 2023-04-29
Simplify test model templates

plypaul:
This PR simplifies the model templates by changing the substitution to only specify the source schema.


tlento (approved):
LOL. We were working so hard on these templates, too.....

PR #485, opened 2023-05-01
Add ObjectToReference converter

WilliamDee:
### Description
With moving all the model objects to dbt-semantic-interfaces and having *Reference objects remain in MetricFlow, all the `Object.reference` properties should be removed from the model objects. This means, we would need to replace all usages of `Object.reference` to instantiating the *Reference object directly. This class organizes and keeps note of where those use cases were.

<!--- 


QMalcolm (approved):
Looks good to me! Let's get this merged 🚀

PR #486, opened 2023-05-01
Move *References objects to dbt-semantic-interfaces

WilliamDee:
### Description
- Move relevant `*Reference` classes from `instances.py` to `references.py`
- Move `references.py` into dbt-semantic-interfaces directory
- Fix imports

<!--- 


metricflow/instances.py
```diff
@@ -22,61 +22,6 @@
 )
 
 
-class ModelReference(SerializableDataclass):
```
tlento, line 25: @QMalcolm @WilliamDee - do we want to move these? As far as I know they're primarily used inside of our internal semantic containers. The DataSourceReference, for example, is only used as a typehint for the property allowing us to access the reference object. If that's the case, it might make sense to keep these in MetricFlow, and have our internal model object implementations (whatever they may look like) add those properties.

If we have some specific use for these in mind - or if we just prefer to move them because then we can have the interfaces package provide a functioning suite of model objects for ease of testing parsing behavior and so forth - that's perfectly fine. Just thought I'd ask before we get too far!

tlento (commented):
Mechanically this looks great. I have a question inline. @QMalcolm if this is on your preferred path, please just approve it and move on with a quick note for posterity about how these Reference objects will be used in the new package (or if they're being moved out of convenience, which is fine too).

metricflow/instances.py
QMalcolm: So if we only look at what is in dbt_semantic_interfaces right now, references are not necessary. However, talking to @plypaul earlier today, it sounds like we'll need either References or Specs in dbt_semantic_interfaces for resolving `{{ dimension('dim_name') }}` and etc in dbt_semantic_interfaces. When weighing Specs and References, references seemed the easier way to go in that we end up moving less to dbt_semantic_interfaces and it is less tangled.

QMalcolm (approved):
Looks good to me!

As a note, we've flip flopped a few times as to whether we move `references` as well as a few other things. Ultimately why we are moving `references` into `dbt_semantic_interfaces` is because they'll be necessary for the resolving elements invoked in the metric filter that look like `{{ dimension('dim_name') }} == ... and {{dimension('another_dim) }} == ... and ...`

metricflow/instances.py
tlento: Ah yes, that makes perfect sense. I think resolving to a Reference is the right choice.

PR #498, opened 2023-05-05
Update DataSource objects to conform with the new property changes

WilliamDee:
resolves https://github.com/dbt-labs/dbt-semantic-interfaces/issues/12

### Description
In the new world of dbt-core x MetricFlow `Models` have become `SemanticModels` and the properties have changed slightly.  Semantic Model objects should have the following properties

| Property Name | Type | Description |
| --- | --- | --- |
| name | str | Name of the semantic model |
| description | str | Description of the semantic model |
| entities | List[Entity] | Entities for the semantic model |
| dimensions | List[Dimension] | Dimensions for the semantic model |
| measures | List[Measure] | Measures for the semantic model  |
| node_relation | NodeRelation | Path object to where the data should be |

Relatedly the NodeRelation object should have the following for this first pass

| Property Name | Type |
| --- | --- |
| alias| str |
| schema_name | str |
| database | Optional[str] |
| relation_name | str |

`schema_name` was initially supposed to be `schema`, but that was a pydantic reserved attribute.
<!--- 


metricflow/test/model/validations/test_common_entities.py
```diff
@@ -12,19 +12,19 @@
 
 
 @pytest.mark.skip("TODO: re-enforce after validations improvements")
-def test_lonely_identifier_raises_issue(simple_model__with_primary_transforms: UserConfiguredModel) -> None:  # noqa: D
+def test_lonely_entity_raises_issue(simple_model__with_primary_transforms: UserConfiguredModel) -> None:  # noqa: D
     model = copy.deepcopy(simple_model__with_primary_transforms)
-    lonely_identifier_name = "hi_im_lonely"
+    lonely_entity_name = "hi_im_lonely"
```
tlento, line 17: 🤣

metricflow/model/transformations/dbt_to_metricflow.py
```diff
@@ -202,7 +202,6 @@ def build_data_source_for_metric(self, dbt_metric: DbtMetric) -> DataSource:
             name=metric_model_ref.name,
             description=metric_model_ref.description,
             sql_table=data_source_table,
-            dbt_model=data_source_table,
```
tlento, line 205: This seems fine to me but I don't know if this affects the conversion tooling @callum-mcdata and @Jstein77 are working on.

If it does it's probably ok - they can fork from an earlier point in the source tree - but just wanted to mention it.

metricflow/test/model/validations/test_reserved_keywords.py
```diff
@@ -19,18 +19,6 @@ def test_no_reserved_keywords(simple_model__with_primary_transforms: UserConfigu
     assert len(issues) == 0
 
 
-def test_reserved_keywords_in_sql_table(simple_model__with_primary_transforms: UserConfiguredModel) -> None:  # noqa: D
```
tlento, line 22: We might want to keep this validation for the string output of the NodeRelation or whatever we wind up calling it, but I think deleting this test is ok for now.

tlento (approved):
I know the NodeRelation thing has something to do with aligning with dbt core's longer term plans, but I don't love the name and I think we have to change that interface away from the `Any` typed inputs.

I have a couple of questions inline that I'm unsure about, but I'm approving now because none of the issues I see here are showstoppers, and I think whatever way we go with these conversations you'll be far better off merging this as-is and then making changes in separate PRs.

metricflow/test/model/validations/test_reserved_keywords.py
WilliamDee: Brought this back and modified to validate NodeRelation

metricflow/model/transformations/dbt_to_metricflow.py
WilliamDee: I'm not aware of the conversion tooling they are working on there, but wanted to point out that the only use case at the moment for the `dbt_model` property was that it is used by the Transform UI purely for showing lineage metadata. So hopefully there's some work around there if they are using it.

dbt_semantic_interfaces/dbt_semantic_interfaces/objects/data_source.py
```diff
@@ -10,6 +11,40 @@
...
+    """Path object to where the data should be"""
+
+    alias: str
+    schema_name: str
+    database: Optional[str] = None
+    relation_name: str = ""
+
+    @validator("relation_name", always=True)
+    @classmethod
+    def create_default_relation_name(cls, value: Any, values: Any) -> str:  # type: ignore[misc]
```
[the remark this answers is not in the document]
WilliamDee: I'm going to go ahead merge this in as is with this and readdress this in the followup PR https://github.com/dbt-labs/metricflow/pull/513 just for the sake of easiness due to the rename of this file in that PR

PR #501, opened 2023-05-05
Move pretty_print and pformat_big_objects to dbt_semantic_interfaces

QMalcolm:
### Description

We weren't originally planning on moving these utility functions. However since we moved over `SerializableDataclass`, we now need to move over these utilities because `SerializableDataclass` depends on them.

<!--- 


dbt_semantic_interfaces/dbt_semantic_interfaces/object_utils.py
```diff
@@ -0,0 +1,104 @@
+from __future__ import annotations
```
tlento, line 1: While we're moving this file can we rename it to `pretty_print.py` or similar, since that's all this module does now? Then we'll be rid of object_bag_of_crap.py once and for all!

metricflow/object_utils.py
```diff
@@ -1,115 +1,13 @@
...
-    items = []
-    for arg in args:
-        items.append(pretty_format(arg))
-    for key, value in kwargs.items():
-        items.append(f"{key}:")
-        items.append(textwrap.indent(pretty_format(value), prefix="    "))
-    return "\n".join(items)
-
-
 def hash_items(items: Sequence[SqlColumnType]) -> str:
```
tlento, line 11: Oh I guess we won't be rid of object_bag_of_crap, but we can move this function someplace more sensible later and THEN it will be gone.

tlento (approved):

dbt_semantic_interfaces/dbt_semantic_interfaces/object_utils.py
QMalcolm: Absolutely!

metricflow/object_utils.py
QMalcolm: I think I can do that 😉

QMalcolm: Moved to `specs`, let me know what you think

PR #502, opened 2023-05-05
Making AggregationType use ExtendedEnum again

QMalcolm:
### Description

When we moved `AggregationType` over, we changed it from an `ExtendedEnum` to an `Enum`. We thought this wouldn't be problematic. However, it turns out that the `_missing_` function of `ExtendedEnum` is actually very important. This is because the custom `_missing_` implementation ensures that using upper-cased or lower-cased values in a config file both work. That is to say that with `ExtendedEnum`, `sum` and `SUM` are both valid, but with `Enum` only `sum` is.

<!--- 


tlento (approved):
So *that's* why validations all started failing on the tutorial model...... my bad.

QMalcolm:
> So _that's_ why validations all started failing on the tutorial model...... my bad.

🙃

Issue #505, opened 2023-05-08
[Feature] Use Templates for Defining Metric Filters


plypaul:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing metricflow functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

When defining a metric, it's possible to specify a filter for the measure. For example, to define the `canada_bookings` metric, you could use the `bookings` measure with the filter that uses the `country` dimension. Previously, in specifying the filter, the user would specify the filter as raw SQL like `listing__country == 'ca'`. The dunder syntax was used to specify the entity path / join path so that the right dimension was used (e.g. not `guest__country`). There were a couple of issues with this approach.

* We're moving away from the dunder syntax as it's difficult to specify additional qualities about the entity path.
* To extract the dimensions that are used, the filter needed to be parsed via a SQL parser, which could be error prone with the variety of SQL dialects and new features. 

Instead, the proposal is to support templating in the filter string, so that future path specifications will be easier to support. e.g. `{{ dimension('listing', entity_path=['listing'] }} == 'ca'`. In addition, the use of a template will eliminate the need for SQL parsing.

### Describe alternatives you've considered

See previous implementation in the description.

### Who will this benefit?

With the templated syntax, there will be fewer issues supporting new SQL engines and enable us to later add improved join path specification. 

### Are you interested in contributing this feature?

Yes

callum-mcdata:
@plypaul similar to other issue, is this one good to be closed out?

Jstein77:
Looks like the https://github.com/dbt-labs/metricflow/pull/506 resolves this issue so closing it out.

PR #506, opened 2023-05-08
Use Templates for Defining Metric Filters

plypaul:
### Description
This PR resolves the issue in #505. Since we renamed `constraint` -> `filter`, there are renames that need to be done to make function signatures more consistent, but that will be handled in a follow up PR.

<!--- 


dbt_semantic_interfaces/dbt_semantic_interfaces/dataclass_serialization.py
```diff
@@ -18,12 +18,10 @@
     get_origin,
     get_type_hints,
 )
-from typing_extensions import TypeAlias
 
 import pydantic
 from pydantic import BaseModel
-
-from dbt_semantic_interfaces.pretty_print import pformat_big_objects
```
tlento, line 26: This commit message is now very confusing, but ok. :)

dbt_semantic_interfaces/dbt_semantic_interfaces/objects/constraints/filter_renderer.py
```diff
@@ -0,0 +1,203 @@
...
+        FilterRenderer.render(
+            templated_filter_sql=templated_filter,
+            call_renderer=_DummyCallRenderer(),
+        )
+
+        return FilterCallParameterSets(
+            dimension_call_parameter_sets=tuple(dimension_call_parameter_sets),
+            time_dimension_call_parameter_sets=tuple(time_dimension_call_parameter_sets),
+            entity_call_parameter_sets=tuple(entity_call_parameter_sets),
+        )
```
tlento, line 149: This is super weird. Maybe this is the only reasonable way to do this? Either way, I guess it'll do for now but it's really odd to overload the renderer for this extraction.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -66,6 +66,7 @@
     ColumnAssociationResolver,
     LinklessEntitySpec,
     InstanceSpecSet,
+    ResolvedWhereFilter,
```
tlento: Checkpoint here.

tlento (commented):
Working on it....

dbt_semantic_interfaces/dbt_semantic_interfaces/dataclass_serialization.py
plypaul: Yeah, this commit is no longer needed since `pformat_big_objects` got moved from `metricflow` to `dbt_semantic_interfaces`.

dbt_semantic_interfaces/dbt_semantic_interfaces/objects/constraints/filter_renderer.py
plypaul: There is a way to do this by parsing the AST that Jinja can generate for you. However, it seemed like a more complex approach so I didn't go into it too deeply. I agree that this is a little odd though.

metricflow/test/model/test_resolved_where_filter.py
```diff
@@ -0,0 +1,91 @@
...
+    where_filter = WhereFilter(
+        where_sql_template="{{ time_dimension('created_at', 'month', entity_path=['listing']) }} = '2020-01-01'"
+    )
+
+    resolved_where_filter = ResolvedWhereFilter.create_from_where_filter(
+        where_filter=where_filter,
+        column_association_resolver=column_association_resolver,
+    )
+
+    assert resolved_where_filter.where_sql == "listing__created_at__month = '2020-01-01'"
```
tlento: Do we want to dunder the month on the end here? Is this because we'll do the dialect-specific DATE_TRUNC() internally and alias the output to the `__month`? I'm worried about that causing issues with optimizer passes, although I think right now it won't be a problem.

I suspect this is a question for later - right now the system works on that rendered dunder so this should match current behavior, right?

metricflow/model/transformations/dbt_to_metricflow.py
```diff
@@ -235,11 +238,11 @@ def build_proxy_metric(self, dbt_metric: DbtMetric) -> Metric:
         if dbt_metric.calculation_method == "derived":
             raise RuntimeError("Cannot build a MetricFlow proxy metric for `derived` DbtMetric")
 
-        where_clause_constraint: Optional[WhereClauseConstraint] = None
+        where_filter: Optional[WhereFilter] = None
         if dbt_metric.filters:
-            where_clause_constraint = WhereClauseConstraint(
-                where=self.build_where_stmt_from_filters(filters=dbt_metric.filters),
-                linkable_names=[filter.field for filter in dbt_metric.filters],
+            # TODO: Figure out how to migrate.
```
tlento, line 243: I think all of this code will go away since we're not supporting this converter in our next release.

metricflow/test/model/validations/helpers.py
```diff
@@ -85,9 +87,9 @@ def data_source_with_guaranteed_meta(
     sql_query: Optional[str] = None,
     dbt_model: Optional[str] = None,
     metadata: Metadata = default_meta(),
-    identifiers: Sequence[Entity] = [],
-    measures: Sequence[Measure] = [],
-    dimensions: Sequence[Dimension] = [],
```
tlento, line 90: Whoops

metricflow/model/semantics/metric_semantics.py
```diff
@@ -145,30 +144,38 @@ def metric_input_specs_for_metric(
...
-                    where_constraint=original_metric_obj.constraint,
+                # original_metric_constraint = WhereConstraintConverter.convert_to_spec_where_constraint(
+                #     data_source_semantics=self._data_source_semantics,
+                #     where_constraint=original_metric_obj.constraint,
+                # )
+                # combined_spec_constraint = (
+                #     combined_spec_constraint.combine(original_metric_constraint)
+                #     if combined_spec_constraint
+                #     else original_metric_constraint
+                # )
```
tlento, line 166: We should probably remove these if you haven't already.

metricflow/query/query_parser.py
```diff
@@ -476,7 +483,7 @@ def _parse_and_validate_query(
             order_by_specs=order_by_specs,
             output_column_name_overrides=tuple(output_column_name_overrides),
             time_range_constraint=time_constraint,
-            where_constraint=spec_where_constraint,
+            where_constraint=resolved_where_filter,
```
tlento: I can't tell if all of the changes here are correct. Clearly, _parse_and_validate_query is due for some tidying. But not now.

metricflow/test/integration/test_configured_cases.py
```diff
@@ -134,6 +134,34 @@ def double_data_type_name(self) -> str:
...
+    def render_entity_template(self, entity_name: str, entity_path: Sequence[str] = ()) -> str:
+        """Similar to render_dimension_template() but for entities."""
+        return f"{{{{ entity('{entity_name}', entity_path={repr(entity_path)}) }}}}"
+
+    def render_time_dimension_template(
+        self, time_dimension_name: str, time_granularity: str, entity_path: Sequence[str] = ()
+    ) -> str:
+        """Similar to render_dimension_template() but for time dimensions."""
+        return (
+            f"{{{{ time_dimension('{time_dimension_name}', '{time_granularity}', entity_path={repr(entity_path)}) }}}}"
```
tlento, line 162: Yikes. Nothing for it..... but that's a mark in favor of not using jinja if we don't end up needing to do so.

metricflow/model/spec_converters.py
```diff
@@ -49,76 +36,3 @@ def convert_to_measure_spec(measure: Measure) -> MeasureSpec:
             element_name=measure.name,
             non_additive_dimension_spec=non_additive_dimension_spec,
         )
-
-
-class WhereConstraintConverter:
```
tlento, line 54: Huzzah!

tlento (approved):
I have a bunch of quibbles inline but all of them can (and probably should, given the merge/rebase you're about to experience) be addressed in follow-ups.

The main concern I have is the FilterRenderer interface is confusing. I think it's a combination of overloading FilterRenderer.render() to do things which are not rendering, plus the use of the not-terribly-pythonic function scoped inner class definitions.

I think the simplest cleanup there is to do some renaming and move those inner classes out of the functions and into a coherent module where they can live together.

Instead of `FilterRenderer` and a `render` method, how about `FilterTransformer` and a `transform` method? We could call it something else if we want, but that's the easiest.

All of this is follow-up-PR territory, though, so let's get this merged (with whatever minor tweaks you see fit to make) and unblock things.

metricflow/model/semantics/metric_semantics.py
plypaul: Oh yeah, will remove.

metricflow/model/transformations/dbt_to_metricflow.py
plypaul: Chatted with @QMalcolm earlier and confirmed that they going away. Also the reason for the skipped tests in a later commit.

metricflow/specs.py
```diff
@@ -731,3 +740,121 @@ class PartitionSpecSet(SerializableDataclass):
...
+        entity_links=parameter_set.entity_path,
+        time_granularity=parameter_set.time_granularity,
+    )
+
+
+def convert_to_entity_spec(parameter_set: EntityCallParameterSet) -> EntitySpec:  # noqa: D
+    return EntitySpec(
+        element_name=parameter_set.entity_reference.element_name,
+        entity_links=parameter_set.entity_path,
+    )
```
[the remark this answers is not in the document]
plypaul: Yeah, I'll just move them into the spec class.

metricflow/specs.py
```diff
@@ -731,3 +740,121 @@ class PartitionSpecSet(SerializableDataclass):
...
+    bind_parameters: SqlBindParameters
+    linkable_spec_set: LinkableSpecSet
+
+    @staticmethod
+    def create_from_where_filter(  # noqa: D
+        where_filter: WhereFilter,
+        column_association_resolver: ColumnAssociationResolver,
+        bind_parameters: SqlBindParameters = SqlBindParameters(),
+    ) -> ResolvedWhereFilter:
+        class _CallRenderer(FilterFunctionCallRenderer):  # noqa: D
```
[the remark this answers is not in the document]
plypaul: Let me see how best to organize that.

metricflow/specs.py
```diff
@@ -731,3 +740,121 @@ class PartitionSpecSet(SerializableDataclass):
...
+                return column_association_resolver.resolve_dimension_spec(
+                    convert_to_dimension_spec(dimension_call_parameter_set)
+                ).column_name
+
+            def render_time_dimension_call(  # noqa: D
+                self, time_dimension_call_parameter_set: TimeDimensionCallParameterSet
+            ) -> str:
+                return column_association_resolver.resolve_time_dimension_spec(
+                    convert_to_time_dimension_spec(time_dimension_call_parameter_set)
+                ).column_name
```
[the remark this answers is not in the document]
plypaul: Yeah, that's correct. We wanted to leave the `ColumnAssociationResolver` out of `db_semantic_interfaces`.

metricflow/test/integration/test_configured_cases.py
plypaul: Up for other ideas!

plypaul:
Yeah, the `FilterRenderer` class is a little odd right now. `FilterTransformer` is a possibility, but also thinking about having two separate classes `FilterParser` and `FilterRenderer` at the expense of some repetition. What do you think?

plypaul:
Actually, maybe `FilterTransformer` is sufficient.

dbt_semantic_interfaces/dbt_semantic_interfaces/objects/filters/where_filter.py
```diff
@@ -0,0 +1,40 @@
...
+        User-provided constraint strings are SQL snippets conforming to the expectations of SQL WHERE clauses,
+        and as such we parse them using our standard parse method below.
+        """
+        if isinstance(input, str):
+            return WhereFilter(where_sql_template=input)
+        else:
+            raise ValueError(f"Expected input to be of type string, but got type {type(input)} with value: {input}")
+
+    @property
+    def call_parameter_sets(self) -> FilterCallParameterSets:
```
plypaul, line 38: @QMalcolm - FYI this was the original implementation.

PR #513, opened 2023-05-09
Rename DataSource -> SemanticModel

WilliamDee:
resolves https://github.com/dbt-labs/dbt-semantic-interfaces/issues/12

### Description
Follow up of [this PR](https://github.com/dbt-labs/metricflow/pull/498). The first PR was to update the objects to the new spec. This PR is to rename all instances of DataSource to SemanticModel.

<!--- 


WilliamDee:
> I don't think this is the interface we want to use here. These Any typed inputs are going to cause problems. I'd be more comfortable with this if it was accepting either a bound generic and requiring a concrete input type at this point, or a Protocol type with a method that does the work we need so we aren't implementing the spec inside of this function.
>
> The reason I think we should push for a better interface is right here. If this isn't an object with get, we hit an exception. If it is a dict or dict-like object of, say, connection properties, we could end up with "metrics_db.None.None"
> 
> That's why this should take a type that does this relation name construction work, rather than assuming it's an arbitrary thing that happens to implement the particular internal behavior we define here.

@tlento wanted to bring the convo here. I'm not quite sure how to avoid the Any typing here. Mainly looking at this https://docs.pydantic.dev/latest/usage/validators/#root-validators. However, we can avoid the `None.None` issue as we can assert that those fields exist and are not None. Realistically, I only wrote this because I wanted to improve the UX by not having to define `relation_name` explicitly if they're all mostly just in the format `db.schema.table` and this can just autopopulate that field given that the `schema_name` and `alias` are already validated as required and not None through the pydantic and json_schema validation.

metricflow/dataflow/builder/node_evaluator.py
```diff
@@ -29,12 +29,12 @@
     ValidityWindowJoinDescription,
 )
 from metricflow.instances import InstanceSet
-from metricflow.model.semantics.semantic_model_join_evaluator import DataSourceJoinEvaluator
+from metricflow.model.semantics.semantic_model_join_evaluator import SemanticModelJoinEvaluator
```
QMalcolm, line 32: Pausing here (at the start of b08305294bc). So far so good. Will pick this back up in the morning.

QMalcolm (commented):
Woof, renaming PRs are so biiiiig. Things are looking good! Not all the names are desirable, thinking of `SemanticModelSemanticsAccessor`, but that can be swallowed for now.

There are some references to `data source` / `Data Source` in error messages, cli outputs, and the like that we should probably take care of as part of this. I believe the relevant files are
- `metricflow/metricflow/cli/main.py`
- `metricflow/metricflow/dataset/convert_semantic_model.py`
- `metricflow/metricflow/inference/runner.py`
- `metricflow/metricflow/model/semantics/semantic_model_semantics.py`
- test fixture object descriptions
- test doc strings

Additionally there are still some references to `DataSource`/`data_source` but they happen in files that I don't think we care about too much: 
- `metricflow/metricflow/test/snapshots/test_dataflow_plan_builder.py/DataflowPlan/test_constrained_measure_plan__dfp_0.xml`
- `metricflow/assets/example_plan.svg`

tlento:
> wanted to bring the convo here. I'm not quite sure how to avoid the Any typing here. Mainly looking at this https://docs.pydantic.dev/latest/usage/validators/#root-validators. However, we can avoid the None.None issue as we can assert that those fields exist and are not None. Realistically, I only wrote this because I wanted to improve the UX by not having to define relation_name explicitly if they're all mostly just in the format db.schema.table and this can just autopopulate that field given that the schema_name and alias are already validated as required and not None through the pydantic and json_schema validation.

Ah, I missed the `@validator` decorator on that method, that explains the signature. I think if you avoid the root validator you can do something a little more natural. There's also the issue, which we might not care about, that we clobber any empty string input.

Will anybody ever pass in a different name structure?  If not, we can make this an @property method that just builds the string from the pieces and returns it, or even use a default factory (although that's in beta).

For now I'd say to make this a property decorated method. Then nobody ever passes in a relation directly.

QMalcolm (approved):
Looks great to me! Thanks again for this work @WilliamDee 🎉 

star-wars-hans-solo

PR #518, opened 2023-05-10
Removing Ratio Metric

### LINKED ISSUE #504: [Bug] Remove expr & ratio metrics and bundle with derived metrics

### Current Behavior

**Valid**:
```yml
metric:
  name: derived_metric
  type: derived
  type_params:
    expr: revenue + customers_7d
    metrics:
      - name: revenue_usd
      - name: customers_7d
```

**Invalid**:
```yml
metric:
  name: derived_metric
  type: derived
  type_params:
    expr: revenue + customers_7d
    metrics:
      - revenue_usd
      - customers_7d
```

### Expected Behavior

The second example working.

### Steps To Reproduce

Metricflow 0.14

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
- metricflow:
```

_No response_

### Additional Context

_No response_

callum-mcdata:
resolves #504 

### Description

This PR removes `ratio` metrics.


tlento (changes requested):
Per discussion, let's sync up on this after the cutover so we can figure out the easiest way to port the functionality so it's based on derived metrics instead of measures.

I think this will need major changes due to all the impending file moves and stuff so pushing this back for now.

tlento:
We decided to keep RATIO but make the inputs different, that work is underway.

PR #525, opened 2023-05-11
Removing Expr Metric 

### LINKED ISSUE #504: [Bug] Remove expr & ratio metrics and bundle with derived metrics

callum-mcdata:
resolves #504 

Closed the old issue for some weird GH issues.

### Description

Removes expr metrics


metricflow/test/model/validations/test_metrics.py
```diff
@@ -254,3 +257,56 @@ def test_derived_metric() -> None:  # noqa: D
     assert (
         len(missing_error_strings) == 0
     ), f"Failed to match one or more expected errors: {missing_error_strings} in {set([x.as_readable_str() for x in build_issues])}"
+
+
+def test_metric_alias_is_set_when_required() -> None:
```
callum-mcdata, line 262: @QMalcolm I am getting an error in this test on line 307 where this isn't triggering an error for some odd reason. 

Did I build the rule wrong? I would think that we wouldn't permit this kind of configuration, unless we do because of the constraint ?

QMalcolm: I'm looking into this now. To sanity check. If you improperly configure an input metric in an example config and then run `mf validate-configs` does it appropriately fail? I only ask to sanity check that whether it's the rule or the test that's failing.

tlento:
This pre-dates the DSI cutover, so it'll need a re-do.

PR #529, opened 2023-05-15
Removing boolean aggregation type

callum-mcdata:
resolves non-created issue

### Description

In dbt-semantic-interfaces, we removed `boolean` as an aggregation type because it was deprecated.


tlento:
Done in dbt-semantic-interfaces

PR #531, opened 2023-05-16
Use a Transform For WhereFilter -> WhereFilterSpec

plypaul:
resolves #

### Description

This PR implements a transform that replaces `WhereFilter.create_from_where_filter()`. This helps address comments from original PR that created `WhereFilterSpec` and also allows resolution of some of the circular dependencies that would result from splitting `specs.py` into different files.

<!--- 


metricflow/aggregation_properties.py
```diff
@@ -2,7 +2,7 @@
 
 from enum import Enum
 
-from dbt_semantic_interfaces.objects.aggregation_type import AggregationType
+from dbt_semantic_interfaces.type_enums.aggregation_type import AggregationType
```
plypaul, line 5: @QMalcolm Mind verifying the package name change?

metricflow/aggregation_properties.py
```diff
@@ -15,7 +15,7 @@ def is_expansive(agg_type: AggregationType) -> bool:
         AggregationType.SUM,
         AggregationType.MIN,
         AggregationType.MAX,
-        AggregationType.BOOLEAN,
+        AggregationType.SUM_BOOLEAN,
```
plypaul, line 18: @QMalcolm Also checking that boolean got removed.

metricflow/aggregation_properties.py
tlento: Looks right. https://github.com/dbt-labs/dbt-semantic-interfaces/pull/33/

metricflow/aggregation_properties.py
tlento: Yup, same PR. https://github.com/dbt-labs/dbt-semantic-interfaces/pull/33/commits/44ea7fba8b6e67f45ed5b8c599109697f872fd23#diff-d7f77c3d98a0690d5778a5a035d0214589902735434bed98c9d13b992609ecc7L11-L12

metricflow/model/semantics/metric_lookup.py
```diff
@@ -105,9 +105,10 @@ def measures_for_metric(
             )
             spec = MetricInputMeasureSpec(
                 measure_spec=measure_spec,
-                constraint=WhereFilterSpec.create_from_where_filter(
-                    where_filter=input_measure.filter,
-                    column_association_resolver=column_association_resolver,
+                constraint=input_measure.filter.transform(
+                    ConvertToWhereSpec(
+                        column_association_resolver=column_association_resolver,
+                    )
```
tlento, line 111: This seems silly given that below we just do `ConvertToWhereSpec(...).transform(filter)` and it also works.

tlento (approved):
This is fine for now - more straightforward than what we had before. I'm a little preoccupied by the dead rat I just found in my water heater closet, so I missed the `transform(Transformer.transform(self))` in the upstream PR, but that's something we should probably re-evaluate when we have some time.  It looks just like the InstanceSpecTransform API layout, which we still haven't fully cleaned up. I didn't see anything that required this construction so maybe we go through and update both of those "transform" APIs in one fell swoop.

metricflow/model/semantics/metric_lookup.py
plypaul: Up for debate, but I actually like the `starting_object.transform(transform_function)` syntax. It flows more naturally when reading.

tlento: I'm suggesting this:

`new_object_type = object_converter.transform(original_object)`

This is also fine, although putting the operation on the instance suggests that in-place mutation might be happening:

`new_object_type = original_object.transform(transform_function)`

This is what we have in this PR:

```
new_object_type = object_converter.transform(original_object)

...

another_new_object = another_original_object.transform(object_converter)
```

We probably shouldn't use both constructions in the same file right out of the gate.

But none of this matters much.

plypaul: > We probably shouldn't use both constructions in the same file right out of the gate.

Agree - think I found those and made it consistent.

PR #533, opened 2023-05-16
Update MFEngine models to contain attributes from pydantic models

courtneyholcomb:
resolves #SEM-202

### Description
Updates `MFEngine` `Metric` and `Dimension` to include the attributes available on their corresponding pydantic models.
This changes the behavior of `simple_dimensions_for_metrics` slightly, since the primary time dim will show up once for each relevant semantic model (instead of just once, total).
This is needed in order to expose metadata queries in MF server with more robust information - specifically `list_metrics` and `list_dimensions`.

<!--- 


metricflow/engine/models.py
```diff
@@ -1,19 +1,35 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import List
+from typing import List, Optional
+
+from dbt_semantic_interfaces.objects.metric import MetricType, MetricTypeParams, Metric as PydanticMetric
```
tlento, line 6: Might be worth calling PydanticMetric something like SemanticManifestMetric or similar. It's so long though. ModelMetric would've been nice. Sigh.

tlento (approved):
Seems reasonable!

tlento:
Oh can you run `changie new` and add an Under the Hood entry? I don't know if Under the Hood is appropriate but I think it's fine.

metricflow/engine/models.py
courtneyholcomb: I don't mind long!

PR #553, opened 2023-05-24
Print warning to parallelize tests when applicable

plypaul:
### Description

This PR adds a warning if a test session could benefit from changing the parallelism.

<!--- 

### DIFF
--- a/metricflow/sql_clients/sql_utils.py
+++ b/metricflow/sql_clients/sql_utils.py
@@ -65,12 +65,17 @@ def create_time_spine_table_if_necessary(time_spine_source: TimeSpineSource, sql
     )
 
 
-def make_sql_client(url: str, password: str) -> SqlClient:
-    """Build SQL client based on env configs. Used only in tests."""
+def dialect_from_url(url: str) -> SqlDialect:
+    """Return the SQL dialect specified in the URL in the configuration."""
     dialect_protocol = make_url(url.split(";")[0]).drivername.split("+")
-    dialect = SqlDialect(dialect_protocol[0])
     if len(dialect_protocol) > 2:
         raise ValueError(f"Invalid # of +'s in {url}")
+    return SqlDialect(dialect_protocol[0])
+
+
+def make_sql_client(url: str, password: str) -> SqlClient:
+    """Build SQL client based on env configs. Used only in tests."""
+    dialect = dialect_from_url(url)
 
     if dialect == SqlDialect.REDSHIFT:
         return RedshiftSqlClient.from_connection_details(url, password)
--- a/metricflow/test/fixtures/setup_fixtures.py
+++ b/metricflow/test/fixtures/setup_fixtures.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import datetime
 import logging
+import warnings
 from dataclasses import dataclass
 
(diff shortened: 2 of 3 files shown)


metricflow/sql_clients/sql_utils.py
```diff
@@ -65,12 +65,17 @@ def make_df(  # type: ignore [misc]
     )
 
 
-def make_sql_client(url: str, password: str) -> AsyncSqlClient:
-    """Build SQL client based on env configs. Used only in tests."""
+def dialect_from_url(url: str) -> SqlDialect:
```
tlento, line 68: Was not expecting this here, it'd be good to put the need for this change in the commit message.

metricflow/test/fixtures/setup_fixtures.py
```diff
@@ -76,7 +80,7 @@ class MetricFlowTestEnvironmentVariables:
 def mf_test_session_state(  # noqa: D
     request: FixtureRequest,
     source_table_snapshot_repository: SqlTableSnapshotRepository,
-) -> MetricFlowTestSessionState:
+) -> Generator[MetricFlowTestSessionState, None, None]:
```
tlento, line 83: I'm very confused as to why this change "just works" - seems to me every call to this fixture should break since you were getting a MetricFlowTestSessionState and now you can't access it without an iterator call.

Is there some Python generator magic I'm missing here?  Or is this pytest fixture + generator magic?

metricflow/test/fixtures/setup_fixtures.py
```diff
@@ -142,6 +146,35 @@ def mf_test_session_state(  # noqa: D
...
+    # 10/30 as the thresholds, but not much thought has been put into it. The following could also be a single
+    # expression, but the nested if structure seemed easier to read.
+    could_benefit_from_parallel_tests = False
+    if num_workers == 1:
+        if dialect is SqlDialect.DUCKDB or dialect is SqlDialect.POSTGRESQL:
+            if num_items > 30:
+                could_benefit_from_parallel_tests = True
+        elif num_items > 10:
+            could_benefit_from_parallel_tests = True
+
```
tlento, line 169: Huh, the threshold layering is a little clunky. Maybe we can do early returns to make those more obvious, something like the below?

Also I think notes about the arbitrary thresholds can go in a docstring instead of inline, that gets picked up by automation a bit better.

```suggestion
    if num_workers > 1 or num_items < 10:
        # Don't warn if we're already using parallelization or test suite is small
        return
    if (dialect is SqlDialect.DUCKDB or dialect is SqlDialect.POSTGRESQL) and num_items < 30:
        # parallelization threshold for DuckDB and Postgres is higher because they are fast
        return
        
```

tlento (approved):
Seems reasonable, left a readability suggestion for you.

I am baffled as to why tests relying on that session state fixture continue to pass with the switch from return to yield.

metricflow/test/fixtures/setup_fixtures.py
plypaul: Updated.

metricflow/sql_clients/sql_utils.py
plypaul: Updated.

PR #561, opened 2023-06-01
duplicate input_measures property to engine.model.Metric

WilliamDee:
### Description
In the DSI implementation of Metric, we populated all the measures associated with the Metric via a property called `input_measures` this is useful and also needed in the public facing object `engine.model.Metric`. So duplicating that logic over to that.

<!--- 


courtneyholcomb (approved):
Beaut ✨

tlento (commented):
Seems reasonable!

We might want to remove the parallel classes altogether at some point and move on to the interface types, but this'll do for now.

PR #562, opened 2023-06-01
Updates to Reflect Changes in DSI 0.1.0.dev4

plypaul:
### Description

This PR updates MF to incorporate the many naming / interface changes in `dbt-semantic-interfaces==0.1.0.dev4`.
<!--- 


plypaul:
I noticed that there are snyk failures whenever there is a change to `pyproject.toml`. I've been having a back-and-forth conversation with snyk support. Anyway, once merged, subsequent PRs don't have the issue.

metricflow/configuration/constants.py
```diff
@@ -16,8 +16,3 @@
 CONFIG_MODEL_PATH = "model_path"
 CONFIG_DWH_HTTP_PATH = "dwh_http_path"
 CONFIG_DWH_ACCESS_TOKEN = "dwh_access_token"
-CONFIG_DBT_REPO = "dbt_repo"
-CONFIG_DBT_PROFILE = "dbt_profile"
-CONFIG_DBT_TARGET = "dbt_target"
```
tlento, line 21: So I'm removing all of the legacy dbt integration locally as well. We will need these three in the future. I'm fine to add them back later, but this is kind of a bummer that we both did this annoying "find all the random integration points and remove them" work.

metricflow/engine/metricflow_engine.py
```diff
@@ -254,31 +248,7 @@ def from_config(handler: YamlFileHandler) -> MetricFlowEngine:
         sql_client = make_sql_client_from_config(handler)
 
         # Ideally we should put this getting of of CONFIG_DBT_REPO in a helper
```
tlento, line 250: Remove comment

metricflow/cli/cli_context.py
```diff
@@ -1,23 +1,17 @@
 from __future__ import annotations
 
 import logging
-from dataclasses import dataclass
```
tlento, line 4: Can we get a "why is this commit here" update to the commit message? It's frustrating to look through git log and see commit messages like "Removed x" or "Updated Y" and then see something odd and have to wonder "why the heck did we do that?"

Also, it's not at all clear to me why this commit is part of this PR. We *should* be removing this code (and I've also done so locally but haven't shipped the stack out for review yet), but it kind of belongs at top level tagged to the relevant issues.

metricflow/model/semantics/semantic_model_lookup.py
```diff
@@ -39,26 +40,26 @@ class SemanticModelLookup(SemanticModelAccessor):
 
     def __init__(  # noqa: D
         self,
-        model: SemanticManifest,
+        model: PydanticSemanticManifest,
```
tlento, line 43: Wait, really? This seems like it'll cause problems later, so I hope it'll get updated somewhere down the line.

tlento (approved):
This seems fine for now. I'll file an issue to move MetricFlow off of concrete implementations - that's going to cause problems at some point.

It's not clear what we need that the interfaces don't already support. Serialization/deserialization is one thing, but that is something that all of the interfaces should have anyway. But also maybe we don't need MetricFlow to do the serialization/deserialization of things internally.

tlento:
> I noticed that there are snyk failures whenever there is a change to pyproject.toml. I've been having a back-and-forth conversation with snyk support. Anyway, once merged, subsequent PRs don't have the issue.

Yeah, this is super annoying. :(

plypaul:
> It's not clear what we need that the interfaces don't already support. Serialization/deserialization is one thing, but that is something that all of the interfaces should have anyway. But also maybe we don't need MetricFlow to do the serialization/deserialization of things internally.

In this PR, I got most of the cases and MF generally uses protocols instead of the implementations. The Pydantic implementation classes are used in the test fixtures because we are parsing them from YAML. But the 2 areas that were problematic to use the protocols were:

* In the data warehouse validations, there were cases where objects were created / mutated, so it needed the concrete implementations. I don't think that should be necessary, but fixing that was a bigger body of work so I left it out.
* There are some useful methods that were needed for the `*Lookup` classes that were only in the Pydantic implementations. Those could be moved to a separate helper class or made a part of the protocol, but left that for later.

metricflow/cli/cli_context.py
plypaul: Context is that this code was going to be removed at some point anyway, and by removing it, it the reduced the dependencies that made the rebase easier, or it enabled more use of protocols vs Pydantic implementations.

tlento:
Cool, makes sense. I filed #567 to track those bigger fixes on the protocol vs implementation thing. Thanks!

metricflow/cli/cli_context.py
plypaul: Updated commit message.

metricflow/model/semantics/semantic_model_lookup.py
plypaul: Addressed in top level comment.

PR #566, opened 2023-06-02
Get dimension values via SQL

WilliamDee:
### Description
There needs to be a way to get dimension values via SQL instead of needing to make a metric query and then process the dimension values in python. In this PR, we introduce a new field to a MetricFlowQueryRequest (that is yet to be surfaced via the CLI/public interfaces). This field is `query_type` which is an enum used to determine whether the query is a metric/dim_vals query. Currently, metric query is how every query works now and dimension values query will basically return ONLY the dimensions requested in the query without the metric data.

<!--- 

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -141,7 +141,7 @@ class DataflowPlanBuilder(Generic[SqlDataSetT]):
         self,
         query_spec: MetricFlowQuerySpec,
         output_sql_table: Optional[SqlTable] = None,
-        filter_column_spec_set: Optional[InstanceSpecSet] = None,
+        output_selection_specs: Optional[InstanceSpecSet] = None,
         optimizers: Sequence[DataflowPlanOptimizer[SqlDataSetT]] = (),
     ) -> DataflowPlan[SqlDataSetT]:
         """Generate a plan for reading the results of a query with the given spec into a dataframe or table."""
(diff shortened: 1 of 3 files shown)


metricflow/engine/metricflow_engine.py
```diff
@@ -523,13 +544,28 @@ def get_dimension_values(  # noqa: D
...
         if result_dataframe is None:
             return []
+        return [str(val) for val in result_dataframe[get_group_by_values]]
 
-        # Process the dimension values
-        result_dataframe.dropna(inplace=True)
-        dimension_values = [str(val) for val in result_dataframe[get_group_by_values]]
-        return dimension_values
+    @log_call(module_name=__name__, telemetry_reporter=_telemetry_reporter)
+    def explain_get_dimension_values(  # noqa: D
```
WilliamDee, line 556: there has to be a better name here, but I can't figure it out. Any advice is appreciated or we can leave it as it 🙈

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -331,25 +333,32 @@ def build_sink_node_from_metrics_output_node(
         order_by_specs: Sequence[OrderBySpec],
         output_sql_table: Optional[SqlTable] = None,
         limit: Optional[int] = None,
+        filter_by_specs: Optional[InstanceSpecSet] = None,
```
plypaul, line 336: Maybe `output_selection_specs`?

plypaul (changes requested):
Looks good overall, but can you add a check so that entities can't be queried as dimension values. Since I'm using github to manage the review queue, sending back with `request changes` even though this is minor.

plypaul (approved):

PR #586, opened 2023-06-08
Specific error for unsupported data platform feature

DevonFulcher:
### Description

This PR replaces a generic `RuntimeError` for a more specific exception, `UnsupportedEngineFeatureError`. This makes it easier for clients to handle this exception. I ran the unit tests locally, but other than that I didn't do much testing.

### DIFF
--- a/metricflow/errors/errors.py
+++ b/metricflow/errors/errors.py
@@ -80,3 +80,12 @@ class ModelCreationException(Exception):
 
 class InferenceError(Exception):
     """Exception to represent errors related to inference."""
+
+class BigQueryApproximatePercentileError(Exception):
+    """Raised when percentile aggregations are used without approximation for BigQuery."""
+
+    def __init__(self) -> None:  # noqa: D
+        super().__init__(
+            "Only approximate continous percentile aggregations are supported for BigQuery. Set "
+            + "use_approximate_percentile and disable use_discrete_percentile in all percentile measures.",
+        )
--- a/metricflow/errors/errors.py
+++ b/metricflow/errors/errors.py
@@ -81,6 +81,7 @@ class ModelCreationException(Exception):
 class InferenceError(Exception):
     """Exception to represent errors related to inference."""
 
+
 class BigQueryApproximatePercentileError(RuntimeError):
     """Raised when percentile aggregations are used without approximation for BigQuery."""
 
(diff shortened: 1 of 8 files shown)


metricflow/errors/errors.py
```diff
@@ -80,3 +80,12 @@ def __init__(self) -> None:  # noqa: D
 
 class InferenceError(Exception):
     """Exception to represent errors related to inference."""
+
+class BigQueryApproximatePercentileError(RuntimeError):
```
tlento, line 85: I prefer to avoid engine-specific errors. Shall we generalize this to something like `UnsupportedEngineFeatureError` instead? The message would have to go back to being passed in at the callsite, as it was before.

This also allows us to use the same custom exception for all of the various percentile cases we have to deal with. For example, [Databricks](https://github.com/dbt-labs/metricflow/blob/main/metricflow/sql/render/databricks.py#L27-L35), [Snowflake](https://github.com/dbt-labs/metricflow/blob/main/metricflow/sql/render/snowflake.py#L39-L43), [Redshift](https://github.com/dbt-labs/metricflow/blob/main/metricflow/sql/render/redshift.py#L31-L40), and [Postgres](https://github.com/dbt-labs/metricflow/blob/main/metricflow/sql/render/postgres.py#L63-L72) all have different levels of support for different percentile aggregation configurations, so we should probably throw the same error for each one.

tlento (commented):
I love the custom exception, RuntimeError for stuff this specific always bothers me, but I think we a more general error will serve us better.

It's probably worth just an empty custom exception type that replaces RuntimeError in all of the percentile mayhem linked in the inline comment. Then we can update that however we like later on.

Note I'm fine with any error type that isn't so engine-specific here. If you'd prefer to limit this to Percentile issues that's reasonable.

metricflow/errors/errors.py
plypaul: +1 for `UnsupportedEngineFeatureError`

plypaul (changes requested):
To move this out of my review queue, marking as `Request changes`.

tlento (approved):
Fantastic! Do you have merge permissions? If so, please use squash and merge with a custom commit message on the squash commit to keep the git log streamlined. Thank you!

DevonFulcher:
Thanks, Tom! I'm not sure if I have merge permissions. Right now, the PR is still blocked by Paul's request for changes. No rush, though!

PR #602, opened 2023-06-13
Clean up legacy constructs in SqlClient interface

tlento:
In order to streamline our addition of a dbt-adapter-backed SqlClient, it's useful to first
tidy up the interface of SqlClient itself. This PR mainly removes unused cruft and forks
certain functionality out into more appropriate locations.

The major news here is dropping support for SqlIsolationLevel, which was a feature
that we never properly supported but could be enabled via custom extensions to a
SqlClient class. Since custom extensions can still enable support for isolation levels
we remove it from the SqlClient as it is outside the scope of MetricFlow's concerns.

PR #601, opened 2023-06-13
Remove introspective queries and on the fly creation of time spine tables

tlento:
MetricFlow uses a time spine table under certain conditions,
such as when cumulative metric computations require us to fill
in missing dates.

The existing time spine logic would, at query planning time, run
some introspective queries to determine if the time spine table
existed, and create one spanning a range from 2000 to 2040.

MetricFlow is moving away from managing this internally. In order
to accelerate our initial integration with dbt, we will remove
this custom introspection and table creation logic from the
MetricFlow runtime.

This will effectively break any cumulative or date offset metrics
in any data warehouse that does not have a pre-existing time
spine data set matching the default MetricFlow configuration (i.e.,
with a sequence of daily granularity dates stored in a column called
`ds` in a table called `mf_time_spine`). We will add the ability for
users to define their own time spine datasets and configure their
MetricFlow execution accordingly in follow-ups.

resolves #592

PR #604, opened 2023-06-14
Eliminate engine attribute structs from SqlClient protocol

tlento:
The SqlClient protocol originally included a struct of engine properties which
was used to indicate certain engine-specific attributes which could be useful
to know about in various parts of the codebase where the SqlClient itself was
not required. The direct impetus of this construction was the need for a
backwards-compatible approach to the pre-MetricFlow implementation inside
Transform.

As time has gone by, this property struct has become increasingly cumbersome,
serving as a living record of all of the engine-specific quirks we've ever
encountered, and conflating various dialect-specific rendering elements with
core engine capabilities and python API implementation quirks.

Now that we are narrowing MetricFlow's scope of responsibility for SqlClient
state and capability management, it has become clear that these engine-specific
property structs have outlived their usefulness. The majority of engine-specific
capability checks are either superfluous (e.g., FULL OUTER JOIN support is now
standard) or were rendered irrelevant (e.g., we no longer support isolation levels).
The remaining attributes are more appropriate as part of dialect rendering support,
leaving only the need for the SqlClient to hold pointers to the engine type and the
dialect renderer.

This PR makes this change, eliminating the cumbersome secondary struct, and moving
capability management to the components most appropriate for handling it.

### DIFF
--- a/.changes/unreleased/Breaking Changes-20230614-173609.yaml
+++ b/.changes/unreleased/Breaking Changes-20230614-173609.yaml
@@ -1,0 +1,7 @@
+kind: Breaking Changes
+body: Remove SqlEngineAttributes construct from SqlClient interface in favor of dialect
+  rendering and engine type properties
+time: 2023-06-14T17:36:09.827638-07:00
+custom:
+  Author: tlento
+  Issue: "577"
--- a/metricflow/cli/cli_context.py
+++ b/metricflow/cli/cli_context.py
@@ -113,7 +113,7 @@ class CLIContext:
                     err_string = str(e)
                     logger.error(f"Health Check Item {step}: failed with error {err_string}")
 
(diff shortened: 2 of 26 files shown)


tlento:
* main
  * **PR #601** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/601" target="_blank">Graphite</a> 
    * **PR #602** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/602" target="_blank">Graphite</a> 
      * **PR #604** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/604" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dbt-labs/metricflow/604?utm_source=stack-comment).

tlento:
Note: I will not be extending override annotations within existing SqlClient implementations because all of them are going away soon.

plypaul (approved):

PR #608, opened 2023-06-20
Use Metrics Instead of Measures To Define Ratio Metrics

### LINKED ISSUE #504: [Bug] Remove expr & ratio metrics and bundle with derived metrics

plypaul:
Resolves #504

### Description

When defining ratio metrics, measures are currently used to define them. To provide the most flexibility, this PR changes ratio metrics to use metrics when defining them instead.

<!--- 

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -6,6 +6,7 @@ import time
 from dataclasses import dataclass
 from typing import DefaultDict, Dict, Generic, List, Optional, Sequence, Set, Tuple, TypeVar, Union
 
+from dbt_semantic_interfaces.enum_extension import assert_values_exhausted
 from dbt_semantic_interfaces.pretty_print import pformat_big_objects
 from dbt_semantic_interfaces.protocols.metric import MetricTimeWindow, MetricType
 from dbt_semantic_interfaces.references import TimeDimensionReference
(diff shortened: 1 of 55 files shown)


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -245,6 +246,8 @@ def _build_metrics_output_node(
                     metric_spec=metric_spec,
                     aggregated_measures_node=aggregated_measures_node,
                 )
+            else:
+                assert_values_exhausted(metric.type)
```
tlento, line 250: YES!

metricflow/engine/models.py
```diff
@@ -47,20 +48,13 @@ def from_pydantic(cls, pydantic_metric: SemanticManifestMetric, dimensions: List
...
-        type_params = self.type_params
-
-        measures: List[MetricInputMeasure] = list(
-            type_params.measures if type_params is not None and type_params.measures is not None else []
+        assert self.type_params.input_measures, (
+            f"Metric {self.name} should have had input_measures populated by "
+            f"{AddInputMetricMeasuresRule.__class__.__name__}"
         )
-        if type_params.measure:
-            measures.append(type_params.measure)
```
tlento, line 58: Did we remove this from DSI? Or is the input_measures property automatically wrapping this?

metricflow/test/fixtures/model_yamls/simple_model/metrics.yaml
```diff
@@ -4,182 +4,160 @@ metric:
   description: "bookings metric"
   type: simple
   type_params:
-    measures:
-      - bookings
+    measure: bookings
```
tlento, line 7: I guess this is getting wrapped by the DSI property accessor or something? Or we're populating input_measures from it?

tlento (approved):
Seems reasonable as long as engine tests pass!

metricflow/engine/models.py
plypaul: Can you clarify what you mean by `remove this`?

tlento: Sorry, I meant move this logic to DSI, I have no idea how that sentence came out that way.

tlento:
Looks like this needs a snapshot update again, but should be good after that.

PR #613, opened 2023-06-21
Move snapshot files to enumerated paths

tlento:
MetricFlow has sets of query planning and dialect rendering tests
that compare against snapshot files representing plans and SQL
queries. For dialect-specific constructs, we rely on a dialect-
specific snapshot file. For historical reasons, these were stored
in a path location bound to the SqlClient class used in the tests.

Now that we are moving to a model where we delegate this work to
dbt adapters, we will be re-using the same SqlClient class for
multiple dialects. More generally, we'd like to separate our
warehouse client from our dialect operations. So this commit
moves the relevant files to a path based on the target engine
property of the SqlClient interface. Since our dialects currently
map on to supported engines 1:1, we can do this for now, and
eventually convert to dialect-only differentiation in the future.

Note - this was done via file moves in the IDE, which seem to
be the equivalent of `git mv` under the hood.


tlento:
* main
  * **PR #613** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/613" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dbt-labs/metricflow/613?utm_source=stack-comment).

tlento:
DuckDB and Postgres passed locally, running other warehouse tests in CI, will fix up anything that goes awry.

plypaul (approved):

PR #614, opened 2023-06-22
Use dbt postgres adapter wrapper for integration tests

tlento:
As part of the effort to provide basic metric querying capabilities
for local installations of dbt core, MetricFlow is removing all
engine-specific dependencies and instead delegating warehouse
connection and query management to dbt adapters for integration
tests and command line operations.

This commit is the first step towards moving onto dbt adapters. It
provides a new wrapper class for dbt adapters, conforming to
MetricFlow's SqlClient protocol, and delegating all operations
down to the underlying adapter. It uses this adapter in test cases
only, and sources it from a dummy project in the test fixtures
directory. The dummy project is configured to allow for different
adapters (configured through test runner environment variables), and
in future it will likely be used as the basis for MetricFlow's
core integration test table management.

Note the implementation of the SqlClient wrapper class has some
highly temporary work-arounds to certain rough edges that exist
between dbt's adapter interface and the current SqlClient requirements.
Most of these arise from table and schema management calls, which are
only used in the MetricFlow CLI tutorial and integration test suite.
As such, we expect those methods to be removed from the SqlClient
interface as we move management of these datasets to dbt.

Further, the dbt adapter fetching is a bit suspect, and not to be
copied - we are relying on the dbtRunner()'s state loading to remain
valid as we access certain internal APIs from dbt core to get the
profile we need. All of this works, and we expect it to continue
working, but we will need to maintain this connection and keep it
up to date as the dbt core project updates its external accessors for
the dbt profile and dbt project which ultimately initialize and house
the adapter instance.

### DIFF
--- a/metricflow/cli/dbt_connectors/adapter_backed_client.py
+++ b/metricflow/cli/dbt_connectors/adapter_backed_client.py
@@ -1,0 +1,308 @@
+from __future__ import annotations
+
+import logging
+import textwrap
+import time
+from typing import Optional, Sequence
+
+import pandas as pd
+from dbt.adapters.base.impl import BaseAdapter
+from dbt_semantic_interfaces.pretty_print import pformat_big_objects
+
+from metricflow.dataflow.sql_table import SqlTable
+from metricflow.errors.errors import SqlBindParametersNotSupportedError
+from metricflow.logging.formatting import indent_log_line
+from metricflow.protocols.sql_client import SqlEngine
+from metricflow.protocols.sql_request import SqlJsonTag, SqlRequestId, SqlRequestTagSet
+from metricflow.random_id import random_id
+from metricflow.sql.render.postgres import PostgresSQLSqlQueryPlanRenderer
+from metricflow.sql.render.sql_plan_renderer import SqlQueryPlanRenderer
+from metricflow.sql.sql_bind_parameters import SqlBindParameters
+from metricflow.sql_clients.sql_statement_metadata import CombinedSqlTags, SqlStatementCommentMetadata
+
+logger = logging.getLogger(__name__)
+
+
+class AdapterBackedSqlClient:
+    """SqlClient implementation which delegates database operations to a dbt BaseAdapter instance.
+
+    This is a generic wrpaper class meant to cover all of our logging, querying, and internal configuration
+    needs while delegating all connection state management and warehouse communication work to an underlying
+    dbt adapter instance. This relies on BaseAdapter, rather than SQLAdapter, because BigQuery is an instance
+    of the more generic BaseAdapter class.
+    """
+
+    def __init__(self, adapter: BaseAdapter):
+        """Initializer sourced from a BaseAdapter instance.
+
+        The dbt BaseAdapter should already be fully initialized, including all credential verification, and
+        ready for use for establishing connections and issuing queries.
+        """
+        if adapter.type() != "postgres":
+            raise ValueError(
+                f"Received dbt adapter with unsupported type {adapter.type()}, but we only support postgres!"
+            )
+        self._adapter = adapter
+        # TODO: normalize from adapter.type()
+        self._sql_engine_type = SqlEngine.POSTGRES
+        # TODO: create factory based on SqlEngine type
+        self._sql_query_plan_renderer = PostgresSQLSqlQueryPlanRenderer()
+
+    @property
+    def sql_engine_type(self) -> SqlEngine:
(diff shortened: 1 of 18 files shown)


tlento:
* main
  * **PR #613** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/613" target="_blank">Graphite</a> 
    * **PR #614** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/614" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dbt-labs/metricflow/614?utm_source=stack-comment).

tlento:
snyk failures are expected due to pyproject.toml changes

plypaul (approved):

plypaul (commented):
Sorry, clicked approve a little too soon. Taking a look still.

metricflow/cli/dbt_connectors/adapter_backed_client.py
```diff
@@ -0,0 +1,308 @@
...
+
+            # Insert rows
+            values = []
+            for row in df.itertuples(index=False, name=None):
+                cells = []
+                for cell in row:
+                    if pd.isnull(cell):
+                        # use null keyword instead of isNA/None/etc.
+                        cells.append("null")
+                    elif type(cell) in [str, pd.Timestamp]:
```
plypaul, line 203: Generally works, but I vaguely recall running into type errors with one SQL engine where specifying a string when a time is expected threw an error. Cross that bridge if / when we get there.

metricflow/cli/dbt_connectors/adapter_backed_client.py
```diff
@@ -0,0 +1,308 @@
...
+            f"\n\n{indent_log_line(stmt)}\n"
+            + (f"\nwith parameters: {dict(sql_bind_parameters.param_dict)}" if sql_bind_parameters.param_dict else "")
+        )
+        # TODO - rely on self._adapter.dry_run() when it is available so this will work for BigQuery.
+        self.execute(f"EXPLAIN {stmt}")
+        stop = time.time()
+        logger.info(f"Finished running the dry_run in {stop - start:.2f}s")
+        return
+
+    def create_table_from_dataframe(
```
plypaul, line 159: This is straightforward, but yet I have a suspicion that we might run into edge cases. Just a comment, no action needed.

pyproject.toml
```diff
@@ -100,6 +100,20 @@ features = [
   "dev-packages",
 ]
 
+[tool.hatch.envs.postgres-env]
+description = "Dev environment for working with Postgres adapter"
+pre-install-commands = [
+  "pip install dbt-postgres",
```
plypaul, line 106: What's the reason for installing via pip instead of specifying it as a dependency?

metricflow/test/sql_clients/test_sql_client.py
```diff
@@ -40,6 +41,8 @@ def test_query(sql_client: SqlClient) -> None:  # noqa: D
 
 def test_query_with_execution_params(sql_client: SqlClient) -> None:
     """Test querying with execution parameters of all supported datatypes."""
+    if isinstance(sql_client, AdapterBackedSqlClient):
```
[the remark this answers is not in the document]
tlento: Once all of our production engines are cut over to adapters I'll either remove these tests or keep them solely for a DuckDB "example" implementation of a SQLClient. I don't think the adapter-backed client will support bind parameters any time soon.

If we don't have one already we should add a test to make sure bind parameters are preserved and propagated through the rendering layers, though. I know they are, generally, but I don't remember what kind of testing we have around that behavior.

pyproject.toml
tlento: I could've sworn I had a comment about this, but clearly there isn't one. I'll add one before merging.

We don't depend on dbt-postgres, or any specific adapter, anywhere in the codebase, so we don't want to force an install in the normal dependency list. We could make it an env-specific dependency but then we're bound to a version range, which means keeping the version ranges in sync between here and where we specify dbt-core, and dealing with things like dbt-semantic-interfaces version requirements.

This way we get the relevant adapter installed in the environment, the way an end user would, and pip takes care of the rest as best it's able.

Once we have all of the packaging laid out with dbt-metricflow we'll likely update this to use an editable dependency on the local package. The pre-install is apparently the only way hatch supports editable dependency installations, so that'll have to be here anyway, but that should allow us to work around dbt-semantic-interfaces version update conflicts in most circumstances.

PR #616, opened 2023-06-22
Add Tooling + Tests to Manage the Persisted Source Schema

plypaul:
### Description

Running tests using an external SQL engine is slow, and a big reason for the slowness is that it takes a while to create the source tables that are queried by the tests. A `--use-persisted-source-schema` flag was previously added to keep the source tables around. When used, the flag would generate a schema name from a hash of all source tables, and see if tables exist in that schema. If tables were missing, tables would be created from snapshots. At the end of the testing session, that schema would not be dropped.

However, there were some potential race conditions when two separate processes create the schema and try to populate tables for the first time. To avoid this, this PR adds a command that can be run by the developer to create and populate the persistent source schema. The developer can ensure that the command is run only by a single process. For a given hash of the total source table data, the command only needs to be run once.

In addition, a test was added to validate that the source schema contains tables with the expected data. A failure of this test will generate warnings suggesting to the developer to run the table population command.

<!--- 


tlento (approved):
Seems reasonable for now!

PR #617, opened 2023-06-23
`explain_get_dimension_values` & `get_dimension_values` takes metrics lists

DevonFulcher:
### Description

Fixes SLA-116 (private to dbt Labs)

This change allows `explain_get_dimension_values` & `get_dimension_values` to take a list of metrics parameters to allow for more flexible integrations, specifically in the Jinja interface.

### DIFF
--- a/metricflow/engine/metricflow_engine.py
+++ b/metricflow/engine/metricflow_engine.py
@@ -269,8 +269,8 @@ class AbstractMetricFlowEngine(ABC):
     @abstractmethod
     def explain_get_dimension_values(  # noqa: D
         self,
-        metric_name: str,
-        get_group_by_values: str,
+        metric_names: List[str],
+        get_group_by_values: List[str],
         time_constraint_start: Optional[datetime.datetime] = None,
         time_constraint_end: Optional[datetime.datetime] = None,
     ) -> MetricFlowExplainResult:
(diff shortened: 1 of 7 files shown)


metricflow/engine/metricflow_engine.py
```diff
@@ -269,8 +269,8 @@ def get_dimension_values(
     @abstractmethod
     def explain_get_dimension_values(  # noqa: D
         self,
-        metric_name: str,
-        get_group_by_values: str,
+        metric_names: List[str],
+        get_group_by_values: List[str],
```
WilliamDee, line 273: I don't think we want to be able to pass multiple group bys.

tlento (commented):
I don't have strong opinions on whether or not we should allow multiple metrics and dimensions in our dimension values accessors, but if we open this up it seems like we should apply it to both the explain command and the real command, at least as far as the external API is concerned.

I think it's worth noting this change is only breaking as far as calling services go - it will not affect any of the core dbt integrations - so we don't necessarily need to get this in before launch unless the feature it supports is part of the launch requirements.

WilliamDee (approved):
Looks good! Idk how much we use the changelog but the one you pushed is out of date from the new changes.

PR #619, opened 2023-06-23
Create dbt-metricflow package

WilliamDee:
### Description
Create the initial `dbt-metricflow` package to be published onto pypi. The use case of this package is to bundle the versioning for `metricflow` and `dbt-core` such that they would work together in parallel and allow the user to use the `mf` and `dbt` CLI. Also, has optional extras installs for the different dbt adapters that are integrated with MetricFlow.

### Short term
This package will work to manage the versioning between `metricflow` and `dbt-core` and the `dbt-adapters` such that they work nicely together. 

### Long term
This package will include the MetricFlow CLI (and possibly the dbt Core CLI) code and allow MetricFlow to rip out the dependency on `dbt-core`

Resolves SEM-392

<!--- 

### DIFF
--- a/dbt-metricflow/README.md
+++ b/dbt-metricflow/README.md
@@ -1,0 +1,2 @@
+# Welcome to dbt-metricflow
+
--- a/dbt-metricflow/README.md
+++ b/dbt-metricflow/README.md
@@ -1,2 +1,18 @@
+<p align="center">
+    <a target="_blank" href="https://twitter.com/dbt_labs">
+    <img src="https://img.shields.io/twitter/follow/dbt_labs?labelColor=image.png&color=163B36&logo=twitter&style=flat">
+  </a>
+    <a target="_blank" href="https://www.getdbt.com/community/">
+    <img src="https://img.shields.io/badge/Slack-join-163B36">
+  </a>
+    <a href="https://github.com/psf/black"><img src="https://img.shields.io/badge/code%20style-black-000000.svg" /></a>
+</p>
+
 # Welcome to dbt-metricflow
 
+This repo encapsulates the dbt-core, MetricFlow, and supported dbt-adapters packages. This package will manage the versioning between these packages such that they are compatible with each other.
+
+## Repo use cases
+- dbt-core and MetricFlow both depend on dbt-semantic-interfaces, which includes the schemas and interfaces for objects related to the semantic layer. Bundled versioning is necessary to ensure that the dbt-core and MetricFlow versions have compatible dbt-semantic-interfaces dependencies.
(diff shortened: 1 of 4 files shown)


dbt-metricflow/README.md
```diff
@@ -0,0 +1,19 @@
...
+  </a>
+    <a target="_blank" href="https://www.getdbt.com/community/">
+    <img src="https://img.shields.io/badge/Slack-join-163B36">
+  </a>
+    <a href="https://github.com/psf/black"><img src="https://img.shields.io/badge/code%20style-black-000000.svg" /></a>
+</p>
+
+# Welcome to dbt-metricflow
+
+This repo encapsulates the dbt-core and MetricFlow and dbt-adapters packages. This package will manage the versioning between the 3 packages such that they are compatible with each other.
```
tlento, line 13:
```suggestion
This repo encapsulates the dbt-core, MetricFlow, and supported dbt-adapters packages. This package will manage the versioning between these packages such that they are compatible with each other. 
```

dbt-metricflow/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+# Welcome to dbt-metricflow
+
+This repo encapsulates the dbt-core and MetricFlow and dbt-adapters packages. This package will manage the versioning between the 3 packages such that they are compatible with each other. 
+
+
+## Repo use cases
+- dbt-core and MetricFlow both depend on dbt-semantic-interfaces which includes the schemas and interfaces for objects related to the semantic layer. If there's a inconsistency between the interfaces, dbt-core and MetricFlow would likely error out and not work. So bundled versioning is necessary.
+- Bundling the installation makes it much easier on the end-users side as you would not need to install `dbt-core` + `metricflow` + `dbt-adapter`, but instead just `dbt-metricflow[adapter]`
+- Because this encapsulates both dbt-core and MetricFlow, this repo can be used to build logic that are shared amongst the 3 packages. For example, the CLI from MetricFlow can live in this repo as it uses logic from all 3 packages which would reduce the dependency cluttering on MetricFlow.
```
tlento, line 19:
```suggestion
- dbt-core and MetricFlow both depend on dbt-semantic-interfaces, which includes the schemas and interfaces for objects related to the semantic layer. Bundled versioning is necessary to ensure that the dbt-core and MetricFlow versions have compatible dbt-semantic-interfaces dependencies.
- Bundling the installation makes it much easier on end users, as they no longer need to install `dbt-core` + `metricflow` + `dbt-adapter` and reconcile versions - instead they can simply install `dbt-metricflow[adapter]`.
- Because this encapsulates both dbt-core and MetricFlow, this repo can be used to build logic that should be shared across the packages. For example, the CLI from MetricFlow can live in this repo, as it uses logic from all of the bundled packages. This can streamline dependency requirements in MetricFlow.
```

tlento (approved):
Looks great, thanks @WilliamDee !

Let's get this deployed and see what happens with the suddenly resurgent resource warnings. It's so odd....

PR #620, opened 2023-06-23
Use Persistent Source Schema for CI Tests

plypaul:
### Description

Since the race condition with the previous setup for `--use-persistent-source-schema` has been resolved, enabling the flag for CI runs. With this change, SQL engine tests can finish in <4 mins:

https://github.com/dbt-labs/metricflow/actions/runs/5360383668/jobs/9725634762

<!--- 

### DIFF
--- a/.github/actions/setup-python-env/action.yaml
+++ b/.github/actions/setup-python-env/action.yaml
@@ -33,11 +33,12 @@ runs:
         ~/.cache/hatch
         ~/.cache/pre-commit
       # >- means combine all lines to a single line
+      # The cache key can be any string. The format used here is just for readability.
       key: >-
-        python_location="${{ env.pythonLocation }}"/
-        pyproject_hash="${{ hashFiles('pyproject.toml') }}"/
-        precommit_config_hash="${{ hashFiles('.pre-commit-config.yaml') }}"/
-        linux_release="${{ env.LINUX_RELEASE }}"}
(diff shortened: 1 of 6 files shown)


.github/actions/setup-python-env/action.yaml
```diff
@@ -33,11 +33,12 @@ runs:
...
+      # The cache key can be any string. The format used here is just for readability.
       key: >-
-        python_location="${{ env.pythonLocation }}"/
-        pyproject_hash="${{ hashFiles('pyproject.toml') }}"/
-        precommit_config_hash="${{ hashFiles('.pre-commit-config.yaml') }}"/
-        linux_release="${{ env.LINUX_RELEASE }}"}
+        (python_location=="${{ env.pythonLocation }}" AND
+        pyproject_hash=="${{ hashFiles('pyproject.toml') }}" AND
+        precommit_config_hash=="${{ hashFiles('.pre-commit-config.yaml') }}" AND
+        linux_release=="${{ env.LINUX_RELEASE }}")
```
tlento, line 41: The update with == makes it look like the final string is going to be TRUE AND TRUE AND TRUE AND TRUE, but I don't particularly care about this since the final string is what really matters.

That said, maybe we don't use = and instead go with `python_location: "${{ env.pythonLocation }}" AND ....`

tlento (approved):

PR #622, opened 2023-06-26
Apply semantic manifest transformations to json parsed output

tlento:
The dbt-generated semantic manifest is not transformed, and as such
it fails at query time with inscrutable errors about missing metrics
which clearly appear in all of the manifest artifacts and runtime
objects.

This change applies the semantic manifest transformations to all
dbt-sourced json-serialized semantic manifests, and embeds the call
inside of the dbt-specific function in order to prevent confusion.

Longer term, we should refine these APIs and unify them across
dbt-semantic-interfaces and MetricFlow so that we can always be
clear about whether or not transformations are needed for a given
serialized object. This change eschews the boolean parameter approach
because we only have one mode for json parsing at this time. If we
add a second approach, this would likely benefit from separate
methods rather than passing in an "apply_transformations" flag.

### DIFF
--- a/.changes/unreleased/Fixes-20230626-110103.yaml
+++ b/.changes/unreleased/Fixes-20230626-110103.yaml
@@ -1,0 +1,7 @@
+kind: Fixes
+body: Apply transformations to dbt-generated serialized model to fix issue with query
+  generation
+time: 2023-06-26T11:01:03.898209-07:00
+custom:
+  Author: tlento
+  Issue: "624"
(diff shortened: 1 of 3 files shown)


tlento:
* main
  * **PR #622** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/622" target="_blank">Graphite</a>  👈
    * **PR #623** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/623" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dbt-labs/metricflow/622?utm_source=stack-comment).

WilliamDee (approved):
Ah good catch. Thanks!

PR #623, opened 2023-06-26
Enable dbt AdapterBackedSqlClient queries from the MetricFlow CLI

tlento:
The CLI was parsing semantic manifests from the dbt project root
but then falling back to MetricFlow's config-backed SqlClient
initialization, which meant users were effectively required to
match their legacy MetricFlow config with their dbt project profile
adapter configuration.

In practice, the `query`, `list dimension-values`, and warehouse
validation steps for the `validate-configs` commands were all
effectively broken for any new user attempting to use the CLI
with the dbt project integration.

This PR enables CLI queries for all supported dbt adapters
(currently Postgres, but more to follow shortly) for any
user with a minimally valid MetricFlow config.

A follow-up PR will remove all dependencies on the legacy
MetricFlow config, which is the last remaining step before
dbt users can simply use the installer, invoke `dbt run`, and
be able to execute `mf` commands from there.

### DIFF
--- a/.changes/unreleased/Features-20230626-133439.yaml
+++ b/.changes/unreleased/Features-20230626-133439.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Use dbt adapter to run queries and warehouse validations from MetricFlow CLI
+time: 2023-06-26T13:34:39.613243-07:00
+custom:
+  Author: tlento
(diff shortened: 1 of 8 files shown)


WilliamDee (approved):
Nice!! Thanks for cleaning up some stuff too

PR #633, opened 2023-06-28
Refactor Tutorial to work with dbt integration

WilliamDee:
### Description
Currently the CLI is disabled due to the integration with dbt. Refactoring the tutorial to work well with a sample dbt project.

#### Changes
- Changed Health checks to only do a `SELECT 1` just to test the connection with the adapter
- Currently, for every `mf` command except `mf tutorial`, will require the user to be in the root directory of a dbt project (this is done by checking if `dbt_projects.yml` exists)
- Added `dbtProjectMetadata` so that we can get the information from `profiles.yml` and `dbt_project.yml` without needing to instantiate the manifest and adapter.

#### Tutorial
The current flow as a user is,
1. call `mf tutorial`
2. if no dbt project exists, prompt user to run `dbt init`
3. once we have a dbt project and the user is in the root directory of the project, run `mf tutorial` again
4. MetricFlow checks the validity of the project (ie., credentials are there) and we prompt the user a confirmation to allow for generating sample models into their project. This includes
   - semantic_manifest.json
   - sample models in the model path
   - sample seeds in the seed path
5. user must run `dbt build` to load the seeds
6. user can now step through the steps in the tutorial to execute `mf` commands

<!--- 
  7. signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
  8. run `changie new` to [create a changelog entry](https://github.com/dbt-labs/metricflow/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)


Jstein77 (approved):
@WilliamDee pushed a commit up with updates to the staging models and tutorial copy.  I ran through the tutorial end to end and everything looks good! A few notes:

1. Can we add a command to delete the sample models and seeds? I added this step at the end of the tutorial `When you're done with the tutorial, run mf tutorial --clean to delete sample models and seeds. ` Is it easy to implement `mf tutorial --clean?` If it's too much work we can tell users to delete the sample directories manually
2. If someone runs `mf tutorial` and there are already sample models they get this error
Screenshot 2023-06-28 at 10 33 13 AM. Should we just throw a warning and ask the user if they want to overwrite them?

metricflow/cli/cli_context.py
```diff
@@ -81,21 +80,8 @@ def sql_client(self) -> SqlClient:  # noqa: D
...
-                (
-                    f"Create table '{schema_name}.{table_name}' with a SELECT",
-                    lambda: self.sql_client.execute(
-                        f"CREATE TABLE {schema_name}.{table_name} AS SELECT 'test' AS test_col"
-                    ),
-                ),
-                (
-                    f"Drop table '{schema_name}.{table_name}'",
-                    lambda: self.sql_client.drop_table(SqlTable(schema_name=schema_name, table_name="health_report")),
-                ),
```
tlento, line 98: 🙌

metricflow/cli/sample_dbt_models/sample_models/countries.sql
```diff
@@ -0,0 +1,3 @@
+select
+    *
+from {{source('tutorial', 'countries_seed')}}
```
tlento, line 3: Huh, I always thought jinja required the spaces, like `{{ source(...) }}`. Is that just a readability thing?

metricflow/cli/utils.py
```diff
@@ -234,3 +234,25 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:  # type: ignore[misc]
...
     return wrapper
+
+
+def dbt_project_file_exists() -> bool:
+    """Check that the cwd is a dbt project root. Currently done by checking for existence of dbt_project.yml."""
+    return pathlib.Path("dbt_project.yml").exists()
+
+
+def error_if_not_in_dbt_project(func: Callable) -> Callable:
+    """Decorator to output an error message and exit if caller is not in a root directory of a dbt project."""
```
tlento, line 245: Nifty, this way we can update the check as more of this functionality is exposed from dbt-core.

metricflow/cli/dbt_connectors/dbt_config_accessor.py
```diff
@@ -23,22 +23,58 @@
 
 
 @dataclasses.dataclass
-class dbtArtifacts:
-    """Container with access to the dbt artifacts required to power the MetricFlow CLI."""
+class dbtPaths:
```
tlento, line 26: Nice, these are handy.

tlento (commented):
I just broke your PR, sorry @WilliamDee .

This all seems reasonable to me. It would be good to address @Jstein77 's comments about cleanup and handling the state where the models already exist. If that's complicated maybe we just update the error message and do the proper fix in a follow-up.

PR #638, opened 2023-06-28
Fix tutorial loading with dbt project

tlento:
During some live testing of the tutorial we discovered two seemingly
separate but closely connected issues. First, running the tutorial
from outside of a dbt project root filled the terminal with error
trace output, and second the dreaded ResourceWarning was appearing
after certain commands.

The base of the issue was due to us having to initialize the dbt project
at the very start of the CLI invocation, before any of our handlers
were active, and then accidentally initializing it a second time
as part of the tutorial operations.

This PR fixes both issues by first ensuring that we will simply not
run the CLI at all if we are not in a dbt project root, and along
the way restructuring our dbt artifact loading stack such that
we only invoke the dbtRunner one time per process.

### DIFF
--- a/metricflow/cli/cli_context.py
+++ b/metricflow/cli/cli_context.py
@@ -8,7 +8,7 @@ from typing import Dict, Optional
 from dbt_semantic_interfaces.protocols.semantic_manifest import SemanticManifest
 
 from metricflow.cli.dbt_connectors.adapter_backed_client import AdapterBackedSqlClient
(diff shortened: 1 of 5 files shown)


tlento:
* main
  * **PR #638** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/638" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dbt-labs/metricflow/638?utm_source=stack-comment).

WilliamDee (approved):

PR #642, opened 2023-06-30
change group-bys to group-by in the CLI

WilliamDee:
### Description
It is a little inconsistent to use `--group-bys` when `--group-by` is used in cases of grouping by single or multiple elements.

<!--- 


tlento (approved):
Is there a similar issue with the API class? If so might be nice to align them, but I don't think it's super important right now - we still need to decide what to do about the Python API wrappers in the longer term.

WilliamDee:
> Is there a similar issue with the API class? If so might be nice to align them, but I don't think it's super important right now - we still need to decide what to do about the Python API wrappers in the longer term.

We haven't even touched the api wrappers. But should definitely align those once we figure out what we want to with thise

PR #644, opened 2023-07-05
Raise `UnknownMetricLinkingError` if a Metric isn't found during linking

DevonFulcher:
### Description

Fixes SLA-158 (internal to dbt Labs)

Raise a more specific exception when a Metric isn't found during linking. This will help with surfacing errors in the Jinja syntax. I only found one other instance of `Unknown metric` [here](https://github.com/dbt-labs/metricflow/blob/5a9e917434b1dbcbd1017ffa2e6635b84dc1c24e/metricflow/query/query_parser.py#L589), but that didn't seem related, so I didn't reuse or change that exception. This change shouldn't cause any breaking changes because `UnknownMetricLinkingError` inherits from `ValueError`


tlento (approved):
LGTM, thanks!

PR #645, opened 2023-07-05
Remove legacy tutorial helper functions

tlento:
The original tutorial relied on dataframe-based table creation,
and cleaned up tables by calling drop_tables on the sql client.

We have switched to managing the tutorial datasets with dbt under
the hood, and we are pushing to deprecate the DDL operations within
the SqlClient interface, so we remove these legacy helpers in
order to cut down on callsites for the soon-to-be-removed SqlClient
methods.

PR #647, opened 2023-07-05
Remove layer of indirection from CLI error messages

tlento:
Currently, error messages in the CLI that result from failing
warehouse health checks or any problem with the model throw a
wrapped exception. In practice, this means a user with a misconfigured
model will see an error of the form:

`ERROR: An error occurred when attempting to create the MetricFlow engine`

along with a pointer to the log file, which will contain a much more
helpful error describing the root cause of the problem.

This change removes the wrapper so that the direct cause of the
exception will be surfaced in the CLI output itself, so users
will instead see exceptions that read something more like:

```
ERROR: Metric `east_coast_order_amount` references measure `MeasureReference(element_name='orderss')` which has not been registered
```

### DIFF
--- a/.changes/unreleased/Fixes-20230705-182137.yaml
+++ b/.changes/unreleased/Fixes-20230705-182137.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Improve error message rendering in MetricFlow CLI
+time: 2023-07-05T18:21:37.717453-04:00
+custom:
+  Author: tlento
+  Issue: "646"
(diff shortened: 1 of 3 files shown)


tlento:
* main
  * **PR #645** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/645" target="_blank">Graphite</a> 
    * **PR #647** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/647" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dbt-labs/metricflow/647?utm_source=stack-comment).

WilliamDee (approved):

PR #656, opened 2023-07-13
Update to Support `dbt-semantic-interfaces==0.1.0rc1`

plypaul:
### Description

This PR makes updates as required to support `dbt-semantic-interfaces==0.1.0rc1`. This shouldn't be merged until there is a version of `dbt-core` that support the same version of `dbt-semantic-interfaces` as there is currently dependency conflict.

<!--- 

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "metricflow"
-version = "0.200.0.dev12"
+version = "0.200.0.dev13"
 description = "Translates a simple metric definition into reusable SQL and executes it against the SQL engine of your choice."
 readme = "README.md"
 requires-python = ">=3.8,<3.12"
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -29,8 +29,9 @@ dependencies = [
   "SQLAlchemy~=1.4.42",
   "click>=7.1.2, <8.1.4",
   "databricks-sql-connector~=2.0",
-  "dbt-core==1.6.0b8",
-  "dbt-semantic-interfaces==0.1.0.dev8",
+  # TODO: Remove once there is a version of dbt-core that supports dbt-semantic-interfaces==0.1.0rc1
+  # "dbt-core==1.6.0b8",
+  "dbt-semantic-interfaces==0.1.0rc1",
```
QMalcolm, line 34: Just as an FYI, in core we're doing `dbt-semantic-interfaces~=0.1.0rc1`. I don't think pinning to `0.1.0rc1` here will be problematic though as we should be past breaking changes at this point

plypaul: Updated to use `~`.

QMalcolm (approved):

PR #657, opened 2023-07-13
Bump dbt and metricflow deps

gwenwindflower:


gwenwindflower:
team is going to cut a fresh RC on monday so we'll close this

Issue #665, opened 2023-07-19
[Feature] Add link to GitHub in PyPI missing for dbt-metricflow


dbeatty10:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing metricflow functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

The PyPI page for [`metricflow`](https://pypi.org/project/metricflow/) includes a link to the repo in GitHub (via [this](https://github.com/dbt-labs/metricflow/blob/10c8feeb8f0666c9d847098aaa248e752f3ed173/pyproject.toml#L65-L67) configuration):

<img width="141" alt="image" src="https://github.com/dbt-labs/metricflow/assets/44704949/624536dc-5f51-4ddf-a41f-244da451907b">

But the page for `dbt-metricflow` does not (since it does not include the relevant config [here](https://github.com/dbt-labs/metricflow/blob/10c8feeb8f0666c9d847098aaa248e752f3ed173/dbt-metricflow/pyproject.toml)).

### Who will this benefit?

Anyone starting at PyPI and trying to trace backwards to the source code repository.

### Are you interested in contributing this feature?

yep 😎 -- will need feedback for the desired URLs tho

dbeatty10:
Resolved via https://github.com/dbt-labs/metricflow/pull/667

PR #667, opened 2023-07-19
URL for dbt-metricflow for PyPI

dbeatty10:
Resolves Issue #665

### Description

See details in #665.

We can optionally change the URL as desired. I did not also add a link to a documentation website like [here](https://github.com/dbt-labs/metricflow/blob/10c8feeb8f0666c9d847098aaa248e752f3ed173/pyproject.toml#L66), but can do so if asked to.

I did not include a changelog entry, but can add one if desired.


Jstein77 (approved):
Thanks for adding this!

PR #674, opened 2023-07-24
Remove unused features in snapshot loader class

tlento:
The SqlTableSnapshotRestorer featured a restore method with an
overwrite parameter and a bypass for any tables that exist.

The former was never used, and the latter can only cause trouble.

The actual mechanics of test snapshot updates are:

1. If use-persistent-source-schema is set, we do not invoke the
snapshot restorer
2. If use-persistent-source-schema is not set, we make a new
schema from scratch and load all tables into it, then drop
the schema when we are done
3. If a developer loads data into a new persistent schema, we
make a new schema and then load all of the tables

The only case where a table might exist is when we are loading data
into a schema that already exists and has a matching table. In theory,
this will never happen. If it does, due to sheer bad luck, there's
also a chance the existing tables will contain the wrong data,
and therefore are not reliable for test purposes.

This change, therefore, removes these extraneous features and renames
the SqlTableSnapshotRestorer to the more descriptive SqlTableSnapshotLoader.
The restore method has also been renamed to load.


tlento:
* main
  * **PR #674** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/674" target="_blank">Graphite</a>  👈
    * **PR #675** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/675" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dbt-labs/metricflow/674?utm_source=stack-comment).

plypaul (approved):

PR #676, opened 2023-07-25
Include metric_time in List Dimensions Output Where Appropriate

### LINKED ISSUE #673: Include metric_time in list dimensions output where appropriate

### Describe the feature

Currently, running `mf list dimensions` looks something like this:

```
$ mf list dimensions --metrics jumbo_orders
✔ 🌱 We've found 12 common dimensions for metrics ['jumbo_orders'].
• location_name
• ds
• is_food_order
• is_drink_order
• order_total_dim
• is_large_order
• customer_name
• count_lifetime_orders
• lifetime_spend
• customer_type
• first_ordered_at
• last_ordered_at
```

The thing is, `ds` is the underlying dimension name for the aggregation time dimension for jumbo_orders, which can also be accessed via the `metric_time` alias, but unless the calling user has read our documentation and remembers that `metric_time` is the name to use, they might not realize that the best way to query this metric - particularly if the query includes other metrics with different aggregation time dimensions - is to use the `metric_time` alias.

Ideally we'd include metric_time in this output list somewhere. For now, it's probably sufficient to simply hard-code the dimension name in the list, but ideally this would be something we could resolve.

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

plypaul:
Resolves #673

### Description

Please see the linked issue for more details.


metricflow/model/semantics/linkable_spec_resolver.py
```diff
@@ -533,10 +578,83 @@ def _get_semantic_models_with_joinable_entity(
...
+        ), f"time_granularity should have been set for {time_dimension_reference} in {semantic_model}"
+
+        return type_params.time_granularity
+
     def _get_linkable_element_set_for_measure(self, measure_reference: MeasureReference) -> LinkableElementSet:
         """Get the valid linkable elements for the given measure."""
         measure_semantic_model = self._get_semantic_model_for_measure(measure_reference)
 
+        # Create elements related to the metric_time
+        measure_agg_time_dimension_reference = measure_semantic_model.checked_agg_time_dimension_for_measure(
```
tlento, line 612: All of this new logic feels like it should be a separate helper method with a good name and a nice docstring. None of it requires anything local and it adds a bunch of complexity here which makes the core method harder to read.

metricflow/model/semantics/linkable_spec_resolver.py
```diff
@@ -561,7 +679,7 @@ def _get_linkable_element_set_for_measure(self, measure_reference: MeasureRefere
                     )
                 )
         all_linkable_elements = LinkableElementSet.merge_by_path_key(
-            [local_linkable_elements]
+            [metric_time_linkable_element_set, local_linkable_elements]
```
tlento, line 682: Added bonus, if we factor the logic into a method where this is defined becomes much more obvious.

metricflow/engine/metricflow_engine.py
```diff
@@ -5,11 +5,13 @@
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from enum import Enum
-from typing import List, Optional, Sequence
+from typing import List, Optional, Sequence, Tuple
```
tlento, line 8: <commit level comment> - it would be nice to have a commit message that provides details on why all of the added complexity is necessary to support jamming a dimension that exists for every metric into the list dimensions call.

I get it because I know the codebase, and because I'm reading the entire change in context, but if someone looks at a subset of this logic and says "that's weird, why is that there" then ideally a blame view would point to a broader commit with a message that explains what's goin gon.

And when I say "someone" I mean "future Tom and future Paul" because I don't know about you, but I'm going to forget all of this context by next week.

tlento (approved):
This is great, thanks! One minor refactoring nit inline. 

Also, in general I heavily favor descriptive commit messages over title-only things for the benefit of Future Us. They don't take long to write.

metricflow/engine/metricflow_engine.py
plypaul: Updated the message.

metricflow/model/semantics/linkable_spec_resolver.py
plypaul: Separated this out into a separate method, but also moved out some other sections as the function was pretty long.

PR #677, opened 2023-07-26
Use ValidLinkableSpecResolver for TimeGranularitySolver Implementation

plypaul:
### Description

With https://github.com/dbt-labs/metricflow/pull/676, it's possible to remove some redundant logic in the `TimeGranularitySolver`. There are plans for additional simplification that will allow us to consolidate `MetricFlowQueryParser`, `ValidLinkableSpecResolver`, and `TimeGranularitySolver`.

### DIFF
--- a/metricflow/time/time_granularity_solver.py
+++ b/metricflow/time/time_granularity_solver.py
@@ -1,30 +1,22 @@
 from __future__ import annotations
 
 import logging
-from collections import OrderedDict, defaultdict
 from dataclasses import dataclass
-from typing import Dict, List, Optional, Sequence, Set, Tuple
+from typing import Dict, Optional, Sequence, Tuple
 
 import pandas as pd
-from dbt_semantic_interfaces.protocols.metric import MetricType
-from dbt_semantic_interfaces.protocols.semantic_manifest import SemanticManifest
+from dbt_semantic_interfaces.pretty_print import pformat_big_objects
 from dbt_semantic_interfaces.references import (
     EntityReference,
     MeasureReference,
-    MetricModelReference,
     MetricReference,
     TimeDimensionReference,
 )
 from dbt_semantic_interfaces.type_enums.time_granularity import TimeGranularity
 
-from metricflow.dataflow.builder.node_data_set import DataflowPlanNodeOutputDataSetResolver
-from metricflow.dataflow.dataflow_plan import BaseOutput
-from metricflow.dataset.semantic_model_adapter import SemanticModelDataSet
 from metricflow.filters.time_constraint import TimeRangeConstraint
 from metricflow.model.semantic_manifest_lookup import SemanticManifestLookup
-from metricflow.query.query_exceptions import InvalidQueryException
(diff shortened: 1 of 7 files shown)


metricflow/time/time_granularity_solver.py
```diff
@@ -151,87 +72,65 @@ def validate_time_granularity(
...
                 )
-                if time_dimension_spec.time_granularity < min_granularity_for_querying:
-                    raise RequestTimeGranularityException(
-                        f"The minimum time granularity for querying metrics {metric_references} is "
-                        f"{min_granularity_for_querying}. Got {time_dimension_spec}"
-                    )
-
-                # If there is a cumulative metric, granularity changes aren't supported. We need to check the granularity
-                # specified in the configs for the cumulative metric alone, since `min_granularity_for_querying` may not be supported.
-                for metric_reference in metric_references:
```
tlento, line 169: Ah, cool, this was moved in to the valid dimensions accessor so we don't need it here.

metricflow/time/time_granularity_solver.py
```diff
@@ -151,87 +72,65 @@ def validate_time_granularity(
...
+                        path_key.time_granularity
+                        if minimum_time_granularity is None
+                        else min(minimum_time_granularity, path_key.time_granularity)
                     )
-            # Handle joined time dimensions.
-            else:
-                # TODO: For joined time dimensions, also compute the minimum granularity for querying.
-                replacement_dict[partial_time_dimension_spec] = TimeDimensionSpec(
+
+            if minimum_time_granularity is not None:
```
tlento, line 121: I can't see it in the PR but I'm assuming this is indented inside of the outer for loop, so we build a dict for each partial_time_dimension_spec. Stupid proportional fonts....

tlento (approved):
I like it!

metricflow/time/time_granularity_solver.py
```diff
@@ -151,87 +72,65 @@ def validate_time_granularity(
...
+                if (
+                    path_key.element_name == time_dimension_spec.element_name
+                    and (path_key.entity_links == time_dimension_spec.entity_links)
+                    and path_key.time_granularity == time_dimension_spec.time_granularity
+                ):
+                    match_found = True
+                    break
+            if not match_found:
+                raise RequestTimeGranularityException(
+                    f"{time_dimension_spec} is not valid for querying {metric_references}"
```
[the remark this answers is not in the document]
plypaul: Should be easy, but since this is on the bottom of 15-20 commits, let me handle that in a follow up.

metricflow/time/time_granularity_solver.py
plypaul: That is correct.

metricflow/time/time_granularity_solver.py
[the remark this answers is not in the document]
plypaul: It might be best to handle it as part of the consolidation.

PR #678, opened 2023-07-26
Cut Redshift client over to dbt-redshift adapter

tlento:
This PR effectively enables redshift for the dbt-metricflow integration.

This removes the sqlalchemy-redshift/postgres implementation
of our redshift SqlClient class in favor of using the
AdapterBackedSqlClient with dbt-redshift as the underlying adapter.

Along with this change came a shift from psycopg2 to redshift-connector,
which is what the dbt adapter uses under the hood. This caused some updates
to types returned by redshift, necessitating a change to one of our
integration tests.

Finally, the dbt adapter appears to be materially slower than the
SqlAlchemy client, and on a smaller sized redshift cluster a random
subset of tests may fail with connection timeouts. The cause of this
is not clear, as configuration changes around timeouts, retries, and
auto-commit behavior made no discernable difference, and timings on
AdapterBackedSqlClient internal behaviors (e.g., converting agate to
pandas) suggest the issue is elsewhere.


tlento:
@plypaul - this is ready for review - the failures are the usual snyk nonsense. It appears the tests are stable when issued from GitHub's CI runner, but even if they're not we'll be cutting our actions over to a larger cluster fairly soon so I think we can live with it for a little while.

plypaul (approved):

PR #681, opened 2023-07-27
Remove `create_a_cycle_in_the_join_graph` Identifier in Test Manifest

plypaul:
### Description

`create_a_cycle_in_the_join_graph` is an identifier in the test model that creates a potential cycle in the join graph if the join resolution logic is improperly handled. However, it makes the test output cluttered and confusing, so moving that out of the main test model into a separate one.


tlento (approved):
Thank you!

PR #684, opened 2023-07-27
Improve CLI error message for generating semantic_manifest

### LINKED ISSUE #683: [Feature] Helpful semantic_manifest JSON does not exist message

### Describe the feature

I upgraded to dbt 1.6.0-rc1 and installed metricflow with pip install "dbt-metricflow[snowflake]" in my exisiting dbt project environment. I pulled a branch where someone else has already converted our existing metrics syntax. When I try to run a `mf query` command on that branch I get the following error:

```
Exception found when parsing manifest from dbt project (An error occurred when attempting to build the semantic model
Unable to find /Users/elizepapineau/dev/internal_analytics/internal-analytics/target/semantic_manifest.json
Please ensure that you are running `mf` in the root directory of a dbt project and that the semantic_manifest JSON exists.)
```

The solution is to execute `dbt parse` prior to the `mf query` command. A more helpful error message might be:

```
Please ensure that you are running `mf` in the root directory of a dbt project and that the semantic_manifest JSON exists. If this is your first time running metric flow, run `dbt parse` to generate the semantic_manifest JSON.
```

Pre-existing knowledge that semantic_manifest.json comes from dbt parsing?

### Who will this benefit?

existing dbt users implementing metricflow for the first time in their local dev environment

Ok :)

### Anything else?

_No response_

epapineau:
Resolves #683

### Description

Communicates how to generate `semantic_manifest.json` if it is missing.

### DIFF
--- a/metricflow/cli/dbt_connectors/dbt_config_accessor.py
+++ b/metricflow/cli/dbt_connectors/dbt_config_accessor.py
@@ -105,7 +105,8 @@ class dbtArtifacts:
             raise ModelCreationException(
                 f"Unable to find {full_path_to_manifest}\n"
                 "Please ensure that you are running `mf` in the root directory of a dbt project "
-                "and that the semantic_manifest JSON exists."
+                "and that the semantic_manifest JSON exists. If this is your first time running "
+                "`mf`, run `dbt parse` to generate the semantic_manifest JSON."
             )
         try:
             with open(full_path_to_manifest, "r") as file:


metricflow/cli/dbt_connectors/dbt_config_accessor.py
```diff
@@ -105,7 +105,8 @@ def build_semantic_manifest_from_dbt_project_root(project_root: Path) -> Semanti
             raise ModelCreationException(
                 f"Unable to find {full_path_to_manifest}\n"
                 "Please ensure that you are running `mf` in the root directory of a dbt project "
-                "and that the semantic_manifest JSON exists."
+                "and that the semantic_manifest JSON exists. If this is your first time running "
+                "`mf`, run `dbt parse` to generate the semantic_manifest JSON."
```
tlento, line 109: dbt n00b question: do you think people might get confused if they run `dbt parse` and then try a query and have it return stale (or even empty) results because they needed a `dbt build` or `dbt run` as well? If so it might be worth adding a bit more detail here.

tlento (approved):
This is great, thanks so much! I have a question inline but unless you feel strongly that we should update we can merge this as-is.

metricflow/cli/dbt_connectors/dbt_config_accessor.py
tlento: Make sense, thanks for the context! I think I'll merge this as-is and we can see if there's added confusion. It's too early to know whether people running this will be managing metrics only, or if they'll be more typically doing end to end model -> metrics development.

In theory the errors they get should be familiar to anyone who has tried to do a `dbt parse` and then run a query in their warehouse on top of the models.

PR #692, opened 2023-07-28
Rename `booking_paid_at` --> `paid_at` in Test Manifest

plypaul:
### Description

 When the primary entity feature is added, this rename will change the name from `booking__booking_paid_at` to `booking__paid_at` for better readability.


tlento (approved):
Good call, thanks for splitting this out and putting it on the bottom.

PR #693, opened 2023-07-29
Use Snapshots for Tests That List Group by Items for Metrics

plypaul:
### Description

There are many tests that check the possible group by items that can be queried for a metric. When a dimension is added to a semantic model used in the tests (or other model changes), many of these tests need to be updated because the new dimension is now available for querying.

To make these updates easier and to allow easier checking of results, this PR changes many of those tests to generate a snapshot that can be updated with the `--overwrite-snapshots` flag.


tlento (approved):
On the one hand, I don't love our ever-growing proliferation of snapshots. On the other hand, these outputs are way way way easier to read than the inlined manual object constructions.

Anyway, I consider this a win.

PR #698, opened 2023-08-02
Remove unused schema property from MetricFlow runtime

tlento:
The schema property for the MetricFlowEngine was no longer used
in any way. This PR removes it from the engine and from all affected
callers, and cleans out the surrounding parameters used solely as
inputs.

PR #699, opened 2023-08-02
Suffix all time dimensions with granularity in name rendering

tlento:
The original implementation of MetricFlow treated DAY granularity
as being particularly special, in part because it's the most common
query granularity and in part because it is the minimum granularity
we currently support.

The special treatment essentially made any time dimension queried
with DAY granularity render as the bare name of the dimension,
regardless of that dimension's base granularity. This is a bit odd,
because "billing_month" would only render as "billing_month" if the
granularity was DAY, but in the normal usage of it it would render
instead as "billing_month__month"

Since we plan to support broader granularities, and the inconsistent
output is confusing, this change makes all time dimensions render
with the granularity suffix across our query rendering and output
data column naming. We do this by removing the special casing around
dimension naming for DAY granularity.


tlento:
* main
  * **PR #698** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/698" target="_blank">Graphite</a> 
    * **PR #699** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/699" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dbt-labs/metricflow/699?utm_source=stack-comment).

plypaul (approved):

PR #710, opened 2023-08-04
Add dbt-duckdb as a supported adapter and use it for local dev/test

### LINKED ISSUE #583: Clean up DuckDB implementation and remove SqlAlchemy dependencies

jwills:
Resolves #583 (well, sort of)

### Description

Adds dbt-duckdb as a supported adapter type and uses it for local dev/test workflows instead of the existing sqlalchemy-based client, which is removed. Note that this only partially addresses #583 in that it does not remove all of the sqlalchemy dependencies because I felt like doing that much surgery to the code was above my paygrade, but if you all have a rough plan of what that would look like I am happy to add it in to this PR to fully close out the issue.


local-data-warehouses/README.md
```diff
@@ -3,9 +3,9 @@
 This folder includes utilities to run data warehouses for local development. See the [Contributing guide](../CONTRIBUTING.md)
 to ensure your environment is setup properly.
 
-## SQLite
+## DuckDB
 
-We assume that you have SQLite installed in your environment. By default, tests will run with SQLite.
+By default, tests will run with DuckDB.
```
tlento, line 8: Nice catch, thank you!

tlento (approved):
LGTM, thanks so much!

I think removing the last straggling dev-only SQLAlchemy dependency is beyond the scope of this PR. I'll file a separate issue to clean up some of this adapter configuration logic, and we can decide what to do about the environment configuration for the dbt profiles (and whether or not to replace the SQLAlchemy URL inputs) at that time.

PR #715, opened 2023-08-08
Add Github workflow to populate the persistent source schema

plypaul:
### Description

When the definitions of the tables used in tests change, the persistent source schema needs to be repopulated manually. This PR adds a Github workflow to do this that can be trigged with the `Reload Test Data in SQL Engines` label.

### DIFF
--- a/.github/workflows/cd-sql-engine-populate-persistent-source-schema.yaml
+++ b/.github/workflows/cd-sql-engine-populate-persistent-source-schema.yaml
@@ -1,0 +1,110 @@
+# See [Persistent Source Schema](/GLOSSARY.md#persistent-source-schema)
+# Populating the source schema via this workflow ensures that it's done with the same settings as the tests.
+
+name: Reload Test Data in SQL Engines
+
+# We don't want multiple workflows trying to create the same table.
+concurrency:
+  group: POPULATE_PERSISTENT_SOURCE_SCHEMA
+  cancel-in-progress: true
+
+on:
+  pull_request:
+    types: [labeled]
+  workflow_dispatch:
+
+env:
+  # Unclear on how to make 'Reload Test Data in SQL Engines' a constant here as it does not work here.
+  PYTHON_VERSION: "3.8"
+
+jobs:
+  snowflake-populate:
+    environment: DW_INTEGRATION_TESTS
+    if: >
+      github.event.action == 'workflow_dispatch'
+      || (github.event.action == 'labeled' && github.event.label.name == 'Reload Test Data in SQL Engines')
+    name: Snowflake
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check-out the repo
+        uses: actions/checkout@v3
+
+      - name: Populate w/Python ${{ env.PYTHON_VERSION }}
+        uses: ./.github/actions/run-mf-tests
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+          mf_sql_engine_url: ${{ secrets.MF_SNOWFLAKE_URL }}
+          mf_sql_engine_password: ${{ secrets.MF_SNOWFLAKE_PWD }}
+          parallelism: 1
+          make-target: "populate-persistent-source-schema-snowflake"
+
+  redshift-populate:
+    environment: DW_INTEGRATION_TESTS
+    name: Redshift
+    if: >
+      github.event.action == 'workflow_dispatch'
+      || (github.event.action == 'labeled' && github.event.label.name == 'Reload Test Data in SQL Engines')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check-out the repo
+        uses: actions/checkout@v3
+
+      - name: Populate w/Python ${{ env.PYTHON_VERSION }}
+        uses: ./.github/actions/run-mf-tests
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+          mf_sql_engine_url: ${{ secrets.MF_REDSHIFT_URL }}
+          mf_sql_engine_password: ${{ secrets.MF_REDSHIFT_PWD }}
+          parallelism: 1
+          make-target: "populate-persistent-source-schema-redshift"
+
+  bigquery-populate:
+    environment: DW_INTEGRATION_TESTS
+    name: BigQuery
+    if: >
+      github.event.action == 'workflow_dispatch'
+      || (github.event.action == 'labeled' && github.event.label.name == 'Reload Test Data in SQL Engines')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check-out the repo
+        uses: actions/checkout@v3
+
+      - name: Populate w/Python ${{ env.PYTHON_VERSION }}
+        uses: ./.github/actions/run-mf-tests
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+          MF_SQL_ENGINE_URL: ${{ secrets.MF_BIGQUERY_URL }}
+          MF_SQL_ENGINE_PASSWORD: ${{ secrets.MF_BIGQUERY_PWD }}
+          parallelism: 1
+          make-target: "populate-persistent-source-schema-bigquery"
+
+  databricks-populate:
+    environment: DW_INTEGRATION_TESTS
+    name: Databricks SQL Warehouse
+    if: >
+      github.event.action == 'workflow_dispatch'
+      || (github.event.action == 'labeled' && github.event.label.name == 'Reload Test Data in SQL Engines')
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -23,22 +26,42 @@ test-postgresql:
...
 # Engine-specific test environments. In most cases you should run these with
 # `make -e ADDITIONAL_PYTEST_OPTIONS="--use-persistent-source-schema" test-<engine_type>`
+#
+# See: [Persistent Source Schema](/GLOSSARY.md#persistent-source-schema)
 .PHONY: test-bigquery
 test-bigquery:
 	hatch -v run bigquery-env:pytest -vv -n $(PARALLELISM) $(ADDITIONAL_PYTEST_OPTIONS) metricflow/test/
 
+.PHONY: populate-persistent-source-schema-bigquery
+populate-persistent-source-schema:
```
WilliamDee, line 36: add `-bigquery` suffix

WilliamDee (approved):

Makefile
plypaul: Nice catch. Apparently, if the name it wrong, `make` still succeeds with `nothing to do`

.github/workflows/cd-sql-engine-populate-persistent-source-schema.yaml
```diff
@@ -0,0 +1,97 @@
...
+
+# We don't want multiple workflows trying to create the same table.
+concurrency:
+  group: POPULATE_PERSISTENT_SOURCE_SCHEMA
+  cancel-in-progress: true
+
+on:
+  push:
+    branches:
+      - main
```
tlento, line 14: Do we really want this to run on every push to main? I feel like that's asking for concurrency problems where one job set is dropping and re-creating persistent source schemas while another job set is trying to read from that schema. The dominant case is going to be the one where a push to main triggers no changes, so doing all of that work in the db feels like it'd be both wasteful and contributing to flaky engine test results.

Given how rare updates to the persistent source schemas are likely to be, I think we'd be better off with this triggered on `label` and `workflow_init` or whatever the workflow page setting is so we can manually kick this off on a per-branch basis via the GitHub UI. That way if main gets busted it's easy to just update the schema without labeling a PR, and for any PR making changes it's easy to update the schema with a label there.

tlento: To put this another way - in theory our dominant use case is to run on push to main. If we settle back into that routine, what's the point of a persistent source schema if we repopulate it every time we push to main?

tlento (changes requested):

.github/workflows/cd-sql-engine-populate-persistent-source-schema.yaml
plypaul: Yeah, the intended behavior was for it to be a manually triggered operation. I got this section from the SQL engine tests workflow, which I thought had the same behavior, but apparently not.

plypaul: Also seems like this would not have taken action generally since we don't push to `main`.

plypaul: NVM push happens on merge.

plypaul:
Updated logic to run the workflow when the label is set, and remove the label when workflow finishes. Hopefully Github's concurrency controls work as expected, and this seems good enough for a try.

.github/workflows/cd-sql-engine-populate-persistent-source-schema.yaml
```diff
@@ -0,0 +1,106 @@
...
+name: Reload Test Data in SQL Engines
+
+# We don't want multiple workflows trying to create the same table.
+concurrency:
+  group: POPULATE_PERSISTENT_SOURCE_SCHEMA
+  cancel-in-progress: true
+
+on:
+  pull_request:
+    types: [labeled]
```
tlento, line 13: Please add `workflow_dispatch` so we don't have to label random PRs if something gets borked off main.

.github/workflows/cd-sql-engine-populate-persistent-source-schema.yaml
```diff
@@ -0,0 +1,106 @@
...
+
+env:
+  # Unclear on how to make 'Reload Test Data in SQL Engines' a constant here as it does not work here.
+  PYTHON_VERSION: "3.8"
+  ADDITIONAL_PYTEST_OPTIONS: "--use-persistent-source-schema"
+
+jobs:
+  snowflake-populate:
+    environment: DW_INTEGRATION_TESTS
+    if: github.event.action == 'labeled' && github.event.label.name == 'Reload Test Data in SQL Engines'
```
tlento, line 23: This may need an update for workflow_dispatch to fire, but it should be simple enough.

Makefile
```diff
@@ -23,22 +26,42 @@ test-postgresql:
...
 # `make -e ADDITIONAL_PYTEST_OPTIONS="--use-persistent-source-schema" test-<engine_type>`
+#
+# See: [Persistent Source Schema](/GLOSSARY.md#persistent-source-schema)
 .PHONY: test-bigquery
 test-bigquery:
 	hatch -v run bigquery-env:pytest -vv -n $(PARALLELISM) $(ADDITIONAL_PYTEST_OPTIONS) metricflow/test/
 
+.PHONY: populate-persistent-source-schema-bigquery
+populate-persistent-source-schema-bigquery:
+	hatch -v run bigquery-env:pytest -vv $(ADDITIONAL_PYTEST_OPTIONS) $(POPULATE_PERSISTENT_SOURCE_SCHEMA)
```
tlento, line 37: I think it's better to define a make command that just works, instead of one that relies on someone knowing that they have to set up the `ADDITIONAL_PYTEST_OPTIONS=--use-persistent-source-schema` environment flag.

I just know I'm going to perpetually be in a loop of:

`make populate-persistent....databricks`
<fail>
`make -e "ADDITIONAL_PYTEST_OPTIONS=--use-persistent-source-schema" populate-persistent...databricks`

.github/workflows/cd-sql-engine-populate-persistent-source-schema.yaml
```diff
@@ -0,0 +1,106 @@
+# See [Persistent Source Schema](/GLOSSARY.md#persistent-source-schema)
+# Populating the source schema via this workflow ensures that it's done with the same settings as the tests.
+
+name: Reload Test Data in SQL Engines
+
+# We don't want multiple workflows trying to create the same table.
+concurrency:
+  group: POPULATE_PERSISTENT_SOURCE_SCHEMA
+  cancel-in-progress: true
```
tlento, line 9: I am very curious to see how this works.

.github/workflows/cd-sql-engine-populate-persistent-source-schema.yaml
```diff
@@ -0,0 +1,106 @@
...
+        uses: ./.github/actions/run-mf-tests
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+          mf_sql_engine_url: ${{ secrets.MF_DATABRICKS_SQL_WAREHOUSE_URL }}
+          mf_sql_engine_password: ${{ secrets.MF_DATABRICKS_PWD }}
+          parallelism: 1
+          additional-pytest-options: ${{ env.ADDITIONAL_PYTEST_OPTIONS }}
+          make-target: "populate-persistent-source-schema-databricks"
+
+  remove-label:
```
tlento, line 97: Oh if this works we should TOTALLY add it to the sql engine tests...... I've got some updates I want to make over there so I can do that once this is in.

.github/workflows/cd-sql-engine-populate-persistent-source-schema.yaml
```diff
@@ -0,0 +1,106 @@
+# See [Persistent Source Schema](/GLOSSARY.md#persistent-source-schema)
+# Populating the source schema via this workflow ensures that it's done with the same settings as the tests.
+
+name: Reload Test Data in SQL Engines
+
+# We don't want multiple workflows trying to create the same table.
```
tlento, line 6: Yeah I guess if this ever happens in different schemas the last one in will be wrong anyway.

tlento (approved):
Sure, let's give it a go. Please address inline comments before merge.

Makefile
plypaul: That's fair. Updated.

.github/workflows/cd-sql-engine-populate-persistent-source-schema.yaml
plypaul: Added, but iterating on this PR has been a huge pain as it can't be tested locally.

PR #717, opened 2023-08-09
Consolidate query interface params

DevonFulcher:
Resolves SLA-218 (internal to dbt Labs)

### Description

This PR makes the `Dimension`, `TimeDimension`, and `Entity` objects that are passed into the where filter Jinja template implement protocols. These protocols allow for different implementations of these objects depending on the context (`group_by`, `where`, `order_by` parameters) while also constraining these implementations to all have the same method signatures. This will create consistency across these contexts.

The `Dimension` protocol is also implemented for parameters in certain methods (i.e. `GroupByOrderByDimension`). This will allow for more complex `Dimension` objects in the future that aren't feasible to serialize into a string, such as `Dimension('demographic').grain('month').alias('monthly_demographics')`. These parameters are added as optional parameters to these methods, so everything should be backward compatible.

I added some unit tests. This change is mostly refactoring existing behavior to implement protocols and adding additional optional parameters. These optional parameters just transform into the existing parameters. So, I relied on the existing tests to ensure no breaking behavior.


metricflow/specs/query_interface.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class QueryInterfaceTimeDimension(ABC):
+    """Time Dimension in a where clause with Jinja."""
+
+    def __init__(  # noqa
+        self,
+        time_dimension_name: str,
+        time_granularity_name: str,
+        entity_path: Sequence[str] = (),
```
tlento, line 40: Are these meant to be accessible properties? It seems like we'd want to access them. Assuming that's the case, I recommend something more like the protocol definitions we use in dbt-semantic-interfaces, with property accessor methods defined on the protocol and everything else left up to the implementing classes.

metricflow/specs/query_interface.py
```diff
@@ -0,0 +1,49 @@
+from __future__ import annotations
+
+from typing import Sequence
+from abc import ABC
+
+
+class QueryInterfaceMetric(ABC):
```
tlento, line 7: Can we make these Protocol types instead of ABC?

Generally speaking, the Protocol is what we'd use for interfaces - especially cross-codebase interfaces, since inheritance is not required.

metricflow/specs/where_filter_dimension.py
```diff
@@ -0,0 +1,64 @@
...
+from metricflow.specs.specs import DimensionSpec
+from dbt_semantic_interfaces.call_parameter_sets import FilterCallParameterSets
+
+
+def get_where_filter_dimension_cls(
+    call_parameter_sets: FilterCallParameterSets,
+    dimension_specs: DimensionSpec,
+    column_association_resolver: ColumnAssociationResolver,
+):
+    class WhereFilterDimension(QueryInterfaceDimension):
```
tlento, line 29: Anonymous class patterns of this sort are not common in Python. It's much more typical - and more efficient in the runtime - to define a class and then initialize it inside a function. Returning a closure for the purposes of having a callable available to jinja template processing is fairly straightforward by comparison.

I might be missing something, but this is a fairly specific class specification. I think we'll find it's most natural to have these classes defined and importable, and then to wrap each one in a closure where necessary, instead of making each class initializer the equivalent of an anonymous function accessed through a separate method like this.

metricflow/specs/query_interface.py
```diff
@@ -0,0 +1,49 @@
...
+    """Metric in the Jinja interface."""
+
+    def __init__(self, name: str) -> None:  # noqa: D
```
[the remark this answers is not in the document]
DevonFulcher: Ya, I realized that. It is up to the implementer to ensure that the same initializer signature is used each time. It is a bit unfortunate, but I couldn't think of a better way.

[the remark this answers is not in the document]
tlento: I'm not sure how much of this is tied up with the anonymous class pattern elsewhere in this PR and how much is required, but if you want to have a concrete interface spec for initializing the class based on specific input parameters, you can require a class factory method. Something like:

```
@abstractmethod
@classmethod
def from_name(cls, name: str) -> Self:
    """Initialize and return an instance of this class from the name."
    raise NotImplementedError
```

Then you always invoke the equivalent of `cls.from_name(some_name)` at the callsite, and never bother with directly invoking the initializer.

This pattern also works with a Protocol type.

It might be worth it for us to have a more interactive chat - I'm realizing I can give feedback on Python mechanics but my understanding of the specific problem you're solving is limited enough that I can't really help out with the bigger questions of what we want from the type spec itself.

metricflow/specs/query_interface.py
DevonFulcher: Can you go into more detail about what you mean by

> inheritance is not required

I am thinking that different parameter types (i.e. group_by, where, order) would inherit from these dimensions, time dimension, and entity interfaces.

metricflow/specs/query_interface.py
DevonFulcher: I don't think this still applies, given my recent refactors. Let me know  if that isn't the case

tlento: Yeah I don't think this comment makes sense anymore.

WilliamDee (commented):
There's this whole thing about inheriting the Protocol classes in the implementation class that we had, don't really remember what we landed at there, but @tlento or @plypaul would know. In DSI, Paul made the `ProtocolHint` class for this, not sure if we would want to use that here too.

DevonFulcher:
Added `ProtocolHint` to the implementations.

metricflow/specs/where_filter_transform.py
```diff
@@ -42,91 +31,29 @@ def __init__(  # noqa: D
...
         try:
             rendered_sql_template = jinja2.Template(
                 where_filter.where_sql_template, undefined=jinja2.StrictUndefined
             ).render(
-                Dimension=_dimension_call,
-                TimeDimension=_time_dimension_call,
-                Entity=_entity_call,
+                {
+                    "Dimension": WhereFilterDimensionFactory(
+                        call_parameter_sets, dimension_specs, self._column_association_resolver
```
tlento, line 48: The implicit update by reference on dimension_specs is challenging, because it's really not obvious that it's happening here.

I haven't come up with a good way of separating it from create, since we kind of need to do it inside of that call. I think that's why Paul originally had everything inlined above this point.

In the meantime, we could just rename the parameter to make it more obvious that the factory invocation will populate the spec list based on user input.

metricflow/specs/where_filter_dimension.py
```diff
@@ -0,0 +1,91 @@
...
+
+from metricflow.specs.column_assoc import ColumnAssociationResolver
+from metricflow.specs.query_interface import (
+    QueryInterfaceDimension,
+    QueryInterfaceDimensionFactory,
+)
+from metricflow.specs.specs import DimensionSpec
+
+
+class WhereFilterDimension(ProtocolHint[QueryInterfaceDimension]):
```
tlento, line 25: I don't have time to review the new classes right now, so I'm checkpointing here and submitting for the time being.

As a broader note, this PR is really big now. In future I'd recommend splitting it up, something along the lines of:

1. Add protocol classes and one example implementation
2. Cut over other implementations one at a time in separate PRs

Then you'd have like 3 or 4, one of which gets the most action and the others are easy quick reads.

I wouldn't bother doing this here unless there's a big hurry to get one of these things merged without the others, I'm like 80-90% through it. But in future it'll be easier to make progress in chunks.

tlento (commented):
Checkpoint review, I'm not quite done.

I'm not sold on the factory class pattern but I have nothing better to offer that will work with jinja, so this is probably one of those cases where it's better for us to proceed with minor readability updates and revisit if we need to.

.changes/unreleased/Features-20230814-111602.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Features
+body: Add optional parameters to pass in objects instead of strings. Add new protocols
```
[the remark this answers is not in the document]
DevonFulcher: Got it, thanks! Tbh, I wasn't sure how changie is used or how user-facing it is.

metricflow/engine/metricflow_engine.py
```diff
@@ -94,39 +97,51 @@ class MetricFlowQueryRequest:
     """
 
     request_id: MetricFlowRequestId
-    metric_names: Sequence[str]
-    group_by_names: Sequence[str]
+    metric_names: Optional[Sequence[str]] = None
+    metrics: Optional[Sequence[QueryInterfaceMetric]] = None
+    group_by_names: Optional[Sequence[str]] = None
+    group_by: Optional[Sequence[GroupByOrderByDimension]] = None
```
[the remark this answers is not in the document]
DevonFulcher: I figured that we would be able to remove the string-based parameters during the next breaking change release. Yes, let's chat

metricflow/engine/metricflow_engine.py
```diff
@@ -94,39 +97,51 @@ class MetricFlowQueryRequest:
...
         where_constraint: Optional[str] = None,
         order_by_names: Optional[Sequence[str]] = None,
+        order_by: Optional[Sequence[GroupByOrderByDimension]] = None,
         output_table: Optional[str] = None,
         sql_optimization_level: SqlQueryOptimizationLevel = SqlQueryOptimizationLevel.O4,
         query_type: MetricFlowQueryType = MetricFlowQueryType.METRIC,
     ) -> MetricFlowQueryRequest:
+        assert_exactly_one_arg_set(metric_names=metric_names, metrics=metrics)
+        assert not (group_by_names and group_by), "Both group_by_names and group_by should not be set"
+        assert not (order_by_names and order_by), "Both order_by_names and order_by should not be set"
```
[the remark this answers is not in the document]
DevonFulcher: That looks great! I modeled these error messages off of [this](https://github.com/dbt-labs/metricflow/blob/main/metricflow/query/query_parser.py#L298) pre-existing message, but I like your improvement instead.

metricflow/engine/metricflow_engine.py
```diff
@@ -94,39 +97,51 @@ class MetricFlowQueryRequest:
...
         time_constraint_start: Optional[datetime.datetime] = None,
         time_constraint_end: Optional[datetime.datetime] = None,
         where_constraint: Optional[str] = None,
         order_by_names: Optional[Sequence[str]] = None,
+        order_by: Optional[Sequence[GroupByOrderByDimension]] = None,
         output_table: Optional[str] = None,
         sql_optimization_level: SqlQueryOptimizationLevel = SqlQueryOptimizationLevel.O4,
         query_type: MetricFlowQueryType = MetricFlowQueryType.METRIC,
     ) -> MetricFlowQueryRequest:
+        assert_exactly_one_arg_set(metric_names=metric_names, metrics=metrics)
```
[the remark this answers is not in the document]
DevonFulcher: This is basically how the Jinja syntax works. I think it makes sense, in that case, to be very permissive given the context that a user would be using it. I chose not to use Union in this case, though, because I figured people who call this method would prefer stricter typing. I think it would be good to remove string-based parameters from these methods eventually, and I thought that would be easier if they were separate parameters.

metricflow/query/query_parser.py
```diff
@@ -281,27 +291,55 @@ def _construct_metric_specs_for_query(
             )
         return tuple(metric_specs)
 
+    def _get_group_by_names(
+        self, group_by_names: Optional[Sequence[str]], group_by: Optional[Sequence[GroupByOrderByDimension]]
+    ) -> Sequence[str]:
+        assert not (group_by_names and group_by), "Both group_by_names and group_by should not be set"
+        return group_by_names if group_by_names else [str(g) for g in group_by] if group_by else []
```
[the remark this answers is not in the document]
DevonFulcher: > Also, if we're going to remove the duality altogether and never have to call str(thing) on these outside jinja in the future a code comment to that effect would be sufficient.

Yes, this is what I had in mind. I will add more comments.

metricflow/specs/group_by_order_by_dimension.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class GroupByOrderByDimensionFactory(ProtocolHint[QueryInterfaceDimensionFactory]):
+    """Creates a GroupByOrderByDimension.
+
+    This is useful as a factory for the Jinja sandbox.
+    """
+
+    @override
+    def _implements_protocol(self) -> GroupByOrderByDimensionFactory:  # noqa: D
```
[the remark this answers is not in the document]
DevonFulcher: Got it! I will make that update! I mostly copy-pasted these `_implements_protocol` from elsewhere, but I must have messed that up.

metricflow/specs/query_interface.py
```diff
@@ -0,0 +1,80 @@
...
+        self,
+        time_dimension_name: str,
+        time_granularity_name: str,
+        entity_path: Sequence[str] = (),
+    ) -> QueryInterfaceTimeDimension:
+        """Create a TimeDimension."""
+        raise NotImplementedError
+
+
+class QueryInterfaceTimeDimension(Protocol):
```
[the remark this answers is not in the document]
DevonFulcher: Do you mean common amongst `Dimension`, `TimeDimension`, and `Entity`? That makes sense! I suppose I was more focused on consolidating the interfaces between `group_by`, `order`, and `where` than consolidating the objects. I will make that update.

metricflow/specs/where_filter_dimension.py
DevonFulcher: Thanks, Tom! I will do that.

metricflow/specs/where_filter_transform.py
DevonFulcher: Ya, that is a good point that I hadn't considered. I think it would be better if they just weren't passed in and instead just properties on `WhereFilterDimensionFactory`.

DevonFulcher: I made that update [here](https://github.com/dbt-labs/metricflow/pull/717/commits/19c79cd80c94d1956540e963f789afbc64e898b3). Let me know if you would prefer a different solution.

metricflow/specs/query_interface.py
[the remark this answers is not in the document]
DevonFulcher: Do you mean that `QueryInterfaceTimeDimension` should implement a common protocol or that `QueryInterfaceTimeDimension` shouldn't exist and the classes that currently implement `QueryInterfaceTimeDimension` should implement a common protocol instead?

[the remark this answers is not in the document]
DevonFulcher: nvm we talked about this in person

tlento (changes requested):
Per our discussion, we are going to do the following here:

1. Clean up the protocols a bit in some way in order to ensure there are property accessors for MetricFlow to use for building specs and so forth behind the API, and property configuration methods for the callers to implement so users can configure things on the client side
2. Leave anything other than minor cleanup for follow-up PRs
3. Figure out if we want to move to something like `group_by: Optional[Sequence[Union[str, NonAggregateParameter]]]` and, if so, have a separate PR where we get that all working nicely

I'm requesting changes just so I can get an alert when this is ready - please request review from me when you update.  GitHub makes you do this manually, which always annoys me, but I haven't found a better mechanism.

metricflow/engine/metricflow_engine.py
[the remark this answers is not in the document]
DevonFulcher: We talked in person about this.

metricflow/specs/where_filter_transform.py
DevonFulcher: We talked in person about this.

metricflow/query/query_parser.py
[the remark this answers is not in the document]
DevonFulcher: I removed `str()` here, so this isn't a problem anymore

DevonFulcher:
I changed the factories to return individualized protocols instead of all sharing the `QueryParameter` protocol. While the parameters are similar in some respects (they all might have an `alias`), they can also be different (only `Dimension` has `grain`). I wanted these protocols to be separate so they could evolve independently. This can be changed later.


Instead, I use `QueryParameter` as the read-only protocol for MF to get `grain` and `name` for an object. Since, both `QueryParameter` and `QueryInterfaceDimension` have a `grain` property, they can't be implemented by the same class. This is somewhat limiting, but I found it an okay tradeoff to have a more Pythonic interface instead of `QueryParameter` having a `get_grain` property or something similar.

metricflow/specs/where_filter_entity.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def alias(self, _alias: str) -> WhereFilterEntity:
+        """Renaming the column."""
+        raise NotImplementedError
+
+    def __str__(self) -> str:
+        """Returns the column name.
+
+        Important in the Jinja sandbox.
+        """
```
tlento, line 41: Is this part of the contract for all QueryInterfaceDimension classes? If so, we might consider marking `__str__` as abstract in the protocol, although it's not clear to me if that'll actually do anything from a typechecking perspective. But if it does do something in the typechecker it'd be nicely self-documenting.

If this is specific to this implementation - meaning we normally don't care about being able to use implementing classes in a jinja sandbox - then the current layout is what we want.

metricflow/specs/where_filter_entity.py
```diff
@@ -0,0 +1,81 @@
...
+        return self.column_name
+
+
+class WhereFilterEntityFactory(ProtocolHint[QueryInterfaceEntityFactory]):
+    """Creates a WhereFilterDimension."""
```
tlento, line 46: nit:
```suggestion
    """Creates a WhereFilterEntity."""
```

metricflow/specs/where_filter_time_dimension.py
```diff
@@ -0,0 +1,86 @@
...
+        return self
+
+    def __init__(  # noqa
+        self,
+        call_parameter_sets: FilterCallParameterSets,
+        column_association_resolver: ColumnAssociationResolver,
+    ):
+        self._call_parameter_sets = call_parameter_sets
+        self._column_association_resolver = column_association_resolver
+        self.time_dimension_specs: List[TimeDimensionSpec] = []
```
tlento, line 57: Might be worth an update to the docstrings on these classes that the specs themselves are updated by reference with each call to create() and therefore are part of the public interface on the class.

It's in the code here in this PR, but it's not clear from looking at this module alone that this is the expected behavior and usage.

metricflow/test/query/test_query_parser.py
```diff
@@ -166,6 +171,45 @@ def test_query_parser() -> None:  # noqa: D
     )
 
 
+class MockQueryParameter:  # noqa: D
+    grain = None
```
tlento, line 175: I think a docstring is in order on this class. As far as I can tell, this class exists because we don't have any QueryParameter implementations inside of MetricFlow itself. So this is mocking an outside caller with a concrete implementation.

Is that correct? If so let's just document it a bit more so future readers of this test file understand why it's here.

tlento (approved):
Looking much better, thank you!

I'm mildly concerned about the layers of indirection being confusing to readers of the future, particularly if they read by module instead of in coordination as these changes appear in this PR.

For now I think this is best addressed via documentation, and I've left a couple of minor pointers inline.

Unless you feel like you want another review pass you may update and merge at your convenience! This might merit a squash and merge operation, or at least a rebase -i and force push to clean up the history in this branch. It's a little gnarly commit-wise.

Thanks!

metricflow/specs/where_filter_entity.py
DevonFulcher: Thank you! Must have been a copy-paste error

metricflow/specs/where_filter_entity.py
DevonFulcher: Actually, I was hoping that I could remove the `__str__` implementation in the Jinja-MFS interface after this PR is merged in.

metricflow/test/query/test_query_parser.py
DevonFulcher: That's right! I added a comment. Even if there were an implementation of `QueryParameter`, I think this mock is appropriate for this test.

metricflow/engine/metricflow_engine.py
```diff
@@ -616,15 +639,19 @@ def get_dimension_values(  # noqa: D
...
         time_constraint_start: Optional[datetime.datetime] = None,
         time_constraint_end: Optional[datetime.datetime] = None,
     ) -> MetricFlowExplainResult:
         return self._create_execution_plan(
             MetricFlowQueryRequest.create_with_random_request_id(
                 metric_names=metric_names,
-                group_by_names=[get_group_by_values],
+                metrics=metrics,
+                group_by_names=[get_group_by_values] if get_group_by_values else None,
+                group_by=[group_by] if group_by else None,
```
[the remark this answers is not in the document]
DevonFulcher: Good catch! I fixed it

metricflow/test/query/test_query_parser.py
tlento: Agreed, thanks!

PR #722, opened 2023-08-15
Remove EnvironmentVariable helper class

tlento:
This class is doing a whole lot to accomplish very little. In
updating the sql engine url environment variable access I noticed
the old EnvironmentVariable helper was hardly being used at all
anymore. Since most of its usage has disappeared into the sands
of time, we remove the last remaining calls - and the class
itself - here.

PR #721, opened 2023-08-15
Make test execution more consistent and convenient

tlento:
MetricFlow's testing configuration currently tags sql engine types
to hatch execution environment. The standard one, which we use for
`make test`, is meant to be duckdb, but a stale environment variable
can cause inscrutable failures on any test requiring an engine
connection.

This change updates the test environment configuration to make duckdb
the standard for any runs in the hatch dev-env, including any
invocation such as `make test`. It also removes the default fallback
to duckdb, which was causing inconsistent behaviors in certain
environment configurations, and replaces it with a consistent error.

Developers with heavily customized environment configurations might
have a bad time with this, but they were already having a bad time
with other things, and at least this gives them an actionable
error message.

This PR also adds a workflow_dispatch trigger to the sql engine tests to
make it easier to test updates to secrets without PR thrash.

PR #720, opened 2023-08-10
Loosen version pins for MetricFlow dependencies

tlento:
Several packages have strict version pins that are not necessary,
as our usage is compatible with a broad range of versions. These
updated version ranges all appear to be safe with MetricFlow's
current usage and dependency requirements.

This also removes our vestigial jsonschema dependencies.

PR #719, opened 2023-08-10
Update Pandas dependency to 1.5.x

tlento:
Our CI jobs were taking forever due to needing to build an old version
of pandas from source for Python 3.11. This updates the dependency to
1.5.x, which allows us to use versions that should have updated
pre-built wheels available.

PR #724, opened 2023-08-15
Render date literals as strings in time range filters

tlento:
MetricFlow's rendering has historically cast date literals to
the underlying engine type in date range comparison filters, so
if a user passed in a start_time of 2021-01-01 and an end_time
of 2021-01-31 we would render something like:

`WHERE ds BETWEEN CAST('2021-01-01' AS TIMESTAMP) AND CAST('2021-01-31' AS TIMESTAMP)`

This turns out to cause problems with time partition columns, because
they typically only apply pruning to literal predicates - any value
processed by a CAST function prevents pruning from kicking in, as the
engine takes the appropriately conservative approach of doing full
row-by-row comparisons between the cast output and the partition value.

In researching this issue two things became clear:

1. We need a holistic answer to ensuring engines apply partition
pruning, one which is beyond the scope of this change.
2. We can solve part of the problem by removing the CAST call
in time range filters based on literal value inputs.

It turns out every engine we support does implicit type coercion
between string values and date/time types. This means we don't have
to care what the appropriate time type column might be in this case,
because the standard behavior is to simply do the comparison correctly
regardless of underlying partition column type.

This change simple removes the CAST calls from this particular rendering
block, which should enable users to at least run queries that will do
some level of pruning, even if the solution is incomplete with respect
to the types of time partition columns it can address.

Note that we assume all engines accept and convert from ISO-standard
date strings, but this is not a change from the prior behavior, which
assumed the same thing but wrapped it in a CAST function call.

### DIFF
--- a/.changes/unreleased/Fixes-20230814-231516.yaml
+++ b/.changes/unreleased/Fixes-20230814-231516.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Remove barrier to partition pruning certain time partition filter predicates
+time: 2023-08-14T23:15:16.658262-07:00
+custom:
+  Author: tlento
+  Issue: "712"
--- a/metricflow/plan_conversion/dataflow_to_sql.py
+++ b/metricflow/plan_conversion/dataflow_to_sql.py
@@ -83,7 +83,6 @@ from metricflow.sql.optimizer.optimization_levels import (
 )
 from metricflow.sql.sql_exprs import (
     SqlBetweenExpression,
-    SqlCastToTimestampExpression,
     SqlColumnReference,
     SqlColumnReferenceExpression,
     SqlDateTruncExpression,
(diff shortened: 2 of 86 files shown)


tlento:
* main
  * **PR #719** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/719" target="_blank">Graphite</a> 
    * **PR #720** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/720" target="_blank">Graphite</a> 
      * **PR #721** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/721" target="_blank">Graphite</a> 
        * **PR #722** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/722" target="_blank">Graphite</a> 
          * **PR #724** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/724" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dbt-labs/metricflow/724?utm_source=stack-comment).

plypaul (approved):
As I recall, this was put in due to an issue with a previously supported engine.

tlento:
> As I recall, this was put in due to an issue with a previously supported engine.

That's what I remember too but I can't find any evidence of it. It's possible the offending engine has updated their type handling. It's also possible we didn't understand BigQuery's coercion logic - their documentation on this is hidden well enough that it appears BigQuery simply doesn't support it, but they do (and they have to because of the way their partition pruning works).

PR #726, opened 2023-08-16
Enable granularities for derived metrics with time offset

courtneyholcomb:
Resolves #SL-767

### Description
With the current placement of the `JoinToTimeSpineNode` in the dataflow plan, we filter down to time columns with the requested granularity BEFORE joining to time spine. Since joining to time spine expects a time column with `DAY` granularity, this means queries with any granularity besides `DAY` don't return expected results. Here, we move the time spine join to happen before filtering elements to fix that issue. So instead of `FilterElementsNode` -> `JoinToTimeSpineNode`, we do `JoinToTimeSpineNode` -> `FilterElementsNode`.
For cumulative metrics, we need to factor in the `JoinOverTimeRangeNode`. This PR updates the dataflow plan from `FilterElementsNode` -> `JoinOverTimeRangeNode` (both required) to `JoinOverTimeRangeNode` (optional, only used if a time dimension is requested) -> `JoinToTimeSpineNode` (optional, only used for offset metrics) -> `FilterElementsNode`. This ensures we'll have the columns needed when joining to time spine.
Unrelated to time offset, this change also removes the unnecessary `JoinOverTimeRangeNode` step for a cumulative metric queried without a time dimension, which 1) does nothing in SQL and 2) might result in inaccurate costing for that query.

Note: there is currently a table missing in the Postgres & BQ test warehouses that is blocking generation of snapshots for some tests.


metricflow/test/fixtures/sql_client_fixtures.py
```diff
@@ -74,7 +74,7 @@ def __configure_test_env_from_url(url: str, password: str, schema: str) -> sqlal
 
 def __configure_bigquery_env_from_credential_string(password: str, schema: str) -> None:
     credential_string = password.replace("'", "")
-    credentials = json.loads(credential_string)
+    credentials = json.loads(credential_string, strict=False)
```
courtneyholcomb, line 77: Needed to work with BQ password characters

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -780,9 +775,35 @@ def _build_aggregated_measures_from_measure_source_node(
                 f"Recipe not found for measure specs: {measure_specs} and linkable specs: {required_linkable_specs}"
             )
 
+        metric_time_dimension_spec: Optional[TimeDimensionSpec] = None
+        for linkable_spec in queried_linkable_specs.time_dimension_specs:
+            if linkable_spec.element_name == self._metric_time_dimension_reference.element_name:
+                metric_time_dimension_spec = linkable_spec
+                break
+
+        time_range_node: Optional[JoinOverTimeRangeNode[SqlDataSetT]] = None
```
courtneyholcomb, line 784: Does this require metric_time? Or any time dim?

metricflow/dataflow/dataflow_plan.py
```diff
@@ -726,6 +727,7 @@ def __init__(
 
         Args:
             parent_node: Node that returns desired dataset to join to time spine.
+            time_dimension_spec: Time dimension requested in query. Used to determine time spine granularity.
```
courtneyholcomb, line 730: need this??

metricflow/test/snapshots/test_dataflow_to_sql_plan.py/SqlQueryPlan/BigQuery/test_cumulative_metric__plan0.sql
```diff
@@ -1,12 +1,12 @@
 -- Compute Metrics via Expressions
 SELECT
-  subq_4.ds__month
-  , subq_4.txn_revenue AS trailing_2_months_revenue
+  subq_3.ds__month
```
courtneyholcomb, line 3: This is the SQL change from removing the unused `JoinOverTimeRangeNode` visit.

metricflow/dataflow/dataflow_plan.py
```diff
@@ -718,6 +718,7 @@ class JoinToTimeSpineNode(Generic[SourceDataSetT], BaseOutput[SourceDataSetT], A
     def __init__(
         self,
         parent_node: BaseOutput[SourceDataSetT],
+        metric_time_dimension_spec: TimeDimensionSpec,
```
tlento, line 721: Do we want this here?

Previously, the granularity was sourced from the parent node for making the time spine dataset, which contained all of the available granularities. Now it's coming in from the query parameters.

Now it's getting threaded through from the callsite, and what we're doing at the callsite is taking one of the potentially many query instances and wiring it through this node.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1343,17 +1297,49 @@ def visit_join_to_time_spine_node(self, node: JoinToTimeSpineNode[SourceDataSetT
...
+        original_time_dim_instance = time_spine_dataset.instance_set.time_dimension_instances[0]
+        time_spine_select_expr: Union[
+            SqlColumnReferenceExpression, SqlDateTruncExpression
+        ] = SqlColumnReferenceExpression(
+            SqlColumnReference(table_alias=time_spine_alias, column_name=original_time_dim_instance.spec.qualified_name)
         )
 
+        # Add requested granularity to time spine spec & column.
+        # Time spine defaults to DAY, so DAY granularity does not require any change here.
+        if node.metric_time_dimension_spec.time_granularity == TimeGranularity.DAY:
```
tlento, line 1314: Let's not hard-code DAY as the default. The time_spine_source should have the granularity, so we can update this logic to match what we had earlier, assuming we don't want to just revert back.

tlento (commented):
I'm having a hard time reasoning about the changes here because this is an old and gnarly part of the codebase, it's been a while, and the PR itself is doing three seemingly independent but possibly interconnected things.

1. Re-ordering the dataflow plan
2. Removing some extra join
3. Re-structuring the granularity adjustment source from the parent node of the time spine join to the first metric_time reference in the query parameters, and moving the adjustment logic inline in the visitor method

The last of these does not seem like it'll be deterministically correct. If a user groups by `metric_time__day` but applies a filter on `metric_time__month` they'll potentially get different results than if they did the opposite, and filtered on day while grouping by month, because the group by ones are added to the spec set first, but the filter happens after the joins.

Are all of these changes necessary to address the core problem with derived metric offsets? And can they be split up into stages so we can see what they're doing independently of each other? Even if it's rough, like if there's a commit pointer in this stack that splits things a bit more clearly, or if you can squash and reorder some of commits and maybe do some `git add --patch` maneuvering to get to a commit breakpoint that divides this up (and especially if you can run the snapshots in between), that'd really help me.

metricflow/test/snapshots/test_dataflow_to_sql_plan.py/SqlQueryPlan/BigQuery/test_derived_metric_with_offset_to_grain__plan0.sql
```diff
@@ -147,33 +147,85 @@ FROM (
...
+          subq_9.metric_time__day
+          , subq_9.bookings
         FROM (
-          -- Pass Only Elements:
-          --   ['bookings', 'metric_time__day']
+          -- Join to Time Spine Dataset
           SELECT
-            subq_6.metric_time__day
-            , subq_6.bookings
+            DATE_TRUNC(subq_7.metric_time__day, day) AS metric_time__day
```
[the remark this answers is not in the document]
courtneyholcomb: Oops this is a bug I fixed and I think something was up with the snapshots script that caused it to skip BQ - so I think I just need to fix the snapshot here!

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -172,92 +173,49 @@ def _make_time_spine_data_set(
...
+                where=_make_time_range_comparison_expr(
+                    table_alias=time_spine_table_alias,
+                    column_alias=time_spine_source.time_column_name,
+                    time_range_constraint=time_range_constraint,
+                )
+                if time_range_constraint
+                else None,
+                order_bys=(),
+            ),
+        )
```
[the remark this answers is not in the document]
courtneyholcomb: It's mostly cleanup. I removed a bunch of redundant code.

courtneyholcomb:
> I'm having a hard time reasoning about the changes here because this is an old and gnarly part of the codebase, it's been a while, and the PR itself is doing three seemingly independent but possibly interconnected things.
> 
> 1. Re-ordering the dataflow plan
> 2. Removing some extra join
> 3. Re-structuring the granularity adjustment source from the parent node of the time spine join to the first metric_time reference in the query parameters, and moving the adjustment logic inline in the visitor method
> 
> The last of these does not seem like it'll be deterministically correct. If a user groups by `metric_time__day` but applies a filter on `metric_time__month` they'll potentially get different results than if they did the opposite, and filtered on day while grouping by month, because the group by ones are added to the spec set first, but the filter happens after the joins.
> 
> Are all of these changes necessary to address the core problem with derived metric offsets? And can they be split up into stages so we can see what they're doing independently of each other? Even if it's rough, like if there's a commit pointer in this stack that splits things a bit more clearly, or if you can squash and reorder some of commits and maybe do some `git add --patch` maneuvering to get to a commit breakpoint that divides this up (and especially if you can run the snapshots in between), that'd really help me.

@tlento I'll take this PR and separate it into smaller PRs to make this all clearer. I didn't totally follow what you meant in #3 which probably means it was not the logic I intended. Hopefully breaking into smaller PRs will help make this more clear to both of us and I can adjust that logic if needed!

tlento:
> @tlento I'll take this PR and separate it into smaller PRs to make this all clearer. I didn't totally follow what you meant in https://github.com/dbt-labs/metricflow/pull/3 which probably means it was not the logic I intended. Hopefully breaking into smaller PRs will help make this more clear to both of us and I can adjust that logic if needed!

That'd be amazing, thanks so much!

There are some fairly large changes I want to make this week so hopefully we can get this all in ahead of those so as to avoid the looming merge conflict pain. I should be back to my normal review cadence this week, assuming no other major disruptions happen over here. Stupid yellow jackets....

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -780,9 +775,38 @@ def _build_aggregated_measures_from_measure_source_node(
                 f"Recipe not found for measure specs: {measure_specs} and linkable specs: {required_linkable_specs}"
             )
 
+        metric_time_dimension_spec: Optional[TimeDimensionSpec] = None
+        for linkable_spec in queried_linkable_specs.time_dimension_specs:
+            if linkable_spec.element_name == self._metric_time_dimension_reference.element_name:
+                metric_time_dimension_spec = linkable_spec
```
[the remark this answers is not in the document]
courtneyholcomb: Interesting point. What if the query included two incompatible granularities like week and month? We might need to take a list of `metric_time_specs` in the time spine builder!

metricflow/dataflow/dataflow_plan.py
courtneyholcomb: The parent dataset will no longer be narrowed down to only the requested `metric_time_dimension_instances` since the measure aggregation is being moved to after joining to time spine in this PR. Because of that, I thought it would be more efficient to pass through the requested granularity and only aggregate time spine columns based on that request. However, I can see a couple of potential issues with that.
1. There might be multiple `metric_time` dimensions in the query. To handle this case, we could accept a list of `metric_time_dimension_instances` in `_make_time_spine_data_set` and create `DATE_TRUNC` columns for each.
2. Maybe we don't want to prematurely optimize this query, and I should leave that to the optimizer. In that case we could create a `DATE_TRUNC` column for every possible granularity in `_make_time_spine_data_set`, and expect the unrequested ones to get filtered out by `FilterElementsNode`.

I'll definitely need to update the logic to handle case #1, but not sure about #2. Thoughts on that one?

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1343,17 +1297,49 @@ def visit_join_to_time_spine_node(self, node: JoinToTimeSpineNode[SourceDataSetT
...
+        # Add requested granularity to time spine spec & column.
+        # Time spine defaults to DAY, so DAY granularity does not require any change here.
+        if node.metric_time_dimension_spec.time_granularity == TimeGranularity.DAY:
+            time_spine_instance_set = time_spine_dataset.instance_set
+            time_spine_column_alias = original_time_dim_instance.associated_column.column_name
+        else:
+            new_time_dim_spec = TimeDimensionSpec(
+                element_name=original_time_dim_instance.spec.element_name,
+                entity_links=original_time_dim_instance.spec.entity_links,
+                time_granularity=node.metric_time_dimension_spec.time_granularity,
```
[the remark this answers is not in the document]
courtneyholcomb: Why's that? I've included some integration tests with granularities larger than day for reference if you want to point to a SQL example.

courtneyholcomb:
Splitting into smaller PRs:
https://github.com/dbt-labs/metricflow/pull/743
https://github.com/dbt-labs/metricflow/pull/744

metricflow/dataflow/dataflow_plan.py
courtneyholcomb: Actually, after some noodling on this I understood more what you meant. Ended up going back to inheriting the lowest granularity available from the parent dataset after all. So you can ignore this thread!

metricflow/dataflow/builder/dataflow_plan_builder.py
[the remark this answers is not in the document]
courtneyholcomb: To wrap up this thread - I ended up implementing the logic to handle multiple metric time dimensions in the dataflow to SQL logic, not the time spine builder.

PR #727, opened 2023-08-16
Fix click bug with mypy

WilliamDee:
### Description
Earlier, all the mypy checks broke when click upgraded to `8.1.4`. This caused us to restrict the version to a lower version. Now, that fix is out on `8.1.6` (info here https://github.com/pallets/click/issues/2558). So reverting the version restriction and setting the restriction for mypy use only in the dev dependencies.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ dependencies = [
   "Jinja2>=2.11.3",
   "MarkupSafe==2.0.1",
   "PyYAML~=6.0",
-  "click>=7.1.2, <8.1.4",
+  "click>=7.1.2",
   "dbt-core~=1.6.0",
   "dbt-semantic-interfaces~=0.2.0",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -28,7 +28,7 @@ dependencies = [
   "Jinja2>=2.11.3",
   "MarkupSafe==2.0.1",
   "PyYAML~=6.0",
-  "click>=7.1.2, <8.1.4",
+  "click>=7.1.2",
```
tlento, line 31: Oh cool, I wasn't sure if we could relax that prod version pin.

tlento (approved):
Thank you!

pyproject.toml
WilliamDee: in hindsight, I probably could've just did that version pin in the dev-dependency, rather than the prod one 😬

PR #735, opened 2023-08-23
Expose measures for metrics on MFEngine

courtneyholcomb:
Resolves #SL-814

### Description
Expose measures for metrics on MFEngine, ensuring that `agg_time_dimension` is populated even when it is inherited from the semantic model default.
We need to know the agg_time_dimension for a given measure so that the API can expose a list of all measures and the name of their `agg_time_dimension` for a given metric. This allows users to understand what `metric_time` actually represents for a given metric.


metricflow/model/semantics/semantic_model_lookup.py
```diff
@@ -141,6 +141,19 @@ def get_measure(self, measure_reference: MeasureReference) -> Measure:  # noqa:
...
 
+    def get_measure_with_agg_time_dimension(self, measure_reference: MeasureReference) -> Measure:  # noqa: D
+        measure = self.get_measure(measure_reference=measure_reference)
+        if not measure.agg_time_dimension:
+            agg_time_dimension = self.get_agg_time_dimension_for_measure(measure_reference=measure_reference)
+            if agg_time_dimension:
+                # Copy original Measure object, updating the `agg_time_dimension` field.
+                # Note: `Measure`` protocol does not include Pydantic methods.
+                # Could remove this `type: ignore` if we added the `copy` method to the protocol in DSI.
+                return measure.copy(  # type: ignore[attr-defined]
```
tlento, line 152: We need to avoid typechecker overrides on internal operations. The contract for these methods is that the type spec is as defined, this inserts an implementation detail that will throw runtime errors for people implementing against the type spec.

If we need this method I'd rather a standard Python copy.copy() -> modify, assuming Pydantic allows it, but I think it'll be easier overall to follow the pattern we use elsewhere of having the engine return objects that are not directly sourced from here. If we do that for now we can add a copy() method to the interface later on and consolidate those secondary object types all at once.

metricflow/model/semantics/semantic_model_lookup.py
```diff
@@ -141,6 +141,19 @@ def get_measure(self, measure_reference: MeasureReference) -> Measure:  # noqa:
             semantic_model=semantic_model, measure_reference=measure_reference
         )
 
+    def get_measure_with_agg_time_dimension(self, measure_reference: MeasureReference) -> Measure:  # noqa: D
```
tlento, line 144: Sorry, I'm not clear on the context behind this change, so this may be a silly question, but do we need this method? For other cases where we need to do some kind of custom stuff at the engine API level we have custom objects. It might be simplest to add a Measure type and then populate the agg_time_dimension there, similar to what we do with Metric: https://github.com/dbt-labs/metricflow/blob/main/metricflow/engine/models.py#L27-L58

metricflow/test/api/test_metricflow_client.py
```diff
@@ -93,6 +93,32 @@ def test_list_dimensions(mf_client: MetricFlowClient) -> None:  # noqa: D
     assert tuple(dim.name for dim in dimensions) == ("metric_time", "metric_time")
 
 
+def test_get_measures_for_metrics(mf_client: MetricFlowClient) -> None:  # noqa: D
+    measures = mf_client.engine.get_measures_for_metrics(["bookings"])
+    assert measures
```
tlento, line 98: Doesn't much matter, but maybe `assert len(measures) == 1` just for consistency?

tlento (commented):
Apart from the typechecker override this all seems reasonable. I think it'll be more expedient to keep the change at the level of the engine and use the existing semantic_model_lookup methods to populate the additional fields you need.

Would you mind also updating the PR description with a little more context on why this change is being made? I'm assuming it's to support metadata lookups on the cloud side, but it's not clear to me how we're expecting this method to be used from the PR itself.

Thanks!

metricflow/model/semantics/semantic_model_lookup.py
courtneyholcomb: Yeah this is a question I had and I think my assumption was just wrong - I assumed that it must be intentional that we were not populating `agg_time_dimension` on `Measure`. If that's not the case, I wonder if we should be populating `agg_time_dimension` whenever we build a `Measure` object? It seems weird to have an `agg_time_dimension` field available on `Measure` anywhere in the code if it is not always properly populated.

courtneyholcomb: Also I'm not sure if it was clear to you why I did this, so I'll explain:
The `Measure` object has an `agg_time_dimension` field, always. This is not a field I'm adding in MFEngine. But that field is not always populated. In the case when the `agg_time_dimension` is inherited from the default specified for the semantic model, that field returns `null`. Is that intentional? If not, my thought is that we should do what I mentioned above and make sure it's populated anytime we build a `Measure` object.

metricflow/model/semantics/semantic_model_lookup.py
courtneyholcomb: The reason I did it this way is that I get a type error if I try to modify that attribute of a Pydantic object:
`Property "agg_time_dimension" defined in "Measure" is read-only`
So I looked up the recommended way to do this in Pydantic.

courtneyholcomb: > I think it'll be easier overall to follow the pattern we use elsewhere of having the engine return objects that are not directly sourced from here.
The reason this doesn't seem like the right solution to me is that there isn't actually anything special or different about the `Measure` object I want to expose. To me it just seems like a bug that the `agg_time_dimension` attribute isn't populated.
The argument against that is if we want the `PydanticMeasure` object to match its YAML representation exactly - so if it's supposed to inherit the default `agg_time_dimension`, that wouldn't show up. I can implement it that way, but it seems very bug-prone to me since any developer who doesn't know that might use `PydanticMeasure` objects with the assumption that the `agg_time_dimension` field is populated with its default (as I did when I jumped into this code).

metricflow/model/semantics/semantic_model_lookup.py
tlento: It's not practical to always populate this parameter because of the way dbt core does these model default -> sub-object transformations, so we pushed the "always populate" logic into the outside accessor you're using. This makes everything difficult, and untangling the whole mess is fairly involved.

What you really need is a version of Measure with agg_time_dimension set to required instead of optional. I think the simplest way to get it is to do what we're doing with Metric and Dimension and make an engine-level dataclass representing the Measure properties for now. You can make agg_time_dimension required on that object, and then populate it from here without needing any API expansion on the SemanticModelLookup class. That way you don't ever need to deal with the Optional nature of the property on the MetricFlow Server side, and we have at least some typechecker support around our contract for the properties we include on these objects.

Perhaps we'll consolidate things later, although I suspect we'll live like this forever. :/

courtneyholcomb: > What you really need is a version of Measure with agg_time_dimension set to required instead of optional.

Yeah that makes sense to me! Cool I can do it that way.

metricflow/model/semantics/semantic_model_lookup.py
courtneyholcomb: I think this conversation is also resolved if we expose a new engine-level dataclass 👍

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "metricflow"
-version = "0.202.0.dev0"
+version = "0.202.0.dev1"
```
tlento, line 7: Let's revert this for now and update separately, we've got a bunch of these version update changes flying around and it messes with the hatch build -> deploy process.

tlento (approved):
LGTM, thanks for your patience!

PR #737, opened 2023-08-24
Modified dimensions_for_metric to allow for returning more details

WilliamDee:
## Context
We want a way to return a list of dimensions while showing the available granularities for time dimensions. Currently, we filter dimensions with `LinkableElementProperties.DERIVED_TIME_GRANULARITY` property and that is generally the behaviour we want where we're just returning unique dimensions. With this change, one can request dimensions with that property and it would return dimensions where each row may repeat but has a different granularity.  Ie., something like this,
```
Dimension("ordered_at", time_granularity="day")
Dimension("ordered_at", time_granularity="month")
Dimension("ordered_at", time_granularity="quarter")
Dimension("ordered_at", time_granularity="year")
Dimension("customer_type", time_granularity=None)
...
```

### DIFF
--- a/metricflow/engine/models.py
+++ b/metricflow/engine/models.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import List, Optional, Sequence
 
+from dbt_semantic_interfaces.implementations.elements.dimension import PydanticDimensionTypeParams
 from dbt_semantic_interfaces.protocols.dimension import (
     Dimension as SemanticManifestDimension,
 )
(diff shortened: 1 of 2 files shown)


metricflow/engine/models.py
```diff
@@ -78,12 +79,18 @@ def from_pydantic(cls, pydantic_dimension: SemanticManifestDimension, path_key:
             element_name=path_key.element_name,
             entity_links=path_key.entity_links,
         ).qualified_name
+        parsed_type_params: Optional[DimensionTypeParams] = None
```
courtneyholcomb, line 82: what's the difference here ?

courtneyholcomb (approved):

metricflow/engine/models.py
WilliamDee: currently, we just pull it from the PydanticDimension which just renders to `TimeGranularity.DAY` (I think that just gets pulled from the config) whereas the `path_key.time_granularity` has the correct rendered granularity.

PR #738, opened 2023-08-25
Bump to 0.201.0.dev0

DevonFulcher:


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "metricflow"
-version = "0.201.0"
+version = "0.201.0.dev0"
```
tlento, line 7: Should this be `0.202.0.dev0`? I thought these dev suffixes were pre-release markers.

tlento (approved):
Let's update the minor version along with this to mark as pre-release.

tlento:
Oh no! Remove the change log please. We have a skip changelog label for that.

pyproject.toml
DevonFulcher: All the previous releases on pypi didn't follow this pattern 
image

DevonFulcher:
Would it help to update this page with the details around not adding a changelog and the nuances of picking a version number https://www.notion.so/dbtlabs/Metricflow-Build-Release-Process-81023685d93a45da8211b05dc36b7bf8?

PR #742, opened 2023-08-28
Updates to snapshot generation script

courtneyholcomb:
### Description
Fixes a couple of small issues with the snapshot generation script:
1. Allow special characters when JSON-parsing BQ password.
2. Don't use persistent source schema for Postgres.

### DIFF
--- a/metricflow/test/fixtures/sql_client_fixtures.py
+++ b/metricflow/test/fixtures/sql_client_fixtures.py
@@ -74,7 +74,8 @@ def __configure_test_env_from_url(url: str, password: str, schema: str) -> sqlal
 
 def __configure_bigquery_env_from_credential_string(password: str, schema: str) -> None:
     credential_string = password.replace("'", "")
-    credentials = json.loads(credential_string)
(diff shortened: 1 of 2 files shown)


metricflow/test/fixtures/sql_client_fixtures.py
```diff
@@ -74,7 +74,8 @@ def __configure_test_env_from_url(url: str, password: str, schema: str) -> sqlal
 
 def __configure_bigquery_env_from_credential_string(password: str, schema: str) -> None:
     credential_string = password.replace("'", "")
-    credentials = json.loads(credential_string)
+    # `strict=False` required to work with BQ password characters.
+    credentials = json.loads(credential_string, strict=False)
```
tlento, line 78: Interesting, the CI credential string doesn't need this. I can't remember what kind of escaping I did to get it working, though. Must be something to do with the nested JSON. Anyway everything seems fine so 🤷

tlento (approved):
Thank you!

PR #743, opened 2023-08-28
Remove unnecessary `JoinOverTimeRange` step from dataflow plan

courtneyholcomb:
### Description
Currently, for all cumulative metrics, we add a `JoinOverTimeRange` step to the dataflow plan. However, some cumulative metrics are able to be queried without `metric_time`, in which case we should not `JoinOverTimeRange`. When we visit that node unnecessarily, there are two repercussions:
1. We skip a SQL table alias. No major consequence here, but it is confusing to look at the SQL and wonder why it's skipped.
2. We assume inaccurate costing, since the `JoinOverTimeRangeNode` assumes `num_aggregations=1`. This results in `num_aggregations` being off by 1 for the overall query.

This PR removes that unneeded step from the dataflow plan.


tlento (approved):
Thanks for splitting this out into such an easily readable PR. The previous silliness is so obvious here!

PR #744, opened 2023-08-29
Enable granularities for derived metrics with time offset

### LINKED ISSUE #725: [Bug] Derived metric offset at yearly granularity returns no results in BQ

### Current Behavior

Say I have a derived metric defined as follow:
```
 name: total_users_yoy_pct_change
    label: Total Users YoY % Change
    type: derived
    type_params:
      expr: (total_users - total_users_prev_year)*100/total_users_prev_year
      metrics:
        - name: total_users
        - name: total_users
          offset_window: 1 year
          alias: total_users_prev_year
```

When I run `mf query --metrics total_users_yoy_pct_change --group-by metric_time__year` I get no results. This is because BigQuery is using ISOYEAR to truncate metric time to year. A quick primer on isoyear; isoyears start on the first Monday of each year. So `DATE_TRUNC(cast('2023-08-01' as datetime), isoyear)`  will return 2023-01-02 and `DATE_TRUNC(cast('2022-08-01' as datetime), isoyear)` will return 2022-01-03. You can see this behavior in the following subquery:

```
(
      SELECT
        metric_time__year
        , COUNT(DISTINCT total_users) AS total_users_prev_year
      FROM (
        SELECT
          **DATE_TRUNC(cast(date as datetime), isoyear) AS metric_time__year**
          , visitor_id AS total_users
        FROM `inq-warehouse`.`dbt_test`.`fct_sessions` sessions_src_0
      ) subq_7
      GROUP BY
        metric_time__year
    ) subq_9
```
The two offset subqueries used to calculate the derived metric are joined via `DATE_SUB(CAST(subq_10.metric_time__year AS DATETIME), INTERVAL 1 year) = subq_9.metric_time__year.` However, subtracting 1 year from a isodate-truncated date, will not give you the isodate-truncated date of the previous year. You can see this behavior in the following example. The join condition MetricFlow renders would compare 2022-01-02 to 2022-01-03. There are no records that match this equality, so we return no results. 
Screenshot 2023-08-15 at 9 19 43 AM

### Expected Behavior

I would expect the metric value from the previous year to be properly joined. If we use `year` instead of `isoyear` as the date type to truncate metric_time, this join would resolve correctly. 

### Steps To Reproduce

1. Check out jaffle-sl-template and update your profile to Bigquery
2. Update the `revenue_growth_mom` query to a year over year comparisom
```
  - name: revenue_growth_yoy
    description: "Percentage growth of revenue compared to 1 year ago. Excluded tax"
    type: derived
    label: Revenue Growth % Y/Y
    type_params:
      expr: (current_revenue - revenue_prev_year)*100/revenue_prev_year
      metrics:
        - name: revenue
          alias: current_revenue
        - name: revenue
          offset_window: 1 year
          alias: revenue_prev_year
```
3. Cast ordered_at to datetime
```
      - name: ordered_at
        expr: cast(ordered_at as datetime)
        type: time
        type_params:
          time_granularity: day
```
4. Run `mf query --metrics revenue_growth_yoy --group-by metric_time__year` and observe that no results are returned.
Screenshot 2023-08-15 at 9 43 18 AM

When I run the same query in snowflake i get the following result:
Screenshot 2023-08-15 at 9 42 54 AM

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
- metricflow:
```

_No response_

### Additional Context

_No response_

courtneyholcomb:
Resolves #725

### Description
With the current placement of the `JoinToTimeSpineNode` in the dataflow plan, we filter down to time columns with the requested granularity BEFORE joining to time spine. Since joining to time spine expects a time column with `DAY` granularity, this means queries with any granularity besides `DAY` don't return expected results. Here, we move the time spine join to happen before filtering elements to fix that issue. So instead of `FilterElementsNode` -> `JoinToTimeSpineNode`, we do `JoinToTimeSpineNode` -> `FilterElementsNode`.
For cumulative metrics, we need to factor in the `JoinOverTimeRangeNode`. This PR updates the dataflow plan from `FilterElementsNode` -> `JoinOverTimeRangeNode` (both required) to `JoinOverTimeRangeNode` (optional, only used if a time dimension is requested) -> `JoinToTimeSpineNode` (optional, only used for offset metrics) -> `FilterElementsNode`. This ensures we'll have the columns needed when joining to time spine.


tlento (commented):
So far this looks great to me but GitHub has completely screwed up the rendering on certain parts of this PR so it'll take me a little longer to piece things together and get through the dataflow to sql section. Will do this after standups today!

metricflow/dataflow/dataflow_plan.py
```diff
@@ -791,13 +799,13 @@ def functionally_identical(self, other_node: DataflowPlanNode[SourceDataSetT]) -
     def with_new_parents(  # noqa: D
         self, new_parent_nodes: Sequence[BaseOutput[SourceDataSetT]]
     ) -> JoinToTimeSpineNode[SourceDataSetT]:
-        # Is this right? how is this method used?
         assert len(new_parent_nodes) == 1
         return JoinToTimeSpineNode[SourceDataSetT](
             parent_node=new_parent_nodes[0],
+            metric_time_dimension_specs=self.metric_time_dimension_specs,
```
tlento, line 805: We need to add a check to the `functionally_identical` method to ensure the metric_time_dimension_specs are the same between other_node and self. 

These methods are used for an impressively difficult-to-reason-about optimizer that prunes out part of the dataflow plan, so if we miss that equivalence check we run the risk of removing a subquery that we need to remain in place.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1339,17 +1339,55 @@ def visit_join_to_time_spine_node(self, node: JoinToTimeSpineNode[SourceDataSetT
...
+            SqlColumnReferenceExpression, SqlDateTruncExpression
+        ] = SqlColumnReferenceExpression(
+            SqlColumnReference(table_alias=time_spine_alias, column_name=original_time_dim_instance.spec.qualified_name)
+        )
+
+        # Add requested granularities (skip for default granularity).
+        metric_time_select_columns = []
+        metric_time_dimension_instances = []
+        for metric_time_dimension_spec in node.metric_time_dimension_specs:
+            if metric_time_dimension_spec.time_granularity == self._time_spine_source.time_column_granularity:
```
tlento, line 1362: Nice!

metricflow/test/snapshots/test_dataflow_to_sql_plan.py/SqlQueryPlan/DuckDB/test_derived_metric_with_offset_to_grain_and_granularity__plan0_optimized.sql
```diff
@@ -0,0 +1,63 @@
...
+      -- Metric Time Dimension 'ds'
+      SELECT
+        ds AS metric_time__day
+        , 1 AS bookings
+      FROM ***************************.fct_bookings bookings_source_src_10001
+    ) subq_20
+    ON
+      DATE_TRUNC('month', subq_22.ds) = subq_20.metric_time__day
+    GROUP BY
+      DATE_TRUNC('week', subq_22.ds)
```
tlento, line 51: I think something's wrong here. Is this supposed to show bookings at start of month for every week in the time range?

Right now, it looks like it'll show bookings at start of month for weeks that include the first day of the month, because the INNER JOIN filters out a bunch of dates. Then there's no matching value in the inner join expression below for any week that's missing from this subset.

tlento (commented):
Sorry, I'm really confused by this SQL, let me think about this more.

If my understanding of the metric configuration is wrong please let me know.

metricflow/test/snapshots/test_dataflow_to_sql_plan.py/SqlQueryPlan/DuckDB/test_derived_metric_with_offset_to_grain__plan0_optimized.sql
```diff
@@ -29,31 +29,26 @@ FROM (
...
+      , SUM(subq_20.bookings) AS bookings_at_start_of_month
+    FROM ***************************.mf_time_spine subq_22
     INNER JOIN (
-      -- Aggregate Measures
-      -- Compute Metrics via Expressions
+      -- Read Elements From Semantic Model 'bookings_source'
+      -- Metric Time Dimension 'ds'
       SELECT
-        metric_time__day
-        , SUM(bookings) AS bookings_at_start_of_month
```
tlento, line 41: Ok, this is where we're hitting the problem. Before, we would aggregate this metric per day and then do the join. Now, we do the join and then compute the metric. But in doing that, we explode the set of input rows to the metric.

I know the description lays out a reason for switching the join order, but what the original code was doing - which the updated code does not handle - was pre-aggregating in order to avoid inappropriate fanout on the time spine join. Without that pre-aggregation you cannot get correct results for expanded granularities for offset_to_grain on additive measures, because you end up over-counting values.

metricflow/test/integration/test_cases/itest_metrics.yaml
```diff
@@ -863,3 +863,142 @@ integration_test:
...
+      INNER JOIN (
+        SELECT
+          ds AS metric_time__day
+          , 1 AS bookings_at_start_of_month
+        FROM {{ source_schema }}.fct_bookings
+      ) d
+      ON {{ render_date_trunc("c.ds", TimeGranularity.MONTH) }} = d.metric_time__day
+      GROUP BY metric_time__week
+    ) b
+    ON a.metric_time__week = b.metric_time__week
```
tlento, line 918: I think we need something like this, or else to do aggregation first and then join as we were doing before, but somehow to propagate the granularity spec down for the offset scenario:

```suggestion
    SELECT
      COALESCE(a.metric_time__week, b.metric_time__week)  AS metric_time__week
      , bookings - bookings_at_start_of_month AS bookings_growth_since_start_of_month
    FROM (
      SELECT
        {{ render_date_trunc("ds", TimeGranularity.WEEK) }} AS metric_time__week
        , SUM(1) AS bookings
      FROM {{ source_schema }}.fct_bookings
      GROUP BY metric_time__week
    ) a
    INNER JOIN (
      SELECT
        {{ render_date_trunc("c.ds", TimeGranularity.WEEK) }} AS metric_time__week
        , SUM(d.bookings_at_start_of_month) AS bookings_at_start_of_month
      FROM {{ mf_time_spine_source }} c
      INNER JOIN (
        SELECT
          ds AS metric_time__day
          , 1 AS bookings_at_start_of_month
        FROM {{ source_schema }}.fct_bookings
      ) d
      ON {{ render_date_trunc("c.ds", TimeGranularity.MONTH) }} = d.metric_time__day
      -- filter time spine join output to the dates we need to carry forward
      WHERE {{ render_date_trunc("c.ds", TimeGranularity.WEEK) }} = c.ds
      GROUP BY metric_time__week
    ) b
    ON a.metric_time__week = b.metric_time__week
```

tlento (commented):
I'm not sure of the best way to fix this, but offset to grain does not appear to be working correctly.

It appears, although I haven't proven it to myself, that offset to grain either requires the aggregation to happen prior to the join, or requires us to apply a filter to the time spine prior to doing the aggregation, because we want to aggregate to a single point in time and then fan that value out to all of the values in the output result set.

metricflow/dataflow/dataflow_plan.py
courtneyholcomb: Good catch!! Updated!

metricflow/test/integration/test_cases/itest_metrics.yaml
```diff
@@ -981,3 +981,24 @@ integration_test:
       GROUP BY metric_time__day
     ) c
     ON {{ render_date_sub("d", "ds", 2, TimeGranularity.DAY) }} = c.metric_time__day
+---
+integration_test:
+  name: derived_metric_offset_window_and_multiple_granularities
+  description: Tests a derived metric offset query with window and 2 metric_time granularities
+  model: SIMPLE_MODEL
+  metrics: ["bookings_5_day_lag"]
+  group_bys: ["metric_time__month", "metric_time__week"]
```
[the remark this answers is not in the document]
courtneyholcomb: Also - is your concern related to this PR? Does that filter work currently?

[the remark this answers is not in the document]
tlento: I am 80% sure that filter works. If it does, it'd be nice to update this test to use it so we can keep an example running in CI. If it does not, we should just file an issue and fix it separately.

metricflow/test/integration/test_cases/itest_metrics.yaml
courtneyholcomb: Had to do a bunch of testing on real SQL results to wrap my head around this but can confirm that that filter works! Question though - does it make a difference if we do the filter in the `WHERE` clause vs the `JOIN` clause? They appear to behave the same way so I'm wondering if one is more efficient.

courtneyholcomb: Also, this complicates things if you query with multiple granularities. So I need to do some more SQL testing to figure out what syntax we want in that case.

metricflow/test/integration/test_cases/itest_metrics.yaml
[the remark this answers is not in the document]
courtneyholcomb: I think I would put that in a separate test - this one is already testing such a specific edge case and it makes it very hard to figure out what we actually expect the results to look like with that filter. But if the case you're looking for is just querying by coarse grain + filtering by fine grain, I can add that!

metricflow/test/integration/test_cases/itest_metrics.yaml
courtneyholcomb: So, we could do an `OR` statement for multiple granularities, like:
`WHERE DATE_TRUNC(MONTH, ds) = ds OR DATE_TRUNC(WEEK, ds) = ds`
 That results in a result set like this for a metric with offset to grain:
image
I haven't thought of a better answer than this.

courtneyholcomb: I went ahead and implemented it in the `WHERE` clause and with the `OR` statement for multiple granularities. We can go through the generated SQL and make sure that all makes the most sense!

tlento: Doing it in the JOIN clause should be more efficient, but I don't know if all engines support that construction. Let's stick with the WHERE for now and we can look into optimizing later.

tlento:
Also, thank you for restructuring the PR, this is great! I recommend a straight merge, rather than a squash, because the commits are all well-formed and it'll be more readable in git diff, but you may commit this in whatever manner is most comfortable for you!

PR #768, opened 2023-09-07
Remove generic source dataset constructs

tlento:
The original implementation of MetricFlow used a generic type
annotation for the "Source DataSet" of the dataflow plan. This was
in place to support some specific implementations of Transform's
proprietary product logic, and as an interface boundary it proved
ineffective. Consequently, attempts to revisit that functionality
will rely on other approaches.

Within MetricFlow, this generic type resolves, to one of two concrete types:
the SqlDataSet, or the SemanticModelDataSet, which inherits from
SqlDataSet. Since it's far simpler to just use SqlDataSet as the
required type than to litter the codebase with generic typehints and
concrete type specs in some places but not others, we make that change
along with some associated cleanup. In particular, the SqlDataSet class
rightly belongs with the other DataSet subclasses, rather than in the
plan_conversion package.


tlento:
* main
  * **PR #768** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/768" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dbt-labs/metricflow/768?utm_source=stack-comment).

plypaul (approved):

PR #770, opened 2023-09-08
Querying with `EXTRACT`

courtneyholcomb:
Resolves #


courtneyholcomb:
Moving these changes to a new PR with more organized commits!

PR #772, opened 2023-09-13
Querying with DATE PART

courtneyholcomb:
### Description
Enable aggregating time dimensions by their date part (e.g., `EXTRACT(month FROM '2020-01-01')` -> `1`).
Completes SL-817.


metricflow/query/query_parser.py
```diff
@@ -529,6 +537,34 @@ def _validate_order_by_specs(
...
+        TODO: figure out expected behavior for date part with these types of metrics.
+        """
+        date_part_requested = False
+        for time_dimension_spec in time_dimension_specs:
+            if time_dimension_spec.date_part:
+                date_part_requested = True
+                if time_dimension_spec.date_part.to_int() < time_dimension_spec.time_granularity.to_int():
+                    raise RequestTimeGranularityException(
+                        f"Date part {time_dimension_spec.date_part.name} is not compatible with time granularity "
+                        f"{time_dimension_spec.time_granularity.name}."
```
tlento, line 554: Shall we update this to indicate what is compatible? It's any date part that's bigger, right? So you can ask for year on daily granularity, but you can't ask for day on yearly granularity.

metricflow/query/query_parser.py
```diff
@@ -529,6 +537,34 @@ def _validate_order_by_specs(
...
+                    )
+        if date_part_requested:
+            for metric_reference in metric_references:
+                metric = self._metric_lookup.get_metric(metric_reference)
+                if metric.type == MetricType.CUMULATIVE:
+                    raise UnableToSatisfyQueryError("Cannot extract date part for cumulative metrics.")
```
tlento, line 560: Huh. I don't think we should do this here, because none of the other validate methods do it, but now I'm wondering if we should validate everything and raise one exception with all the problems, instead of this piecemeal thing.

metricflow/query/query_parser.py
```diff
@@ -643,59 +679,90 @@ def _parse_metric_names(
...
+        linkable_elements: Optional[Sequence[QueryParameter]] = None,
     ) -> QueryTimeLinkableSpecSet:
         """Convert the linkable spec names into the respective specification objects."""
-        qualified_linkable_names = [x.lower() for x in qualified_linkable_names]
+        assert not (qualified_linkable_names and linkable_elements)
+
+        structured_names: List[StructuredLinkableSpecName] = []
+        if qualified_linkable_names:
+            qualified_linkable_names = [x.lower() for x in qualified_linkable_names]
+            structured_names = [StructuredLinkableSpecName.from_name(name) for name in qualified_linkable_names]
```
tlento, line 694: So at the moment these would never contain a date_part appendage, right? Is it possible for us to validate this?

metricflow/query/query_parser.py
```diff
@@ -643,59 +679,90 @@ def _parse_metric_names(
...
+        elif linkable_elements:
+            for linkable_element in linkable_elements:
+                parsed_name = StructuredLinkableSpecName.from_name(linkable_element.name)
+                if parsed_name.time_granularity:
+                    raise ValueError(
+                        "Time granularity must be passed in the `grain` attribute for `group_by` query param."
```
tlento, line 700: Will this always be a group_by? I think at some point it might be a filter or order by or something, but I'm honestly not sure. @DevonFulcher, do you know?

metricflow/query/query_parser.py
```diff
@@ -643,59 +679,90 @@ def _parse_metric_names(
...
+        assert not (qualified_linkable_names and linkable_elements)
+
+        structured_names: List[StructuredLinkableSpecName] = []
+        if qualified_linkable_names:
+            qualified_linkable_names = [x.lower() for x in qualified_linkable_names]
+            structured_names = [StructuredLinkableSpecName.from_name(name) for name in qualified_linkable_names]
+        elif linkable_elements:
+            for linkable_element in linkable_elements:
+                parsed_name = StructuredLinkableSpecName.from_name(linkable_element.name)
+                if parsed_name.time_granularity:
```
tlento, line 698: Interesting. Are we hitting this?

Seems like this validation should be done in the query param objects but I guess we're not there yet.

metricflow/query/query_parser.py
```diff
@@ -643,59 +679,90 @@ def _parse_metric_names(
...
                 dimension_specs.append(DimensionSpec(element_name=element_name, entity_links=entity_links))
             elif EntityReference(element_name=element_name) in self._known_entity_element_references:
                 entity_specs.append(EntitySpec(element_name=element_name, entity_links=entity_links))
             else:
+                valid_group_bys_for_metrics = self._metric_lookup.element_specs_for_metrics(list(metric_references))
                 valid_group_by_names_for_metrics = sorted(
-                    x.qualified_name for x in self._metric_lookup.element_specs_for_metrics(list(metric_references))
+                    list(
+                        set(
+                            x.qualified_name if qualified_linkable_names else x.element_name
```
tlento, line 746: I'm really confused by this. It looks like these are different names. The qualified name includes the spec path but the element name does not, right?

metricflow/query/query_parser.py
```diff
@@ -643,59 +679,90 @@ def _parse_metric_names(
...
+                            for x in valid_group_bys_for_metrics
+                        )
+                    )
                 )
 
+                # If requested by name, show qualified name. If requested as object, show element name.
+                display_name = structured_name.qualified_name if qualified_linkable_names else element_name
```
tlento, line 753: Oh, is this the same as above? How does the output differ? Is this fixing that thing where our suggested names just don't work correctly?

If so, we might want to factor this out into a function so we can put a docblock around it.

metricflow/specs/where_filter_time_dimension.py
```diff
@@ -86,4 +94,6 @@ def _convert_to_time_dimension_spec(
             element_name=parameter_set.time_dimension_reference.element_name,
             entity_links=parameter_set.entity_path,
             time_granularity=parameter_set.time_granularity,
+            # TODO: add date_part to TimeDimensionCallParameterSet in DSI
+            # date_part=parameter_set.date_part,
```
tlento, line 98: I don't think we want to do this. That makes date_part a piece of the metric spec, so if we want this as a query-only construct it's probably best to keep it in MetricFlow.

metricflow/test/integration/test_cases/itest_metrics.yaml
```diff
@@ -1043,3 +1043,77 @@ integration_test:
...
+        ds AS metric_time__day
+        , 1 AS bookings_5_day_lag
+      FROM {{ source_schema }}.fct_bookings
+    ) b
+    ON {{ render_date_sub("a", "ds", 5, TimeGranularity.DAY) }} = b.metric_time__day
+    GROUP BY metric_time__extract_month
+---
+integration_test:
+  name: date_part_overrides_granularity
+  description: Test query using date_part with incompatible granularity; should override granularity
```
tlento, line 1110: Wait, really? Shouldn't this just be an error? I'd rather throw an exception here, it's hard to know what output the user is expecting.

Like the only time I can think of where I would use this is I want to do some comparison for monthly metrics based on which day of the week the month starts on, in which case making these daily is incorrect. Or I might just be doing something silly, in which case an error is appropriate.

tlento (commented):
Overall this seems reasonable although it's hard for me to see what issues might crop up.

I've left a bunch of small things inline, but my main concerns are with validation vs overrides for mismatched grain/extract parameters, and with the way qualified_name gets used.

For the latter, looking at the codebase, I think we're safe for now, but I'm wondering if there's some scenario where we update some code that accidentally tries to parse a qualified name with a date_part appended and then weird stuff happens. I'll think about this a bit more but if we can't come up with a reasonable safeguard (maybe we have a different property for "select names" or something?) I guess we'll have to fall back on documentation for qualified_name.

metricflow/query/query_parser.py
DevonFulcher: Yes, you will soon be able to specify `Dimension('metric_time').grain('WEEK')` in the where param or filter field.

metricflow/sql/render/big_query.py
```diff
@@ -129,6 +130,17 @@ def visit_date_trunc_expr(self, node: SqlDateTruncExpression) -> SqlExpressionRe
...
         )
 
+    @override
+    def render_date_part(self, date_part: DatePart) -> str:
+        if date_part == DatePart.DOY:
+            return "dayofyear"
+        if date_part == DatePart.DOW:
+            return "dayofweek"
+        if date_part == DatePart.WEEK:
+            return "isoweek"
```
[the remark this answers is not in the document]
courtneyholcomb: The other engines return the value of the enum (so `dow`, `doy`, and `week`). BQ is the only one that doesn't support `dow` and `doy`, and the only one that we use `isoweek` for so it gets the special logic!

metricflow/naming/linkable_spec_name.py
```diff
@@ -44,18 +47,30 @@ def from_name(qualified_name: str) -> StructuredLinkableSpecName:
...
-        """Return the full name form. e.g. ds or listing__ds__month."""
+        """Return the full name form. e.g. ds or listing__ds__month.
+
+        If date_part is specified, don't include granularity in qualified_name since it will not impact the result.
+        """
         items = list(self.entity_link_names) + [self.element_name]
-        if self.time_granularity:
+        if self.date_part:
+            items.append(self.date_part_suffix(date_part=self.date_part))
+        elif self.time_granularity:
```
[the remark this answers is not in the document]
courtneyholcomb: > does it blow up with a TypeError

It won't blow up with a `TypeError` - it should parse to something like `ds__extract_year`, and it will think that's the `entity__dimension` name. Then the user will get an error because that dimension doesn't exist.

> does it work for any values that happen to be the same across the two enums?

The values won't match granularity values because all the date part column names include `extract_` (no dunder).

Initially I added support for parsing `date_part` in the `from_name` function, but then decided to remove it because I know we're trying to move away from using the dunder syntax for anything except `entity__dimension`!

metricflow/engine/metricflow_engine.py
```diff
@@ -456,6 +456,7 @@ def _create_execution_plan(self, mf_query_request: MetricFlowQueryRequest) -> Me
                 query_spec = self._query_parser.parse_and_validate_query(
                     metric_names=mf_query_request.metric_names,
                     group_by_names=mf_query_request.group_by_names,
+                    group_by=mf_query_request.group_by,
```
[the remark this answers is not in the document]
courtneyholcomb: Yep.

metricflow/query/query_parser.py
courtneyholcomb: Correct!
If `qualified_linkable_names` is passed, we're dealing with the `group_by_names` query param - a list of string `qualified_names`. In that case, we want to check those against the qualified names allowed for the metric.
Otherwise, we know `linkable_elements` was passed into this function, which means the user used the `group_by` query param. That means we have objects instead of strings, and the object name is the `element_name`, not the `qualified_name`. So we want to check those names against the allowed `element_names` for the metric.
Also used a `set` here since we might have repeated `element_names` in the same query.

metricflow/query/query_parser.py
courtneyholcomb: This shouldn't change the previous behavior, so nope not a bug fix!
All the refactors to this function are to enable passing object `group_by` into `parse_and_validate_query`. Previously, at the start of `parse_and_validate_query`, we would parse the `group_by` objects into `qualified_names` and stop dealing with objects at all. That doesn't work with `date_part` since we can't parse `date_part` into a qualified name and still retain the granularity, so I refactored these functions to handle the full objects (figuring that is the direction we want to move in anyway).

metricflow/specs/where_filter_time_dimension.py
courtneyholcomb: Sounds gooood!

metricflow/test/integration/test_cases/itest_metrics.yaml
courtneyholcomb: Oooh interesting. I was not imagining we would want to support querying a date part smaller than the granularity requested. In that case I figured we would just error (which is what I initially implemented). But then I ran into some issues with setting a default granularity based on what was defined in the metric, and ended up deciding to just override what was requested since, if the granularity is >= the date part, changing the granularity should not impact the date part.

metricflow/query/query_parser.py
courtneyholcomb: Yes! I can add that!

metricflow/query/query_parser.py
courtneyholcomb: I put this here because it was similar to what's happening in `_validate_no_time_dimension_query()`. Where else would you expect this to live?
I don't disagree about raising one exception, but not sure it's in scope for this PR.

metricflow/query/query_parser.py
courtneyholcomb: I can add validation to `StructuredLinkableSpecName.from_name()` to error if there is a `date_part`, if that's what you mean!

metricflow/query/query_parser.py
courtneyholcomb: Yeah... This is something I can fix. I'm just learning that there aren't actually any implementations of the `QueryParam` protocol yet, so I can add an implementation class to this PR and put the validation in there.

metricflow/query/query_parser.py
courtneyholcomb: I can update this to say `Time granularity must be passed in the `grain` attribute of the {type({linkable_element})} parameter`, but I'm not sure the type we receive here will match the type of the parameter they're seeing on the client side 🤔

courtneyholcomb: Updating this to say `Must use object syntax for `grain` parameter if `date_part` is requested.` We will probably want to change that error message when we stop supporting dundered granularity syntax.

metricflow/test/integration/test_cases/itest_metrics.yaml
courtneyholcomb: IMO, we should not support changing the granularity and then requesting a date part on that new granularity. The integration we are building for currently only allows one or the other. If we were to support both in combination, it would greatly complicate the code & SQL - we would need a new column for each _combination_ of granularity & date part in the source dataset. It seems like a very unlikely edge case, but one that we could later decide to support if customers ask for it.
I'm now leaning toward this option: disallow querying `granularity` and `date_part` on the same time dimension. Instead, if `date_part` is requested, we assume the default granularity from the metric (or time dimension if no metric is requested).
WDYT?

metricflow/sql/render/big_query.py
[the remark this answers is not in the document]
tlento: Sorry, I meant if the other engines use the equivalent of isoweek vs some other week numbering standard.

And indeed they do not, because everything is terrible:

https://docs.snowflake.com/en/sql-reference/functions/year#usage-notes

metricflow/test/integration/test_cases/itest_metrics.yaml
tlento: All of this would be so much easier if it didn't have to exist inside MetricFlow itself, but I don't think we have that option at all.

I have to be honest, I'm not entirely sure why we need all of the combinations of all of the granularities in the rendered SQL. I mean we clearly don't, because the optimized SQL doesn't include them all, and it's not like the optimizer has more information than the dataflow plan builder. However, fixing *that* bit of silliness is way beyond the scope of this PR.

Since our core entry-point only allows one or the other I think it's fine to have partial support. We do have some interfaces where users must specify granularity (e.g., the `TimeDimension` query parameter object), so I think we can just add validation that the granularity matches the default, and we can infer grain on the `Dimension()` parameter objects.

Just to make sure I get it, we'll then have:

1. Inference for the default granularity if extract is used without granularity specified
2. Validation that granularity matches the default for cases where both are provided

Ideally we'd put a docblock on a validation function explaining why this exists, because I will definitely forget all about this conversation by the time someone gets annoyed by the limitation.

metricflow/naming/linkable_spec_name.py
```diff
@@ -37,8 +37,13 @@ def from_name(qualified_name: str) -> StructuredLinkableSpecName:
         if len(name_parts) == 1:
             return StructuredLinkableSpecName(entity_link_names=(), element_name=name_parts[0])
 
+        for date_part in DatePart:
+            if name_parts[-1] == StructuredLinkableSpecName.date_part_suffix(date_part=date_part):
+                raise ValueError(
+                    "Dunder syntax not supported for querying date_part. Use `group_by` object syntax instead."
+                )
+
```
tlento, line 45: Nice! Now if we mess up we'll get exceptions and we can restructure the interfaces at that time.

tlento (commented):
Looking good! The one big change we discuss inline about moving from granularity overrides to requiring default granularity. I think your plan makes the most sense.

If you'd prefer to merge this as-is and follow up with that change that's fine with me, just request review with a note. I don't have strong preferences but I'd like to make sure I get to read the change just so I understand how it works.

Thanks!

metricflow/query/query_parser.py
courtneyholcomb: Sounds good, will do!

metricflow/test/integration/test_cases/itest_metrics.yaml
courtneyholcomb: Does that mean we should error if a granularity is requested that does not match the default? Even if we _could_ technically get the requested date part with that granularity?

tlento: I don't know? Let's chat about this tomorrow, I think we're well into the stage where talking will be a lot faster.

metricflow/query/query_parser.py
```diff
@@ -754,6 +758,25 @@ def _parse_linkable_elements(
...
+        requested_dimension_structured_name: StructuredLinkableSpecName,
+        partial_time_dimension_spec: PartialTimeDimensionSpec,
+        metric_references: Sequence[MetricReference],
+    ) -> None:
+        resolved_granularity = list(
+            self._time_granularity_solver.resolve_granularity_for_partial_time_dimension_specs(
+                metric_references=metric_references,
+                partial_time_dimension_specs=(partial_time_dimension_spec,),
+            ).values()
+        )[0].time_granularity
```
tlento, line 772: Instead of doing the list(dict.values())[0], can we just use the index? The dict is keyed on the partial time dimension spec, and any spec that doesn't resolve will raise, so we should be able to do this:

```
resolved_granularity = self._time_granularity_solver.resolve_granularity_for_partial_time_dimension_specs(
    ...
)[partial_time_dimension_spec]
```

tlento: Even cleaner would be to split the TimeGranularitySolver method into two - one that does a dict for multiple specs and one called something like `find_minimum_granularity_for_partial_time_dimension_spec` that returns a granularity value.

But I think it's totally reasonable to not do that in this PR and just use the dict access instead.

metricflow/test/integration/test_cases/itest_metrics.yaml
```diff
@@ -1043,3 +1043,64 @@ integration_test:
...
+---
+integration_test:
+  name: simple_query_with_date_part
+  description: Test query using date_part
+  model: SIMPLE_MODEL
+  metrics: ["bookings"]
+  group_by_objs: [{"name": "metric_time", "date_part": "year"}]
+  check_query: |
+    SELECT
+      SUM(1) AS bookings
```
tlento, line 1055: Question - what happens with distinct metrics? What should happen?

I'm assuming it'll replace the SUM(1) here with a COUNT(DISTINCT guest_id) or whatever, which seems right, but it might be nice to double check this.

tlento (approved):
I'm going to approve this because after looking at the update I think it's appropriately restrictive.

With this set of changes, we currently allow date_part queries against default granularity, but the default granularity effectively aggregates to date_part. What I mean is, if a metric is DAILY  and we request YEAR for date_part, we'll just add the whole thing up across the year and return a single row. The granularity effectively gets overridden, and the user gets one row per date_part value.

From an end user perspective, they can't alter this behavior. So they can't say "I want distinct users, which is day grain, but computed at month grain, grouped by date_part YEAR" because there's no reasonable way for us to do that - we'd end up returning 12 rows per YEAR but with the same YEAR grain. The way for them to get what they want is to group by both date_part YEAR and date_part MONTH.

Therefore, we only allow them to set a granularity equal to the base granularity of the metric, and we only allow that because certain interfaces require it. That's a rough edge we can clean up later, but it does feel like any granularity adjustments should be made via the date_part values, and we should simply block everything else.

metricflow/query/query_parser.py
```diff
@@ -754,6 +758,25 @@ def _parse_linkable_elements(
             entity_specs=tuple(entity_specs),
         )
 
+    def _verify_resolved_granularity_for_date_part(
+        self,
+        requested_dimension_structured_name: StructuredLinkableSpecName,
+        partial_time_dimension_spec: PartialTimeDimensionSpec,
+        metric_references: Sequence[MetricReference],
+    ) -> None:
```
tlento, line 766: Almost forgot, can we add a docblock explaining why this is here? Something like:

"""
Enforce that any granularity value associated with a date part query is the minimum.

By default, we will always ensure that a date_part query request uses the minimum granularity.
However, there are some interfaces where the user must pass in a granularity, so we need a check to
ensure that the correct value was passed in.
"""

metricflow/query/query_parser.py
courtneyholcomb: Happy to do the latter!

metricflow/test/integration/test_cases/itest_metrics.yaml
courtneyholcomb: I'll add a test case!

metricflow/query/query_parser.py
courtneyholcomb: > there are some interfaces where the user must pass in a granularity

Which interface(s) require this ?

PR #773, opened 2023-09-14
Support Saved Queries in MetricFlow

### LINKED ISSUE #765: [SL-915] [Feature] Support Saved Queries
MetricFlow should have a saved query construct. Saved queries extend governance beyond the semantic objects (metrics, dimensions, and entities) to their consumption by enabling users to pre-define a query in a governed format (yaml/git) and then reference it in downstream interfaces. This construct will use the same structure as the existing querying interface in the Metricflow CLI:  metrics, group_bys, a where clause, limit, and order. Saved queries should be referenceable through semantic layer APIs, including by orchestration in dbt cloud for exports and caching. 

In Cloud, Saved queries are the foundation for supporting exports and a declarative cache. Exports will allow users to write the results of saved queries back to their data warehouse. Exports would be a set of configurations you apply to a saved query that defines how the query should be executed, including destination, materialization strategy, and table name.

### **Saved Query Design**

The saved query should provide a matching interface to the query interfaces used in all other APIs but in YAML form. I’m in favor of nesting these fields into a sub-param called `query` as shown below:

```
saved_queries:
  - name: my_query
    description:
    query:
      metrics:
        -
      group_bys:
        -
      where:
      limit:
      order:
```

<sub>From [SyncLinear.com](https://synclinear.com) | [SL-915](https://linear.app/dbt-labs/issue/SL-915/[feature]-support-saved-queries-in-metricflow)</sub>

plypaul:
Resolves #765 

### Description

Please see the associated issue for more details.

Note: This PR is dependent on https://github.com/dbt-labs/dbt-semantic-interfaces/pull/148, and will be broken until that's in.


plypaul:
This implemented saved queries with the initial (now outdated) spec.

PR #775, opened 2023-09-18
Support for sort order in query interface

DevonFulcher:
Linear: SL-847 (internal to dbt Labs)

### Description

This PR adds support for syntax like `Dimension(...).descending(True)`. This will be used in the order_by parameter. If a user tries to use this syntax in the where clause, they will be presented with an error.


metricflow/specs/query_parameter.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import annotations
```
tlento, line 1: Will these (and the ones in query_interface.py) be accessed externally at all? If so, we might want to put these into the metricflow/protocols directory instead. That way we don't forget about them when we go through and establish API boundaries and clarify our contracts for what we expect and what might break at any time.

tlento (approved):
Seems reasonable!

Note - the commit history on this PR isn't all that helpful for future `git log` readers so I suggest a history rewrite before merging. Easiest thing to do is to squash and merge, but do whatever you're comfortable with.

PR #777, opened 2023-09-18
Validate Semantic Manifests Used in Tests

plypaul:
### Description

When DSI was cut out from MF, it seems that the validation for the semantic manifests used tests was removed. This PR updates the test fixtures so that validations are run on those manifests and makes fixes as needed so that those validations pass.


QMalcolm (approved):
Looks good to me! 🚀 

Also all I could think of was "find and replace goes brrrrrrrrr"

PR #778, opened 2023-09-19
Bump dev version

courtneyholcomb:
Resolves #SL-978

### Description
Bump dev version so we can do a dev release to get the new DATE PART feature into our APIs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "metricflow"
-version = "0.202.0"
+version = "0.202.0.dev4"
 description = "Translates a simple metric definition into reusable SQL and executes it against the SQL engine of your choice."
 readme = "README.md"
 requires-python = ">=3.8,<3.12"


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "metricflow"
-version = "0.202.0"
+version = "0.202.0.dev4"
```
tlento, line 7:
```suggestion
version = "0.203.0.dev1"
```

tlento (commented):
I think the version is off?

pyproject.toml
courtneyholcomb: Ah that was not my understanding of how we move to dev versions. I was confused to see we had moved from 0.202.0.dev3 to 0.202.0, because it seems like a reversion? Is that how we've always done it and I'm just being crazy?

courtneyholcomb: Also, don't we start with dev0?

courtneyholcomb: I just went to check commit history and at least in recent bumps (excluding the last one you bumped to 0.202.0) we have been following the pattern `of 0.200.0 -> 0.200.0.dev0 -> 0.201.0`. There seems have been some fluctuation over time about what pattern we use.

tlento: Oops, yes, it should be `0.203.0.dev0` although I'm not too concerned with the dev version value itself, as long as nobody ever goes backwards.

We've actually always followed the `<existing release version> -> <new version with pre-release> -> <new version>` standard.  We [moved from 0.140.0 to the 0.200.0.dev0 pre-release](https://github.com/dbt-labs/metricflow/commit/e58cc36959076bab843de732b7fea41a60a9d592) and then [moved from the 0.200.0 pre-release to 0.200.0](https://github.com/dbt-labs/metricflow/commit/bba366973b43074306c8578b5f0681ac24c9cfd7) 

Among the confusing things in our history we have:

1. Devon's PR for the 0.202.0.dev0 pre-release has the wrong title, which I'll fix, although the git commit history might just be busted and I don't want to rewrite it just for this
2. There was no pre-release for 0.201.0 - that was a straight release of the new version.

All of this confusion (and you're not the only one, it happened to Devon and to tell you the truth I need to look it up every time myself) tells me I need to move our version updating to be something scriptable, add that to the goofy shell script I wrote for making release PRs, and maybe put that up for review so all of us can use it for at least these mechanical dev version PRs.

PR #780, opened 2023-09-21
Proposed protocols & implementations for query parameters

courtneyholcomb:
Completes SL-993

### Description
Proposed updates to the protocols and implementations used for queries.

IMO, we are using the same protocols for too many things. For instance, we had a `QueryParameterDimension` as the type for the `group_by` param. `QueryParameterDimension` was required to have `grain`, and `date_part` attributes. That means that if the client wants to pass an entity or a categorical dimension into the `group_by` param, they still need to use an object with `grain` and `date_part` attributes even though those are not valid attributes for entities or categorical dimensions. This leaves us with 2 options:
1. Force the client to always include those attributes, even when they aren't valid. This simplifies type checking a lot in MF, but it seems pretty unreasonable on the client side.
2. Separate the group by options into two different protocols, one of which includes those attributes (`TimeDimensionQueryParameter`). This creates some awkwardness because we have to accept a `Union` `group_by` parameter and figure out which type the user passed.

In this PR I implement option 2, because to me it seems much cleaner to handle that logic in one place (here in MF) than to put the onus on all clients to handle the awkwardness of passing objects around with invalid attributes.

Second, we were using `DimensionQueryParameter` as the type for both `group_by` and `order_by` inputs. This forces us to add a `descending` parameter to `DimensionQueryParameter`, even though the `descending` param is not valid as a `group_by` input. Instead, I added a new OrderByQueryParameter protocol that includes a `descending` attribute and a nested `order_by` item, which can be a `MetricQueryParameter`, `GroupByQueryParameter`, or `TimeDimensionQueryParameter`.

One other note about the `group_by` param - I changed this param to expect a `Tuple`, since that's the only type hint that allows you to have multiple datatypes in one iterable. (This is just an annoying mypy limitation.)

Let me know if y'all agree with these changes or if you have other ideas!

### DIFF
--- a/metricflow/engine/metricflow_engine.py
+++ b/metricflow/engine/metricflow_engine.py
@@ -5,7 +5,7 @@ import logging
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from enum import Enum
-from typing import List, Optional, Sequence, Tuple, Union
+from typing import List, Optional, Sequence, Tuple
 
 import pandas as pd
 from dbt_semantic_interfaces.implementations.elements.dimension import PydanticDimensionTypeParams
@@ -47,12 +47,7 @@ from metricflow.plan_conversion.dataflow_to_execution import (
     DataflowToExecutionPlanConverter,
 )
 from metricflow.plan_conversion.dataflow_to_sql import DataflowToSqlQueryPlanConverter
-from metricflow.protocols.query_parameter import (
-    GroupByQueryParameter,
-    MetricQueryParameter,
-    OrderByQueryParameter,
-    TimeDimensionQueryParameter,
-)
+from metricflow.protocols.query_parameter import GroupByParameter, MetricQueryParameter, OrderByQueryParameter
 from metricflow.protocols.sql_client import SqlClient
 from metricflow.query.query_exceptions import InvalidQueryException
 from metricflow.query.query_parser import MetricFlowQueryParser
@@ -90,13 +85,15 @@ class MetricFlowQueryRequest:
     """Encapsulates the parameters for a metric query.
 
     metric_names: Names of the metrics to query.
+    metrics: Metric objects to query.
     group_by_names: Names of the dimensions and entities to query.
+    group_by: Dimension or entity objects to query.
     limit: Limit the result to this many rows.
     time_constraint_start: Get data for the start of this time range.
     time_constraint_end: Get data for the end of this time range.
     where_constraint: A SQL string using group by names that can be used like a where clause on the output data.
-    order_by_names: metric and group by names to order by. A "-" can be used to specify reverse order e.g. "-ds"
-    order_by: metric and group by objects to order by
+    order_by_names: metric and group by names to order by. A "-" can be used to specify reverse order e.g. "-ds".
+    order_by: metric, dimension, or entity objects to order by.
     output_table: If specified, output the result data to this table instead of a result dataframe.
     sql_optimization_level: The level of optimization for the generated SQL.
     query_type: Type of MetricFlow query.
@@ -106,7 +103,7 @@ class MetricFlowQueryRequest:
     metric_names: Optional[Sequence[str]] = None
     metrics: Optional[Sequence[MetricQueryParameter]] = None
     group_by_names: Optional[Sequence[str]] = None
-    group_by: Optional[Tuple[Union[GroupByQueryParameter, TimeDimensionQueryParameter], ...]] = None
+    group_by: Optional[Tuple[GroupByParameter, ...]] = None
     limit: Optional[int] = None
     time_constraint_start: Optional[datetime.datetime] = None
     time_constraint_end: Optional[datetime.datetime] = None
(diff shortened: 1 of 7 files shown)


metricflow/engine/metricflow_engine.py
```diff
@@ -91,38 +96,39 @@ class MetricFlowQueryRequest:
     time_constraint_end: Get data for the end of this time range.
     where_constraint: A SQL string using group by names that can be used like a where clause on the output data.
     order_by_names: metric and group by names to order by. A "-" can be used to specify reverse order e.g. "-ds"
+    order_by: metric and group by objects to order by
```
DevonFulcher, line 99: Kind of unrelated, but it seems weird to pass in parameters that are called `group by` into an `order_by`. I noticed that this is the case in GQL, not just this comment.

Also, it looks like I forgot to add a comment here for `group_by` & `metrics` too. Do you mind adding it?

metricflow/protocols/query_parameter.py
```diff
@@ -1,28 +1,42 @@
...
     @property
-    def grain(self) -> Optional[TimeGranularity]:
-        """The time granularity."""
+    def name(self) -> str:
+        """The name of the metric."""
         raise NotImplementedError
 
+
+@runtime_checkable
+class TimeDimensionQueryParameter(Protocol):  # noqa: D
```
DevonFulcher, line 31: I don't think we need `TimeDimensionQueryParameter`. I think it would be simpler if `GroupByQueryParameter` just has the `grain` property with the `Optional[TimeGranularity]` type.

courtneyholcomb: Even though that puts the onus on all clients to use objects with invalid attributes ? I think that simplifies typing a little bit here but complicates things everywhere else. (Not sure if you read my PR description!)

DevonFulcher: I read the PR description. I guess I'm just not understanding the burden for the client. If the object doesn't have a grain then the value would just be `None`.

courtneyholcomb: So I see the problem as this: if I'm using typed inputs as a client (like both GQL and JDBC are), then I have a `Metric` object, a `Dimension` object, an `Entity` object, and a `TimeDimension` object. If I want to pass most of those objects into an MF query, they will fail. Instead I'll have to create some new object to convert `Metric`, `Dimension`, or `Entity` objects into a new, compatible `GroupBy` object with `grain` and `date_part`, even though `grain` and `date_part` aren't relevant for those objects.

And also, on the MF side we then have to validate that a `Metric/Dimension/Entity` doesn't have a `time_granularity` or `date_part`, when the type checking should be able to do that for us.

courtneyholcomb: This is really a matter of opinion, so if you and @tlento both agree, then I'm happy to revert to that for the `group_by` param.

tlento: > if I'm using typed inputs as a client (like both GQL and JDBC are), then I have a Metric object, a Dimension object, an Entity object, and a TimeDimension object. If I want to pass most of those objects into an MF query, they will fail. Instead I'll have to create some new object to convert Metric, Dimension, or Entity objects into a new, compatible GroupBy object with grain and date_part, even though grain and date_part aren't relevant for those objects.

This seems like a thing we should fix. How, I don't know. Maybe with this PR, maybe some other way? Or maybe it's not that big of a deal?

tlento (commented):
I made a comment on @DevonFulcher's PR over in dbt-semantic-interfaces that this kind of protocol separation might be good, but in the context of that change it seemed fine to keep on with the current layout.

My default is to prefer strongly typed things to weakly typed things, and having optional parameters to satisfy the interface is, from a design perspective, unsatisfying. That said, I'm not at all sure what the right balance is here because I'm having a super hard time keeping track of all of the interfaces and where these things will be used.

Can we set up a 30 minute chat where we just list out the different query param interfaces, where they're defined, where they get called/how they get invoked, who is accessing them, and how they interact with types (like where filter users get no benefit from typing because they give us a string and we resolve it, but GraphQL callers should get support from the API)? I think if we have that list it'll be more obvious how to lay things out.

courtneyholcomb:
> Can we set up a 30 minute chat where we just list out the different query param interfaces, where they're defined, where they get called/how they get invoked, who is accessing them, and how they interact with types (like where filter users get no benefit from typing because they give us a string and we resolve it, but GraphQL callers should get support from the API)? I think if we have that list it'll be more obvious how to lay things out.

Sounds good - I set up 30 min for the three of us on Monday!

courtneyholcomb:
Updated based on our convo today!

metricflow/protocols/query_parameter.py
```diff
@@ -45,11 +45,15 @@ def date_part(self) -> Optional[DatePart]:
         raise NotImplementedError
 
 
+GroupByParameter = Union[DimensionOrEntityQueryParameter, TimeDimensionQueryParameter]
+InputOrderByParameter = Union[MetricQueryParameter, GroupByParameter]
```
DevonFulcher, line 49: Non-blocking: should these both have the `Input` prefix? Or neither?

DevonFulcher (approved):
Looks great!

metricflow/protocols/query_parameter.py
courtneyholcomb: Basically I just wanted to differentiate that `InputOrderByParameter` is a nested input to the `OrderByQueryParameter`. So I followed the naming convention we use for `InputMetrics` on derived metrics, if that makes sense! But I don't think that applies to the `GroupByParameter` since there's no nesting.

DevonFulcher: Got it, thanks!

PR #781, opened 2023-09-21
Expose `granularity_free_qualified_name` on `StructuredLinkableSpecName`

courtneyholcomb:
### Description
Sometimes in the API we need to parse a qualified name into its `time_granularity` and its `name`. For the `name`, we want to include entity links, not just the `element_name`. Currently, `StructuredLinkableSpecName` only supports parsing `element_name`, `entity_links`, and `time_granularity`. Adding this here to ensure we always use the same logic to get the qualified name without granularity.


tlento (approved):
Seems reasonable!

PR #789, opened 2023-09-28
Change `OrderBySpec` to Use a Single Spec

plypaul:
### Description

Thanks to some earlier changes, the `OrderBySpec` class can be made easier / cleaner to use by having a single `InstanceSpec` field instead of more concrete types.


tlento (approved):
Seems reasonable!

To be honest, I forget why we needed the spec type separation at all, so I'm glad I won't need to wonder about it anymore.

PR #790, opened 2023-09-28
Add `Mergeable` to Simplify Merging Collections

plypaul:
### Description

This PR adds the utility class `MergeBuilder` that allows you to combine collections that implements `Mergeable`. This is useful as there are a few spec-related classes that are collections, and there are upcoming cases where they need to be merged. This replaces the collection of merge-related methods in the spec set classes.

This PR also includes a minor update to the signature of `InstanceSpec` to make it more flexible.

### DIFF
--- a/metricflow/collections/merger.py
+++ b/metricflow/collections/merger.py
@@ -1,0 +1,66 @@
+from __future__ import annotations
+
+import functools
+from abc import ABC, abstractmethod
+from typing import Iterable, Type, TypeVar
+
+from typing_extensions import Self
+
+MergeableT = TypeVar("MergeableT", bound="Mergeable")
+
+
+class Mergeable(ABC):
+    """Objects that can be merged together to form a superset object of the same type.
+
+    Merging objects are frequently needed in MetricFlow as there are several recursive operations where the output is
+    the superset of the result of the recursive calls.
+
+    e.g.
+    * The validation issue set of a derived metric includes the issues of the parent metrics.
+    * The output of a node in the dataflow plan can include the outputs of the parent nodes.
+    * The query-time where filter is useful to combine with the metric-defined where filter.
+
+    Having a common interface also gives a consistent name to this operation so that we don't end up with multiple names
+    to describe this operation (e.g. combine, add, concat).
+
+    This is used to streamline the following case that occurs in the codebase:
+
+        items_to_merge: List[ItemType] = []
+        ...
+        if ...
+            items_to_merge.append(...)
+        ...
+        if ...
+            items_to_merge.append(...)
+        ...
+        if ...
+            ...
+            items_to_merge.append(...)
+            return merge_items(items_to_merge)
+        ...
+        return merge_items(items_to_merge)
+
+    This centralizes the definition of the merge_items() call.
+    """
+
+    @abstractmethod
+    def merge(self: Self, other: Self) -> Self:
+        """Return a new object that is the result of merging self with other."""
+        raise NotImplementedError
+
+    @classmethod
+    @abstractmethod
+    def empty_instance(cls: Type[MergeableT]) -> MergeableT:
+        """Create an empty instance to handle merging of empty sequences of items.
+
+        As merge_iterable() returns an empty instance for an empty iterable, there needs to be a way of creating one.
+        """
+        raise NotImplementedError
+
+    @classmethod
+    def merge_iterable(cls: Type[MergeableT], items: Iterable[MergeableT]) -> MergeableT:
+        """Merge all items into a single instance.
+
+        If an empty iterable has been passed in, this returns an empty instance.
+        """
+        return functools.reduce(cls.merge, items, cls.empty_instance())
--- a/metricflow/specs/specs.py
+++ b/metricflow/specs/specs.py
@@ -28,8 +28,10 @@ from dbt_semantic_interfaces.references import (
 from dbt_semantic_interfaces.type_enums.aggregation_type import AggregationType
 from dbt_semantic_interfaces.type_enums.date_part import DatePart
 from dbt_semantic_interfaces.type_enums.time_granularity import TimeGranularity
+from typing_extensions import override
 
 from metricflow.aggregation_properties import AggregationState
+from metricflow.collections.merger import Mergeable
 from metricflow.filters.time_constraint import TimeRangeConstraint
 from metricflow.naming.linkable_spec_name import StructuredLinkableSpecName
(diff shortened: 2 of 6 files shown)


metricflow/specs/merge_builder.py
```diff
@@ -0,0 +1,50 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import Generic, Iterable, TypeVar
+
+SelfTypeT = TypeVar("SelfTypeT", bound="Mergeable")
```
tlento, line 6: `from typing_extensions import Self` is a thing. Let's use that instead.

metricflow/specs/merge_builder.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class MergeBuilder(Generic[MergeableT]):
+    """Implementation of the builder pattern for merging many mergeable objects."""
+
+    def __init__(self, initial_item: MergeableT) -> None:  # noqa: D
+        self._accumulator = initial_item
+
+    def add(self, other: MergeableT) -> None:  # noqa: D
+        if self._accumulator is None:
```
tlento, line 28: Do we need this? Can the generic actually be `Optional[MergeableThing]`? I don't think it can, with a bound of "Mergeable" here.

metricflow/specs/merge_builder.py
```diff
@@ -0,0 +1,50 @@
...
+            self._accumulator = other
+            return
+
+        self._accumulator = self._accumulator.merge(other)
+
+    def add_all(self, *others: MergeableT) -> None:  # noqa: D
```
tlento, line 34: Ok. I realize this is my not-Python developer side showing, but I prefer `others: Sequence[...]` to `*others`.

Given add_iterable is a thing, do we really need this method? If they can't be combined, some docstrings with explanations providing that context would be nice.

metricflow/specs/merge_builder.py
```diff
@@ -0,0 +1,50 @@
...
+            self.add(other)
+
+    @staticmethod
+    def merge_iterable(initial_item: MergeableT, other_iterable: Iterable[MergeableT]) -> MergeableT:  # noqa: D
+        builder = MergeBuilder[MergeableT](initial_item)
+        builder.add_iterable(other_iterable)
+        return builder.build_result
+
+    @property
+    def build_result(self) -> MergeableT:  # noqa: D
```
tlento, line 49: This sounds like an action rather than a property. Maybe `merged_result` or `merged_output` would be better?

metricflow/test/specs/test_merge_builder.py
```diff
@@ -0,0 +1,32 @@
...
+        NumberList([4]),
+        NumberList([5]),
+    )
+    number_list_builder.add_iterable([NumberList([6]), NumberList([7])])
+
+    assert number_list_builder.build_result == NumberList([1, 2, 3, 4, 5, 6, 7])
+
+
+def test_merge_iterable() -> None:  # noqa: D
+    assert MergeBuilder.merge_iterable(NumberList(), [NumberList([1, 2, 3])]) == NumberList([1, 2, 3])
```
tlento, line 32: You might want a test that merges a non-empty thing as well, because `return other_iterable.next()` passes this test.

metricflow/specs/specs.py
```diff
@@ -522,28 +522,33 @@ class LinkableSpecSet(SerializableDataclass):
...
-                    entity_specs.append(entity_spec)
+    @override
+    def merge(self, other: LinkableSpecSet) -> LinkableSpecSet:
+        return LinkableSpecSet(
+            dimension_specs=self.dimension_specs + other.dimension_specs,
+            time_dimension_specs=self.time_dimension_specs + other.time_dimension_specs,
+            entity_specs=self.entity_specs + other.entity_specs,
+        )
+
+    def dedupe(self) -> LinkableSpecSet:  # noqa: D
```
tlento, line 533: Why are we separating these? Is there ever a case where we want duplicates in this output?

If not, this feels like we're adding an old bug factory back to the codebase.

metricflow/specs/specs.py
```diff
@@ -594,6 +594,37 @@ def linkable_specs(self) -> LinkableSpecSet:  # noqa: D
             entity_specs=self.entity_specs,
         )
 
+    @override
+    def merge(self, other: MetricFlowQuerySpec) -> MetricFlowQuerySpec:  # noqa: D
+        # Can't merge if both properties are set.
```
tlento, line 599: Wait, why can we merge when these don't match at all?

Can you please add a docstring explaining this?

tlento (commented):
In a vacuum where all implementation choices are reasonable and correct this all seems fine. However, we don't operate in such a vacuum, and there are a lot of "but WHY is it like this?" implementation choices in here. Every single one of them is preceded by a `#noqa: D` override on docstrings, and every commit message is completely devoid of the crucial "why" for each change.

Please write docstrings explaining the context behind these decisions, and then I'll be able to evaluate this better. I'm sure it's fine, but I also know that I'm going to screw up using this and when I go through the code (and commit history) I'll be quite irritated to find that nothing has been explained anywhere.

plypaul (commented):
Updated the commit messages and docstrings, and found a fix for type issues to allow removal of the `MergeBuilder` and just have the `Mergeable` interface.

tlento (approved):
LGTM apart from that last commit and the hanging If in the docstring. 

You built the thing so you get to choose whether to opt for more flexibility here.

metricflow/collections/merger.py
```diff
@@ -0,0 +1,66 @@
...
+
+        As merge_iterable() returns an empty instance for an empty iterable, there needs to be a way of creating one.
+        """
+        raise NotImplementedError
+
+    @classmethod
+    def merge_iterable(cls: Type[MergeableT], items: Iterable[MergeableT]) -> MergeableT:
+        """Merge all items into a single instance.
+
+        If
```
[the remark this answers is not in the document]
plypaul: Updated.

mypy.ini
```diff
@@ -56,3 +56,8 @@ ignore_missing_imports = True
 
 [mypy-update_checker]
 ignore_missing_imports = True
+
+# Added to resolve an issue with mypy in CI:
+# ... python3.8/site-packages/referencing/jsonschema.py:426: error: Invalid "type: ignore" comment  [syntax]
+[mypy-referencing.jsonschema]
+follow_imports = skip
```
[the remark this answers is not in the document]
plypaul: Removed.

PR #793, opened 2023-10-03
Upgrade DSI to 0.2.1dev1

DevonFulcher:
### Description

Upgrades DSI to support backported functionality including support for `Dimension(...).grain(...)` syntax


DevonFulcher:
We are going to upgrade to 0.2.1 instead

PR #794, opened 2023-10-05
Support Saved Queries

### LINKED ISSUE #765: [SL-915] [Feature] Support Saved Queries

plypaul:
Resolves #765 

### Description

Please see the linked issue for more details. As this PR depends on `dbt-semantic-interfaces~=0.3.0a2` which includes the interface definition for saved queries, and it's currently not possible to specify that dependency due to a conflict with `dbt-core~=1.6.0` as required by this package. Once `dbt-core` is updated, this PR will be updated so that CI checks can pass. Locally, a custom `pyproject.yaml` was used to test this feature.


metricflow/query/query_parser.py
```diff
@@ -172,6 +176,65 @@ def _top_fuzzy_matches(
...
+        limit: Optional[int],
+        order_by_parameters: Optional[Sequence[OrderByQueryParameter]],
+    ) -> MetricFlowQuerySpec:
+        """Parse and validate a query using parameters from a pre-defined / saved query.
+
+        Additional parameters act in conjunction with the parameters in the saved query.
+        """
+        saved_query = self._get_saved_query(saved_query_parameter)
+
+        # This logic could be encapsulated in the WhereFilter through a merge interface.
```
tlento, line 192: Yes, and even more so now that the object is going to hold a list of WhereFilters.

I haven't added it yet, but it might be something we should add in to the base protocol. If we don't think that's worth doing we can re-type everything internally to take an MFWhereFilter that just extends the WhereFilter protocol with merge or combine or whatever.

metricflow/query/query_parser.py
```diff
@@ -172,6 +176,65 @@ def _top_fuzzy_matches(
...
+
+        if len(where_conditions) == 0:
+            combined_where_filter = None
+        if len(where_conditions) == 1:
+            combined_where_filter = PydanticWhereFilter(where_sql_template=where_conditions[0])
+        else:
+            where_conditions_with_parenthesis = tuple(f"({where_condition})" for where_condition in where_conditions)
+            combined_where_filter = PydanticWhereFilter(
+                where_sql_template=" AND ".join(where_conditions_with_parenthesis)
+            )
```
tlento, line 209: eeeewwwwww......

I was thinking about eventually moving this to the renderers, since we won't always render the combined filter. The container class I created should help with this.

metricflow/specs/python_object.py
```diff
@@ -0,0 +1,90 @@
...
+def parse_object_builder_naming_scheme(group_by_item_name: str) -> GroupByParameter:
+    """Parses a string following the object-builder naming scheme into the corresponding GroupByParameter.
+
+    The implementation of the query parameter classes seems incomplete and there needs to be follow up with the author
+    of the query interface classes for the best approach. Right now, it seems like using the where filter is the only
+    way to handle this conversion. However, it seems like this functionality should be abstracted into a module that
+    handles operations related to the object-builder naming scheme.
+
+    Additional issues:
+    * The call parameter sets in DSI does not support date part.
```
tlento, line 23: Let's also not support date part in saved queries, it's a partially supported feature built solely for use in interactive sessions through the Tableau connector. As we get more experience with how the date_part mechanics are used in practice we'll push more robust support down to MetricFlow.

metricflow/specs/python_object.py
```diff
@@ -0,0 +1,90 @@
...
+from dbt_semantic_interfaces.call_parameter_sets import ParseWhereFilterException
+from dbt_semantic_interfaces.implementations.filters.where_filter import PydanticWhereFilter
+
+from metricflow.naming.linkable_spec_name import StructuredLinkableSpecName
+from metricflow.protocols.query_parameter import GroupByParameter
+from metricflow.query.query_exceptions import InvalidQueryException
+from metricflow.specs.query_param_implementations import DimensionOrEntityParameter, TimeDimensionParameter
+
+
+def parse_object_builder_naming_scheme(group_by_item_name: str) -> GroupByParameter:
```
tlento, line 14: I don't know how much of this still applies and what's been improved. Agreed we should check in with @DevonFulcher, probably worth doing that before merge since you'll need to update some stuff due to me mucking about with WhereFilter interfaces in dbt-semantic-interfaces.

tlento (commented):
This seems like a reasonable first cut.

How were you planning to address the parameterizable where filter? What we really want out of a saved query is something where the filter is more like:

```
saved_queries:
  - name: date_filter_example
    filter: metric_time BETWEEN <start_date> AND <end_date>
```

Then on invocation we want people to be able to do something like:

`mf query --saved-query date_filter_example --params "start_date='2021-01-01',end_date='2021-01-05'"`

metricflow/specs/python_object.py
DevonFulcher: Hey, sorry I am just catching up here. In general, I agree that query interface objects should share more code. Now that they all share an interface, this should be easier. I'm happy to chat about that more about that.

metricflow/specs/python_object.py
```diff
@@ -0,0 +1,90 @@
...
+    Additional issues:
+    * The call parameter sets in DSI does not support date part.
+    * Conversion from the element name / entity path to the name field in the query parameter objects requires going
+    through StructuredLinkableSpecName.
+
+    TODO: Replace this method once the aforementioned issues are resolved.
+    """
+    try:
+        call_parameter_sets = PydanticWhereFilter(
+            where_sql_template="{{ " + group_by_item_name + " }}"
```
DevonFulcher, line 31: Is it possible for `group_by_item_name` to contain malicious code? We should consider validating the string before calling this constructor. That is probably a concern of MFS though

metricflow/cli/utils.py
```diff
@@ -46,7 +46,8 @@ def query_options(function: Callable) -> Callable:
     )(function)
     function = click.option(
         "--metrics",
-        type=click_custom.SequenceParamType(min_length=1),
+        # Validity checks for this parameter was moved to the MetricFlowEngine.
```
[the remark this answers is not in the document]
plypaul: Updated.

metricflow/engine/metricflow_engine.py
```diff
@@ -130,15 +135,9 @@ def create_with_random_request_id(  # noqa: D
         sql_optimization_level: SqlQueryOptimizationLevel = SqlQueryOptimizationLevel.O4,
         query_type: MetricFlowQueryType = MetricFlowQueryType.METRIC,
     ) -> MetricFlowQueryRequest:
-        assert_exactly_one_arg_set(metric_names=metric_names, metrics=metrics)
-        assert not (
-            group_by_names and group_by
-        ), "Both group_by_names and group_by were set, but if a group by is specified you should only use one of these!"
-        assert not (
-            order_by_names and order_by
-        ), "Both order_by_names and order_by were set, but if an order by is specified you should only use one of these!"
```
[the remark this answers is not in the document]
plypaul: Yeah, it's a redundant validation that's handled in `MetricFlowQueryParser`.

metricflow/query/query_parser.py
plypaul: Yeah, we could add `Mergeable` to the base class, but we can visit that once `Mergeable` is in.

metricflow/query/query_parser.py
plypaul: The container class helped, but I didn't see a place to render it?

plypaul: It's currently here: https://github.com/dbt-labs/metricflow/pull/809/files#diff-14f344e3a9bd66bc0d46537a8017023c16cd242d65ace611485b8277e0719b74

tlento: Yeah I was thinking eventually we'd just use the container and pass it all the way through but maybe that's not the right way to do it. What you did in #809 seems reasonable for now.

metricflow/specs/python_object.py
plypaul: Chatted with @DevonFulcher - using this for now until an improved setup is ready.

metricflow/specs/python_object.py
plypaul: This is mainly for the CLI, and verified with @DevonFulcher that it's handled on the MFS side.

plypaul:
> How were you planning to address the parameterizable where filter?

In this initial version, there is support for additional where filters when running a saved query, so users would have to pass in an appropriately constructed where filter to do something like - run the saved query for a different set of dates. We haven't discussed nor designed the parameterization interface, so that will have to be handled later.

metricflow/test/fixtures/semantic_manifest_yamls/simple_manifest/saved_queries.yaml
```diff
@@ -0,0 +1,12 @@
+---
+saved_query:
+  name: p0_booking
+  description: Booking-related metrics that are of the highest priority.
+  metrics:
+    - bookings
+    - instant_bookings
+  group_bys:
```
Jstein77, line 8: @plypaul should this be group_by?

metricflow/test/fixtures/semantic_manifest_yamls/simple_manifest/saved_queries.yaml
```diff
@@ -0,0 +1,12 @@
...
+    - TimeDimension('metric_time', 'day')
+    - Dimension('listing__capacity_latest')
+  where:
```
Jstein77, line 11: @plypaul I think this should be `filter` to keep it consistent with metrics.

metricflow/test/fixtures/semantic_manifest_yamls/simple_manifest/saved_queries.yaml
plypaul: Was there a spec change? This is following https://github.com/dbt-labs/metricflow/issues/765

metricflow/specs/python_object.py
tlento: Also, this should all go away when we finish cleaning up the parameter parsing stuff, right?

metricflow/test/fixtures/semantic_manifest_yamls/simple_manifest/saved_queries.yaml
tlento: That's a typo in the spec. We use `group_by` everywhere else, so it should be `group_by` for saved queries.

We used to call them `group_bys` but it was horribly confusing for everybody so we switched to `group_by`.

metricflow/test/fixtures/semantic_manifest_yamls/simple_manifest/saved_queries.yaml
tlento: @Jstein77 we decided that `where` was what we'd use for saved queries because they're queries, and `where` matches the query interfaces for the CLI, JDBC and GraphQL.

`filter` is what we use for metrics and other non-query objects in the spec. 

I'd prefer to pick one and use it everywhere, but alas that ship has sailed. Or we can go with `filter` here and commit to changing the other query interfaces.

One argument in favor of `where` - we may need to introduce a post-aggregation filter (a `having` input) for query time specification and that can be a natural distinction for people accustomed to SQL. That said, I don't like calling anything `having` because it's a goofy name for a filter expression.

tlento (approved):
So I think this is good to merge at this point, but it seems like we need a dbt-semantic-interfaces 0.3.1 bugfix to roll out for the `group_by` thing.

PR #796, opened 2023-10-05
Clean up SQL time interval math expressions

tlento:
Our time interval math expressions were originally represented in a class called
SqlTimeDeltaExpression. There were two issues with this:

1. The class does not represent a time delta. It represents an interval subtraction
from an input time value.
2. The class includes an unused parameter called grain_to_date, which would change
it's behavior from rendering an interval subtraction to rendering date_trunc

This PR addresses both of these, and improves some of the documentation and class
labeling properties as part of these updates.

PR #797, opened 2023-10-05
Coerce time granularity to configured value in all cases

tlento:
MetricFlow's initial time dimension select statement rendering
was built on the assumption that the granularity configred for a given
time dimension would always match the granularity of the values stored.
For example, if a time dimension was defined with granularity DAY, we
assumed that the values in the warehouse would always correspond to
daily granularity data, and would never be offset from a day boundary.

This assumption is flawed for a variety of reasons - yearly or quarterly
granularities might use different offset boundaries, users may
misconfigure dimensions or allow offset values into the dataset which
could be safely corrected, etc.

This change updates our rendering to apply the date_trunc operation on
the base time dimension column, just as we do for the other, coarser
granularities. The one exception is for the base granularity column
for validity window time dimensions, which must retain their underlying
granularity in order to remain viable as validity window boundaries
for any DAILY granularity contexts, as we do not support smaller than
daily granularities at this time.

Note this may cause queries with start and end time constraints to
no longer trigger partition pruning, an issue which will be addressed
in subsequent updates.

### DIFF
--- a/.changes/unreleased/Fixes-20231005-124722.yaml
+++ b/.changes/unreleased/Fixes-20231005-124722.yaml
@@ -1,0 +1,7 @@
+kind: Fixes
+body: Coerce time granularity to configured value to prevent finer-grained timestamps
+  from causing unexpected query behavior
+time: 2023-10-05T12:47:22.662371-07:00
+custom:
+  Author: tlento
+  Issue: "714"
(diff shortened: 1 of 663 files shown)


tlento:
* main
  * **PR #796** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/796" target="_blank">Graphite</a> 
    * **PR #797** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/797" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dbt-labs/metricflow/797?utm_source=stack-comment).

tlento:
Warehouse test action run: https://github.com/dbt-labs/metricflow/actions/runs/6423205270/job/17441226665?pr=797

plypaul (approved):

PR #802, opened 2023-10-10
Case insensitivity for query params

courtneyholcomb:
### Description
Support case-insensitive params in `group_by` and `order_by` parameters.
A [community member recently ran into a bug](https://getdbt.slack.com/archives/C046L0VTVR6/p1696602554947709) where they uppercased a dimension in the `group_by`, which worked, but then the same syntax didn't work in the `order_by`. This should fix that issue.

### DIFF
--- a/.changes/unreleased/Fixes-20231010-144137.yaml
+++ b/.changes/unreleased/Fixes-20231010-144137.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Enables case insensitivity for various query params.
+time: 2023-10-10T14:41:37.181704-07:00
+custom:
+  Author: courtneyholcomb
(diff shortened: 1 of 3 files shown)


DevonFulcher (approved):

tlento (approved):
Oh interesting, yeah, we require all names to be lower-case so we shouldn't care much what casing gets put in at query time.

What happens with the object syntax though? Is that case-insensitive?

courtneyholcomb:
> Oh interesting, yeah, we require all names to be lower-case so we shouldn't care much what casing gets put in at query time.
> 
> What happens with the object syntax though? Is that case-insensitive?

@tlento this PR makes the object syntax case-insensitive!

PR #803, opened 2023-10-11
Support dundered granularity syntax with object params

courtneyholcomb:
### Description
Allow users to pass dundered granularity syntax even in object parameters. Originally, I intentionally disallowed this. But it creates a weird inconsistency across parameters that users were running into, so it seems better to allow it.

### DIFF
--- a/.changes/unreleased/Features-20231010-172053.yaml
+++ b/.changes/unreleased/Features-20231010-172053.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Enable dundered granularity in object parameters.
+time: 2023-10-10T17:20:53.123738-07:00
+custom:
+  Author: courtneyholcomb
+  Issue: "803"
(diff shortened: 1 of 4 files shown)


DevonFulcher (approved):

tlento:
Ooof, really? I thought we agreed to deprecate this access pattern. The reason it's confusing right now is because sometimes it works and sometimes it doesn't - if it never works, then it will no longer be confusing.

If we allow it everywhere now we live with it forever, and I would dearly love to be rid of it forever instead.

tlento:
Oh wait actually I remember now, this was an open question and we were leaning towards allowing granularity as a property of the time dimension. So essentially we have dimensions of time dimensions and you can ask for the day/month/year/etc.

The thing is, we don't have a design for that at all. Can we hold off on this for a bit until we have some space to deal with it?

courtneyholcomb:
> Oh wait actually I remember now, this was an open question and we were leaning towards allowing granularity as a property of the time dimension. So essentially we have dimensions of time dimensions and you can ask for the day/month/year/etc.
> 
> The thing is, we don't have a design for that at all. Can we hold off on this for a bit until we have some space to deal with it?

@tlento Actually, I thought this was needed for a fix in MFS, but turns out it's not. So I'll close it!

tlento:
> @tlento Actually, I thought this was needed for a fix in MFS, but turns out it's not. So I'll close it!

😌

PR #804, opened 2023-10-11
Querying dimensions without metrics

courtneyholcomb:
Completes #SL-949

### Description
Enable querying for dimensions without metrics.
If a user submits a query with dimensions and not metrics, we'll assume they want the distinct values for those dimensions. 
This PR adds a new `DataFlowPlan` for distinct values that will allow those queries. The plan is essentially a simple `ReadSqlSourceNode` -> optional `JoinToBaseOutputNode` -> `FilterElementsNode`.
For simplicity's sake, I added an attribute to `FilterElementsNode` called `distinct` to indicate if we should group by the filtered elements (therefore getting the distinct values). I updated the costing for that node to reflect an aggregation in that case.
A lot of the other changes here are just renaming or restructuring things to not assume that metrics / measures are included in the query, and removing errors that enforce that requirement.

Note: something is up with the Databricks credentials, so I still need to fix that & update those snapshots.


metricflow/query/query_parser.py
```diff
@@ -573,6 +583,8 @@ def _adjust_time_range_constraint(
                 self._time_granularity_solver.resolve_granularity_for_partial_time_dimension_specs(
                     metric_references=metric_references,
                     partial_time_dimension_specs=(partial_metric_time_spec,),
+                    read_nodes=self._read_nodes,
+                    node_output_resolver=self._node_output_resolver,
```
tlento, line 587: Checkpoint, this must be changed in a later commit one way or the other.....

metricflow/time/time_granularity_solver.py
```diff
@@ -116,35 +128,75 @@ def resolve_granularity_for_partial_time_dimension_specs(
...
+        for read_node in read_nodes:
+            output_data_set = node_output_resolver.get_output_data_set(read_node)
+            for time_dimension_instance in output_data_set.instance_set.time_dimension_instances:
+                if time_dimension_instance.spec.date_part:
+                    continue
+                time_dim_name_without_granularity = StructuredLinkableSpecName.from_name(
+                    time_dimension_instance.spec.qualified_name
+                ).granularity_free_qualified_name
+                if time_dim_name_without_granularity == granularity_free_qualified_name:
+                    return time_dimension_instance.spec.time_granularity
```
tlento, line 196: This is the minimum because it comes from the semantic manifest definition directly, is that right?

There's an assumption here that we'll never have duplicated read nodes or the ability to have a set of granularities rather than just one defined in a given semantic model, but I don't think that'll hold - at some point we will need derived semantic models or container time dimensions or something and this logic might not hold up.

For now - and this can be done in a follow-up - let's update this logic to collect all matches and assert that there's exactly one (or just return the minimum of that set).

metricflow/time/time_granularity_solver.py
```diff
@@ -116,35 +128,75 @@ def resolve_granularity_for_partial_time_dimension_specs(
...
+    ) -> Optional[TimeGranularity]:
+        """Find the minimum."""
+        granularity_free_qualified_name = StructuredLinkableSpecName(
+            entity_link_names=tuple(
+                [entity_link.element_name for entity_link in partial_time_dimension_spec.entity_links]
+            ),
+            element_name=partial_time_dimension_spec.element_name,
+        ).granularity_free_qualified_name
+        for read_node in read_nodes:
+            output_data_set = node_output_resolver.get_output_data_set(read_node)
```
tlento, line 188: This feels like potentially a lot of iterating through datasets to me since we'll call this function once for every partial time dimension spec, but I think it's fine for now.

metricflow/test/query/test_query_parser.py
```diff
@@ -249,6 +249,9 @@ def test_query_parser_case_insensitivity(bookings_query_parser: MetricFlowQueryP
         ),
     )
 
+    with pytest.raises(UnableToSatisfyQueryError):
+        bookings_query_parser.parse_and_validate_query(group_by_names=["random_stuff"])
```
tlento, line 253: Why is this here? Should this be a separate test case?

metricflow/time/time_granularity_solver.py
```diff
@@ -116,35 +128,75 @@ def resolve_granularity_for_partial_time_dimension_specs(
         return result
 
     def find_minimum_granularity_for_partial_time_dimension_spec(
-        self, partial_time_dimension_spec: PartialTimeDimensionSpec, metric_references: Sequence[MetricReference]
+        self,
+        partial_time_dimension_spec: PartialTimeDimensionSpec,
+        metric_references: Sequence[MetricReference],
+        read_nodes: Sequence[BaseOutput],
+        node_output_resolver: DataflowPlanNodeOutputDataSetResolver,
```
tlento, line 135: Ok. I think it might be better to keep this totally separate from the new function and branch at the callsite for now. 

There are only two callsites, one of which is in this class, and that way we aren't tempted to keep adding responsibility to the node output resolver.

What do you think? Otherwise it might be worth it to mark the new method with the leading _ for the time being, I'm not sure we'll ever really want to call it on its own if we keep this method signature.

tlento (approved):
Ok, let's go with this!

There are no blocking concerns but I do have a lot of small quality considerations that we should address one way or another in the very near future. 

Whatever you feel is worth fixing prior to merge, please go ahead and update. I'm happy to take another look if you'd like, but I don't need to. 

After merge I'll go through and file issues for whatever follow-ups we still need to make.

Thank you!

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -103,6 +144,7 @@ class DataflowPlanBuilder:
     def __init__(  # noqa: D
         self,
         source_nodes: Sequence[BaseOutput],
+        read_nodes: Sequence[BaseOutput],
```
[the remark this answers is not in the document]
courtneyholcomb: Yeah I can update that!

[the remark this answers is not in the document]
courtneyholcomb: ...hopefully

metricflow/query/query_parser.py
```diff
@@ -118,13 +118,15 @@ def __init__(  # noqa: D
         self,
         column_association_resolver: ColumnAssociationResolver,
         model: SemanticManifestLookup,
-        source_nodes: Sequence[BaseOutput],
+        read_nodes: Sequence[BaseOutput],
```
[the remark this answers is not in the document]
courtneyholcomb: Ya I think it was just missed in a refactor!

metricflow/time/time_granularity_solver.py
courtneyholcomb: > This is the minimum because it comes from the semantic manifest definition directly, is that right?

Correct!

Sounds good I'll make that update.

metricflow/time/time_granularity_solver.py
courtneyholcomb: Yeah, it would probably be better to create a dict of time dimension instances to their output datasets on init (or something like that) so we only need to do this loop once. Not sure how complicated that change will be, let me look into it and see if it's worth doing before merge

metricflow/time/time_granularity_solver.py
courtneyholcomb: Not totally sure I understand what you mean. Is it this? We don't want to use the `node_output_resolver` here, so we should pass in the output datasets already resolved.

I'm thinking maybe it would be better to do this all on init for this class. Iterate through all the datasets and store a dict of all `time_dimension_instances` to a set of `granularities` (currently should only be one per instance). Then it will be super easy to grab that at query time.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -509,36 +542,23 @@ def _find_measure_recipe(
...
-        for node in self._sort_by_suitability(potential_measure_nodes):
-            data_set = self._node_data_set_resolver.get_output_data_set(node)
-
-            missing_specs = [spec for spec in measure_specs if spec not in data_set.instance_set.spec_set.measure_specs]
-            if missing_specs:
-                logger.debug(
-                    f"Skipping evaluation for node since it does not have all of the measure specs {missing_specs}:"
-                    f"\n\n{dataflow_dag_as_text(node)}"
-                )
-                continue
```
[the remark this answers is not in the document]
courtneyholcomb: Hmmm maybe I didn't actually need to remove this, since it should just be an empty list if no metrics are requested! I'll test adding it back

metricflow/test/query/test_query_parser.py
courtneyholcomb: ...laziness. I'll make a new test case LOL

metricflow/query/query_parser.py
courtneyholcomb: What do you mean??

metricflow/dataflow/builder/dataflow_plan_builder.py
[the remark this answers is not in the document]
courtneyholcomb: Updated!

metricflow/dataflow/builder/dataflow_plan_builder.py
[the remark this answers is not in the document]
courtneyholcomb: Added back and works fine!

metricflow/time/time_granularity_solver.py
courtneyholcomb: Actually, this behaves slightly different than I thought. The dataset has a time dimension for every supported granularity, so here I'll grab them all and take the minimum one.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -259,67 +302,50 @@ def _build_metrics_output_node(
...
-        distinct_values_node = FilterElementsNode(
-            parent_node=metrics_output_node,
-            include_specs=InstanceSpecSet.create_from_linkable_specs((linkable_spec,)),
-        )
+        where_constraint_node: Optional[WhereConstraintNode] = None
+        if query_spec.where_constraint:
+            where_constraint_node = WhereConstraintNode(
+                parent_node=distinct_values_node,
+                where_constraint=query_spec.where_constraint,
+            )
```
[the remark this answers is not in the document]
courtneyholcomb: Updated!

metricflow/time/time_granularity_solver.py
courtneyholcomb: FWIW, it turns out we are caching the output dataset so `get_output_data_set()` is just a dictionary lookup after the first call for a given node!

metricflow/time/time_granularity_solver.py
courtneyholcomb: ^ implemented it this way!

courtneyholcomb:
Note that Databricks is failing due to connection errors unrelated to this PR.

kingduxia:
Has this feature already been released? In which version is it available?

courtneyholcomb:
@kingduxia use the `--group_by` flag.

Please use the [community slack](https://www.getdbt.com/community/join-the-community/) for future questions.

PR #808, opened 2023-10-12
Bump dev version

courtneyholcomb:
### Description
Bump dev version so we can get dimensions without metrics into MFS.


WilliamDee (approved):
🚀

PR #809, opened 2023-10-12
Update Dependencies to `dbt-semantic-interfaces~=0.3.0`

plypaul:
### Description

As per title.


metricflow/specs/dimension_spec_resolver.py
```diff
@@ -50,5 +50,10 @@ def resolve_time_dimension_spec(
         return TimeDimensionSpec(
             element_name=call_parameter_set.time_dimension_reference.element_name,
             entity_links=call_parameter_set.entity_path,
-            time_granularity=call_parameter_set.time_granularity,
+            time_granularity=(
+                call_parameter_set.time_granularity
+                # TODO: This should be updated once resolution of unspecified grain is supported.
```
tlento, line 55: Ok. DAY could be wrong, and then it'll throw a less helpful error, but at least it'll work for daily granularities and error out otherwise.

pyproject.toml
```diff
@@ -114,9 +114,13 @@ exclude = [
 description = "Environment for development. Includes a DuckDB-backed client."
 features = [
   "dev-packages",
-  "dbt-duckdb",
   "sql-client-packages",
 ]
+# Temporary workaround for installing the dbt-duckdb adapter until it's updated to support dbt-core~=1.7.0rc1.
+post-install-commands = [
+  "pip install duckdb>=0.7.0",
```
tlento, line 121: Oh, we need to manually install the adapter dependencies? That's a bummer.

tlento (approved):
Ok, so Postgres (and DuckDB) work but SqlEngineTests will not, that seems fine for now. We should be able to use the same trick we used for DuckDB here.

Thanks for doing this!

PR #813, opened 2023-10-17
Remove error for date part with non-metric time dimension

courtneyholcomb:
### Description
Found a bug where date part queries error for any time dimension that isn't `metric_time`. This wasn't caught because all of our tests used `metric_time` (in both MF and MFS) - lesson learned! This creates an instant bug if you try to use any time dimension besides `metric_time` in our Tableau integration.
This is not the ideal fix. Ideally we'll need to put `date_part` in the validation path the way that `time_granularity` is, but we don't have time to do that before the fix needs to ship.


tlento (approved):
Seems reasonable. We definitely need a better approach in the query parser logic.

PR #817, opened 2023-10-17
Re-categorize `TypeErrors` that arise from `create_from_where_filter` into `InvalidQueryException`

DevonFulcher:
Resolves https://dbt-labs.slack.com/archives/C04TRSFQN75/p1697568904350429 (internal to dbt Labs)

### Description

The intent of this PR is to re-categorize `TypeErrors` that arise from `create_from_where_filter` into `InvalidQueryException`. That way, these errors can be caught and handled properly. Additionally, I move the try-catch block for `ParseWhereFilterException` to be within the body of `create_from_where_er` that way all call sites can benefit from that catch.


metricflow/specs/where_filter_transform.py
```diff
@@ -67,23 +67,41 @@ def create_from_where_filter(self, where_filter: WhereFilter) -> WhereFilterSpec
...
-            where_sql=rendered_sql_template,
-            bind_parameters=self._bind_parameters,
-            linkable_spec_set=LinkableSpecSet(
-                dimension_specs=tuple(dimension_specs),
-                time_dimension_specs=tuple(time_dimension_factory.time_dimension_specs),
-                entity_specs=tuple(entity_factory.entity_specs),
-            ),
-        )
+    def create_from_where_filter(self, where_filter: WhereFilter) -> WhereFilterSpec:  # noqa: D
+        # Used to check that call parameter sets generated by DSI match those generated below.
```
tlento, line 71: nit: this comment is hard to understand.

Can we replace it, and the triple-quoted inline below, with a docstring for this method that explains why it's doing what it's doing? It's super odd to me to see a `#noqa: D` on a method annotated with a bunch of inline comments, including one in triple-quoted blocks, when most of those are describing the purpose of the function.

I realize none of this is due to your change, but while we are refactoring a bit updating and improving documentation on this method seems in bounds to me.

tlento (approved):
Nice, thanks for fixing the error classification on this one!

PR #823, opened 2023-10-26
Update minimum typing-extensions version to 4.4

tlento:
The `@override` decorator was added to typing-extensions with version
4.4, but our minimum version is 4.0, which means there are scenarios
where this will suddenly start throwing errors.

Since we require 4.4, let's make that true in the build directive.


courtneyholcomb (approved):
Good call!!

PR #824, opened 2023-10-28
Add missing snapshots for saved queries

tlento:
The new saved query CLI tests rely on snapshot generation, but
the generate snapshot script will not update them and so our engine
tests are difficult to maintain.

This adds the relevant tests to the update list and generates the
associated snapshot files by way of illustrating that everything works.

We might consider moving to a pytest.mark attribute for the modules
and test cases that rely on snapshots, but for the time being the list
of test paths seems fine.


courtneyholcomb (approved):
Looks good to me!

PR #827, opened 2023-10-30
Update sql-engine-test workflow to remove label on completion.

plypaul:
### Description

Currently, when iterating on a PR, the PR author would need to remove and then re-add the label to rerun the SQL-engine tests. This PR updates the Github Action workflow to remove the label for running the SQL engine tests after the tests are complete.


courtneyholcomb (approved):
LGTM!

PR #831, opened 2023-10-31
Fix random failures with parallel CLI tests

tlento:
We use pytest-xdist for parallel test execution, which means multiple
worker threads fire up tests which are loosely grouped by class/module.

Unfortunately, there are no guarantees around any of this, so the CLI
tests, which are all in one module, may execute on different workers.
We had a bad fixture layout on the CLI tests which would touch a file
in the CLI directory and then unlink it after the test was done. If
every test ran on the same worker this was fine - only one test would
ever be accessing this path location at a time. But when the tests get
split across workers bad things happen with FileNotFound and things
of that nature.

This change moves us off of having multiple tests creating and
destroying the same file in the same file path location, instead
relying on a fixed file that already exists in a static location in
the repository.

### DIFF
--- a/metricflow/test/cli/test_cli.py
+++ b/metricflow/test/cli/test_cli.py
@@ -15,6 +15,7 @@ from dbt_semantic_interfaces.parsing.dir_to_model import (
 from dbt_semantic_interfaces.parsing.objects import YamlConfigFile
 from dbt_semantic_interfaces.test_utils import base_semantic_manifest_file
 
+from metricflow.cli.cli_context import CLIContext
 from metricflow.cli.main import (
     dimension_values,
     dimensions,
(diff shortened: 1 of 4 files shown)


tlento:
* main
  * **PR #831** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/831" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dbt-labs/metricflow/831?utm_source=stack-comment).

WilliamDee (approved):

courtneyholcomb (approved):
Ahh thanks for resolving this!! Thought I had something wrong with my local setup that was causing these errors 🫠

plypaul (approved):
I was wondering about that issue - thanks for the fix!

PR #834, opened 2023-11-02
Add ability to use distinct select in nodes

WilliamDee:
### Description
As part of reviving the PR for conversion metrics (https://github.com/dbt-labs/metricflow/pull/352). This was needed and was buried in that PR for the sake of convenience. Pulling that out to this PR so we can shrink that PR's coverage.

### DIFF
--- a/metricflow/test/sql/optimizer/test_sub_query_reducer.py
+++ b/metricflow/test/sql/optimizer/test_sub_query_reducer.py
@@ -266,3 +266,65 @@ def test_rewrite_order_by_with_a_join_in_parent(
         sql_plan_node=sub_query_reducer.optimize(rewrite_order_by_statement),
         plan_id="after_reducing",
     )
+
+
+def test_distinct_select_node_is_not_reduced(
+    request: FixtureRequest,
+    mf_test_session_state: MetricFlowTestSessionState,
+) -> None:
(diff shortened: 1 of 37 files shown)


metricflow/test/sql/optimizer/test_sub_query_reducer.py
```diff
@@ -266,3 +266,65 @@ def test_rewrite_order_by_with_a_join_in_parent(
         sql_plan_node=sub_query_reducer.optimize(rewrite_order_by_statement),
         plan_id="after_reducing",
     )
+
+
+def test_rewrite_distinct_select_node(
```
tlento, line 271: Maybe `test_reduce_distinct_select_node`?

Also, I prefer the test name to show what should happen, like `test_distinct_select_node_is_not_removed` or something, but if that's not how the test files do their naming this is fine as-is.

tlento (approved):
@WilliamDee FYI @courtneyholcomb added a group_by in the select node in https://github.com/dbt-labs/metricflow/pull/804/commits/4f0d5d2552fde763928aee5b932cd9dc772d8667 which could be switched over to use distinct instead.

Since DISTINCT is apparently the ANSI-standard keyword maybe it's worth it to make that update. I think a separate PR would be preferable just to keep these changes isolated.

PR #835, opened 2023-11-02
Bug fix: CLI saved queries dropping `order_by_names` param

courtneyholcomb:
### Description
Bug fix. In the CLI, we are using the `MetricFlowQueryRequest.order_by_names` param. When passing that query to the query parser, we end up dropping that param in favor of order by objects, instead. This meant the order by param was dropped entirely from CLI saved queries, returning unexpected results. You can see one of the hardcoded result sets has the incorrect order. This PR makes sure both get passed through to the final query, fixing the ordering issue.


courtneyholcomb:
Realized I need to regenerate snapshots for the other SQL engines before merging - working on that now.

tlento (approved):
Nice catch! It is wild to me that this wasn't just failing all the time.....

PR #842, opened 2023-11-03
Switch to `FULL OUTER JOIN` for derived & ratio metrics

courtneyholcomb:
Resolves #SL-1172

### Description
We've decided to switch to using `FULL OUTER JOIN` when combining input metrics for derived & ratio metrics. This is a step needed to avoid removing rows unexpectedly when filling nulls.
Users should note that this may change the output for derived & ratio metric queries, resulting in new output `NULL` rows that previously would have been removed.

### DIFF
--- a/metricflow/plan_conversion/dataflow_to_sql.py
+++ b/metricflow/plan_conversion/dataflow_to_sql.py
@@ -959,7 +959,7 @@ class DataflowToSqlQueryPlanConverter(DataflowPlanNodeVisitor[SqlDataSet]):
         ), "All parent nodes should have the same set of linkable instances since all values are coalesced."
 
         linkable_spec_set = from_data_set.data_set.instance_set.spec_set.transform(SelectOnlyLinkableSpecs())
-        join_type = SqlJoinType.CROSS_JOIN if len(linkable_spec_set.all_specs) == 0 else node.join_type
+        join_type = SqlJoinType.CROSS_JOIN if len(linkable_spec_set.all_specs) == 0 else SqlJoinType.FULL_OUTER
 
         joins_descriptions: List[SqlJoinDescription] = []
         # TODO: refactor this loop into SqlQueryPlanJoinBuilder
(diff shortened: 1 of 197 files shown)


metricflow/test/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/BigQuery/test_derived_metric__plan0.sql
```diff
@@ -441,13 +441,7 @@ FROM (
     ) subq_8
   ) subq_9
   ON
-    (
```
courtneyholcomb, line 444: @tlento do you have context on why for INNER JOIN we add this NULL column comparison, but for FULL OUTER JOIN we don't? Wondering if that's intentional or something I should fix!

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -906,12 +906,10 @@ def visit_combine_metrics_node(self, node: CombineMetricsNode) -> SqlDataSet:
...
         This node may exist in one of two situations: when metrics need to be combined in order to produce a single
-        dataset with all required inputs for a derived metric (in which case the join type is INNER), or when
-        metrics need to be combined in order to produce a single dataset of output for downstream consumption by
-        the end user, in which case we will use FULL OUTER JOIN.
+        dataset with all required inputs for a derived metric, or when metrics need to be combined in order to produce
+        a single dataset of output for downstream consumption by the end user.
 
-        In the case of a multi-data-source FULL OUTER JOIN the join key will be a coalesced set of all previously
-        seen dimension values. For example:
+        The join key will be a coalesced set of all previously seen dimension values. For example:
```
tlento, line 912: I don't think we need to do this here, but if there's logic in this function for rendering the INNER JOIN case maybe we can remove the join type parameter AND that logic.

tlento (approved):
Looks good, thanks! If we can simplify the code a bit further let's do that in a follow-up.

metricflow/plan_conversion/dataflow_to_sql.py
courtneyholcomb: I'm fairly certain the INNER JOIN case is not used at all! I can remove the join_type param before merging!

courtneyholcomb: ...I take it back, it's more complicated than I thought 😅  so will deal with that in a follow up!

courtneyholcomb: Actually, I was just on the wrong branch. Easy to add here after all! Removed the param and all it changes is the displayed properties in the dataflow plans 👍

PR #844, opened 2023-11-03
`Makefile` updates

courtneyholcomb:
### Description
`Makefile` updates:
- Removes quotes from variables. The commands work without the quotes. Sometimes they work with the quotes, but there are some cases where the quotes cause errors.
- Adds `test-snap` target, something I use often to see snapshot changes without waiting for all SQL engines to run.


tlento (approved):
astronaut dog

PR #848, opened 2023-11-06
Fill nulls for joined metrics

courtneyholcomb:
Resolves #SL-1173

### Description
- Updates the logic when filling nulls for input measures. When joining multiple metrics, you may end up with null rows post-join if 1) there are dimensions in the query and 2) one of the metrics has dimension values that the other metric does not have. In that case, if the metric's input measure fills nulls, we want to fill those nulls with the requested value. This PR implements that behavior.
- Tangentially, this updates `MetricInstance.defined_from` to contain a single `MetricModelReference` instead of multiple. It doesn't seem like there is ever a time we would want multiple (multiple are never used in the codebase), and we need to know just one to get the correct element name.


metricflow/test/snapshots/test_fill_nulls_with_rendering.py/SqlQueryPlan/DuckDB/test_derived_fill_nulls_for_one_input_metric__plan0.sql
```diff
@@ -6,7 +6,7 @@ FROM (
   -- Combine Metrics
   SELECT
     COALESCE(subq_7.metric_time__day, subq_15.metric_time__day) AS metric_time__day
-    , MAX(subq_7.bookings_fill_nulls_with_0) AS bookings_fill_nulls_with_0
+    , COALESCE(MAX(subq_7.bookings_fill_nulls_with_0), 0) AS bookings_fill_nulls_with_0
```
courtneyholcomb, line 9: Can you aggregate within COALESCE for all engines?

PR #850, opened 2023-11-06
Fill nulls for multi-metric queries

courtneyholcomb:
Resolves #SL-1173

### Description
- Updates the logic when filling nulls for input measures. When joining multiple metrics, you may end up with null rows post-join if 1) there are dimensions in the query and 2) one of the metrics has dimension values that the other metric does not have. In that case, if the metric's input measure fills nulls, we want to fill those nulls with the requested value. This PR implements that behavior.
- Tangentially, this updates `MetricInstance.defined_from` to contain a single `MetricModelReference` instead of multiple. It doesn't seem like there is ever a time we would want multiple (multiple are never used in the codebase), and we need to know just one to get the correct element name.


metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -706,7 +706,7 @@ def visit_compute_metrics_node(self, node: ComputeMetricsNode) -> SqlDataSet:
             metric_instances.append(
                 MetricInstance(
                     associated_columns=(output_column_association,),
-                    defined_from=(MetricModelReference(metric_name=metric_spec.element_name),),
+                    defined_from=MetricModelReference(metric_name=metric_spec.element_name),
```
tlento, line 709: LOL, thanks. We have so many of these kind of "let's allow a sequence of things and then have exactly one callsite that manually enforces there's only one" interfaces lying around...

metricflow/specs/specs.py
```diff
@@ -436,7 +436,7 @@ def from_reference(reference: MetricReference) -> MetricSpec:
 
     @property
     def alias_spec(self) -> MetricSpec:
-        """Returns a MetricSpec represneting the alias state."""
+        """Returns a MetricSpec representing the alias state."""
```
tlento, line 439: 😸

metricflow/test/integration/test_cases/itest_metrics.yaml
```diff
@@ -1312,4 +1312,107 @@ integration_test:
...
+---
+integration_test:
+  name: fill_nulls_with_0_multi_metric_query
+  description: Test a multi-metric query that fills nulls
+  model: SIMPLE_MODEL
+  metrics: ["bookings_fill_nulls_with_0", "views"]
+  group_by_objs: [{"name": "metric_time"}]
+  check_query: |
+    SELECT
+      COALESCE(subq_7.metric_time__day, subq_12.metric_time__day) AS metric_time__day
```
tlento, line 1326: Are there missing values in this date sequence for the bookings_fill_nulls_with metric? If so that's great. An output test would be nice but at least we can inspect the data frame from here and see it.

If not, it might be worth adding a dimension that has a gap for a given metric_time, dimension_value pair.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -894,6 +895,27 @@ def _make_select_columns_for_metrics(
...
+                # At this point, the MetricSpec might have the alias in place of the element name, so we need to look
+                # back at where it was defined from to get the metric element name.
+                metric_name = metric_instance.defined_from.metric_name
+                input_measures = self._metric_lookup.measures_for_metric(
+                    metric_reference=MetricReference(metric_name),
+                    column_association_resolver=self._column_association_resolver,
+                )
+                # If multiple input measures, this is a query with a nested derived/ratio metric. In that case, we can
+                # skip this step because it has already occurred when rendering the nested metric.
+                # TODO: this logic might need updating for conversion metrics, which will allow multiple input measures
```
tlento, line 907: Can we make this condition based on metric type (or some is_derived property or similar) instead of number of measures?

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -894,6 +895,27 @@ def _make_select_columns_for_metrics(
                 else:
                     select_expression = column_reference_expression
 
+                # At this point, the MetricSpec might have the alias in place of the element name, so we need to look
+                # back at where it was defined from to get the metric element name.
+                metric_name = metric_instance.defined_from.metric_name
+                input_measures = self._metric_lookup.measures_for_metric(
+                    metric_reference=MetricReference(metric_name),
+                    column_association_resolver=self._column_association_resolver,
+                )
```
tlento, line 904: Another way to deal with the issue below is to add a different, internal API here that only gives the measures that are direct inputs into the metric. That way, if we add another metric type that takes measures as input anything derived from it won't have anything to do here, and we ensure we only do the coalesce once within a derived metric chain.

tlento (commented):
I don't know if this works, but if it does, can we either:

1. Have an accessor on the metric_lookup object that only returns the immediate measure inputs for the metric in question?
2. Have an is_derived or similar property that is based on the metric type rather than the number of measure inputs?

I'm concerned about edge case bugs creeping into the rendering here, and it'll be hard to reason about them because the typechecker can't help out at all. Like if I add a new metric type and we have logic that does an assert_values_exhausted if/else block switching over all metric types, that'll fail the typechecker.

metricflow/plan_conversion/dataflow_to_sql.py
courtneyholcomb: I like that idea! I'll add that.

courtneyholcomb: Added that!

metricflow/test/integration/test_cases/itest_metrics.yaml
courtneyholcomb: Added a new test case & updated the source data to make sure there were some nulls filled for this case.
Will plan to add some output tests tomorrow!

courtneyholcomb:
@tlento This should be good to go, with the exception of updating Databricks snapshots. Wrapping that up now - repopulating the source schema is just taking forever.

courtneyholcomb:
> @tlento This should be good to go, with the exception of updating Databricks snapshots. Wrapping that up now - repopulating the source schema is just taking forever.

And now done! Kicking off SQL engine tests now

metricflow/test/integration/query_output/test_fill_nulls_with_0.py
```diff
@@ -0,0 +1,145 @@
...
+from _pytest.fixtures import FixtureRequest
+
+from metricflow.engine.metricflow_engine import MetricFlowQueryRequest
+from metricflow.protocols.sql_client import SqlClient
+from metricflow.test.fixtures.setup_fixtures import MetricFlowTestSessionState
+from metricflow.test.integration.conftest import IntegrationTestHelpers
+from metricflow.test.snapshot_utils import assert_object_snapshot_equal
+
+
+@pytest.mark.sql_engine_snapshot
```
tlento, line 15: These snapshot tests are great, thanks for doing this!

tlento (approved):
Let's go! Thank you!

metricflow/model/semantics/metric_lookup.py
```diff
@@ -105,6 +106,23 @@ def add_metric(self, metric: Metric) -> None:
                 )
         self._metrics[metric_reference] = metric
 
+    def yaml_input_measure_for_metric(self, metric_reference: MetricReference) -> Optional[MetricInputMeasure]:
```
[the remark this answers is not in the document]
courtneyholcomb: Updated!

PR #852, opened 2023-11-08
Implemented date_part in where filter

DevonFulcher:
### Description

This PR allows users to specify `date_part` in the where filter. This is useful for the Jinja syntax as well as integrations such as Tableau.


tlento (commented):
This looks like two PRs in one. Is there something wonky in the underlying branch, or should this be split up?

DevonFulcher:
> This looks like two PRs in one. Is there something wonky in the underlying branch, or should this be split up?

I fixed it now. I'm sorry about that.

metricflow/specs/where_filter_dimension.py
```diff
@@ -37,32 +38,48 @@ def __init__(  # noqa
...
-        self.dimension_spec = self._dimension_spec_resolver.resolve_dimension_spec(name, entity_path)
-        self.time_dimension_spec: Optional[TimeDimensionSpec] = None
+        self.dimension_spec: DimensionSpec = self._dimension_spec_resolver.resolve_dimension_spec(
+            self._name, self._entity_path
+        )
+        self.date_part_name: Optional[str] = None
+        self.time_granularity_name: Optional[str] = None
+
+    @property
+    def time_dimension_spec(self) -> TimeDimensionSpec:
```
tlento, line 48: Ah, so this returns a TimeDimension with DEFAULT_TIME_GRANULARITY if neither grain nor date_part are set. Interesting. I think that makes the most sense with where we're heading with time dimension expressions.

metricflow/test/model/test_where_filter_spec.py
```diff
@@ -98,6 +99,94 @@ def test_time_dimension_in_filter(  # noqa: D
...
+        entity_specs=(),
+    )
+
+
+@pytest.mark.parametrize(
+    "where_sql",
+    (
+        ("{{ TimeDimension('metric_time', 'WEEK', date_part_name='year') }} = '2020'"),
+        ("{{ Dimension('metric_time').date_part('year').grain('WEEK') }} = '2020'"),
+        ("{{ Dimension('metric_time').grain('WEEK').date_part('year') }} = '2020'"),
```
tlento, line 131: I hope these aren't case-sensitive. I may add some additional tests on after we merge.

tlento (commented):
Looks great, thanks! I'll hold this open until after 0.203.0 is out (unless that drags past Monday, in which case maybe we just merge this).

tlento (approved):

PR #854, opened 2023-11-08
Dataflow Plan for Min & Max of Distinct Values Query

courtneyholcomb:
Resolves #SL-1080

### Description
Adds a new Dataflow Plan Node to get the min & max of a single-column distinct values query. This is a very limited use case node, only designed to get metadata about group by inputs for integrations.
I opted not to add a changelog entry for this because I don't think we should document it as something for users.


tlento:
Oh I sure did forget all about this one, will take a look tomorrow if nobody else gets to it first.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1348,3 +1349,36 @@ def visit_join_to_time_spine_node(self, node: JoinToTimeSpineNode) -> SqlDataSet
...
+                expr=SqlAggregateFunctionExpression(
+                    sql_function=SqlFunction.from_aggregation_type(aggregation_type=agg_type),
+                    sql_function_args=[parent_column_alias_expr],
+                ),
+                column_alias=agg_type.value,
+            )
+            for agg_type in (AggregationType.MIN, AggregationType.MAX)
+        ]
+        return SqlDataSet(
+            instance_set=parent_data_set.instance_set,
```
plypaul, line 1374: We haven't fleshed out the `MetadataSpec` too much, but the min / max value might be better associated with that.

courtneyholcomb: Are you thinking - instead of adding a `MinMaxNode`, use the `MetadataSpec` and add a new `InstanceSpecVisitor` to visit it?

tlento: Oh please don't add more InstanceSpecVisitors.....

tlento: FWIW, this would be right in line with the documented purpose of the MetadataSpec, although it's a bit fuzzier since the column is actually part of the output.

It would be nice if the contract around instance spec and SqlSelectStatementNode.select_columns was more clear and easier to inspect, but it's complicated.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1348,3 +1349,36 @@ def visit_join_to_time_spine_node(self, node: JoinToTimeSpineNode) -> SqlDataSet
...
+        parent_column_alias_expr = SqlStringExpression(parent_data_set.sql_select_node.select_columns[0].column_alias)
+
+        # Build aggregate columns, labeled "min" & "max"
+        select_columns = [
+            SqlSelectColumn(
+                expr=SqlAggregateFunctionExpression(
+                    sql_function=SqlFunction.from_aggregation_type(aggregation_type=agg_type),
+                    sql_function_args=[parent_column_alias_expr],
+                ),
+                column_alias=agg_type.value,
```
tlento, line 1369: `min` and `max` are reserved keywords in a lot of dialects, aren't they? They're also kind of nondescript. I wonder if we want to do `min_<name>` and `max_<name>` instead.

tlento (commented):
Overall this seems reasonable but I think it'll cause problems if we ever end up wrapping this in a subquery for some reason. Updating the naming and spec should help with this.

Separately, I know @plypaul is working on cleaning up our query param handling. With the current state I don't think there's a great alternative to having this be part of the spec and then doing runtime validation to make sure we don't allow it when it shouldn't be included, but hopefully you all can sort out the smoothest transition.

metricflow/plan_conversion/dataflow_to_sql.py
courtneyholcomb: This is how it's currently implemented in MFS for the Tableau integration (with just `min` and `max` as column names). Seems to be working as expected. I wanted to make sure SLG would be able to know what the column name would be to select it - but maybe in SLG we can do a `startswith` check on the column name instead to see if it starts with `min_` or `max_`. Let me sync with @aiguofer on this.

courtneyholcomb: Diego's on board. @tlento would it make more sense to do `<name>__min` and `<name>__max` to follow our usual dunder pattern? Not sure where we're at in terms of getting rid of dunders these days...

courtneyholcomb: I implemented this with `min_<name>` and `max_<name>` for now - LMK if you think that should change!

tlento (commented):
I have so many review-related things to do today.....

See inline. Short answer is I don't know what we should do regarding to dunder or not to dunder, but I'll have more space for that on Monday.

I suspect we'll end up with a more general way of putting the min/max as dunder suffixes on the output columns. If so that doesn't necessarily need to happen before this PR goes in, but we'd at least want the output columns to be formatted accordingly.

metricflow/specs/specs.py
```diff
@@ -126,14 +126,15 @@ class MetadataSpec(InstanceSpec):
     """A specification for a specification that is built during the dataflow plan and not defined in config."""
 
     element_name: str
+    agg_type: Optional[AggregationType] = None
 
     @property
     def qualified_name(self) -> str:  # noqa: D
-        return self.element_name
+        return f"{self.agg_type.value}_{self.element_name}" if self.agg_type else self.element_name
```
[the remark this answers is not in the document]
courtneyholcomb: Sounds good - will leave PR this until next week!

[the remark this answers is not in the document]
courtneyholcomb: Following up on this - have you thought about the above yet?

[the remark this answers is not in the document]
courtneyholcomb: Sounds good! Just updated to use those names.

Anything else missing for this PR? Would love to get it merged if not!

metricflow/dataflow/dataflow_plan.py
```diff
@@ -1248,6 +1253,36 @@ def with_new_parents(self, new_parent_nodes: Sequence[BaseOutput]) -> ConstrainT
...
+class MinMaxNode(BaseOutput):
+    """Calculate the min and max of a single instance data set."""
+
+    def __init__(self, parent_node: BaseOutput) -> None:  # noqa: D
+        self._parent_node = parent_node
+        super().__init__(node_id=self.create_unique_id(), parent_nodes=[parent_node])
+
+    @classmethod
+    def id_prefix(cls) -> str:  # noqa: D
+        return DATAFLOW_NODE_MIN_MAX_ID_PREFIX
```
tlento, line 1265: @plypaul is this still ok given your other changes to an enumerated ID prefix type?

metricflow/query/query_parser.py
```diff
@@ -397,7 +399,17 @@ def _parse_and_validate_query(
         where_constraint_str: Optional[str] = None,
         order_by_names: Optional[Sequence[str]] = None,
         order_by: Optional[Sequence[OrderByQueryParameter]] = None,
+        min_max_only: bool = False,
```
tlento, line 402: @courtneyholcomb please coordinate with @plypaul , this file changed dramatically in his pending stack of changes.

metricflow/specs/specs.py
```diff
@@ -126,14 +126,15 @@ class MetadataSpec(InstanceSpec):
     """A specification for a specification that is built during the dataflow plan and not defined in config."""
 
     element_name: str
+    agg_type: Optional[AggregationType] = None
 
     @property
     def qualified_name(self) -> str:  # noqa: D
-        return self.element_name
+        return f"{self.element_name}{DUNDER}{self.agg_type.value}" if self.agg_type else self.element_name
```
tlento, line 133: @plypaul - as we discussed, this is an example of where we want to have structured annotations. The plan is to add an better abstraction to the spec interface so we can unify this with the similar thing we're doing with non-additive time dimensions, and then build out support for auto-aliasing metric offsets on top of the same set of spec interfaces.

tlento (approved):

PR #856, opened 2023-11-08
Expand test coverage for cumulative and derived offset metrics

tlento:
Current test coverage for cumulative and derived offset metrics does not include
any tests of how we render queries with filters that do not allow us to do any
expansion on metric_time. This will be necessary as we add support for
detecting metric_time filter expressions and full predicate pushdown, as we will need
a way to verify that our logic is still rendering queries that will produce the correct results.

Worth noting, in this PR, that we currently have a number of rendering scenarios for
cumulative metrics where the queries are incorrect. As these are meant to be blocked by
query validation, I am leaving most of them intact, with the idea that when we properly
support those scenarios the rendering tests will update to reflect the fixes.

PR #857, opened 2023-11-08
Fix query validation for metric_time requirements

### LINKED ISSUE #825: Cumulative metric validation allows queries which will not render correctly
### Updated task scope

See sections below for original context.

This task is now to fix the broken validation that allows users to issue cumulative metric queries with time dimensions that our rendering logic does not properly support.

There are three other follow-ups which we need to track:

1. Allowing queries for cumulative and derived offset metrics where metric_time is not specified OR where metric_time is specified by its underlying agg_time_dimension name
2. Allowing queries for cumulative and derived offset metrics to be grouped by other time dimensions even if the metric_time equivalent is not requested directly in the query. This is a much larger adjustment than it seems, as it effectively opens up the question about allowing for query-time agg_time_dimension overrides
3. Providing users with a mechanism for giving us structured information about time ranges in order to enable query optimization via more aggressive initial filtering (for engines which benefit from such things)

### Original issue details

Cumulative metrics in MetricFlow have two different time axes that they need to consider:

1. The metric_time axis, which can be subject to a filter applied by a derived metric or a query-time `where` input
4. The cumulative window time axis, which is defined in the cumulative metric spec itself

Today, there are three ways to specify a metric_time filter constraint:

1. As a raw filter expression assigned to one of the `filter` parameters available for a derived metric that uses one or more cumulative metrics as input
2. As a raw `where` expression at query time
5. via the deprecated and all but inaccessible `start_time` and `end_time` query inputs

Of these, only the third one has any awareness of the relationship between the cumulative metric and the metric_time axis - we will automatically expand the time range filter such that all relevant inputs are included. For the other two approaches - which are the current and future standard - we truncate input data according to the filter range specified. This results in confusing outputs.

Imagine a 7-day cumulative metric adding up events per day for the first week of January, 2023. For simplicity, we will assume that there is 1 event per day, which means the 7 day cumulative value for 2023-01-07 should be 7.

### Cumulative metrics using raw filter expressions

Input: `mf query --metrics event_count_7d --where "{{TimeDimension('metric_time', 'day')}} = '2023-01-07'"`
Output: 

| metric_time | event_count_7d |
| ------------ | ---------------- |
| 2023-01-07 | 1 |

### Cumulative metrics using start-time/end-time

Input: `mf query --metrics event_count_7d --start-time '2023-01-07' --end-time '2023-01-07'
Output: 

| metric_time | event_count_7d |
| ------------ | ---------------- |
| 2023-01-01 | 1 |
| 2023-01-02 | 2 |
| 2023-01-03 | 3 |
| 2023-01-04 | 4 |
| 2023-01-05 | 5 |
| 2023-01-06 | 6 |
| 2023-01-07 | 7 |

Since this is confusing and not in accordance with any reasonable user expectation, we are treating this as a high priority bug for our 1.7 release. The plan for the fix is:

1. Ensure correctness in all cases. This will be expensive, but we are going to adjust the query plan to use a post-aggregation filter for cumulative metrics by default. End users should still be able to get the input range expansion and pre-aggregation filter behavior with the start-time/end-time parameters, but derived metric filter expressions will trigger full input data scans.
2. Update our filter handling (mechanism still TBD) to allow us to do pre-aggregation filtering on expanded ranges for raw filter expressions in a manner that works with derived metrics. This will be tracked in a separate issue.

tlento:
Resolves #825

### Description

As of right now we require metric_time in the group by expression
for any metric query against cumulative metrics or derived metrics
with time offset windows or an offset_to_grain parameter set.

The current query validation was only asserting that some time
dimension was included in the group by. In the case of derived
metrics, this simply failed later in the query building phase for
any query with a different time dimension in the group by expression.

In the case of cumulative metrics, this would run an incorrectly
rendered query and then return incorrect results.

This validation change ensures that users have the proper
configuration on queries based on our current limitations.

### DIFF
--- a/.changes/unreleased/Fixes-20231108-150708.yaml
+++ b/.changes/unreleased/Fixes-20231108-150708.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Fix query validation for metric_time requirements
+time: 2023-11-08T15:07:08.681102-08:00
+custom:
+  Author: tlento
+  Issue: "825"
--- a/metricflow/query/query_parser.py
+++ b/metricflow/query/query_parser.py
@@ -270,8 +270,13 @@ class MetricFlowQueryParser:
         finally:
             logger.info(f"Parsing the query took: {time.time() - start_time:.2f}s")
 
-    def _validate_no_time_dimension_query(self, metric_references: Sequence[MetricReference]) -> None:
-        """Validate if all requested metrics are queryable without a time dimension."""
+    def _validate_no_metric_time_dimension_query(
+        self, metric_references: Sequence[MetricReference], time_dimension_specs: Sequence[TimeDimensionSpec]
+    ) -> None:
+        """Validate if all requested metrics are queryable without grouping by metric_time."""
+        if any([spec.reference == DataSet.metric_time_dimension_reference() for spec in time_dimension_specs]):
+            return
+
         for metric_reference in metric_references:
             metric = self._metric_lookup.get_metric(metric_reference)
             if metric.type == MetricType.CUMULATIVE:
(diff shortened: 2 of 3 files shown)


tlento:
* main
  * **PR #856** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/856" target="_blank">Graphite</a> 
    * **PR #857** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/857" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dbt-labs/metricflow/857?utm_source=stack-comment).

WilliamDee (approved):

courtneyholcomb (approved):
LGTM!

PR #858, opened 2023-11-09
Allow CombineMetricsNode to combine the outputs from aggregated measures

WilliamDee:
### Description
Currently, all metrics have only 1 input_measure linked to it, this was a result from ratio metric switching to use input_metrics rather than input_measures. This resulted in the removal of `JoinAggregatedMeasuresByGroupByColumnsNode` which was then unused. However, this node is now useful again for conversion metrics as we need to have the ability to join multiple aggregated measures node together. After a few discussions, tweaking `CombineMetricsNode` to work for joining metrics or measures outputs was a better alternative than bringing back `JoinAggregatedMeasuresByGroupByColumnsNode`.

### Changes
- Added `fill_nulls_with` to `MeasureSpec`
  - populate the `fill_nulls_with` during measure aggregation such that nodes upstream (such as the combine node) can know about what values should be filled for nulls
- Added the ability to combine `AggregateMeasuresNode` in `CombineMetricsNode`
- Rename `CombineMetricsNode` to `CombineAggregatedOutputsNode`


metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -838,6 +840,28 @@ def _make_select_columns_for_multiple_metrics(
...
+                        aggregation_type=aggregation_type, sql_column_expression=column_reference_expression
+                    )
+                else:
+                    select_expression = column_reference_expression
+                select_columns.append(
+                    SqlSelectColumn(
+                        expr=select_expression,
+                        column_alias=measure_column_name,
+                    )
+                )
```
tlento, line 864: Ok, I see. Since this is in fact identical to the metric logic I think we can do a little refactor to have a MetricInput Protocol type (or whatever we want to call it) and work off of that.

Do we need to apply the fill nulls logic here as well? I suspect we do, since this is a post-aggregation join, right? Assuming that's the case, we can add an is_measure property to the protocol if we need to.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -877,12 +901,18 @@ def visit_combine_metrics_node(self, node: CombineMetricsNode) -> SqlDataSet:
...
             parent_sql_data_set = parent_node.accept(self)
             table_alias = self._next_unique_table_alias()
             parent_data_sets.append(AnnotatedSqlDataSet(data_set=parent_sql_data_set, alias=table_alias))
-            table_alias_to_metric_instances[table_alias] = parent_sql_data_set.instance_set.metric_instances
+            if parent_sql_data_set.instance_set.metric_instances:
+                table_alias_to_metric_instances[table_alias] = parent_sql_data_set.instance_set.metric_instances
+            elif parent_sql_data_set.instance_set.measure_instances:
+                table_alias_to_measure_instances[table_alias] = parent_sql_data_set.instance_set.measure_instances
+            else:
+                raise RuntimeError("Attempting to combine output nodes without any measures/metrics upstream")
```
tlento, line 915: If we use a shared protocol we don't need to do this, although it might be cleaner if we update the type on table_alias_to_metric_instances to take the specs instead. We only use the instance to get the spec in the one callsite here.

tlento (commented):
I think combining the logic makes sense but let's add a Protocol type so (or, since the properties are the same, we can just take a Union[] and add a single switch on an isinstance() check - I prefer the protocol type as it eliminates the need for an isinstance() check and makes it a little more obvious which properties need to change in lock-step).

metricflow/plan_conversion/dataflow_to_sql.py
WilliamDee: this was recently changed to use instance due to the fill null https://github.com/dbt-labs/metricflow/pull/850/files#diff-488be53d5a33d76e334c75668b62f8c961f169996b1fba1cca5daaf177e1d81bR900

metricflow/plan_conversion/dataflow_to_sql.py
WilliamDee: > Do we need to apply the fill nulls logic here as well?

This is a good point actually. Need to figure out how we can propagate that information for combining AggregateMeasureNodes as it's easy for ComputeMetricNodes when you have the metric_spec and you can get the measure_inputs from that, but in this case we're just combining the AggregateMeasureNodes which means we need to somehow pass the metric_spec without actually using it during the combine process

WilliamDee: Oh wait, the `AggregateMeasureNodes` has `metric_input_measure_specs` in it.

metricflow/plan_conversion/instance_converters.py
```diff
@@ -586,6 +593,47 @@ def transform(self, instance_set: InstanceSet) -> InstanceSet:  # noqa: D
         )
 
 
+class UpdateMeasureFillNullsWith(InstanceSetTransform[InstanceSet]):
```
courtneyholcomb, line 596: Not totally relevant to the PR, but this whole "instance converter" pattern is something I need to understand better... i.e., when is it appropriate to add a new instance converter vs. just putting that code in `dataflow_to_sql` or wherever it's relevant? Maybe a question for @tlento!

courtneyholcomb (dismissed):
Looks good to me! With the note that I would love for us to establish some clear logic about when to use instance converters.

courtneyholcomb:
Also, why aren't checks running for this PR 🤔

WilliamDee:
> Also, why aren't checks running for this PR 🤔

@courtneyholcomb there's been some discussion on how we wanted to do this, so I wanted an initial review on the implementation before I rename the node itself and then run the tests

metricflow/protocols/semantics.py
```diff
@@ -187,3 +187,13 @@ def metric_input_specs_for_metric(
     ) -> Sequence[MetricSpec]:
         """Returns the metric input specs required to compute the metric."""
         raise NotImplementedError
+
+    def configured_input_measure_for_metric(self, metric_reference: MetricReference) -> Optional[MetricInputMeasure]:
```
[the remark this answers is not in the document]
WilliamDee: ah right, missed that

metricflow/plan_conversion/instance_converters.py
WilliamDee: yea abit hazy on that as well. I had it here just cause it felt cleaner. My take on this is that we would put it in the converter when we're mutating the instance spec set itself. But I guess there's also cases where we;re converting the spec set into select columns, etc...

plypaul:
> single switch on an isinstance() check 

Let's not do that.

WilliamDee:
> > single switch on an isinstance() check

> Let's not do that.

@plypaul there's no need to do this anymore since we added the fill_nulls_with property to MeasureSpec.

metricflow/plan_conversion/instance_converters.py
plypaul: @courtneyholcomb The original context on that is - let's say we created a new type of instance and had to update the code base. How do we find all the cases that need to be updated? The thought was that centralizing that logic into these classes would make it easy to do so. We haven't had to add a new instance in a while though. However, `dataflow_to_sql` is already large so separating things into smaller functional chucks is preferred.

plypaul (approved):

PR #866, opened 2023-11-14
Remove RudderStack logging from MetricFlow telemetry

tlento:
Prior to its acquisition by dbt Labs, Transform used Rudderstack
for analytics event log handling. As a result, the MetricFlow
telemetry system included the option to log data to Rudderstack.

Nowadays there is no longer a need to support this route, and indeed
the Rudderstack client itself causes some small problems, such as
the occasional irrepressible warning about SSL sockets failing
to close properly after running queries in the MetricFlow CLI.


courtneyholcomb (approved):
🪓 🪓 🪓

marcodamore (approved):
Had to dust off the old stamp for this one

PR #871, opened 2023-11-15
Fix incorrect cumulative metric output when start-time and end-time are used

tlento:
When users submit CLI queries with the --start-time and --end-time
parameters, MetricFlow will automatically adjust the time constraint
window to include all relevant cumulative metric input.

Unfortunately, we got over-zealous with fencepost adjustments and
introduced an off-by-one error that meant cumulative metric
queries would only be correct when the specified granularity was
DAILY.

We never caught this issue because the error in the rendered SQL is
subtle, and all of our test input data on the integration tests
happened to fit within the artificially shortened constraint windows
in all of the cases where we were testing non-daily granularity
queries.

This change updates the integration test configurations to ensure
boundary conditions will be caught in the future. In addition, the
fix alters the values of the query output snapshot tests, which,
when compared with the relevant input test data, will reveal the
nature of the original bug and the effects of this fix.


metricflow/filters/time_constraint.py
```diff
@@ -80,7 +80,7 @@ def adjust_time_constraint_for_cumulative_metric(
     ) -> TimeRangeConstraint:
         """Given a time constraint for the overall query, adjust it to cover the time range for this metric."""
         if granularity is not None:
-            return self._adjust_time_constraint_start_by_window(granularity, count - 1)
+            return self._adjust_time_constraint_start_by_window(granularity, count)
```
tlento, line 83: This is the actual fix.

metricflow/test/integration/test_cases/itest_cumulative_metric.yaml
```diff
@@ -6,7 +6,7 @@ integration_test:
   metrics: ["trailing_2_months_revenue"]
   group_bys: ["metric_time"]
   order_bys: ["metric_time"]
-  time_constraint: ["2020-01-05", "2021-01-04"]
+  time_constraint: ["2020-03-05", "2021-01-04"]
```
tlento, line 9: I left these tests in place for a couple of reasons, but one is to illustrate how the constraint adjustments need to be applied here vs in the where clause.

A number of these were producing correct results on inspection of input data only because the input data didn't start until within the too-short filter window.

metricflow/test/snapshots/test_cumulative_metrics.py/str/BigQuery/test_grain_to_date_cumulative_metric__query_output.txt
```diff
@@ -1,5 +1,5 @@
   metric_time__day  revenue_mtd
-0       2021-01-03         1000
-1       2021-01-04         2000
-2       2021-01-05         2000
-3       2021-01-06         2000
+0       2021-01-03         3000
```
tlento, line 2: These updated numbers are correct, while the old ones were missing inputs. See https://github.com/dbt-labs/metricflow/blob/main/metricflow/test/fixtures/source_table_snapshots/simple_model/fct_revenue.yaml#L15-L18

courtneyholcomb (approved):
Nice find 🙏 🙏  Thank you for fixing!!

PR #874, opened 2023-11-15
Organize DataflowPlanBuilder Recursive Case Handling

plypaul:
### Description

This PR separates the recursive handling of derived metrics into cases 
using different methods that are shorter / easier to understand:

* Build a node to compute a simple / base metric.
* Build a node to compute a derived metric.
* Build a node to compute a metric of any type.
* Build a node to compute many metrics.

In addition, this also moves MetricLookUp.measures_for_metric() into the
DataflowPlanBuilder because it seemed a little out of place in MetricLookUp
due to the dependence on the column association resolver.

This is mostly a shuffling of the code w/o logic changes.

### DIFF
--- a/metricflow/model/semantics/metric_lookup.py
+++ b/metricflow/model/semantics/metric_lookup.py
@@ -20,8 +20,6 @@ from metricflow.protocols.semantics import MetricAccessor
 from metricflow.specs.column_assoc import ColumnAssociationResolver
 from metricflow.specs.specs import (
     LinkableInstanceSpec,
-    MeasureSpec,
-    MetricInputMeasureSpec,
     MetricSpec,
 )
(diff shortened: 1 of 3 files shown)


metricflow/model/semantics/metric_lookup.py
```diff
@@ -123,35 +121,6 @@ def configured_input_measure_for_metric(self, metric_reference: MetricReference)
         else:
             assert_values_exhausted(metric.type)
 
-    def measures_for_metric(
```
tlento, line 126: Oh yeah I think this was used somewhere else, and isn't anymore. Agree it makes more sense in the one place it's used, although making the DataflowPlanBuilder longer is a bit of a bummer.

tlento (approved):
Seems reasonable!

PR #878, opened 2023-11-16
Test nested derived metric with multiple offsets

plypaul:
### Description

Test case for discussion.


courtneyholcomb:
I've added these test cases to a different PR, so closing this one.

PR #879, opened 2023-11-16
Consolidate Measure Time Spine Join Conditional Behavior Into `MetricInputMeasureSpec`

plypaul:
### Description

I found the `DataflowPlanBuilder` hard to follow in regards to when the measure would be joined to the time spine. There was conditional logic that depended on several parameters scattered throughout methods. This PR encapsulates the join condition in `MetricInputMeasureSpec`, and the consolidates filling out those fields when that spec is constructed. Once the behavior is encapsulated, the signature for methods `_build_aggregated_measure_from_measure_source_node()` can be simplified.

There's additional consolidation that can be done for cumulative measure parameters, but putting that off for now as consolidating this logic has been complex as it is. Also, there's a potential edge case with nested derived offset metrics that was found, but it is not addressed here.

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -52,7 +52,9 @@ from metricflow.plan_conversion.column_resolver import DunderColumnAssociationRe
 from metricflow.plan_conversion.node_processor import PreJoinNodeProcessor
 from metricflow.specs.column_assoc import ColumnAssociationResolver
 from metricflow.specs.specs import (
+    CumulativeMeasureDescription,
     InstanceSpecSet,
+    JoinToTimeSpineDescription,
     LinkableInstanceSpec,
     LinkableSpecSet,
     LinklessEntitySpec,
@@ -174,12 +176,19 @@ class DataflowPlanBuilder:
         """Builds a node to compute a metric that is not defined from other metrics."""
         metric_reference = metric_spec.reference
         metric = self._metric_lookup.get_metric(metric_reference)
-        metric_input_measure_specs = self._measures_for_metric(
+        metric_input_measure_spec = self._build_input_measure_spec_for_base_metric(
             metric_reference=metric_reference,
             column_association_resolver=self._column_association_resolver,
+            query_contains_metric_time=queried_linkable_specs.contains_metric_time,
+            child_metric_offset_window=metric_spec.offset_window,
+            child_metric_offset_to_grain=metric_spec.offset_to_grain,
+            culmination_description=CumulativeMeasureDescription(
+                cumulative_window=metric.type_params.window,
+                cumulative_grain_to_date=metric.type_params.grain_to_date,
+            )
+            if metric.type is MetricType.CUMULATIVE
+            else None,
         )
-        assert len(metric_input_measure_specs) == 1, "Simple and cumulative metrics must have one input measure."
-        metric_input_measure_spec = metric_input_measure_specs[0]
 
         logger.info(
             f"For {metric_spec}, needed measure is:\n"
@@ -190,19 +199,13 @@ class DataflowPlanBuilder:
             combined_where = (
                 combined_where.combine(metric_spec.constraint) if combined_where else metric_spec.constraint
             )
+
         aggregated_measures_node = self.build_aggregated_measure(
             metric_input_measure_spec=metric_input_measure_spec,
-            metric_spec=metric_spec,
             queried_linkable_specs=queried_linkable_specs,
             where_constraint=combined_where,
             time_range_constraint=time_range_constraint,
-            cumulative=metric.type == MetricType.CUMULATIVE,
-            cumulative_window=metric.type_params.window if metric.type == MetricType.CUMULATIVE else None,
-            cumulative_grain_to_date=(
-                metric.type_params.grain_to_date if metric.type == MetricType.CUMULATIVE else None
-            ),
         )
-
         return self.build_computed_metrics_node(
             metric_spec=metric_spec,
             aggregated_measures_node=aggregated_measures_node,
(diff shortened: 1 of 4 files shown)


metricflow/specs/specs.py
```diff
@@ -521,6 +532,20 @@ class LinkableSpecSet(Mergeable, SerializableDataclass):
     time_dimension_specs: Tuple[TimeDimensionSpec, ...] = ()
     entity_specs: Tuple[EntitySpec, ...] = ()
 
+    @property
+    def contains_metric_time(self) -> bool:
```
courtneyholcomb, line 536: Helpful thank you!!

metricflow/specs/specs.py
```diff
@@ -469,25 +472,33 @@ def reference(self) -> MetricReference:
         return MetricReference(element_name=self.element_name)
 
 
+@dataclass(frozen=True)
+class MeasureCulminationDescription:
```
courtneyholcomb, line 476: Did you mean `Cumulation` instead of `Culmination`? If not, why `Culmination`? 🤔

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -645,54 +677,87 @@ def build_computed_metrics_node(
...
         metric_reference: MetricReference,
         column_association_resolver: ColumnAssociationResolver,
-    ) -> Sequence[MetricInputMeasureSpec]:
-        """Return the measure specs required to compute the metric."""
+        child_metric_offset_window: Optional[MetricTimeWindow],
+        child_metric_offset_to_grain: Optional[TimeGranularity],
+        query_contains_metric_time: bool,
+        culmination_description: Optional[MeasureCulminationDescription],
+    ) -> MetricInputMeasureSpec:
+        """Return the measure specs required to compute the base metric."""
```
courtneyholcomb, line 689: Nit: update to singular `measure spec`

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -645,54 +677,87 @@ def build_computed_metrics_node(
             metric_specs=[metric_spec],
         )
 
-    def _measures_for_metric(
+    def _build_metric_input_measure_for_base_metric(
```
courtneyholcomb, line 680: Note that we're about to add conversion metrics, which will have multiple input metrics - so this might need to be rewritten by @WilliamDee

courtneyholcomb (approved):
Looks good to me!

metricflow/dataflow/builder/dataflow_plan_builder.py
WilliamDee: +1 I need this, so we should sync to see whats the best path forward

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -225,13 +228,42 @@ def _build_derived_metric_output_node(
             f"For {metric.type} metric: {metric_spec}, needed metrics are:\n"
             f"{pformat_big_objects(metric_input_specs=metric_input_specs)}"
         )
+
+        parent_nodes: List[BaseOutput] = []
+
+        for metric_input_spec in metric_input_specs:
+            # TODO: See: https://github.com/dbt-labs/metricflow/issues/881
+            if (metric_spec.offset_to_grain is not None or metric_spec.offset_to_grain is not None) and (
+                metric_input_spec.offset_window is not None or metric_input_spec.offset_to_grain is not None
```
tlento, line 237: This looks incorrect. Maybe this?
```suggestion
            if (metric_spec.offset_window is not None or metric_spec.offset_to_grain is not None) and (
                metric_input_spec.offset_window is not None or metric_input_spec.offset_to_grain is not None
```

I'm not sure it matters if @courtneyholcomb will remove this and fix up immediately afterwards but we might as well fix it in case something has to get reverted or deployed quickly.

metricflow/dataflow/builder/dataflow_plan_builder.py
tlento: we mean multiple input measures, not metrics, but yes.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -174,12 +176,19 @@ def _build_base_metric_output_node(
...
         metric_reference = metric_spec.reference
         metric = self._metric_lookup.get_metric(metric_reference)
-        metric_input_measure_specs = self._measures_for_metric(
+        metric_input_measure_spec = self._build_metric_input_measure_for_base_metric(
             metric_reference=metric_reference,
             column_association_resolver=self._column_association_resolver,
+            query_contains_metric_time=queried_linkable_specs.contains_metric_time,
+            child_metric_offset_window=metric_spec.offset_window,
+            child_metric_offset_to_grain=metric_spec.offset_to_grain,
+            culmination_description=MeasureCulminationDescription(
```
tlento, line 185: Agree with @courtneyholcomb, this name is odd. I'd go with:

`cumulative_description=CumulativeMeasureDescription`

Or else

`accumulation_description=MeasureAccumulationDescription`

tlento (commented):
There's what appears to be a bad conditional which may or may not matter if you coordinate merging with @courtneyholcomb's work on the nested offset metric fix, and the word `culmination` does not mean what it's referring to in the code.

metricflow/dataflow/builder/dataflow_plan_builder.py
WilliamDee: im not gonna lie, i always get so confused visually and acoustically. Having to say/see metric input measure instead of just input measure messes with my eyes and ears thinking we're talking about input metrics

metricflow/specs/specs.py
plypaul: I think I need to update the dictionary for my IDE:

image

But yeah, that's off.

metricflow/dataflow/builder/dataflow_plan_builder.py
plypaul: Updated.

metricflow/dataflow/builder/dataflow_plan_builder.py
plypaul: Updated.

metricflow/dataflow/builder/dataflow_plan_builder.py
plypaul: Updated to `input_measure`.

metricflow/dataflow/builder/dataflow_plan_builder.py
plypaul: Updated.

PR #880, opened 2023-11-16
Fix tutorial

WilliamDee:
### Description
With the addition of unique labels validation, the tutorial model fails on that. Update the labels to resolve that. Additionally, cleaned up the tutorial message to remove outdated info and added link to adding time spine model.


Jstein77 (approved):
Thank for updating this will!

PR #883, opened 2023-11-16
Add a Way to Compare `TimeDimensionSpec`s While Ignoring the Grain

plypaul:
### Description

This adds a key class for `TimeDimensionSpec` that can be used for comparing the specs while ignoring the grain. Useful for ambiguous group-by-item resolution where we want to match an ambiguous time dimension that the user has specified from a list of available time dimension specs.


courtneyholcomb (approved):
LGTM!

metricflow/specs/specs.py
```diff
@@ -292,6 +295,67 @@ def accept(self, visitor: InstanceSpecVisitor[VisitorOutputT]) -> VisitorOutputT
         return visitor.visit_dimension_spec(self)
 
 
+class TimeDimensionSpecField(Enum):
+    """Fields of the time dimension spec.
+
+    The value corresponds to the name of the field in the dataclass. This should contain all fields, but implementation
```
tlento, line 301: We might consider switching to `dataclasses.fields` or some other utility in the `dataclasses` module for this. Depends on what we're trying to do, I guess.

metricflow/specs/specs.py
```diff
@@ -292,6 +295,67 @@ def accept(self, visitor: InstanceSpecVisitor[VisitorOutputT]) -> VisitorOutputT
...
+            source_spec: The spec that this key is based on.
+            exclude_fields: The fields to ignore when determining equality.
+        """
+        self._excluded_fields = frozenset(exclude_fields)
+        self._source_spec = source_spec
+
+        # This is a list of field values of TimeDimensionSpec that we should use for comparison.
+        spec_field_values_for_comparison: List[
+            Union[str, Tuple[EntityReference, ...], TimeGranularity, Optional[DatePart]]
+        ] = [self._source_spec.element_name, self._source_spec.entity_links]
```
tlento, line 331: Would it be worth it to instead allow for a `unique_name and type` style equality comparison for all of these spec classes? I can see this being useful for more than just time dimension specs, although right now they're the only ones that have all of these extra properties.

tlento (approved):
Seems fine, might want a different approach, hard to say without seeing how (or if) it'll be used more broadly.

metricflow/specs/specs.py
plypaul: @tlento - mind elaborating? Would those be public properties?

tlento: They don't have to be public properties. I just mean doing a comparison based on the name/type of spec. Everything else we layer into specs as properties is transient information or something we need to propagate along, like annotations about aggregation state, aliases, stuff like that.

I did just realize that you're using the entire entity links, so my original suggestion doesn't work. That said, our comparison key is currently entity link path and name for linkable specs, and I suspect would be name for everything else. Will we need other properties? If we don't have specific things we'll need to add to that comparison then maybe we just add a method that does the simple thing.

metricflow/specs/specs.py
plypaul: It's a little unclear how those would be used since those are dynamic, and I was looking for something specific I can reference in code.

plypaul:
@courtneyholcomb - I change the name of the module `collections -> collection_helpers` here, but we can change it to something else as well.

PR #884, opened 2023-11-16
Bug fix: apply offset for nested derived metrics

### LINKED ISSUE #882: [Bug] Time offset doesn't work for nested derived metric

### Current Behavior

Currently, if you apply an `offset_window` or `offset_to_grain` to the outer metric of a nested derived metric, the offset will not be applied.
Example:
```
metric:
  name: "bookings_offset_once"
  description: bookings metric offset once.
  type: derived
  type_params:
    expr: 2 * bookings
    metrics:
      - name: bookings
        offset_window: 5 days
---
metric:
  name: "bookings_offset_twice"
  description: bookings metric offset twice.
  type: derived
  type_params:
    expr: 2 * bookings_offset_once
    metrics:
      - name: bookings_offset_once
        offset_window: 5 days
```
Querying `bookings_offset_twice` with `metric_time__day` produces this SQL:
```
SELECT
  metric_time__day
  , 2 * bookings_offset_once AS bookings_offset_twice
FROM (
  SELECT
    metric_time__day
    , 2 * bookings AS bookings_offset_once
  FROM (
    -- Join to Time Spine Dataset
    -- Pass Only Elements:
    --   ['bookings', 'metric_time__day']
    -- Aggregate Measures
    SELECT
      subq_12.ds AS metric_time__day
      , SUM(subq_10.bookings) AS bookings
    FROM ***************************.mf_time_spine subq_12
    INNER JOIN (
      -- Read Elements From Semantic Model 'bookings_source'
      -- Metric Time Dimension 'ds'
      SELECT
        DATE_TRUNC('day', ds) AS metric_time__day
        , 1 AS bookings
      FROM ***************************.fct_bookings bookings_source_src_10001
    ) subq_10
    ON
      subq_12.ds - INTERVAL 5 day = subq_10.metric_time__day
    GROUP BY
      subq_12.ds
  ) subq_16
) subq_17
```
Above you can see that only one offset is happening, when two would be expected.

### Expected Behavior

The SQL should have an offset for the outer metric (`bookings_offset_twice`):
```
SELECT
  subq_9.metric_time__day
  , 2 * bookings_offset_once AS bookings_offset_twice
FROM {{ source_schema }}.mf_time_spine subq_9
INNER JOIN (
  SELECT
    metric_time__day
    , 2 * bookings AS bookings_offset_once
  FROM (
    SELECT
      subq_3.ds AS metric_time__day
      , SUM(subq_1.bookings) AS bookings
    FROM {{ source_schema }}.mf_time_spine subq_3
    INNER JOIN (
      SELECT
        DATE_TRUNC('day', ds) AS metric_time__day
        , 1 AS bookings
      FROM {{ source_schema }}.fct_bookings bookings_source_src_1
    ) subq_1
    ON
      subq_3.ds - INTERVAL 5 day = subq_1.metric_time__day
    GROUP BY
      subq_3.ds
  ) subq_7
) subq_8
ON
  subq_9.ds - INTERVAL 5 day = subq_8.metric_time__day
```

### Steps To Reproduce

Add a nested derived metric where the outer metric has an offset, then query it with `metric_time`.

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
- metricflow:
```

_No response_

### Additional Context

_No response_

courtneyholcomb:
Resolves #882

### Description
Apply time offset for nested ratio or derived input metrics. Previously, we were only applying offset during measure aggregation. For these metrics, the offset needs to be applied post-metric aggregation.


metricflow/test/plan_conversion/test_dataflow_to_sql_plan.py
```diff
@@ -1035,3 +1036,56 @@ def test_dimensions_requiring_join(
         sql_client=sql_client,
         node=dataflow_plan.sink_output_nodes[0].parent_node,
     )
+
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.sql_engine_snapshot
+def test_nested_offsets(  # noqa: D
```
tlento, line 1104: Can we put these into the test_derived_metric_rendering.py module instead? I'd like to keep this one more strictly limited to simple component conversion rather than query rendering.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -234,14 +234,32 @@ def _build_derived_metric_output_node(
             ),
             metric_specs=[metric_spec],
         )
+        # For nested ratio / derived metrics with time offset, apply offset after metric computation.
+        join_to_time_spine_node: Optional[JoinToTimeSpineNode] = None
+        if metric_spec.offset_window or metric_spec.offset_to_grain:
```
tlento, line 239: What happens if there's a filter on metric_time?

Today we generally do:

1. join
2. filter
3. aggregate

All of this is in ComputeMetricsNode, which comes post-aggregation, so I think we'll end up joining the time spine against the filtered input set here, effectively truncating the input data on the second offset.

If I'm right about this, we'll need to propagate the offset information so we can expand that window. For now it might be adequate to apply the filter expression after the final calculation if there's a nested derived offset, and then we can optimize the predicate handling later.

tlento (commented):
I have a concern about where this new time spine join is happening. I'll also leave this in other people's review queues so @plypaul can take a look. He's been spending a lot of time in this derived metric stack, so he may spot things I'm missing.

tlento:
Also, #879 looks relevant.

metricflow/test/plan_conversion/test_dataflow_to_sql_plan.py
courtneyholcomb: Yes my bad!

metricflow/dataflow/builder/dataflow_plan_builder.py
courtneyholcomb: Moving this discussion over to [a new PR](https://github.com/dbt-labs/metricflow/pull/886), based off of one of Paul's branches

PR #885, opened 2023-11-16
Remove collections test module

courtneyholcomb:
### Description
Small thing for internal scripts! Adding this `collections` module to the `test` directory overrode our ability to use the built-in Python `collections` module from `test`. This broke at least  the ability to run `metricflow/test/generate_snapshots.py`, and likely any other scripts in the `test` directory. As far as I can tell, we don't need this to be a functioning module (nothing imports from it) so we can just remove the `__init__.py` to resolve this.


tlento (approved):
This seems fine, but maybe we should just rename the `collections` directory here and in the main metricflow sub-directory instead. Given what's in there `collections` is a strange name, but I don't know what @plypaul has planned in upcoming PRs.

courtneyholcomb:
> This seems fine, but maybe we should just rename the `collections` directory here and in the main metricflow sub-directory instead. Given what's in there `collections` is a strange name, but I don't know what @plypaul has planned in upcoming PRs.

I think the intent was to copy the same file structure that exists in `metricflow/` into `metricflow/test/` for the mocks! @plypaul do you have a different preference for how to resolve this? Happy to change the approach if so

tlento:
> I think the intent was to copy the same file structure that exists in metricflow/ into metricflow/test/ for the mocks! @plypaul do you have a different preference for how to resolve this? Happy to change the approach if so

Right, I'm suggesting renaming both directories to preserve the parallelism (and remove the possibility of ambiguity in `from collections import....` inside of MetricFlow proper). But naming is hard.

plypaul:
I didn't realize that there was a naming collision - in which case, we should definitely rename it. Let me put in a commit since it'll make it easier to manage my stack.

courtneyholcomb:
Closing this, @plypaul will put up a different fix.

PR #888, opened 2023-11-18
Add `SpecPattern` Classes to Model Ambiguous Group-By-Item Inputs

plypaul:
### Description

This PR implements the pattern classes for ambiguous group-by-items as described in https://github.com/dbt-labs/metricflow/pull/new/plypaul--58.7--spec-pattern.

### DIFF
--- a/metricflow/specs/patterns/entity_link_pattern.py
+++ b/metricflow/specs/patterns/entity_link_pattern.py
@@ -1,0 +1,104 @@
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from enum import Enum
+from typing import Any, List, Optional, Sequence, Tuple
+
+from dbt_semantic_interfaces.references import EntityReference
+from dbt_semantic_interfaces.type_enums import TimeGranularity
+from dbt_semantic_interfaces.type_enums.date_part import DatePart
+from more_itertools import is_sorted
+from typing_extensions import override
+
+from metricflow.specs.patterns.spec_pattern import SpecPattern
+from metricflow.specs.specs import InstanceSpec, InstanceSpecSet, LinkableInstanceSpec
+
+logger = logging.getLogger(__name__)
+
+
+class ParameterSetField(Enum):
+    """The fields of the EntityLinkPatternParameterSet class used for matching in the EntityLinkPattern.
+
+    Considering moving this to be a part of the specs module / classes.
+    """
+
+    ELEMENT_NAME = "element_name"
+    ENTITY_LINKS = "entity_links"
+    TIME_GRANULARITY = "time_granularity"
+    DATE_PART = "date_part"
+
+    def __lt__(self, other: Any) -> bool:  # type: ignore[misc]
+        """Allow for ordering so that a sequence of these can be consistently represented for test snapshots."""
+        if self.__class__ is other.__class__:
+            return self.value < other.value
+        return NotImplemented
+
+
+@dataclass(frozen=True)
+class EntityLinkPatternParameterSet:
+    """See EntityPathPattern for more details."""
+
+    # Specify the field values to compare. None can't be used to signal "don't compare" because sometimes a pattern
+    # needs to match a spec where the field is None. These should be sorted.
+    fields_to_compare: Tuple[ParameterSetField, ...]
+
+    # The name of the element in the semantic model
+    element_name: Optional[str] = None
+    # The entities used for joining semantic models.
+    entity_links: Optional[Tuple[EntityReference, ...]] = None
+    # Properties of time dimensions to match.
+    time_granularity: Optional[TimeGranularity] = None
+    date_part: Optional[DatePart] = None
+
+    @staticmethod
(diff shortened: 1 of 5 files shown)


metricflow/specs/patterns/spec_pattern.py
```diff
@@ -0,0 +1,22 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import Sequence
+
+from metricflow.specs.specs import LinkableInstanceSpec
+
+
+class SpecPattern(ABC):
```
tlento, line 9: nit: LinkableSpecPattern? It only works on LinkableInstanceSpecs.

metricflow/specs/patterns/entity_link_pattern.py
```diff
@@ -0,0 +1,174 @@
...
+
+        # Using some Python introspection magic to figure out specs that match the listed fields.
+        keys_to_check = set(field_to_compare.value for field_to_compare in self.parameter_set.fields_to_compare)
+        asdict(self.parameter_set)
+        # Checks that EntityLinkPatternParameterSetField is valid wrt to the parameter set.
+        parameter_set_values = tuple(getattr(self.parameter_set, key_to_check) for key_to_check in keys_to_check)
+
+        for spec in candidate_specs:
+            spec_values = tuple(
+                (getattr(spec, key_to_check) if hasattr(spec, key_to_check) else None) for key_to_check in keys_to_check
```
tlento, line 78: getattr [allows for a default return](https://docs.python.org/3.8/library/functions.html#getattr):

`getattr(spec, key_to_check, default=None) for key_to_check in keys_to_check`

metricflow/specs/patterns/entity_link_pattern.py
```diff
@@ -0,0 +1,174 @@
...
+        time_dimension_call_parameter_set: TimeDimensionCallParameterSet,
+    ) -> TimeDimensionPattern:
+        fields_to_compare: List[ParameterSetField] = [
+            ParameterSetField.ELEMENT_NAME,
+            ParameterSetField.ENTITY_LINKS,
+            ParameterSetField.DATE_PART,
+        ]
+
+        if time_dimension_call_parameter_set.time_granularity is not None:
+            fields_to_compare.append(ParameterSetField.TIME_GRANULARITY)
```
tlento, line 138: It's a little puzzling to me why one of these is using an is not None check but not the other. Is there any reason not to just compare both of these every time? Why do we care if the granularity is set for the purposes of the `match` call?

It appears the reason is so we can use None as a value for "any time granularity" here. Regardless, adding a docstring to the method explaining this instead of suppressing lint would be great.

tlento (approved):
This seems reasonable. See inlines for small improvement requests.

Going forward, please do write out brief commit messages and PR descriptions explaining what is changing and why - pointing to issue links is more indirection for readers to wade through when they often just need a local summary.

metricflow/specs/patterns/entity_link_pattern.py
plypaul: Yeah, that's the case to follow the behavior that's defined for how the time grain / date part are used in the object builder naming scheme. Added a comment.

metricflow/specs/patterns/spec_pattern.py
plypaul: Actually, this was updated later in the stack to apply for any spec. Moved those changes into this commit.

metricflow/specs/specs.py
```diff
@@ -697,6 +700,24 @@ def from_specs(specs: Sequence[LinkableInstanceSpec]) -> LinkableSpecSet:  # noq
...
+
+    Similar to InstanceSpecSetTransform, but for LinkableSpecSets. TBD: There's a lot of similarity between
+    InstanceSpecSet and LinkableSpecSet, so there's probably a better class structure so that we don't need to duplicate
+    this.
+    """
+
+    @abstractmethod
+    def transform(self, spec_set: LinkableSpecSet) -> LinkableSpecSetTransformOutputT:  # noqa: D
+        pass
+
```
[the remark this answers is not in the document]
plypaul: Removing this interface as later in the stack, this turned out to not be necessary.

metricflow/specs/patterns/entity_link_pattern.py
```diff
@@ -0,0 +1,174 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
plypaul: Expanded the commit message, but also left the link in there as I thought it would still be useful. It's now:

```
Add SpecPattern classes.

These pattern classes are used to model user inputs when group-by items are
specified through the query interface or specified in a filter. These patterns
allow for ambiguous user inputs of group-by items e.g. a time dimension with
an unknown grain. For that case, the ambiguous input makes it easier for the
user to author queries as figuring out the time grain requires inspection of
the configs. For more details, please see:

https://github.com/dbt-labs/metricflow/issues/887
```

PR #893, opened 2023-11-21
Add Naming Schemes to Represent Different Input Formats

plypaul:
### Description
These naming scheme classes will be used in the query parser to convert string inputs into patterns. The patterns will be used later to resolve ambiguous group-by-items.

### DIFF
--- a/metricflow/naming/dunder_scheme.py
+++ b/metricflow/naming/dunder_scheme.py
@@ -1,0 +1,169 @@
+from __future__ import annotations
+
+import re
+from typing import Optional, Sequence, Tuple
+
+from dbt_semantic_interfaces.naming.keywords import DUNDER
+from dbt_semantic_interfaces.references import EntityReference
+from dbt_semantic_interfaces.type_enums import TimeGranularity
+from dbt_semantic_interfaces.type_enums.date_part import DatePart
+from typing_extensions import override
+
+from metricflow.naming.naming_scheme import QueryItemNamingScheme
+from metricflow.specs.patterns.entity_link_pattern import (
+    EntityLinkPattern,
+    EntityLinkPatternParameterSet,
+    ParameterSetField,
+)
+from metricflow.specs.specs import (
+    InstanceSpec,
+    InstanceSpecSet,
+    InstanceSpecSetTransform,
+)
+
+
+class DunderNamingScheme(QueryItemNamingScheme):
+    """A naming scheme using the dundered name syntax.
+
+    TODO: Consolidate with StructuredLinkableSpecName / DunderedNameFormatter.
+    """
+
+    _INPUT_REGEX = re.compile(r"\A[a-z]([a-z0-9_])*[a-z0-9]\Z")
+
+    @staticmethod
+    def date_part_suffix(date_part: DatePart) -> str:
+        """Suffix used for names with a date_part."""
+        return f"extract_{date_part.value}"
+
+    @override
+    def input_str(self, instance_spec: InstanceSpec) -> Optional[str]:
+        spec_set = InstanceSpecSet.from_specs((instance_spec,))
+
+        for time_dimension_spec in spec_set.time_dimension_specs:
+            # From existing comment in StructuredLinkableSpecName:
+            #
+            # Dunder syntax not supported for querying date_part
+            #
+            if time_dimension_spec.date_part is not None:
+                return None
+        names = _DunderNameTransform().transform(spec_set)
+        if len(names) != 1:
+            raise RuntimeError(f"Did not get 1 name for {instance_spec}. Got {names}")
+
+        return names[0]
+
+    @override
+    def spec_pattern(self, input_str: str) -> EntityLinkPattern:
+        if not self.input_str_follows_scheme(input_str):
+            raise ValueError(f"{repr(input_str)} does not follow this scheme.")
+
+        input_str = input_str.lower()
+
+        input_str_parts = input_str.split(DUNDER)
+        fields_to_compare: Tuple[ParameterSetField, ...] = (
+            ParameterSetField.ELEMENT_NAME,
+            ParameterSetField.ENTITY_LINKS,
+            ParameterSetField.DATE_PART,
+        )
+
+        time_grain = None
+
+        # No dunder, e.g. "ds"
+        if len(input_str_parts) == 1:
+            return EntityLinkPattern(
+                parameter_set=EntityLinkPatternParameterSet.from_parameters(
+                    element_name=input_str_parts[0],
+                    entity_links=(),
+                    time_granularity=time_grain,
+                    date_part=None,
+                    fields_to_compare=tuple(fields_to_compare),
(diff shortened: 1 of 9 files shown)


metricflow/naming/dunder_scheme.py
```diff
@@ -0,0 +1,169 @@
...
+        names = _DunderNameTransform().transform(LinkableSpecSet.from_specs((instance_spec,)))
+        if len(names) != 1:
+            raise RuntimeError(f"Did not get 1 name for {instance_spec}. Got {names}")
+
+        return names[0]
+
+    @override
+    def spec_pattern(self, input_str: str) -> EntityLinkPattern:
+        if not self.input_str_follows_scheme(input_str):
+            raise RuntimeError(f"{repr(input_str)} does not follow this scheme.")
```
tlento, line 58: What exception gets thrown in this case today? I assume this is almost always going to be a query input error, which means we may want to throw an exception we can handle and re-cast for things like alert management.

If we don't have a custom exception we can go to here I'd tag this as ValueError since it's an invalid value. RuntimeError is really broad. Also, we use ValueError below.

metricflow/naming/object_builder_scheme.py
```diff
@@ -0,0 +1,200 @@
...
+
+        if len(names) != 1:
+            raise RuntimeError(f"Did not get exactly 1 name from {instance_spec}. Got {names}")
+
+        return names[0]
+
+    @override
+    def spec_pattern(self, input_str: str) -> SpecPattern:
+        if not self.input_str_follows_scheme(input_str):
+            raise ValueError(
```
tlento, line 47: Nice, ValueError. Same question about a custom exception applies, and we should be consistent here.

metricflow/naming/naming_scheme.py
```diff
@@ -0,0 +1,40 @@
...
+        """
+        pass
+
+    @abstractmethod
+    def spec_pattern(self, input_str: str) -> SpecPattern:
+        """Given an input that follows this scheme, return a spec pattern that matches the described input."""
+        pass
+
+    @abstractmethod
+    def input_str_follows_scheme(self, input_str: str) -> bool:
```
tlento, line 35: It might be worth adding an enforcing version of this that's implemented to raise a consistent error, since mismatches are likely to all share the same root cause and error type/response messaging, something like:

```
def assert_input_str_follows_scheme(self, input_str: str) -> None:
    if not self.input_str_follows_scheme(input_str):
        raise ....
```

Then the implementations can just call the assert method when they need it instead of handling the exception info itself.

metricflow/naming/object_builder_scheme.py
```diff
@@ -0,0 +1,200 @@
...
+                        ParameterSetField.ENTITY_LINKS,
+                    ),
+                )
+            )
+
+        raise RuntimeError("There should have been a return associated with one of the CallParameterSets.")
+
+    @override
+    def input_str_follows_scheme(self, input_str: str) -> bool:
+        if ObjectBuilderNamingScheme._NAME_REGEX.match(input_str) is None:
```
tlento, line 118: Do we need this regex, or can we jump to the parsing? It's not clear to me what the regex is adding, honestly.

metricflow/naming/object_builder_scheme.py
```diff
@@ -0,0 +1,200 @@
...
+        if time_granularity is not None:
+            initializer_parameters.append(
+                f"'{time_granularity.value}'",
+            )
+        if date_part is not None:
+            initializer_parameters.append(f"date_part_name={repr(date_part.value)}")
+        if len(entity_link_names) > 1:
+            initializer_parameters.append(f"entity_path={repr(entity_link_names[:-1])}")
+
+        return ", ".join(initializer_parameters)
```
tlento, line 165: I feel like there has to be a better way to do this but without overhauling our spec property interfaces I don't know what it is. The logic here is a little brittle because of the assumption that all the parameters passed are correct for the spec type but since it's all in this one small class I think it's fine.

metricflow/naming/object_builder_scheme.py
```diff
@@ -0,0 +1,200 @@
...
+                    date_part=time_dimension_call_parameter_set.date_part,
+                    fields_to_compare=tuple(fields_to_compare),
+                )
+            )
+
+        for entity_call_parameter_set in call_parameter_sets.entity_call_parameter_sets:
+            return EntityLinkPattern(
+                EntityLinkPatternParameterSet(
+                    element_name=entity_call_parameter_set.entity_reference.element_name,
+                    entity_links=entity_call_parameter_set.entity_path,
```
tlento, line 104: Tangentially related - do we actually allow linked entities, like `user__listing`, in a query? Or is this forward-looking for things we might add, or for using this for internal cases where we may need that entity path resolution in an intermediate query step?

metricflow/test/naming/test_dunder_naming_scheme.py
```diff
@@ -0,0 +1,112 @@
...
+from metricflow.specs.specs import DimensionSpec, EntitySpec, LinkableInstanceSpec, TimeDimensionSpec
+from metricflow.test.time.metric_time_dimension import MTD_SPEC_MONTH, MTD_SPEC_WEEK, MTD_SPEC_YEAR
+
+
+@pytest.fixture(scope="session")
+def dunder_naming_scheme() -> DunderNamingScheme:  # noqa: D
+    return DunderNamingScheme()
+
+
+def test_input_str(dunder_naming_scheme: DunderNamingScheme) -> None:  # noqa: D
```
tlento, line 20: nit: I'm not a fan of tests that do multi-step assertions against different logic triggers, I'd rather they be parameterized on input/output or else split into separate methods - that way when I inevitably change something that breaks a bunch of stuff I don't have to go through this `run test, fix bug one, run test, fix bug 2, run test, fix bug 3` workflow because all of the errors are exposed on the first run.

tlento (approved):
Seems reasonable! I left a bunch of minor comments. At the very least, please do clean up the error types before merging.

metricflow/naming/dunder_scheme.py
```diff
@@ -0,0 +1,169 @@
...
+    def input_str(self, instance_spec: LinkableInstanceSpec) -> Optional[str]:
+        spec_set = LinkableSpecSet.from_specs((instance_spec,))
+
+        for time_dimension_spec in spec_set.time_dimension_specs:
+            # From existing comment in StructuredLinkableSpecName:
+            #
+            # Dunder syntax not supported for querying date_part
+            #
+            if time_dimension_spec.date_part is not None:
+                return None
```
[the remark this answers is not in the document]
tlento: I had an idea a while back but it pushed too much stuff into the common interface. We'll come up with something.

metricflow/naming/object_builder_scheme.py
tlento: Really? TIL. I don't think I've ever seen it.

PR #894, opened 2023-11-21
Add a Way to Print Object + Fields in a Compact and Pretty Format

plypaul:
### Description

This is needed to aid debugging via logs and to make tests easier to write by allowing for easy generation of descriptive snapshots. The previous method `pformat_big_objects()` was too verbose with the paths that needed to be displayed as a part of the ambiguous group-by-item resolution feature.


courtneyholcomb (approved):

metricflow/collection_helpers/pretty_print.py
```diff
@@ -0,0 +1,451 @@
...
+            A string representation of the sequence like (1,) or [1, 2].
+        """
+        if isinstance(list_like_obj, list):
+            left_enclose_str = "["
+            right_enclose_str = "]"
+        elif isinstance(list_like_obj, tuple):
+            left_enclose_str = "("
+            right_enclose_str = ")"
+        else:
+            raise RuntimeError(f"Unhandled type: {type(list_like_obj)}")
```
tlento, line 58: This should be unreachable although it's fine to have it in case someone decides to override the typechecker.

Relatedly, sets are a thing, so we might consider adding support for those. They're represented as `"{1, 2, 3}"`

plypaul:
> how much value does this custom class provide us over just using Python's prettyprinter for the internal rendering? It feels like we're doing an awful lot of work to manage indentation and junk like that.

Yeah, I had the same thought. If I had known that the implementation would require this much work beforehand / we were using a later version of Python, I probably would not have done this. The ideal state is to use that or some other library. On the plus side, it'll be easy to replace later.

PR #911, opened 2023-11-29
Add `DatePart` Support to `ValidLinkableSpecResolver`

plypaul:
### Description

`ValidLinkableSpecResolver` is used to list specs that are available for querying metrics. Since we added `date_part` to `TimeDimensionSpec`s, this PR updates `ValidLinkableSpecResolver` to include them as well.


metricflow/model/semantics/linkable_spec_resolver.py
```diff
@@ -476,7 +504,19 @@ def __init__(
                     assert_values_exhausted(metric.type)
 
             self._metric_to_linkable_element_sets[metric.name] = linkable_sets_for_measure
-        logger.info(f"Building the [metric -> valid linkable element] index took: {time.time() - start_time:.2f}s")
+
+        # If no metrics are specified, the query interface supports distinct dimension values from a single semantic
+        # model.
```
tlento, line 509: Is this true? Because build_plan_for_distinct_values doesn't behave as if this is true. For that matter, neither does this code. Maybe we just remove this comment?

metricflow/model/semantics/linkable_spec_resolver.py
```diff
@@ -627,13 +669,24 @@ def _get_metric_time_elements(self, measure_reference: MeasureReference) -> Link
...
-            path_key_to_linkable_dimensions={
-                ElementPathKey(
+
+        path_key_to_linkable_dimensions: Dict[ElementPathKey, List[LinkableDimension]] = defaultdict(list)
+        for time_granularity in possible_metric_time_granularities:
+            possible_date_parts: Sequence[Optional[DatePart]] = (
+                # No date part, just the metric time at a different grain.
+                (None,)
+                # date part of a metric time at a different grain.
+                + tuple(date_part for date_part in DatePart if time_granularity.to_int() <= date_part.to_int())
```
tlento, line 679: I see this granularity/date_part comparison all over the place and it's confusing every time. Maybe we should add a method to DatePart that does this and document the mechanics of the comparison there.

metricflow/model/semantics/linkable_spec_resolver.py
```diff
@@ -627,13 +669,24 @@ def _get_metric_time_elements(self, measure_reference: MeasureReference) -> Link
...
+            )
```
tlento, line 680: nit: I find this hard to read because it looks like a tuple of tuples and that leading `+` gets elided a bit by the commentary. Maybe it's a "Tom, you need reading glasses" problem, but there it is. There are lots of ways to restructure this so it's easier to see. Here's one:

```
# No date part (i.e., just metric time) plus all date parts for metric time at a different grain
possible_date_parts = (None,) + tuple([date_part for date_part....])
```

In theory the typechecker should infer that correctly but if it doesn't it's more contained and easier to reorient. Note you can also keep the multi-line commentary by using locals and just doing the assignment at the end, whatever works.

tlento (approved):
Seems reasonable! Minor suggestions inline.

metricflow/model/semantics/linkable_spec_resolver.py
plypaul: Actually, it does:

https://github.com/dbt-labs/metricflow/blob/868748ae7f16ddf2479594b7ea43d9a33759c4e4/metricflow/test/query_rendering/test_query_rendering.py#L196

metricflow/model/semantics/linkable_spec_resolver.py
plypaul: Will update before merging.

PR #912, opened 2023-11-29
Add Group-By-Item Resolver

plypaul:
### Description

This PR adds the `GroupByItemResolver` and associated classes to resolve the spec for an ambiguous group-by-item in a query or in a filter. The resolved specs can be later used in `MetricFlowQuerySpec`.


metricflow/query/group_by_item/candidate_push_down/push_down_visitor.py
```diff
@@ -0,0 +1,408 @@
...
+
+        e.g.
+
+            def visit_some_node(self, node) -> Result:
+                with track_node_visit(node) as current_traversal_path:
+                    # current_traversal_path contain the nodes visited from the start.
+        """
+        self._current_path.append(node)
+        yield MetricFlowQueryResolutionPath(tuple(self._current_path))
+        self._current_path.pop(-1)
```
tlento, line 109: Maybe add a bit to the docblock to illustrate the need for this last line, because on its own it looks like this doesn't maintain state. Something like:

```
def visit_some_node(self, node) -> Result:
    with track_node_visit(node as current_traversal_path):
        # current_traversal_path contain the nodes visited before this
        # do things

    # context manager removes this node from current_traversal_path so sibling branches don't include it
```

metricflow/query/group_by_item/candidate_push_down/push_down_visitor.py
```diff
@@ -0,0 +1,408 @@
...
+
+
+class DagTraversalPathTracker:
+    """Helps track the path traveled by a visitor through the nodes in a group-by-tem resolution DAG."""
+
+    def __init__(self) -> None:  # noqa: D
+        self._current_path: List[GroupByItemResolutionNode] = []
+
+    @contextmanager
+    def track_node_visit(self, node: GroupByItemResolutionNode) -> Iterator[MetricFlowQueryResolutionPath]:
```
tlento, line 98: Can we make track_node_visit a decorator since we use it everywhere, or does the argument passing get dodgy?

We might also want to bake this into the push down operator class itself somehow, since it seems like every node type is going to require this tracking. I'm not sure if that's possible, though. Maybe the manual context manager is better then, it's more obvious if someone forgets to use it.

metricflow/query/group_by_item/candidate_push_down/group_by_item_candidate.py
```diff
@@ -0,0 +1,97 @@
...
+        common_specs = set.intersection(*specs_as_sets)
+        if len(common_specs) == 0:
+            return GroupByItemCandidateSet.empty_instance()
+
+        measure_paths = tuple(
+            itertools.chain.from_iterable(candidate_set.measure_paths for candidate_set in candidate_sets)
+        )
+
+        return GroupByItemCandidateSet(
+            specs=tuple(common_specs), measure_paths=measure_paths, path_from_leaf_node=path_from_leaf_node
```
tlento, line 56: So common_specs removes specs from the candidate input through the intersection, so there may be measure_paths that are no longer relevant to the output set. Just to make sure - we want to track the global measure_paths in all cases, correct?

Also I'm assuming order doesn't matter - like we want the measure_paths to preserve the chained order regardless of how the intersection re-orders the common_specs.

metricflow/query/group_by_item/candidate_push_down/group_by_item_candidate.py
```diff
@@ -0,0 +1,97 @@
...
+    This candidate set is refined as it is passed from the root node (representing a measure) to the leaf node
+    (representing a query) in the resolution DAG.
+    """
+
+    specs: Tuple[LinkableInstanceSpec, ...]
+    # The path from the leaf node to the measure node where the specs originated. There can be multiple paths as
+    # when candidate sets are merged, there can be multiple measures were a spec could have come from.
+    measure_paths: Tuple[MetricFlowQueryResolutionPath, ...]
+    # The path from the leaf node to the node where this set was generated.
+    path_from_leaf_node: MetricFlowQueryResolutionPath
```
tlento, line 29: Two things:

1. I think these inline comments are better placed in the class docblock for a host of reasons
2. It would be nice to have an indication of what these last two properties are used for - if it's error reporting, that warrants a different level of scrutiny (and possibly a separate container class) than if these interact with the valid spec set we're generating in these push down operations

metricflow/query/group_by_item/candidate_push_down/push_down_visitor.py
```diff
@@ -0,0 +1,408 @@
...
+            # For metrics with offset_to_grain, don't allow date_part group-by-items
+            patterns_to_apply: Sequence[SpecPattern] = ()
+            if metric.type is MetricType.SIMPLE or metric.type is MetricType.CUMULATIVE:
+                pass
+            elif metric.type is MetricType.RATIO or metric.type is MetricType.DERIVED:
+                for input_metric in metric.input_metrics:
+                    if input_metric.offset_to_grain:
+                        # From comment in previous query parser:
+                        # "Cannot extract date part for metrics with offset_to_grain."
+                        patterns_to_apply = (NoneDatePartPattern(),)
```
tlento, line 300: @courtneyholcomb this is relevant to your interests. :)

metricflow/query/group_by_item/candidate_push_down/push_down_visitor.py
```diff
@@ -0,0 +1,408 @@
...
+                    + indent_log_line(mf_pformat(matched_specs))
+                )
+            )
+
+            # There were candidates that were common from the ones passed from parents, but after applying the filters,
+            # none of the candidates were valid.
+            if len(matched_specs) == 0:
+                issue_sets_to_merge.append(
+                    MetricFlowQueryResolutionIssueSet.from_issue(
+                        MetricExcludesDatePartIssue.from_parameters(
```
tlento, line 338: Ok, this is a little weird, because all of the intervening logic is general to any arbitrary collection of patterns but this issue is highly specific to date part support.

I don't think we need to fix this now - and it might be tricky to do this cleanly - but it would be nice to align the error handling with the intervening logic. Otherwise, someone will add a pattern and end up with a date part issue even when no date part is in the group by, and that'll be confusing.

metricflow/query/group_by_item/candidate_push_down/push_down_visitor.py
```diff
@@ -0,0 +1,408 @@
...
+                    without_any_of=self._without_any_property,
+                )
+            )
+
+            # The following is needed to handle limitation of cumulative metrics. Filtering could be done at the measure
+            # node, but doing it here makes it a little easier to generate the error message.
+            metric = self._semantic_manifest_lookup.metric_lookup.get_metric(node.child_metric_reference)
+
+            patterns_to_apply: Tuple[SpecPattern, ...] = ()
+            if metric.type is MetricType.SIMPLE or metric.type is MetricType.RATIO or metric.type is MetricType.DERIVED:
```
tlento, line 161: Should DERIVED be an error? Or will that propagate from the leaf to here?

metricflow/query/group_by_item/candidate_push_down/push_down_visitor.py
```diff
@@ -0,0 +1,408 @@
...
+
+            if len(patterns_to_apply) == 0:
+                return PushDownResult(
+                    candidate_set=GroupByItemCandidateSet(
+                        specs=tuple(candidate_specs),
+                        measure_paths=merged_result_from_parents.candidate_set.measure_paths,
+                        path_from_leaf_node=current_traversal_path,
+                    ),
+                    issue_set=MetricFlowQueryResolutionIssueSet.merge_iterable(issue_sets_to_merge),
+                )
```
tlento, line 316: This is fine here, but we don't actually need it, right?

The cumulative metrics logic in the measure visitor doesn't include the early return, so we'll get those "after applying patterns" log entries for all measures but not for all metrics. Does that matter?

tlento (approved):
It's a little hard to evaluate without the live use context from the next couple of PRs, but so far this looks great.

I've left a big pile of comments about documentation and naming just to streamline the explanation of this logic. The change itself is clear and well-structured, but it's inherently implementing a complex operation so those little bits and pieces will help people understand it.

There are also a couple of bigger questions that might or might not warrant changes. However, there's nothing here that needs to be done at this point in the stack, so feel free to fix up whatever's easy and push whatever else needs doing to the top of the stack.

metricflow/query/group_by_item/candidate_push_down/push_down_visitor.py
plypaul: It was originally here to cut down on the log output, but I think it's better to remove it for consistency in log output.

metricflow/query/group_by_item/candidate_push_down/push_down_visitor.py
plypaul: Yeah, this could be improved, but I couldn't think of a great solution in time. Should be addressed though.

metricflow/query/group_by_item/group_by_item_resolver.py
```diff
@@ -0,0 +1,198 @@
...
+
+        return GroupByItemResolution(spec=push_down_result.candidate_set.specs[0], issue_set=push_down_result.issue_set)
+
+    def resolve_available_items(
+        self,
+        resolution_node: Optional[ResolutionDagSinkNode] = None,
+    ) -> AvailableGroupByItemsResolution:
+        """Return all available group-by-items at a given node.
+
+        By default, the query node is used, so this will return the available group-by-items that can be in a query.
```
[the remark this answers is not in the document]
plypaul: Updated.

PR #918, opened 2023-12-05
Bug Fix: `Unable To Satisfy Query Error` with Cumulative Metrics in Saved Queries

### LINKED ISSUE #917: [SL-1344] [Bug] `Unable To Satisfy Query Error` with Cumulative Metrics in Saved Queries

### Current Behavior

Running a saved query with a cumulative metric throws an error:

```
ERROR: Unable To Satisfy Query Error: Recipe not found for linkable specs: LinkableSpecSet(dimension_specs=(), time_dimension_specs=(), entity_specs=())
```

See: https://getdbt.slack.com/archives/C046L0VTVR6/p1701434102038599

### Expected Behavior

Saved queries with cumulative metrics should not throw that error.

### Steps To Reproduce

See: https://getdbt.slack.com/archives/C046L0VTVR6/p1701434102038599

### Relevant log output

_No response_

### Environment

```markdown
- metricflow: 0.203.1
```

_No response_

### Additional Context

_No response_

<sub>[SL-1344](https://linear.app/dbt-labs/issue/SL-1344/[bug]-unable-to-satisfy-query-error-with-cumulative-metrics-in-saved)</sub>

plypaul:
Resolves #917

### Description

This PR resolves a bug with cumulative metrics in saved queries. There is logic to adjust the time constraint for cumulative metrics, but for saved queries, the adjustment is handled incorrectly, resulting empty parameters getting passed around.


metricflow/specs/specs.py
```diff
@@ -719,6 +719,19 @@ def linkable_specs(self) -> LinkableSpecSet:  # noqa: D
             entity_specs=self.entity_specs,
         )
 
+    def with_time_range_constraint(self, time_range_constraint: Optional[TimeRangeConstraint]) -> MetricFlowQuerySpec:
```
courtneyholcomb, line 722: Helpful 👍

metricflow/engine/metricflow_engine.py
```diff
@@ -320,13 +320,20 @@ def __init__(
         sql_client: SqlClient,
         time_source: TimeSource = ServerTimeSource(),
         column_association_resolver: Optional[ColumnAssociationResolver] = None,
+        enable_default_time_constraint: bool = True,
```
courtneyholcomb, line 323: Will this be expected for customers if we default to an adjusted time constraint? They might expect all time instead, in which case we should default to False. Definitely something we should make sure the docs reflect correctly.

courtneyholcomb (approved):

metricflow/engine/metricflow_engine.py
plypaul: Actually, after looking at the old code, it seems like the adjustment feature was broken from the start, and it doesn't work correctly with derived metrics. I'm thinking that we remove it and add it as a query validation instead. Let me handle that in a follow up PR though.

PR #914, opened 2023-11-30
Resolve Ambiguous Group-By-Items in Queries

plypaul:
### Description

This PR updates MF to resolve ambiguous group-by-items in queries by incorporating the foundation that was set up in prior PRs:

* Update the `DataflowPlanBuilder` to use a lookup to resolve ambiguous group-by-items in filters.
* Map existing input types (strings that specify group-by-items in different ways, `*QueryParameter` classes) into a set of common intermediate input classes (`ResolverInput*`) that contain spec patters.
* Implement a query resolver that resolves the patterns in input classes into concrete specs.
* Rewrite `MetricFlowQueryParser` to only parse the current inputs into the common intermediate input classes, and use the query resolver to validate inputs and resolve the query items into concrete specs.

In addition, the way that query validation errors were collected and reported has been changed. Insted of raising an exception when an error is encountered, issues are collected and returned together. This makes resolving those errors easier for the user as all errors can be resolved in a single pass vs. fixing one error, re-running the validation to find the next error, etc.

PR #926, opened 2023-12-05
Allow querying `metric_time` without metrics

### LINKED ISSUE #928: [SL-1376] [Feature] Enable querying `metric_time` without metrics

### Describe the feature

Allow dimension-only queries to include `metric_time`. Since `metric_time` only exists in relation to metrics, this feature will query the time spine table and use a cross join if other dimensions are queried.

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

<sub>[SL-1376](https://linear.app/dbt-labs/issue/SL-1376/[feature]-enable-querying-metric-time-without-metrics)</sub>

courtneyholcomb:
Resolves #928
Resolves SL-1098

### Description
Allow dimension-only queries to include `metric_time`. Since `metric_time` only exists in relation to metrics, this feature will query the time spine table. If queried with other dimensions, it will use a cross join since no entities are included in the time spine table.
Also adds the ability to use a `TimeRangeConstraint` in distinct values queries. This previously wasn't possible because `TimeRangeConstraint` requires `metric_time`, which was not available in distinct values queries.


tlento (commented):
I think @plypaul should take a look here, as the core of this logic is operating on sections that have been completely overhauled in #914 so I'm not sure how best to merge the two changes.

courtneyholcomb:
I resolved all the issues between the 2 PRs - would be great if one of you could take a look again! @tlento @plypaul 🙏

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -787,24 +789,34 @@ def _find_dataflow_recipe(
...
+            # In that case, we'll use the time spine table to represent `metric_time` values.
+            requested_metric_time_specs = [
+                time_dimension_spec
+                for time_dimension_spec in linkable_spec_set.time_dimension_specs
+                if time_dimension_spec.element_name == self._metric_time_dimension_reference.element_name
+            ]
+            if requested_metric_time_specs:
+                # Add time_spine source node to potential source nodes
+                potential_source_nodes = list(potential_source_nodes) + [self._time_spine_source_node]
+                use_time_spine_source_node = True
```
tlento, line 816: This "tag a boolean local and then use it down the line" thing feels really brittle. I suspect we need a little refactor in this method to account for it properly, or maybe we should push this selection down into the node processor.

A smaller improvement would be to have the boolean just be an optional pointer to the time spine source node and then use it if set, something like:

```
input_time_spine_source_node: Optional[ReadSqlSourceNode] = None  # or whatever type it is

....

    if requested_metric_time_specs:
        ...
        input_time_spine_source_node = self._time_spine_source_node
```

metricflow/dataflow/builder/node_evaluator.py
```diff
@@ -283,10 +298,24 @@ def _find_joinable_candidate_nodes_that_can_satisfy_linkable_specs(
...
+                candidates_for_join.append(
+                    JoinLinkableInstancesRecipe(
+                        node_to_join=right_node,
+                        join_on_entity=None,
+                        satisfiable_linkable_specs=list(needed_metric_time_specs),
+                        join_on_partition_dimensions=(),
+                        join_on_partition_time_dimensions=(),
+                        join_type=SqlJoinType.CROSS_JOIN,
+                    )
+                )
```
tlento, line 317: I feel like this belongs at the top of the loop, and with a `continue` here to indicate that this is a short-circuit.

Should we ever have the time spine where it's valid to go through the rest of the logic in this loop over the set of possible `right nodes`? If not let's move this up and make the special case around time spine handling more obvious.

tlento (commented):
Sorry, I know this is has been sitting waiting for review for a LONG time, and this feedback should've come much sooner, but I'm having a hard time reasoning about what's happening in here and the first two commits aren't helping me much. There's just a lot of movement in this PR and it's not clear why all of it is needed.

It seems like this is basically doing the following:

1. Add a time spine source node that just exists in the DataflowPlanBuilder
2. Thread that through everywhere so we can access it for metric_time joins when other entities or metrics are not present
3. Do some light refactoring and updating of associated functions

If we could see those three classes of changes broken out the review process here would be a lot cleaner, but I understand at this point that might be hard.

Separately, there's a larger question here:

Are we assuming the metric_time dimension will not carry linkable specs, and by extension the time spine will not either? We are talking about doing things like making metric_time more like an entity, and I wonder if the special case selection we're doing around this time spine join can reasonably support that.

Finally, I left some inline suggestions. I haven't thoroughly read all of the pieces of this PR, but those should help streamline the logic for a second pass. The most important is actually about the join rendering, I don't know what is supposed to happen to this if the join links to an SCD somehow.

metricflow/time/time_granularity_solver.py
```diff
@@ -0,0 +1,220 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
courtneyholcomb: Whoops, bad merge!

courtneyholcomb:
> Are we `assuming` the metric_time dimension will not carry linkable specs, and by extension the time spine will not either?
@tlento I think that should only be true when no metrics are specified with `metric_time`, since that's the case when we would have no way to determine what's linkable. But if metrics are specified, I don't see why `metric_time` can't be treated as linkable.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -715,18 +717,18 @@ def _select_source_nodes_with_measures(
 
     def _select_read_nodes_with_linkable_specs(
         self, linkable_specs: LinkableSpecSet, read_nodes: Sequence[ReadSqlSourceNode]
-    ) -> Dict[BaseOutput, Set[LinkableInstanceSpec]]:
+    ) -> List[ReadSqlSourceNode]:
         """Find source nodes with requested linkable specs and no measures."""
-        nodes_to_linkable_specs: Dict[BaseOutput, Set[LinkableInstanceSpec]] = {}
-        linkable_specs_set = set(linkable_specs.as_tuple)
+        selected_nodes: List[ReadSqlSourceNode] = []
```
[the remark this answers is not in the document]
courtneyholcomb: Ah good catch, I'll make it a set to dedupe!

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -402,32 +397,43 @@ def visit_join_to_base_output_node(self, node: JoinToBaseOutputNode) -> SqlDataS
             right_data_set: SqlDataSet = right_node_to_join.accept(self)
             right_data_set_alias = self._next_unique_table_alias()
 
-            sql_join_descs.append(
-                SqlQueryPlanJoinBuilder.make_base_output_join_description(
+            if join_description.join_type == SqlJoinType.CROSS_JOIN:
```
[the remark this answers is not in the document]
courtneyholcomb: Good call! Updated.

metricflow/dataflow/builder/dataflow_plan_builder.py
courtneyholcomb: Will do!

metricflow/dataflow/builder/node_evaluator.py
courtneyholcomb: Will do!

metricflow/dataflow/builder/node_evaluator.py
```diff
@@ -69,14 +71,22 @@ class JoinLinkableInstancesRecipe:
...
+    def __post_init__(self) -> None:  # noqa: D
+        if self.join_on_entity is None and self.join_type != SqlJoinType.CROSS_JOIN:
+            raise RuntimeError("`join_on_entity` is required unless using CROSS JOIN.")
+
     @property
     def join_description(self) -> JoinDescription:
         """The recipe as a join description to use in the dataflow plan node."""
         # Figure out what elements to filter from the joined node.
         include_specs: List[LinkableInstanceSpec] = []
-        assert all([len(spec.entity_links) > 0 for spec in self.satisfiable_linkable_specs])
```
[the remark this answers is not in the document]
courtneyholcomb: Updated!

metricflow/dataflow/builder/node_evaluator.py
```diff
@@ -231,9 +247,8 @@ def _find_joinable_candidate_nodes_that_can_satisfy_linkable_specs(
 
                 satisfiable_linkable_specs = []
                 for needed_linkable_spec in needed_linkable_specs:
-                    assert (
-                        len(needed_linkable_spec.entity_links) != 0
-                    ), f"Invalid needed linkable spec passed in {needed_linkable_spec}"
+                    if len(needed_linkable_spec.entity_links) == 0:
+                        continue
```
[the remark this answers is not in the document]
courtneyholcomb: Updated!

courtneyholcomb:
@tlento addressed all comments - LMK if you have any other feedback!

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -73,7 +73,7 @@
     ColumnEqualityDescription,
     SqlQueryPlanJoinBuilder,
 )
-from metricflow.plan_conversion.time_spine import TimeSpineSource
+from metricflow.plan_conversion.time_spine import TIME_SPINE_DATA_SET_DESCRIPTION, TimeSpineSource
```
tlento, line 76: Now I'm reading this file! Have to stop for the night though, will pick up tomorrow.

metricflow/plan_conversion/sql_join_builder.py
```diff
@@ -158,30 +158,34 @@ def make_base_output_join_description(
...
-        # We have the columns that we need to "join on" in the query, so add it to the list of join descriptions to
-        # use later.
         column_equality_descriptions = []
-        for idx in range(len(left_data_set_entity_cols)):
-            column_equality_descriptions.append(
-                ColumnEqualityDescription(
-                    left_column_alias=left_data_set_entity_cols[idx],
-                    right_column_alias=right_data_set_entity_cols[idx],
-                )
+        if join_on_entity:
```
tlento, line 162: Wow. That just works? I'm not sure if I should be impressed or disappointed. 🤣 

(I looked, it does just work.)

metricflow/test/generate_snapshots.py
```diff
@@ -73,7 +73,7 @@ class MetricFlowTestCredentialSetForAllEngines(FrozenBaseModel):  # noqa: D
     big_query: MetricFlowTestCredentialSet
     databricks: MetricFlowTestCredentialSet
     postgres: MetricFlowTestCredentialSet
-    trino: MetricFlowTestCredentialSet
+    # trino: MetricFlowTestCredentialSet
```
tlento, line 76: Please revert this and regenerate the Trino things. I think it should all be working now.

metricflow/test/generate_snapshots.py
```diff
@@ -102,10 +102,10 @@ def as_configurations(self) -> Sequence[MetricFlowTestConfiguration]:  # noqa: D
...
                 credential_set=self.postgres,
             ),
-            MetricFlowTestConfiguration(
-                engine=SqlEngine.TRINO,
-                credential_set=self.trino,
-            ),
+            # MetricFlowTestConfiguration(
+            #     engine=SqlEngine.TRINO,
+            #     credential_set=self.trino,
+            # ),
```
tlento, line 108: These too, of course.

metricflow/test/plan_conversion/test_dataflow_to_sql_plan.py
```diff
@@ -1126,3 +1127,111 @@ def test_dimension_with_joined_where_constraint(
...
+
+@pytest.mark.sql_engine_snapshot
+def test_metric_time_only(
+    request: FixtureRequest,
+    mf_test_session_state: MetricFlowTestSessionState,
+    dataflow_plan_builder: DataflowPlanBuilder,
+    dataflow_to_sql_converter: DataflowToSqlQueryPlanConverter,
+    sql_client: SqlClient,
+) -> None:
+    """Tests querying only metric time."""
```
tlento, line 1140: nit: do you mind moving these to a new module in the query_rendering directory? I know it's a little annoying but these are really query rendering tests, and I'd like to keep this file from becoming the accumulator of all the SQL snapshot tests again.

metricflow/test/integration/test_cases/itest_dimensions.yaml
```diff
@@ -188,6 +188,83 @@ integration_test:
...
+integration_test:
+  name: metric_time_only
+  description: Query metric_time alone
+  model: SIMPLE_MODEL
+  group_bys: ["metric_time"]
+  check_query: |
+    SELECT
+      {{ render_date_trunc("ds", TimeGranularity.DAY) }} AS metric_time__day
+    FROM {{ source_schema }}.mf_time_spine
+    GROUP BY metric_time__day
```
tlento, line 200: There's a good chance Trino won't like this. You can either wrap the aliasing in a subquery or substitute positional numbers for the group by, so either this:

```
SELECT metric_time__day
FROM (
  SELECT {{ render_date_trunc.... }} AS metric_time__day
  FROM ...
) a
GROUP BY metric_time__day
```

Or this:
```suggestion
    GROUP BY 1
```

metricflow/test/integration/test_cases/itest_dimensions.yaml
```diff
@@ -188,6 +188,83 @@ integration_test:
...
+integration_test:
+  name: metric_time_week
+  description: Query metric_time alone with non-default granularity
+  model: SIMPLE_MODEL
+  group_bys: ["metric_time__week"]
+  check_query: |
+    SELECT
+      {{ render_date_trunc("ds", TimeGranularity.WEEK) }} AS metric_time__week
+    FROM {{ source_schema }}.mf_time_spine
+    GROUP BY metric_time__week
```
tlento, line 211: Same thing here. The Trino PR just used the numbers, FWIW. The subquery approach is marginally more future-proof.

metricflow/test/generate_snapshots.py
courtneyholcomb: Oops, don't think I meant to commit that one!

PR #932, opened 2023-12-07
Bug fix: incorrect time constraint for time offset metrics

### LINKED ISSUE #925: [SL-1357] [Bug] Time constraint should be removed in metric subquery for time offset metrics

### Current Behavior

In time offset queries, the time constraint is applied to both the time spine subquery and the metric subquery. Here's the current SQL:
```
SELECT
  metric_time__day
  , bookings_2_weeks_ago AS bookings_2_weeks_ago
FROM (
  -- Join to Time Spine Dataset
  -- Pass Only Elements:
  --   ['bookings', 'metric_time__day']
  -- Aggregate Measures
  SELECT
    subq_12.metric_time__day AS metric_time__day
    , SUM(subq_11.bookings) AS bookings_2_weeks_ago
  FROM (
    -- Date Spine
    SELECT
      ds AS metric_time__day
    FROM ***************************.mf_time_spine subq_13
    WHERE ds BETWEEN '2020-01-01' AND '2020-01-01'
  ) subq_12
  INNER JOIN (
    -- Read Elements From Semantic Model 'bookings_source'
    -- Metric Time Dimension 'ds'
    -- Constrain Time Range to [2020-01-01T00:00:00, 2020-01-01T00:00:00]
    SELECT
      DATE_TRUNC('day', ds) AS metric_time__day
      , 1 AS bookings
    FROM ***************************.fct_bookings bookings_source_src_10001
    WHERE DATE_TRUNC('day', ds) BETWEEN '2020-01-01' AND '2020-01-01'
  ) subq_11
  ON
    subq_12.metric_time__day - INTERVAL 14 day = subq_11.metric_time__day
  GROUP BY
    subq_12.metric_time__day
) subq_17
```

### Expected Behavior

Since the metric dates will be offset, the time constraint should not be applied in that subquery. The inner join to the time spine query will effectively apply that time constraint, instead.
```
SELECT
  metric_time__day
  , bookings_2_weeks_ago AS bookings_2_weeks_ago
FROM (
  -- Join to Time Spine Dataset
  -- Pass Only Elements:
  --   ['bookings', 'metric_time__day']
  -- Aggregate Measures
  SELECT
    subq_12.metric_time__day AS metric_time__day
    , SUM(subq_11.bookings) AS bookings_2_weeks_ago
  FROM (
    -- Date Spine
    SELECT
      ds AS metric_time__day
    FROM ***************************.mf_time_spine subq_13
    WHERE ds BETWEEN '2020-01-01' AND '2020-01-01'
  ) subq_12
  INNER JOIN (
    -- Read Elements From Semantic Model 'bookings_source'
    -- Metric Time Dimension 'ds'
    -- Constrain Time Range to [2020-01-01T00:00:00, 2020-01-01T00:00:00]
    SELECT
      DATE_TRUNC('day', ds) AS metric_time__day
      , 1 AS bookings
    FROM ***************************.fct_bookings bookings_source_src_10001
  ) subq_11
  ON
    subq_12.metric_time__day - INTERVAL 14 day = subq_11.metric_time__day
  GROUP BY
    subq_12.metric_time__day
) subq_17
```

### Steps To Reproduce

Query a time offset metric with a time constraint.

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
- metricflow:
```

_No response_

### Additional Context

_No response_

<sub>[SL-1357](https://linear.app/dbt-labs/issue/SL-1357/[bug]-time-constraint-should-be-removed-in-metric-subquery-for-time)</sub>

courtneyholcomb:
Resolves SL-1357
Resolves #925

### Description
Fixes time constraint issues with time offset metrics. The time constraint was being applied in several places:
1. In the time spine subquery
2. In the input metric subquery
3. After joining the two subqueries
4. In the outer query for the outer derived metric

This PR removes 2 and 4. 
2 is incorrect because the input metric subquery has not had the time offset applied yet, so the constrained values will be incorrect. Plus, one the input metric subquery is joined to the time spine subquery, the time constraint will effectively be applied since the queries use `INNER JOIN` on the time constraint column.
3 is not incorrect, it's just redundant.
4 may appear redundant for cases when all input metrics have offsets, but in the case when one input metric is offset and another is not, it is necessary.

### DIFF
--- a/metricflow/test/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/BigQuery/test_cumulative_time_offset_metric_with_time_constraint__plan0_optimized.sql
+++ b/metricflow/test/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/BigQuery/test_cumulative_time_offset_metric_with_time_constraint__plan0_optimized.sql
@@ -1,0 +1,42 @@
+-- Compute Metrics via Expressions
+SELECT
+  metric_time__day
+  , every_2_days_bookers_2_days_ago AS every_2_days_bookers_2_days_ago
+FROM (
+  -- Join to Time Spine Dataset
+  -- Pass Only Elements:
+  --   ['bookers', 'metric_time__day']
+  -- Constrain Time Range to [2019-12-19T00:00:00, 2020-01-02T00:00:00]
+  -- Aggregate Measures
+  -- Compute Metrics via Expressions
+  SELECT
+    subq_17.metric_time__day AS metric_time__day
+    , COUNT(DISTINCT subq_16.bookers) AS every_2_days_bookers_2_days_ago
+  FROM (
+    -- Date Spine
+    SELECT
+      ds AS metric_time__day
+    FROM ***************************.mf_time_spine subq_18
+    WHERE ds BETWEEN '2019-12-19' AND '2020-01-02'
+  ) subq_17
+  INNER JOIN (
+    -- Join Self Over Time Range
+    SELECT
+      subq_15.ds AS metric_time__day
+      , bookings_source_src_10001.guest_id AS bookers
+    FROM ***************************.mf_time_spine subq_15
+    INNER JOIN
+      ***************************.fct_bookings bookings_source_src_10001
+    ON
+      (
+        DATE_TRUNC(bookings_source_src_10001.ds, day) <= subq_15.ds
+      ) AND (
+        DATE_TRUNC(bookings_source_src_10001.ds, day) > DATE_SUB(CAST(subq_15.ds AS DATETIME), INTERVAL 2 day)
+      )
+  ) subq_16
+  ON
+    DATE_SUB(CAST(subq_17.metric_time__day AS DATETIME), INTERVAL 2 day) = subq_16.metric_time__day
+  WHERE subq_17.metric_time__day BETWEEN '2019-12-19' AND '2020-01-02'
+  GROUP BY
+    metric_time__day
+) subq_23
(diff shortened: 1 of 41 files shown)


WilliamDee (approved):

metricflow/test/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/BigQuery/test_cumulative_time_offset_metric_with_time_constraint__plan0_optimized.sql
```diff
@@ -0,0 +1,42 @@
...
+      ***************************.fct_bookings bookings_source_src_10001
+    ON
+      (
+        DATE_TRUNC(bookings_source_src_10001.ds, day) <= subq_15.ds
+      ) AND (
+        DATE_TRUNC(bookings_source_src_10001.ds, day) > DATE_SUB(CAST(subq_15.ds AS DATETIME), INTERVAL 2 day)
+      )
+  ) subq_16
+  ON
+    DATE_SUB(CAST(subq_17.metric_time__day AS DATETIME), INTERVAL 2 day) = subq_16.metric_time__day
```
tlento, line 38: This all makes my head hurt, but I think it is correct - we query everything with a 2 day expansion at the start window, then we offset everything by 2 days and use the INNER JOIN to constrain the output times. And also have a where clause for that, but that's fine.

tlento (commented):
Thanks for the fix!

I'm a little wary of these changes because the inlined special casing is hard to reason about. Looking at the snapshots it seems ok, but I think future readers will struggle.

That said, it seems like a lot of this is going to conflict with the approach @plypaul is using in his recent PRs, so some coordination seems in order there. In theory it should be straightforward to control when these time constraint filters get applied using the resolvers, but I don't know if all of the relevant handlers are built out for time constraints specifically.

PR #934, opened 2023-12-08
Remove Default Time Constraint for Queries with Cumulative Metrics

### LINKED ISSUE #917: [SL-1344] [Bug] `Unable To Satisfy Query Error` with Cumulative Metrics in Saved Queries

plypaul:
Resolves #917

### Description

A previous change (https://github.com/dbt-labs/metricflow/pull/918) updated the logic to apply a default time constraint for queries with cumulative metrics if one was not supplied. However, the fix revealed that the feature was not working for some time, possibly since the public release. I also realized that the original logic (and also the fix) is not correct with derived metrics that use cumulative metrics. There is future work to better handle time constraints / predicate push down, so removing this for now.

### DIFF
--- a/metricflow/test/cli/test_cli.py
+++ b/metricflow/test/cli/test_cli.py
@@ -271,7 +271,17 @@ def test_saved_query_with_cumulative_metric(  # noqa: D
     sql_client: SqlClient,
 ) -> None:
     resp = cli_runner.run(
-        query, args=["--saved-query", "saved_query_with_cumulative_metric", "--order", "metric_time__day"]
+        query,
+        args=[
+            "--saved-query",
+            "saved_query_with_cumulative_metric",
+            "--order",
+            "metric_time__day",
+            "--start-time",
+            "2020-01-01",
+            "--end-time",
+            "2020-01-01",
+        ],
     )
 
     assert_object_snapshot_equal(
--- a/.changes/unreleased/Fixes-20231207-170704.yaml
+++ b/.changes/unreleased/Fixes-20231207-170704.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Remove default time constraint for queries with cumulative metrics.
+time: 2023-12-07T17:07:04.040509-08:00
+custom:
+  Author: plypaul
+  Issue: "917"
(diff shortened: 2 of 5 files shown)


metricflow/test/cli/test_cli.py
```diff
@@ -82,7 +82,7 @@ def test_validate_configs(cli_context: CLIContext) -> None:
 
     This test is special, because the CLI bypasses the semantic manifest read into the CLIContext and
     validates the config files on disk. It's not entirely clear why we do this, so we should probably
-    figure that out and, if possible, stop doing it so we can have this test depend on an injectable
+    figure that out and, if possible, stop doing it so we can have this test depen on an injectable
```
tlento, line 85: Dude, put that d back where it was!

tlento (approved):
Seems reasonable assuming this doesn't change behavior from what it was before #918 beyond making saved queries work. It's kind of hard to tell with the back and forth.

metricflow/test/cli/test_cli.py
courtneyholcomb, line 85: this looks like it was an accident!

courtneyholcomb (approved):
Much cleaner! Thank you! Looks good except one spelling error

metricflow/test/cli/test_cli.py
plypaul: Yep, fixed.

PR #939, opened 2023-12-15
Add a Validation Rule for Duplicate Metrics in a Query

### LINKED ISSUE #943: [SL-1455] [Feature] Add a Query Validation Rule for Repeated Metrics in a Query

### Describe the feature

A query should not contain the same metric more than once as it would result in a duplicate column in the generated SQL.

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

<sub>[SL-1455](https://linear.app/dbt-labs/issue/SL-1455/[feature]-add-a-query-validation-rule-for-repeated-metrics-in-a-query)</sub>

plypaul:
Resolves #943 

### Description

This PR add a validation to check that queries do not contain the same metric more than once, as that would result in a duplicate column in the generated SQL.

### DIFF
--- a/.changes/unreleased/Features-20231218-121506.yaml
+++ b/.changes/unreleased/Features-20231218-121506.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Add a Query Validation Rule for Repeated Metrics in a Query
+time: 2023-12-18T12:15:06.224541-08:00
+custom:
+  Author: plypaul
+  Issue: "943"
--- a/metricflow/query/issues/parsing/duplicate_metric.py
+++ b/metricflow/query/issues/parsing/duplicate_metric.py
@@ -1,0 +1,50 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Sequence, Tuple
+
+from dbt_semantic_interfaces.references import MetricReference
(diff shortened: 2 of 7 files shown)


tlento (approved):
This is an awful lot of code to do what amounts to `assert len(set(metrics)) == len(metrics)`.  We are certainly working very hard to accommodate this use of the visitor pattern, so I suppose there's that. 

I'd very much prefer to dispense with yet another set of recursive DAG walks in favor of just validating the resolver output for correctness, seems like that'd be cleaner, simpler to build and update, and easier to reason about. I cannot come up with a use case where we'd need to do any kind of stateful dependency validation at this stage in the query processor.

That said, all of this..... exists, so I guess we may as well proceed and see if the hypothesis holds that this will really end up being the equivalent of `map(resolver.resolved_metrics, lambda x: validator.validate(x))`

plypaul:
> This is an awful lot of code to do what amounts to assert len(set(metrics)) == len(metrics)

As previously mentioned, the value of the query validation framework is not there since most of the intended rules for it were moved out. I decided to keep it around for a little bit in case new use cases pop up and if not, I was going to take it out. If you want to go ahead and do that early, go ahead.

PR #940, opened 2023-12-15
Fix Optimizer to Dedupe Cases Where a Metric Is Used More Than Once

plypaul:
# Resolves #

### Description

If there are multiple derived metrics in a query that use the same metric, it's
possible that the same metric is used multiple times. In such cases, the metric
should be deduped so that we don't get a collision in the colum name.

### DIFF
--- a/.changes/unreleased/Fixes-20231215-141744.yaml
+++ b/.changes/unreleased/Fixes-20231215-141744.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Optimizer Does Not Deduplicate Common Metrics
+time: 2023-12-15T14:17:44.7035-08:00
+custom:
+  Author: plypaul
+  Issue: "941"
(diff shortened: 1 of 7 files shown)


tlento (approved):
Seems reasonable, but I have a question here:

Out of interest, will the base (unoptimized) render still have this issue? Or will the original sql rendering always work because these things are always compartmentalized into separate subqueries, and the optimizer merge is what pushes them together and selects the same thing multiple times?

plypaul:
> Out of interest, will the base (unoptimized) render still have this issue? Or will the original sql rendering always work because these things are always compartmentalized into separate subqueries, and the optimizer merge is what pushes them together and selects the same thing multiple times?

That is correct - the unoptimized SQL won't have this issue because it would have been compartmentalized into subqueries. There is a related error for these column name collisions: https://github.com/dbt-labs/metricflow/pull/939

PR #949, opened 2023-12-19
Allow Partial Entity Paths in Group-By-Items

plypaul:
### Description

This PR allows for partial entity paths to be specified for group-by-items. e.g. `user__country` instead of `listing__user__country` if there is a unique shortest path available.

### DIFF
--- a/metricflow/query/issues/group_by_item_resolver/multiple_join_paths.py
+++ b/metricflow/query/issues/group_by_item_resolver/multiple_join_paths.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import List
+from typing import Sequence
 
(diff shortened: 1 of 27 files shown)


tlento (approved):
I agree this should work but the tests seem incomplete here. We effectively only test that we can find a longer path than the one specified given that the longer path is unique, because we cap join hops to 2.

I think we want one more test case. Specifically, we want to show we find the shortest path given two valid paths of different lengths (e.g., `entity_1__some_dimension` will match directly in some kind of `entity_1_entity_0_metric` instead of going through some mapping join like `entity_0__entity_1_some_dimension`).

Separately, this needs a rebase.

plypaul:
@tlento Added some additional cases if you want to take a look.

PR #961, opened 2024-01-05
Expand `MetricFlowEngine` Initializer Signature

plypaul:
Resolves #

### Description

This PR updates the initializer for `MetricFlowEngine` to allow a query parser to be passed in. A different query parser can be used to alter how inputs to the engine are handled.


courtneyholcomb (approved):
Looks good! THANK YOU for the quick turnaround! 🙏

PR #964, opened 2024-01-05
Add Decorator To Log Slow Functions

plypaul:
### Description

This PR adds a decorator that logs how long a function took and if it was slow. This decorator was also added to some functions of interest. This PR also includes some reorganization / renaming, so please view by commit.


courtneyholcomb (approved):
I like it 👍 👍

PR #977, opened 2024-01-13
Don't use the persistent source schema for Trino in `generate_snapshots.py`

plypaul:
### Description

Since Trino is run locally for testing, this PR configures the `generate_snapshots.py` script to not use the persistent source schema for that engine. Instead, it will load tables for each testing session.


courtneyholcomb (approved):

tlento:
Just FYI this was already fixed and merged on Friday so you can probably drop this one.

PR #978, opened 2024-01-13
Update metricflow package version to `0.204.0.dev8`

plypaul:
### Description

As per title.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "metricflow"
-version = "0.204.0.dev7"
+version = "0.204.0.dev8"
 description = "Translates a simple metric definition into reusable SQL and executes it against the SQL engine of your choice."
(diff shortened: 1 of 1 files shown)


courtneyholcomb (approved):

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "metricflow"
-version = "0.204.0.dev7"
+version = "0.204.0.dev8"
```
WilliamDee, line 7: `0.204.0` was just released, this should be `0.204.1.dev0`?

WilliamDee (changes requested):
comment on the version bump

pyproject.toml
tlento: +1

I've got a tiny change I'd like to make for an 0.204.1 after this as well so 0.204.1.dev0 or similar sounds good to me.

PR #982, opened 2024-01-18
Add a make target to run Trino warehouse locally

courtneyholcomb:
### Description
Quick dev improvement. Adds a make target to run a Trino warehouse locally. This makes it so you can run `make trino` instead of needing to remember the docker command to run it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,10 @@ lint:
 postgresql postgres:
 	make -C local-data-warehouses postgresql
 
+.PHONY: trino
+trino:
+	docker run --name trino -d -p 8080:8080 trinodb/trino
+
 # Re-generate test snapshots using all supported SQL engines.
 .PHONY: regenerate-test-snapshots
 regenerate-test-snapshots:
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ postgresql postgres:
 
 .PHONY: trino
 trino:
-	docker run --name trino -d -p 8080:8080 trinodb/trino
+	make -C local-data-warehouses trino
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -72,6 +72,10 @@ lint:
 postgresql postgres:
 	make -C local-data-warehouses postgresql
 
+.PHONY: trino
+trino:
+	docker run --name trino -d -p 8080:8080 trinodb/trino
```
tlento, line 77: Note: this fails if you've already run it once and haven't removed the existing installation. Or at least that happened to me when I tried to add this (long story, but note `trinodb/trino` and `trinodb/trinodb` are not the same thing and it took me like an hour to figure that one out).

Anyway, maybe it's worth adding a comment here to indicate that you can just start your existing docker hub install if you already have this working locally with the `trino` name.

I don't think Postgres has the same issue because of the docker file but I admit I can't be bothered to try it.

tlento (approved):
Ok! 

If you want to get fancy I suppose you could write a docker file so that we don't hit failures the second time you run the command.

Makefile
courtneyholcomb: Ended up adding the docker-compose file and its working!

local-data-warehouses/Makefile
```diff
@@ -5,3 +5,7 @@
 .PHONY: postgresql
 postgresql:
 	docker-compose -f postgresql/docker-compose.yaml up
+
+.PHONY: trino
+trino:
+	docker-compose -f trino/docker-compose.yaml up
```
tlento, line 11: YES! Thank you!

PR #983, opened 2024-01-18
Potential bug fix: Don't use `defaultdict` for `SemanticModelLookup` indexes

courtneyholcomb:
### Description
`SemanticModelLookup` keeps track of all the valid measures, dimensions, and entities for a given semantic manifest using dictionaries `_measure_index`, `_dimension_index`, `_entity_index`, and `_linkable_reference_index`. These are used for various purposes, one of which is to check if a given measure/dimension/entity name is valid (i.e., if someone queries a dimension name that's not in the index, we will raise an error telling them the dimension doesn't exist).

Previously, these were created as `defaultdicts`. This simplifies the code, but creates a potential bug. If someone queries `get_dimension()` with a dimension that doesn't exist, we will raise the appropriate error, but with an unintended side effect. When you index into a `defaultdict` using a key that doesn't exist yet, that key gets automatically added to the dict. Therefore, when calling `get_dimension()` with a dimension that doesn't exist, that non-existent dimension will be added to `_dimension_index`. This becomes a problem when you call `get_dimension_references()`, which returns a list of all keys in `_dimension_index`; you would end up with a non-existent dimension in the list of what should be all valid dimensions.

This PR converts all those `defaultdicts` into normal dicts to avoid that potential bug.

### DIFF
--- a/metricflow/model/semantics/semantic_model_lookup.py
+++ b/metricflow/model/semantics/semantic_model_lookup.py
@@ -52,7 +52,7 @@ class SemanticModelLookup(SemanticModelAccessor):
         model: SemanticManifest,
     ) -> None:
         self._model = model
-        self._measure_index: Dict[MeasureReference, List[SemanticModel]] = defaultdict(list)
+        self._measure_index: Dict[MeasureReference, List[SemanticModel]] = {}
         self._measure_aggs: Dict[
             MeasureReference, AggregationType
         ] = {}  # maps measures to their one consistent aggregation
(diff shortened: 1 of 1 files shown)


metricflow/model/semantics/semantic_model_lookup.py
```diff
@@ -202,7 +205,11 @@ def _add_semantic_model(self, semantic_model: SemanticModel) -> None:
 
         for measure in semantic_model.measures:
             self._measure_aggs[measure.reference] = measure.agg
-            self._measure_index[measure.reference].append(semantic_model)
+
+            semantic_models_for_measure = self._measure_index.get(measure.reference, [])
+            semantic_models_for_measure.append(semantic_model)
+            self._measure_index[measure.reference] = semantic_models_for_measure
```
WilliamDee, line 198:
```python
self._measure_index[measure.reference] = self._measure_index.get(measure.reference, []) + [semantic_model]
```
Just throwing this one out there as a possibility, pick whichever you'd prefer tho!

metricflow/model/semantics/semantic_model_lookup.py
```diff
@@ -92,34 +92,23 @@ def get_dimension_from_semantic_model(
 
     def get_dimension(self, dimension_reference: DimensionReference) -> Dimension:
         """Retrieves a full dimension object by name."""
-        for dimension_source in self._dimension_index[dimension_reference]:
-            dimension = SemanticModelLookup.get_dimension_from_semantic_model(
-                semantic_model=dimension_source, dimension_reference=dimension_reference
+        semantic_models = self._dimension_index.get(dimension_reference)
+        if not semantic_models:
```
WilliamDee, line 94: could this be `if semantic_models is None`? or are we expecting `[]` as a return value?

WilliamDee (approved):
crazy catch! left a few aesthetic suggestions, feel free to drop those.

metricflow/model/semantics/semantic_model_lookup.py
courtneyholcomb: I intentionally left it as a falseness check! I don't EXPECT `[]` as a return value but if for some reason that were to happen this would still handle that, so it seemed preferable.

metricflow/model/semantics/semantic_model_lookup.py
courtneyholcomb: good idea!

PR #995, opened 2024-01-25
Methods to look up `agg_time_dimensions`

courtneyholcomb:
### Description
Adds some methods that will be used to enable querying with `agg_time_dimension` instead of `metric_time`.
`MetricLookup.get_valid_agg_time_dimensions_for_metric()` will be used in validations, while `SemanticModelLookup.get_agg_time_dimension_specs_for_measure()` will be used in query building.

### DIFF
--- a/metricflow/model/semantics/linkable_spec_resolver.py
+++ b/metricflow/model/semantics/linkable_spec_resolver.py
@@ -26,7 +26,6 @@ from metricflow.errors.errors import UnknownMetricLinkingError
 from metricflow.mf_logging.pretty_print import mf_pformat
 from metricflow.model.semantics.linkable_element_properties import LinkableElementProperties
 from metricflow.model.semantics.semantic_model_join_evaluator import SemanticModelJoinEvaluator
-from metricflow.model.semantics.semantic_model_lookup import SemanticModelLookup
 from metricflow.protocols.semantics import SemanticModelAccessor
 from metricflow.specs.specs import (
     DEFAULT_TIME_GRANULARITY,
@@ -524,9 +523,7 @@ class ValidLinkableSpecResolver:
         linkable_element_sets_to_merge: List[LinkableElementSet] = []
 
         for semantic_model in semantic_manifest.semantic_models:
-            linkable_element_sets_to_merge.append(
-                ValidLinkableSpecResolver._get_elements_in_semantic_model(semantic_model)
-            )
+            linkable_element_sets_to_merge.append(self._get_elements_in_semantic_model(semantic_model))
 
         metric_time_elements_for_no_metrics = self._get_metric_time_elements(measure_reference=None)
         self._no_metric_linkable_element_set = LinkableElementSet.merge_by_path_key(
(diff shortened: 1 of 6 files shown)


metricflow/model/semantics/linkable_spec_resolver.py
```diff
@@ -524,9 +523,7 @@ def __init__(
         linkable_element_sets_to_merge: List[LinkableElementSet] = []
 
         for semantic_model in semantic_manifest.semantic_models:
-            linkable_element_sets_to_merge.append(
-                ValidLinkableSpecResolver._get_elements_in_semantic_model(semantic_model)
-            )
+            linkable_element_sets_to_merge.append(self._get_elements_in_semantic_model(semantic_model))
```
courtneyholcomb, line 526: Updated this to use `self` to avoid a circular import.

metricflow/model/semantics/linkable_spec_resolver.py
```diff
@@ -550,8 +547,7 @@ def _get_semantic_model_for_measure(self, measure_reference: MeasureReference) -
             )
         return semantic_models_where_measure_was_found[0]
 
-    @staticmethod
-    def _get_elements_in_semantic_model(semantic_model: SemanticModel) -> LinkableElementSet:
+    def _get_elements_in_semantic_model(self, semantic_model: SemanticModel) -> LinkableElementSet:
```
courtneyholcomb, line 550: See comment re: circular import

metricflow/model/semantics/linkable_spec_resolver.py
```diff
@@ -568,7 +564,7 @@ def _get_elements_in_semantic_model(semantic_model: SemanticModel) -> LinkableEl
                     properties=frozenset({LinkableElementProperties.LOCAL, LinkableElementProperties.ENTITY}),
                 )
             )
-            for entity_link in SemanticModelLookup.entity_links_for_local_elements(semantic_model):
+            for entity_link in self._semantic_model_lookup.entity_links_for_local_elements(semantic_model):
```
courtneyholcomb, line 567: See comment re: circular import

metricflow/model/semantics/linkable_spec_resolver.py
courtneyholcomb, line 526: Changes in this file were made to avoid a circular import.

metricflow/model/semantics/semantic_model_lookup.py
```diff
@@ -339,3 +338,37 @@ def get_element_spec_for_name(self, element_name: str) -> LinkableInstanceSpec:
...
+        semantic_models = self.get_semantic_models_for_measure(measure_reference)
+        assert (
+            len(semantic_models) == 1
+        ), f"Expected exactly one semantic model for measure {measure_reference}, but found semantic models {semantic_models}."
+        semantic_model = semantic_models[0]
+
+        entity_link = self.resolved_primary_entity(semantic_model)
+        assert (
+            entity_link is not None
+        ), f"Expected semantic model {semantic_model} to have a primary entity since is contains dimensions, but found none."
```
tlento, line 357: Ok, this is implicit, right? We know it contains a measure, which means it must have an agg_time_dimension, which means it has at least a time dimension.

Maybe update the message to `to have a primary entity since it has a measure requiring an agg_time_dimension`?

Regardless, micro-nit - `it` not `is`.

tlento (approved):
Seems reasonable! Can we add some test cases for the new methods before merge? The tests for these classes seem to live in `metricflow/test/model/test_semantic_model_container.py` so we should be able to add a few cases pretty easily.

PR #996, opened 2024-01-25
Bug fix: always use candidate join type when evaluating nodes to join

courtneyholcomb:
### Description
Resolves [a bug](https://dbtlabsmt.datadoghq.com/logs?query=service%3Ametricflow-server%20status%3A%28error%20OR%20info%20OR%20warn%29%20cross%20join%20&cols=host%2Cservice&event=AgAAAY1CURh7OqhiHgAAAAAAAAAYAAAAAEFZMUNVU1BaQUFCbDVWMXY5LWllSUFBMQAAACQAAAAAMDE4ZDQyNTEtMjQ0ZC00ZGNhLTg1YTUtMGM2ZDBiMjI0OTll&index=%2A&messageDisplay=inline&refresh_mode=sliding&stream_sort=desc&viz=stream&from_ts=1706210388237&to_ts=1706213988237&live=true) found during Tableau complaint storm. We should be using the join type specified for the join candidate here instead of the default join type. This caused an error when joining to the time spine table, since the default join type for distinct values queries is `FULL OUTER JOIN` but time spine joins require `CROSS JOIN`.

### DIFF
--- a/metricflow/dataflow/builder/node_evaluator.py
+++ b/metricflow/dataflow/builder/node_evaluator.py
@@ -339,7 +339,6 @@ class NodeEvaluatorForLinkableInstances:
     def _update_candidates_that_can_satisfy_linkable_specs(
         candidates_for_join: List[JoinLinkableInstancesRecipe],
         already_satisfisfied_linkable_specs: List[LinkableInstanceSpec],
-        join_type: SqlJoinType,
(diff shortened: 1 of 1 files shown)


WilliamDee (approved):
LGTM! Should probably run the engine tests?

PR #1005, opened 2024-01-26
Allow measures that join to time spine to use `agg_time_dim`

### LINKED ISSUE #1009: [SL-1619] [Feature] Allow measures that join to time spine to use `agg_time_dimension`

### Describe the feature

Some measures specify that they should be joined to time spine in the YAML. We only execute this join if metric time is queried, since that's the only time joining to time spine is allowed. Since we're relaxing that restriction, we now can execute the join if `agg_time_dimension` is queried.

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

<sub>[SL-1619](https://linear.app/dbt-labs/issue/SL-1619/[feature]-allow-measures-that-join-to-time-spine-to-use-agg-time)</sub>

courtneyholcomb:
Resolves #1009

### Description
Some measures specify that they should be joined to time spine in the YAML. We only execute this join if metric time is queried, since that's the only time joining to time spine is allowed. Since we're relaxing that restriction, we now can execute the join if `agg_time_dimension` is queried.


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -1103,15 +1103,25 @@ def _build_input_measure_spec_for_base_metric(
...
+            query_contains_agg_time_dimension = queried_linkable_specs.contains_metric_time
+            if not query_contains_agg_time_dimension:
+                valid_agg_time_dimensions = self._semantic_model_lookup.get_agg_time_dimension_specs_for_measure(
+                    measure_spec.reference
+                )
+                query_contains_agg_time_dimension = bool(
+                    set(queried_linkable_specs.time_dimension_specs).intersection(set(valid_agg_time_dimensions))
+                )
+
+            if query_contains_agg_time_dimension:
```
tlento, line 1119: Yeah if we have some kind of helper I think it'll clean up a lot of these conditionals as well, because it'll hide the wonky logic inside of the helper and this can become something like:

`if query_contains_agg_time_dimension(queried_linkable_specs, measure_spec):`

metricflow/test/integration/test_cases/itest_metrics.yaml
```diff
@@ -1194,6 +1194,26 @@ integration_test:
...
+  description: Test a simple query that joins to time spine and fills nulls, with agg_time_dimension
+  model: SIMPLE_MODEL
+  metrics: ["bookings_fill_nulls_with_0"]
+  group_by_objs: [{"name": "booking__ds__day"}]
+  check_query: |
+    SELECT
+      subq_5.ds AS booking__ds__day
+      , COALESCE(subq_3.bookings, 0) AS bookings_fill_nulls_with_0
+    FROM {{ source_schema }}.mf_time_spine subq_5
+    LEFT OUTER JOIN (
```
tlento, line 1208: Ok, so previously this would have just run the query but skipped the join, correct? That seems strange and incorrect to me.

tlento (commented):
Checking in before approving - are there any cases where this might result in an ambiguous plan?

I'm thinking of a situation where someone has multiple metrics, or some combination nested metric, and they group by this agg_time_dimension and also metric_time. I'm not even sure if that's a valid query ordinarily, honestly.

metricflow/dataflow/builder/dataflow_plan_builder.py
courtneyholcomb: Updated!

metricflow/test/integration/test_cases/itest_metrics.yaml
courtneyholcomb: The logic there was that if there's no time dimension to join to, we can't join to time spine anyway.
And then, if there is a time dimension that isn't agg time, should we time spine join to that dimension? Seems inconsistent if we're treating it like any other categorical dimension in all other ways.

courtneyholcomb:
> Checking in before approving - are there any cases where this might result in an ambiguous plan?
> 
> I'm thinking of a situation where someone has multiple metrics, or some combination nested metric, and they group by this agg_time_dimension and also metric_time. I'm not even sure if that's a valid query ordinarily, honestly.

@tlento This clause should be able to execute for as many measures as are requested, each join will happen in its own subquery. We're checking for the `agg_time_dimension` specific to the measure here. So we would join to time spine only for measures whose `agg_time_dimension` was queried (or if `metric_time` was queried).
Does that answer the question?

tlento (approved):
> This clause should be able to execute for as many measures as are requested, each join will happen in its own subquery. 

Duh, of course it does, thank you.

Also, in re-reading that comment, I think the right answer for people duplicating their requested time dimensions across metrics is "don't do that" - I'm sure I'm missing some super critical use case where that's a totally reasonable thing to do, but it doesn't sound reasonable when I read it out loud.

PR #1007, opened 2024-01-26
Enable querying time offset metrics with `agg_time_dimension`

### LINKED ISSUE #1006: [SL-1611] [Feature] Enable querying time offset metrics with their agg_time_dimension

### Describe the feature

Currently, time offset metrics can only be queried with `metric_time`. Since metric_time represents the `agg_time_dimensions` associated with the metric's measures, you should be able to replace `metric_time` with the `agg_time_dimension` and get the same behavior.
Note: if a metric has multiple input metrics with different `agg_time_dimensions`, `metric_time` will still be required.

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

<sub>[SL-1611](https://linear.app/dbt-labs/issue/SL-1611/[feature]-enable-querying-time-offset-metrics-with-their-agg-time)</sub>

courtneyholcomb:
Resolves #1006

### Description
Enable querying time offset metrics with their `agg_time_dimension` in place of `metric_time`.
Should only work for when all input metrics have the same `agg_time_dimension`.

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -515,7 +515,7 @@ class DataflowPlanBuilder:
                 )
             assert (
                 queried_agg_time_dimension_specs
-            ), "Joining to time spine requires querying with metric_timeor the appropriate agg_time_dimension."
+            ), "Joining to time spine requires querying with metric_time or the appropriate agg_time_dimension."
             output_node = JoinToTimeSpineNode(
                 parent_node=output_node,
                 requested_agg_time_dimension_specs=queried_agg_time_dimension_specs,
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -420,12 +420,14 @@ class DataflowPlanBuilder:
             query_contains_metric_time=queried_linkable_specs.contains_metric_time,
             child_metric_offset_window=metric_spec.offset_window,
             child_metric_offset_to_grain=metric_spec.offset_to_grain,
-            cumulative_description=CumulativeMeasureDescription(
-                cumulative_window=metric.type_params.window,
-                cumulative_grain_to_date=metric.type_params.grain_to_date,
-            )
-            if metric.type is MetricType.CUMULATIVE
-            else None,
+            cumulative_description=(
+                CumulativeMeasureDescription(
+                    cumulative_window=metric.type_params.window,
+                    cumulative_grain_to_date=metric.type_params.grain_to_date,
+                )
+                if metric.type is MetricType.CUMULATIVE
+                else None
+            ),
             descendent_filter_specs=metric_spec.filter_specs,
         )
 
@@ -490,9 +492,9 @@ class DataflowPlanBuilder:
                         offset_window=metric_input_spec.offset_window,
                         offset_to_grain=metric_input_spec.offset_to_grain,
                     ),
-                    queried_linkable_specs=queried_linkable_specs
-                    if not metric_spec.has_time_offset
-                    else required_linkable_specs,
+                    queried_linkable_specs=(
+                        queried_linkable_specs if not metric_spec.has_time_offset else required_linkable_specs
+                    ),
                     filter_spec_factory=filter_spec_factory,
                     time_range_constraint=time_range_constraint if not metric_spec.has_time_offset else None,
                 )
@@ -514,7 +516,7 @@ class DataflowPlanBuilder:
             output_node = JoinToTimeSpineNode(
                 parent_node=output_node,
                 requested_agg_time_dimension_specs=queried_agg_time_dimension_specs,
-                query_includes_metric_time=queried_linkable_specs.contains_metric_time,
+                use_custom_agg_time_dimension=not queried_linkable_specs.contains_metric_time,
                 time_range_constraint=time_range_constraint,
                 offset_window=metric_spec.offset_window,
                 offset_to_grain=metric_spec.offset_to_grain,
@@ -1154,12 +1156,14 @@ class DataflowPlanBuilder:
                 element_name=input_metric.name,
                 filter_specs=tuple(filter_specs),
                 alias=input_metric.alias,
-                offset_window=PydanticMetricTimeWindow(
-                    count=input_metric.offset_window.count,
-                    granularity=input_metric.offset_window.granularity,
-                )
-                if input_metric.offset_window
-                else None,
+                offset_window=(
+                    PydanticMetricTimeWindow(
+                        count=input_metric.offset_window.count,
+                        granularity=input_metric.offset_window.granularity,
+                    )
+                    if input_metric.offset_window
+                    else None
+                ),
                 offset_to_grain=input_metric.offset_to_grain,
             )
             input_metric_specs.append(spec)
@@ -1274,9 +1278,11 @@ class DataflowPlanBuilder:
             find_recipe_start_time = time.time()
             measure_recipe = self._find_dataflow_recipe(
                 measure_spec_properties=measure_properties,
-                time_range_constraint=(cumulative_metric_adjusted_time_constraint or time_range_constraint)
-                if not before_aggregation_time_spine_join_description
-                else None,
+                time_range_constraint=(
+                    (cumulative_metric_adjusted_time_constraint or time_range_constraint)
+                    if not before_aggregation_time_spine_join_description
+                    else None
+                ),
                 linkable_spec_set=required_linkable_specs,
             )
             logger.info(
@@ -1309,9 +1315,9 @@ class DataflowPlanBuilder:
                 time_dimension_spec_for_join=agg_time_dimension_spec_for_join,
                 window=cumulative_window,
                 grain_to_date=cumulative_grain_to_date,
-                time_range_constraint=time_range_constraint
-                if not before_aggregation_time_spine_join_description
-                else None,
+                time_range_constraint=(
+                    time_range_constraint if not before_aggregation_time_spine_join_description else None
+                ),
             )
 
         # If querying an offset metric, join to time spine before aggregation.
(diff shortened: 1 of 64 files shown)


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -505,12 +505,20 @@ def _build_derived_metric_output_node(
...
+                    metric_spec.reference
+                )
+                queried_agg_time_dimension_specs = list(
+                    set(queried_linkable_specs.time_dimension_specs).intersection(set(valid_agg_time_dimensions))
+                )
             assert (
-                queried_linkable_specs.contains_metric_time
-            ), "Joining to time spine requires querying with metric_time."
+                queried_agg_time_dimension_specs
+            ), "Joining to time spine requires querying with metric_timeor the appropriate agg_time_dimension."
```
tlento, line 518: nit `metric_time or`

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -787,9 +795,7 @@ def _build_measure_spec_properties(self, measure_specs: Sequence[MeasureSpec]) -
                 f"semantic models: {semantic_models}. This suggests the measure_specs were not correctly filtered."
             )
 
-        agg_time_dimension = agg_time_dimension = self._semantic_model_lookup.get_agg_time_dimension_for_measure(
```
tlento, line 790: LOL. I did not notice that.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1244,56 +1245,77 @@ def visit_join_to_time_spine_node(self, node: JoinToTimeSpineNode) -> SqlDataSet
...
-        ), "Can't join to time spine without metric time. Validations should have prevented this."
-        metric_time_dimension_column_name = self.column_association_resolver.resolve_spec(
-            metric_time_dimension_instance.spec
+            len(node.requested_agg_time_dimension_specs) > 0
+        ), "Must have at least one value in requested_agg_time_dimension_specs for JoinToTimeSpineNode."
+
+        # Determine if the time spine join should use metric_time or the agg_time_dimension (metric_time takes priority).
+        agg_time_dimension_for_join = node.requested_agg_time_dimension_specs[0]
+        for spec in node.requested_agg_time_dimension_specs[1:]:
+            if spec.element_name == METRIC_TIME_ELEMENT_NAME:
```
tlento, line 1255: This is no longer guaranteed to be the lowest granularity, right? Is that intentional? It seems like it doesn't matter, but it's not clear at this point if that's really the case or not.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1244,56 +1245,77 @@ def visit_join_to_time_spine_node(self, node: JoinToTimeSpineNode) -> SqlDataSet
...
-            metric_time_dimension_instance
-        ), "Can't join to time spine without metric time. Validations should have prevented this."
-        metric_time_dimension_column_name = self.column_association_resolver.resolve_spec(
-            metric_time_dimension_instance.spec
+            len(node.requested_agg_time_dimension_specs) > 0
+        ), "Must have at least one value in requested_agg_time_dimension_specs for JoinToTimeSpineNode."
+
+        # Determine if the time spine join should use metric_time or the agg_time_dimension (metric_time takes priority).
+        agg_time_dimension_for_join = node.requested_agg_time_dimension_specs[0]
+        for spec in node.requested_agg_time_dimension_specs[1:]:
```
tlento, line 1254: This is a little magical - how do we know the first one is always not-metric-time and the others are always metric time? What if the user just requested metric_time, wouldn't the first one be potentially lowest-granularity metric_time?

I think it'll be cleaner to add a property to the node that gives us back the spec we need here, then we can be intentional with the naming and establish what people should care about. Like if all we need is the name/link pattern of an agg time dimension spec we can even return just those things.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1244,56 +1245,77 @@ def visit_join_to_time_spine_node(self, node: JoinToTimeSpineNode) -> SqlDataSet
...
-                    instance.spec.time_granularity < metric_time_dimension_instance.spec.time_granularity
-                ):
-                    metric_time_dimension_instance = instance
         assert (
-            metric_time_dimension_instance
-        ), "Can't join to time spine without metric time. Validations should have prevented this."
-        metric_time_dimension_column_name = self.column_association_resolver.resolve_spec(
-            metric_time_dimension_instance.spec
+            len(node.requested_agg_time_dimension_specs) > 0
+        ), "Must have at least one value in requested_agg_time_dimension_specs for JoinToTimeSpineNode."
```
tlento, line 1250: Can we put this assertion in the JoinToTimeSpineNode? The return type is Sequence so we shouldn't be able to append to it, but do we ever update that property in other ways?

metricflow/query/validation_rules/metric_time_requirements.py
```diff
@@ -65,6 +46,18 @@ def _group_by_items_include_metric_time(self, query_resolver_input: ResolverInpu
 
         return False
 
+    def _group_by_items_include_agg_time_dimension(
+        self, query_resolver_input: ResolverInputForQuery, metric_reference: MetricReference
```
tlento, line 50: This change was made in the other PR too, right?

tlento (commented):
The main thing I'm seeing here is that this logic is kind of the same everywhere.

In this case, the two tests I think we'd like to have are:

1. Rendering the case where the metric uses multiple measures with the same agg_time_dimension (if that isn't already here, kind of hard to tell)
2. Throwing the error in the case where the metric uses mulitple agg_time_dimensions

I'm assuming case 2 should throw here as well.

If we're able to factor out the repeated logic we may be able to test the conditions that component handles directly and then keep our overall query rendering test suite a little more streamlined, but it's not clear how viable that is.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -505,12 +505,20 @@ def _build_derived_metric_output_node(
...
         # For nested ratio / derived metrics with time offset, apply offset & where constraint after metric computation.
         if metric_spec.has_time_offset:
+            queried_agg_time_dimension_specs = list(queried_linkable_specs.metric_time_specs)
+            if not queried_agg_time_dimension_specs:
+                valid_agg_time_dimensions = self._metric_lookup.get_valid_agg_time_dimensions_for_metric(
+                    metric_spec.reference
+                )
+                queried_agg_time_dimension_specs = list(
+                    set(queried_linkable_specs.time_dimension_specs).intersection(set(valid_agg_time_dimensions))
+                )
```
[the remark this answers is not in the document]
courtneyholcomb: Was planning to resolve this in merge conflict resolution once one of the PRs merges, but I like the idea of adding a property to make this cleaner 🙏🏼 (My initial PR had the logic consolidated, but I ended up duplicating when I split into smaller PRs.)

metricflow/plan_conversion/dataflow_to_sql.py
courtneyholcomb: We don't know if the first one is always not-metric-time, I just overlooked checking the first one 🤦 

Re: granularity - the purpose of this check is just to see if the user passed `metric_time` or `agg_time_dimension` - it doesn't care about the granularity, so it stops as soon as it finds ANY `metric_time`. The granularity check happens a few lines down when going through all time dimension instances in the parent data set.

But I agree with your point about more intentional naming - will clean that up!

metricflow/query/validation_rules/metric_time_requirements.py
courtneyholcomb: Right. Planning to resolve that in merge conflict resolution.

courtneyholcomb: Next time will probably base the branches off of each other instead... Didn't realize they still had overlap when I made that choice.

metricflow/dataflow/builder/dataflow_plan_builder.py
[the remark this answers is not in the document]
courtneyholcomb: Updated!

metricflow/plan_conversion/dataflow_to_sql.py
courtneyholcomb: The granularity check happens a few lines down!

metricflow/plan_conversion/dataflow_to_sql.py
courtneyholcomb: Alright, I added a new property to the node called `query_contains_metric_time` and removed this logic!

tlento (commented):
Submitting commentary for now, be back later. :)

metricflow/dataflow/dataflow_plan.py
```diff
@@ -673,6 +673,7 @@ def __init__(
         self,
         parent_node: BaseOutput,
         requested_agg_time_dimension_specs: List[TimeDimensionSpec],
+        query_includes_metric_time: bool,
```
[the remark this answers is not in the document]
courtneyholcomb: Updated!

courtneyholcomb:
> In this case, the two tests I think we'd like to have are:
> 
> Rendering the case where the metric uses multiple measures with the same agg_time_dimension (if that isn't already here, kind of hard to tell)
> Throwing the error in the case where the metric uses mulitple agg_time_dimensions

Added these tests! See the two latest commits.

metricflow/test/query/test_query_parser.py
```diff
@@ -584,3 +608,16 @@ def test_duplicate_metric_query(bookings_query_parser: MetricFlowQueryParser) ->
...
         bookings_query_parser.parse_and_validate_query()
+
+
+def test_offset_metric_with_diff_agg_time_dims_error() -> None:  # noqa: D
+    metrics_yaml_file = YamlConfigFile(filepath="inline_for_test_1", contents=METRICS_YAML)
+    revenue_yaml_file = YamlConfigFile(filepath="inline_for_test_1", contents=REVENUE_YAML)
+    query_parser = query_parser_from_yaml(
+        [EXAMPLE_PROJECT_CONFIGURATION_YAML_CONFIG_FILE, revenue_yaml_file, metrics_yaml_file]
+    )
+    with pytest.raises(InvalidQueryException, match="does not match any of the available group-by-items"):
```
tlento, line 619: Out of curiosity, does this raise a more specific metric-time related error somewhere? Might be a nice additional improvement.

tlento (approved):
Huzzah!

PR #1015, opened 2024-01-30
Add Support for Consistent SQL Query Generation

### LINKED ISSUE #1020: [SL-1642] [Feature] Add Support for Consistent SQL Query Generation

### Describe the feature

Currently, SQL generated by `MetricFlowEngine` is not completely consistent between identical queries because there are SQL elements (e.g. sub-query aliases) that are generated via a counter that carries state from query to query. To make use of query caching as provided by some SQL engines, the proposed feature would make the ID generation consistent between otherwise identical MF queries.

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

<sub>[SL-1642](https://linear.app/dbt-labs/issue/SL-1642/[feature]-add-support-for-consistent-sql-query-generation)</sub>

plypaul:
Resolves #1020

### Description
This PR includes updates to make the SQL generated by `MetricFlowEnigne` more consistent. Previously, they were not consistent because there were SQL elements (e.g. sub-query aliases) were generated via a counter that carried state from query to query. This PR adds facilities for resetting the ID generation counters, and also streamlines / simplifies how IDs are handled. Please view by commit.


tlento:
Nit: I think we should have a changelog entry for this since it was a community request. Also it affects output, however slightly.

metricflow/dag/id_prefix.py
```diff
@@ -9,9 +9,61 @@ class IdPrefix(Enum):
     TODO: Move all ID prefixes here.
     """
 
+    DATAFLOW_NODE_AGGREGATE_MEASURES_ID_PREFIX = "am"
```
tlento, line 12: This commit is a thing of beauty.

metricflow/dataset/convert_semantic_model.py
```diff
@@ -423,8 +424,8 @@ def create_sql_source_data_set(self, semantic_model: SemanticModel) -> SemanticM
         all_entity_instances: List[EntityInstance] = []
 
         all_select_columns: List[SqlSelectColumn] = []
-        from_source_alias = IdGeneratorRegistry.for_class(self.__class__).create_id(f"{semantic_model.name}_src")
-
+        # from_source_alias = IdGeneratorRegistry.for_class(self.__class__).create_id(f"{semantic_model.name}_src")
```
tlento, line 427: ???

metricflow/dag/id_prefix.py
```diff
@@ -67,3 +67,7 @@ class IdPrefix(Enum):
     EXEC_PLAN_PREFIX = "ep"
 
     MF_DAG = "mfd"
+
+    TIME_SPINE_SOURCE = "spine"
```
tlento, line 71: Why make this change? Wouldn't keeping it `time_spine_src` reduce snapshot thrash considerably?

metricflow/query/group_by_item/resolution_dag/dag.py
```diff
@@ -31,9 +31,7 @@ class GroupByItemResolutionDag(MetricFlowDag[GroupByItemResolutionNode]):
 
     def __init__(self, sink_node: ResolutionDagSinkNode) -> None:  # noqa: D
         super().__init__(
-            dag_id=DagId.from_str(
-                IdGeneratorRegistry.for_class(self.__class__).create_id(IdPrefix.GROUP_BY_ITEM_RESOLUTION_DAG.value)
-            ),
+            dag_id=DagId.from_str(PrefixIdGenerator.create_next_id(IdPrefix.GROUP_BY_ITEM_RESOLUTION_DAG.value)),
```
tlento: Is this right, or should it be `create_next_id(IdPrefix.GROUP_BY_ITEM_RESOLUTION_DAG)` instead?

metricflow/dag/id_prefix.py
```diff
@@ -67,3 +67,7 @@ class IdPrefix(Enum):
...
+    SEMANTIC_MODEL_SOURCE = "src"
```
tlento, line 72: Hm. This is really nondescript. It's also not used as a prefix.

metricflow/dataset/convert_semantic_model.py
```diff
@@ -423,8 +424,8 @@ def create_sql_source_data_set(self, semantic_model: SemanticModel) -> SemanticM
...
+        from_source_alias = PrefixIdGenerator.create_next_id(IdPrefix.SEMANTIC_MODEL_SOURCE).str_value
```
tlento: This is a big change since the constant now formats in `src` instead of `{semantic_model.name}_src`

metricflow/engine/metricflow_engine.py
```diff
@@ -376,6 +388,7 @@ def __init__(
     @log_call(module_name=__name__, telemetry_reporter=_telemetry_reporter)
     def query(self, mf_request: MetricFlowQueryRequest) -> MetricFlowQueryResult:  # noqa: D
         logger.info(f"Starting query request:\n{indent(mf_pformat(mf_request))}")
+        logger.info(f"Setting ID generation to start at: {MetricFlowEngine._QUERY_ID_START_VALUE}")
```
tlento, line 391: There's no reset call anywhere that I can find. Either remove the logging (and the constant, which is otherwise unused) or add the reset, whichever is appropriate.

tlento (commented):
For the most part the changes look reasonable apart from a few things that don't seem right (the enum.value call, the semantic_model.name removal, and the info log with no corresponding action).

Given the extent of the snapshot changes due to renames (several files will simply move to new locations and then get updated) I don't see how we can evaluate the output difference from them.

Do you have concrete examples you can point to/selectively add to this PR that illustrate the expected nature of these changes? It'd be nice to double-check things like "most prefixes aren't changing" and "semantic model names aren't really important" and "numbering is shifting"

plypaul (commented):
Actually, maybe it's simpler to see if I separated out the snapshot changes. I've added commits that revert all of them except the numbering change and we can take a look at the other changes in a later PR.

plypaul:
FYI - looking into the Trino issue.

metricflow/dag/id_prefix.py
```diff
@@ -1,14 +1,39 @@
...
+        Metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all
+        its bases
+
+    since Enum has a special metaclass.
     """
 
+    pass
+
+
+class StaticIdPrefix(IdPrefix, Enum, metaclass=EnumMetaClassHelper):
```
tlento, line 34: Oh I see.

The other option is to make IdPrefix a protocol, but we've been over the tradeoffs there and I think having it be a superclass makes sense even though enum-type subclasses are a bit weird.

tlento (approved):
Nice, the updated form of this is a lot easier to reason about in terms of the changes I'm seeing.

PR #1016, opened 2024-01-30
add exports to MFE SavedQuery object

WilliamDee:
### Description
When using the MFE, when we retrieve saved queries object they don't have the exports field as it doesn't exist in the MFE model object.


tlento (approved):
Well that's annoying, since MetricFlow is just a pass through for these configs, but I guess it can't be helped.

courtneyholcomb (approved):

PR #1024, opened 2024-02-05
Split `multi_hop_join_manifest` Into Separate Manifests

plypaul:
### Description

The test manifest / fixture `multi_hop_join_manifest` had 2 subdirectories that contained a non-partitioned and partitioned version of a manifest that could be used for testing multi-hop joins. The directory structure is different from the other test manifests, so this PR moves the subdirectories to `multi_hop_join_manifest` and `partitioned_multi_hop_join_manifest` at the same level as the other test manifests.  This makes the directory structure consistent, and allows for simpler logic in later utility functions.


tlento (approved):
Seems reasonable

PR #1025, opened 2024-02-05
Improve ID Consistency in Fixtures / Snapshots.

plypaul:
### Description

I found some cases where there were run-to-run ID inconsistencies in some fixture objects. Previously, the test fixtures included a `ConsistentIdObjectRepository` that handled this, but there have since been a few additions. This PR adds / migrates existing fixtures to use `mf_engine_test_fixture_mapping` (may be renamed) that is a more comprehensive version of `ConsistentIdObjectRepository`. This will make snapshots generated using those fixtures (added in later PRs) consistent.

To recap:

* Many tests compare snapshots of objects / SQL queries.
* Those objects with IDs may have been created through a fixture, and IDs are assigned sequentially in the order of creation.
* When running tests, the order of execution of fixtures for a given scope can vary from run to run.
* When the run order changes, IDs for objects may change since the IDs are assigned sequentially.
* ID changes will cause unexpected snapshot diffs. 
* To resolve this, we create a single session-level fixture that creates all affected objects in a consistent order.


metricflow/test/fixtures/manifest_fixtures.py
```diff
@@ -3,7 +3,8 @@
 import logging
 from dataclasses import dataclass
 from enum import Enum
-from typing import Dict, Mapping, OrderedDict, Sequence, Tuple
+from typing import Dict, Mapping, OrderedDict, Sequence
+from typing import Tuple
```
tlento: Weird, but whatever....

metricflow/test/fixtures/model_fixtures.py
```diff
@@ -1,116 +1,17 @@
...
-from metricflow.dataset.semantic_model_adapter import SemanticModelDataSet
 from metricflow.model.semantic_manifest_lookup import SemanticManifestLookup
-from metricflow.test.fixtures.id_fixtures import IdNumberSpace, patch_id_generators_helper
-
 
 logger = logging.getLogger(__name__)
 
 
-@dataclass(frozen=True)
-class ConsistentIdObjectRepository:
```
tlento, line 79: Huzzah!

metricflow/test/fixtures/id_fixtures.py
```diff
@@ -19,7 +19,7 @@ class IdNumberSpace:
     CONSISTENT_ID_REPOSITORY = 10000
 
 
-@pytest.fixture(autouse=True)
+@pytest.fixture(autouse=True, scope="function")
```
tlento, line 40: Huh, the function-scope here may be what imposes id number consistency, but this shouldn't be a change in behavior since scope defaults to function so I'm not really sure what you were running into or what this fixes.

metricflow/test/fixtures/dataflow_fixtures.py
```diff
@@ -45,19 +46,21 @@ def query_parser(  # noqa: D
 def extended_date_dataflow_plan_builder(  # noqa: D
     mf_engine_test_fixture_mapping: Mapping[SemanticManifestName, MetricFlowEngineTestFixture]
 ) -> DataflowPlanBuilder:
+    # Scope needs to be function as the DataflowPlanBuilder contains state.
```
tlento, line 49: Shall we explicitly tag these with "function" then?

metricflow/test/fixtures/id_fixtures.py
```diff
@@ -10,13 +11,30 @@
...
+
+        This is useful for creating unique / non-overlapping IDs for different groups of objects.
+        """
+        if not block_index >= 0:
+            raise RuntimeError(f"block_index should be >= 0. Got: {block_index}")
+        if not block_size > 1:
+            raise RuntimeError(f"block_size should be > 1. Got: {block_size}")
+        if not first_block_offset >= 0:
+            raise RuntimeError(f"first_block_offset should be >= 0. Got: {first_block_offset}")
+        return IdNumberSpace(first_block_offset + block_size * block_index)
```
tlento: In future, can we please avoid mixing large mechanical changes like renames with structural or functional changes (or even other renames)?

It's hard to convey just how much harder it is to review this commit as a result, and in other contexts this kind of thing leads to preventable incidents because reviewers get confused about the intersection of these changes and fail to recognize when some small thing is out of place in the mass of mechanical updates.

metricflow/test/fixtures/manifest_fixtures.py
```diff
@@ -37,22 +37,64 @@
...
+    """Enumeration of semantic manifests that defined in YAML files and used for testing."""
+
+    AMBIGUOUS_RESOLUTION_MANIFEST = SemanticManifestSetupPropertySet(
+        semantic_manifest_name="ambiguous_resolution_manifest",
+        id_number_space=IdNumberSpace.for_block(0),
+    )
+    # Not including CONFIG_LINTER_MANIFEST as it has intentional errors for running validations.
+    CYCLIC_JOIN_MANIFEST = SemanticManifestSetupPropertySet(
+        semantic_manifest_name="cyclic_join_manifest", id_number_space=IdNumberSpace.for_block(1)
+    )
```
tlento, line 62: Ok, this refines the consistency further, but I'm still not sure if this solves the consistency problem altogether.

tlento (approved):
A few things:

1. I can't load the snapshots except the fop few. Those look fine, so I'm trusting your assertion about the remainder being limited to number changes is correct.
2. I do not see how this solves the problem you were having, in part because I'm not clear on where you were encountering inconsistencies.
3. We might need to do even more here because xdist doesn't do test-suite-execution-level fixtures

All that said, the general cleanup in here is worth a lot - the fixtures are cleaner and better organized, so we should merge this and see if we still need to do more later.

plypaul:
> I do not see how this solves the problem you were having, in part because I'm not clear on where you were encountering inconsistencies.

The issue I saw was similar to the following:
* There are 2 fixtures (`fixture_0` and `fixture_1`) that assign IDs.
* When all tests are run in a testing session, `fixture_0` runs first, followed by `fixture_1`. The ID for the object in `fixture0` is `id_0` and the ID for the object in `fixture1` is `id_1`.
* When running a single test that only depends on `fixture1`, the ID for the object is now `id_0` since `fixture_0` didn't run.


> We might need to do even more here because xdist doesn't do test-suite-execution-level fixtures

There shouldn't need to be anything special done for `xdist` since it's effectively the same as running a subset of tests.

metricflow/test/fixtures/manifest_fixtures.py
```diff
@@ -0,0 +1,201 @@
...
+            column_association_resolver=DunderColumnAssociationResolver(multihop_semantic_manifest_lookup)
+        )
+
+        for semantic_model in semantic_models:
+            data_sets[semantic_model.name] = converter.create_sql_source_data_set(semantic_model)
+
+        return data_sets
+
+
+@pytest.fixture(scope="session")
```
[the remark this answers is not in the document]
plypaul: Replied as a top-level comment.

> Also, since this is apparently not part of the stack with https://github.com/dbt-labs/metricflow/pull/1015 (or maybe that one was moved on top of this somewhere?) it's kind of hard for me to reason about how those two things will work together.

Should have outlined this better, but yes, this PR is part of the same stack and is under #1015.

metricflow/test/fixtures/manifest_fixtures.py
plypaul: I've been using the IDE to add imports, so maybe a bug there? I'll update.

metricflow/test/fixtures/id_fixtures.py
plypaul: This was left over from some testing iteration - I'll remove.

metricflow/test/fixtures/id_fixtures.py
plypaul: Yeah, I've been meaning to avoid such cases as I do get that it makes for a more challenging review.

PR #1028, opened 2024-02-08
New test cases for coalescing nulls

courtneyholcomb:
### Description
Tests for some edge cases that have come up in discussions about coalescing nulls. Getting these SQL examples merged will allow us to see the SQL changes more clearly when working on follow ups like [SL-1677](https://linear.app/dbt-labs/issue/SL-1677/coalesce-nulls-is-not-being-applied-to-conversion-metric-input).


tlento:
I can barely load the snapshots because there are so many. Are we expecting all of those ID number changes to happen here? Or is the merge base of this PR not including all of Paul's stable ID generation changes?

courtneyholcomb:
Sorry, bad merge... moving to a new PR

PR #1029, opened 2024-02-08
Encapsulate Input Nodes for the `DataflowPlanBuilder` Into `SourceNodeSet`

plypaul:
### Description

This PR encapsulates the nodes used by the `DataflowPlanBuilder` into `SourceNodeSet` so that there's a single place where those nodes are generated and can be passed around. This will later be important for generating consistent SQL queries as the queries contain sequentially generated IDs.

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -29,6 +29,7 @@ from metricflow.dataflow.builder.node_evaluator import (
     LinkableInstanceSatisfiabilityEvaluation,
     NodeEvaluatorForLinkableInstances,
 )
+from metricflow.dataflow.builder.source_node import SourceNodeSet
 from metricflow.dataflow.dataflow_plan import (
     AddGeneratedUuidColumnNode,
     AggregateMeasuresNode,
@@ -43,10 +44,8 @@ from metricflow.dataflow.dataflow_plan import (
     JoinOverTimeRangeNode,
     JoinToBaseOutputNode,
     JoinToTimeSpineNode,
-    MetricTimeDimensionTransformNode,
     MinMaxNode,
     OrderByLimitNode,
-    ReadSqlSourceNode,
     SemiAdditiveJoinNode,
     SinkOutput,
     WhereConstraintNode,
@@ -121,20 +120,15 @@ class DataflowPlanBuilder:
 
     def __init__(  # noqa: D
         self,
-        source_nodes: Sequence[BaseOutput],
-        read_nodes: Sequence[ReadSqlSourceNode],
-        time_spine_source_node: MetricTimeDimensionTransformNode,
+        source_node_set: SourceNodeSet,
         semantic_manifest_lookup: SemanticManifestLookup,
         node_output_resolver: Optional[DataflowPlanNodeOutputDataSetResolver] = None,
         column_association_resolver: Optional[ColumnAssociationResolver] = None,
     ) -> None:
         self._semantic_model_lookup = semantic_manifest_lookup.semantic_model_lookup
         self._metric_lookup = semantic_manifest_lookup.metric_lookup
-        self._time_spine_source = semantic_manifest_lookup.time_spine_source
-        self._time_spine_source_node = time_spine_source_node
         self._metric_time_dimension_reference = DataSet.metric_time_dimension_reference()
-        self._source_nodes = source_nodes
-        self._read_nodes = read_nodes
+        self._source_node_set = source_node_set
         self._column_association_resolver = (
             DunderColumnAssociationResolver(semantic_manifest_lookup)
             if not column_association_resolver
@@ -745,20 +739,21 @@ class DataflowPlanBuilder:
                 nodes.append(source_node)
         return nodes
 
-    def _select_read_nodes_with_linkable_specs(
-        self, linkable_specs: LinkableSpecSet, read_nodes: Sequence[ReadSqlSourceNode]
-    ) -> List[ReadSqlSourceNode]:
+    def _select_source_nodes_with_linkable_specs(
+        self, linkable_specs: LinkableSpecSet, source_nodes: Sequence[BaseOutput]
+    ) -> Sequence[BaseOutput]:
         """Find source nodes with requested linkable specs and no measures."""
-        selected_nodes: Set[ReadSqlSourceNode] = set()
+        # Use a dictionary to dedupe for consistent ordering.
+        selected_nodes: Dict[BaseOutput, None] = {}
         requested_linkable_specs_set = set(linkable_specs.as_tuple)
-        for read_node in read_nodes:
-            output_spec_set = self._node_data_set_resolver.get_output_data_set(read_node).instance_set.spec_set
+        for source_node in source_nodes:
+            output_spec_set = self._node_data_set_resolver.get_output_data_set(source_node).instance_set.spec_set
             all_linkable_specs_in_node = set(output_spec_set.linkable_specs)
             requested_linkable_specs_in_node = requested_linkable_specs_set.intersection(all_linkable_specs_in_node)
             if requested_linkable_specs_in_node:
-                selected_nodes.add(read_node)
+                selected_nodes[source_node] = None
 
-        return list(selected_nodes)
+        return tuple(selected_nodes.keys())
 
     def _find_non_additive_dimension_in_linkable_specs(
         self,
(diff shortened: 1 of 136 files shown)


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -745,20 +739,21 @@ def _select_source_nodes_with_measures(
...
 
-    def _select_read_nodes_with_linkable_specs(
-        self, linkable_specs: LinkableSpecSet, read_nodes: Sequence[ReadSqlSourceNode]
-    ) -> List[ReadSqlSourceNode]:
+    def _select_source_nodes_with_linkable_specs(
+        self, linkable_specs: LinkableSpecSet, source_nodes: Sequence[BaseOutput]
+    ) -> Sequence[BaseOutput]:
         """Find source nodes with requested linkable specs and no measures."""
-        selected_nodes: Set[ReadSqlSourceNode] = set()
+        # Use a dictionary to deduoe for consistent ordering.
```
courtneyholcomb, line 746: nit: `dedupe` spelling

courtneyholcomb: Also curious about the use of a dict here instead of a set. Seems odd to use a dict where the values are always `None`. Set would get you the deduping, but maybe there is some ordering behavior I'm not aware of that differs between set & dict?

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -817,78 +812,78 @@ def _find_dataflow_recipe(
...
-                for time_dimension_spec in linkable_spec_set.time_dimension_specs
-                if time_dimension_spec.element_name == self._metric_time_dimension_reference.element_name
-            ]
-            if requested_metric_time_specs:
-                input_time_spine_source_node = self._time_spine_source_node
-                # Add time_spine source node to potential source nodes
-                potential_source_nodes = list(potential_source_nodes) + [self._time_spine_source_node]
+            # For the case where we want to find a recipe for a group-by-item and metric_time, we add the time spine
+            # node (which has metric_time for all date values), and set the default join type between nodes to
+            # FULL_OUTER. This will generate a recipe where you join the node containing the group-by-item and the
```
courtneyholcomb, line 827: This comment is a bit confusing and I think not quite accurate. This `else` clause applies to `all` queries without metrics, but the comment makes it sound like it only applies to queries without metrics that also include `metric_time`.
When querying only group by items, the default join type will _always_ be `FULL OUTER`. This makes it sound like that is only the case when `metric_time` is included in the group by items.

This comment is extra confusing because when `metric_time` is included, the join type is likely going to end up being `CROSS JOIN` (but that probably shouldn't be mentioned here since it is not the _default_ join type, and that logic comes later).

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -817,78 +812,78 @@ def _find_dataflow_recipe(
...
+            # time spine node to get the group-by-item with metric_time.
+            candidate_nodes_for_right_side_of_join = list(self._source_node_set.source_nodes_for_group_by_item_queries)
```
courtneyholcomb, line 829: This is actually a behavior change, though it might not result in a change in output. We previously only added time spine to the source nodes if `metric_time` was queried without metrics. Now time spine will be included in the source nodes any time only group by items are queried.
I think that should be fine, but just want to double check that the time spine wouldn't be used unless `metric_time` is queried. Would be something to double check in node evaluation.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -817,78 +812,78 @@ def _find_dataflow_recipe(
...
             metric_time_dimension_reference=self._metric_time_dimension_reference,
+            time_spine_node=self._source_node_set.time_spine_node,
         )
-        nodes_available_for_joins = tuple(nodes_available_for_joins)
-        if input_time_spine_source_node:
-            nodes_available_for_joins += (input_time_spine_source_node,)
         logger.info(
-            f"After removing unnecessary nodes, there are {len(nodes_available_for_joins)} nodes available for joins"
+            f"After removing unnecessary nodes, there are {len(candidate_nodes_for_right_side_of_join)} candidate "
+            f"nodes for the right side of the join"
```
courtneyholcomb, line 862: nit: f string with no variables (I'm surprised the linter didn't catch this)

metricflow/plan_conversion/node_processor.py
```diff
@@ -332,10 +334,19 @@ def remove_unnecessary_nodes(
...
             element_names_in_data_set = ToElementNameSet().transform(data_set.instance_set.spec_set)
+            element_names_intersection = element_names_in_data_set.intersection(relevant_element_names)
+            if len(element_names_intersection) > 0:
+                logger.debug(f"Including {node} since `element_names_intersection` is {element_names_intersection}")
+                relevant_nodes.append(node)
+                continue
 
-            if len(element_names_in_data_set.intersection(relevant_element_names)) > 0:
+            # Used for group-by-item-values queries.
+            if node == time_spine_node:
```
courtneyholcomb, line 347: Also a behavior change - does this impact the output?

metricflow/plan_conversion/node_processor.py
```diff
@@ -297,7 +297,7 @@ def add_multi_hop_joins(
 
         return list(x.node_with_multi_hop_elements for x in all_multi_hop_join_candidates) + list(nodes)
 
-    def remove_unnecessary_nodes(
+    def heuristically_prune_nodes(
```
courtneyholcomb, line 300: What's the reason for the rename? Seems harder to understand the new name 😅

courtneyholcomb (approved):
Appreciate the organizational changes - thank you!

Some questions about the potential behavior changes here. I think they should be covered by integration tests, but wanted to confirm you had checked that there aren't any changes in output, mainly from the change of always including the time spine node when evaluating nodes for group by-only queries.

metricflow/dataflow/builder/dataflow_plan_builder.py
tlento: Sets are unordered, so if order matters we can't use them as-is.

Do we care about duplicates at all? If so maybe we count them and log an error message if there are duplicates.

metricflow/dataflow/builder/dataflow_plan_builder.py
tlento: Huh. Maybe because it's multi-line? We're on an old version of Ruff, too, so it could be a false negative.

metricflow/plan_conversion/node_processor.py
tlento: Yeah, that would definitely benefit from a separate commit with a commit message explaining it.

tlento (commented):
Don't mind me, just looking through to see what this will do to my plans for predicate pushdown. Looks like a lot of merge conflicts, makes me glad I'm a little slow getting started. 😆 

+1 all of courtney's comments.

metricflow/dataflow/builder/dataflow_plan_builder.py
courtneyholcomb: Googling answered my question - I was thinking sets were equally as ordered as dicts in python3 but that is not the case. TIL! https://stackoverflow.com/questions/1653970/does-python-have-an-ordered-set

metricflow/dataflow/builder/dataflow_plan_builder.py
plypaul: It's fine to always have it there as if `metric_time` is not queried, then the `NodeEvaluatorForLinkableInstances` won't select it as a node to join with.

Also, there were no changes to the SQL snapshots as a result.

metricflow/dataflow/builder/dataflow_plan_builder.py
plypaul: Yeah, this is auto-formatted and f-strings without variables are done for multi-line cases for consistency.

metricflow/dataflow/builder/dataflow_plan_builder.py
plypaul: Duplicates are possible as a dimension with the same name could be defined in multiple semantic models.

metricflow/dataflow/builder/dataflow_plan_builder.py
plypaul: Good call, this was an incorrect comment that I added while I was trying to understand the logic. I've removed it.

metricflow/plan_conversion/node_processor.py
plypaul: This does not as the node it won't be selected unless required.

metricflow/plan_conversion/node_processor.py
plypaul: I thought it would be more helpful, but either works for me. Let me revert it.

PR #1030, opened 2024-02-09
Precompute / Cache Outputs for Nodes in `SourceNodeSet`

plypaul:
### Description

When building a `DataflowPlan`, nodes from `SourceNodeSet` are used as common building blocks for different queries using the same semantic manifest. When the output for a `DataflowPlanNode` is computed, the output is lazily computed and then cached since the output will be required many times for the same nodes (e.g. between queries, and between common join candidates). The output contains generated IDs, and so precomputing the output nodes for nodes from `SourceNodeSet` will have the following effects:

* The runtime of building the `DataflowPlan` should be a little more consistent between queries at the expense of initialization time.
* The IDs in the generated output will be consistent regardless of the order queries seen by the `DataflowPlanBuilder`.

### DIFF
--- a/metricflow/engine/metricflow_engine.py
+++ b/metricflow/engine/metricflow_engine.py
@@ -14,6 +14,7 @@ from dbt_semantic_interfaces.references import EntityReference, MeasureReference
 from dbt_semantic_interfaces.type_enums import DimensionType
 
 from metricflow.dataflow.builder.dataflow_plan_builder import DataflowPlanBuilder
+from metricflow.dataflow.builder.node_data_set import DataflowPlanNodeOutputDataSetResolver
 from metricflow.dataflow.builder.source_node import SourceNodeBuilder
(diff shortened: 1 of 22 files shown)


metricflow/engine/metricflow_engine.py
```diff
@@ -294,8 +295,10 @@ def explain_get_dimension_values(  # noqa: D
         """Returns the SQL query for get_dimension_values.
 
         Args:
-            metric_name: Names of metrics that contain the group_by.
+            metric_names: Names of metrics that contain the group_by.
+            metrics: Similar to `metric_names`, but specified via parameter objects.,
```
courtneyholcomb, line 299: nit: remove comma at the end of this line

courtneyholcomb (approved):
LGTM!

metricflow/engine/metricflow_engine.py
plypaul: Removed.

PR #1032, opened 2024-02-14
Streamline Changelog generation to remove invalid issue links

tlento:
The `Issue` entry in our changelogs is required to be non-empty,
and if it is we format the value into a GitHub link in the final
changelog. This leads to broken links pointing at things like
issue `None` or `0`.

This change updates the changelog formatting to skip over those
entries when forming issue links, and skip formatting in issue
lists if no issues are included. We still include the contributor
and change body, so all due credit remains, just minus the need
for validating common circumventions to our request for issue
numbers.

Running `changie batch 0.205.0  -d` on the underlying trunk commit
and comparing it here illustrates the new behavior with the
docs entry - and the associated author change list - no longer
including bad links.

Since the logic involved was fairly complicated, the initial commits
in this PR cut down on certain sources of complexity in the
changelog formatting. Specifically:

- Remove custom changeFormat from Docs changelog entries
- Consolidate PR/Issue division into Issue requests
- Remove unused duplicate splitList on changelog authors
- Suppress changelog links for missing Issues

### DIFF
--- a/.changes/unreleased/Dependencies-20240119-174213.yaml
+++ b/.changes/unreleased/Dependencies-20240119-174213.yaml
@@ -3,4 +3,4 @@ body: Remove unused numpy dependency
 time: 2024-01-19T17:42:13.46617-08:00
 custom:
   Author: tlento
-  PR: "984"
+  Issue: "984"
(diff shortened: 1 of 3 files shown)


courtneyholcomb (approved):
I love Tom's productive procrastination 🥲 THANK YOU!

PR #1037, opened 2024-02-15
New test cases for `fill_nulls_with`

courtneyholcomb:
### Description
Tests for some edge cases that have come up in discussions about coalescing nulls. Getting these SQL examples merged will allow us to see the SQL changes more clearly when working on follow ups like [SL-1677](https://linear.app/dbt-labs/issue/SL-1677/coalesce-nulls-is-not-being-applied-to-conversion-metric-input).

### DIFF
--- a/metricflow/test/fixtures/semantic_manifest_yamls/simple_manifest/metrics.yaml
+++ b/metricflow/test/fixtures/semantic_manifest_yamls/simple_manifest/metrics.yaml
@@ -704,3 +704,39 @@ metric:
     expr: bookings
     metrics:
       - name: bookings
+---
+metric:
+  name: twice_bookings_fill_nulls_with_0_without_time_spine
+  description: 2x twice_bookings_fill_nulls_with_0_without_time_spine
+  type: derived
+  type_params:
+    expr: 2 * bookings_fill_nulls_with_0_without_time_spine
+    metrics:
+      - name: bookings_fill_nulls_with_0_without_time_spine
+---
+metric:
+  name: nested_fill_nulls_without_time_spine
+  description: 2x twice_bookings_fill_nulls_with_0_without_time_spine but then x3
+  type: derived
+  type_params:
+    expr: 3 * twice_bookings_fill_nulls_with_0_without_time_spine
+    metrics:
+      - name: twice_bookings_fill_nulls_with_0_without_time_spine
+---
+metric:
+  name: visit_buy_conversion_rate_7days_fill_nulls_with_0
+  description: conversion rate on visits-buys on a 7 day window, filling nulls with 0 for both input measures
+  type: conversion
+  type_params:
+    conversion_type_params:
+      base_measure:
+        name: visits
+        fill_nulls_with: 0
+        join_to_timespine: true
+      conversion_measure:
+        name: buys
+        fill_nulls_with: 0
+        join_to_timespine: true
+      window: 7 days
+      entity: user
+      calculation: conversion_rate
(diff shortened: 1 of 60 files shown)


metricflow/test/fixtures/semantic_manifest_yamls/simple_manifest/metrics.yaml
```diff
@@ -704,3 +704,39 @@ metric:
     expr: bookings
     metrics:
       - name: bookings
+---
+metric:
+  name: twice_bookings_fill_nulls_with_0_without_time_spine
+  description: 2x twice_bookings_fill_nulls_with_0_without_time_spine
```
tlento, line 710: nit: `2x bookings_fill_nulls_with_0_without_time_spine`

metricflow/test/fixtures/semantic_manifest_yamls/simple_manifest/metrics.yaml
```diff
@@ -704,3 +704,39 @@ metric:
...
+  type: derived
+  type_params:
+    expr: 2 * bookings_fill_nulls_with_0_without_time_spine
+    metrics:
+      - name: bookings_fill_nulls_with_0_without_time_spine
+---
+metric:
+  name: nested_fill_nulls_without_time_spine
+  description: 2x twice_bookings_fill_nulls_with_0_without_time_spine but then x3
```
tlento, line 719: Same nit, but this time maybe strike the `2x `

metricflow/test/plan_conversion/dataflow_to_sql/test_conversion_metrics_to_sql.py
```diff
@@ -168,3 +168,32 @@ def test_conversion_rate_with_constant_properties(
...
+
+@pytest.mark.sql_engine_snapshot
+def test_conversion_metric_fill_nulls_with_0(
+    request: FixtureRequest,
+    mf_test_session_state: MetricFlowTestSessionState,
+    dataflow_plan_builder: DataflowPlanBuilder,
+    dataflow_to_sql_converter: DataflowToSqlQueryPlanConverter,
+    sql_client: SqlClient,
+) -> None:
+    """Test conversion metric with constant properties by data flow plan rendering."""
```
tlento, line 181: Let's update the docstring here and in the other test cases to indicate any weirdness. Like if this won't fill nulls at every level, we can put that in with a note that this should highlight the relevant changes if/when we alter that behavior.

metricflow/test/snapshots/test_conversion_metrics_to_sql.py/SqlQueryPlan/BigQuery/test_conversion_metric_fill_nulls_with_0__plan0_optimized.sql
```diff
@@ -0,0 +1,84 @@
+-- Compute Metrics via Expressions
+SELECT
+  metric_time__day
+  , CAST(buys AS FLOAT64) / CAST(NULLIF(visits, 0) AS FLOAT64) AS visit_buy_conversion_rate_7days_fill_nulls_with_0
```
tlento, line 4: Oh interesting. If we want to support recursively filling nulls with something we'll need to apply it on the metric, because conversion and ratio metrics will return NULL for any zero denominator.

This is the default behavior we want - NULL is the correct result - but users may override when they know, for example, that any zero denominator implies a 0% conversion rate for their analysis purposes.

Nothing to do about that now except maybe make a note of this in the test docstring.

tlento (approved):
Seems reasonable!

PR #1044, opened 2024-02-23
Bug fix: querying multiple granularities with offset metrics

courtneyholcomb:
Resolves SL-1777
Resolves SL-1810


metricflow/test/snapshots/test_offset_metrics.py/str/DuckDB/test_offset_to_grain_with_multiple_granularities__query_output.txt
```diff
@@ -0,0 +1,3 @@
+  metric_time__day metric_time__month  bookings_at_start_of_month
+0       2020-01-01         2020-01-01                           5
```
courtneyholcomb, line 2: But why are there 2 rows for a one-date filter? 🙃

PR #1047, opened 2024-02-26
Bug fix: Apply `join_to_timespine` for conversion metric input measures

### LINKED ISSUE #1048: [Bug] `join_to_timespine` not applied for conversion metrics

### Current Behavior

If a conversion metric specifies that either of it's input measures should join to time spine, that join never occurs.

### Expected Behavior

The join to time spine should occur, ensuring that there is a row for each date in the query.

### Steps To Reproduce

Add `join_to_timespine: true` to either measure in a conversion metric's YAML. Query the metric, and you'll see that the output SQL never joins to time spine. Example YAML:
```
metric:
  name: visit_buy_conversion_rate_7days_fill_nulls_with_0
  description: conversion rate on visits-buys on a 7 day window, filling nulls with 0 for both input measures
  type: conversion
  type_params:
    conversion_type_params:
      base_measure:
        name: visits
        fill_nulls_with: 0
        join_to_timespine: true
      conversion_measure:
        name: buys
        fill_nulls_with: 0
        join_to_timespine: true
      window: 7 days
      entity: user
      calculation: conversion_rate
```

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
- metricflow:
```

_No response_

### Additional Context

_No response_

courtneyholcomb:
Resolves #1048
Completes SL-1677

### Description
Previously, if a conversion metric had `join_to_timespine: true` on either of its input measures, that join never happened. This fixes that.
I'd recommend reviewing by commit. Steps taken:
- Refactor dataflow plan logic for input measures so that it's reusable for conversion metrics. You can see that all tests passed for this first commit - it should not change any output.
- Reuse that base logic when building input measures for conversion metrics, ensuring that we don't skip any steps that should be applied to all input measures (e.g., joining to time spine). In this commit you can more clearly see how the SQL has changed. (The test gets renamed in the next commit, generating new files so it's harder to see what changed.)
- Rename relevant test to make it clear it should include joining to time spine.


WilliamDee (approved):
Thanks! The conversion code looks a lot cleaner too!

PR #1054, opened 2024-02-28
Bug fix: querying multiple granularities with offset metrics

### LINKED ISSUE #1052: [SL-1810] [Bug] Querying a time offset metric with multiple `metric_time` dimensions returns incorrect results

### Current Behavior

If you query a time offset metric with one instance of `metric_time` (or `agg_time_dimension`), the results are offset as expected. But if you query with multiple instances of `metric_time`, only one of the `metric_time` instances will be offset appropriately.

### Expected Behavior

Every instance of `metric_time` should be offset in the query results.

### Steps To Reproduce

Build a derived metric that uses offset_window or offset_to_grain. Query that metric with `metric_time__day` and `metric_time__week`.

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
- metricflow:
```

_No response_

### Additional Context

_No response_

<sub>[SL-1810](https://linear.app/dbt-labs/issue/SL-1810/[bug]-querying-a-time-offset-metric-with-multiple-metric-time)</sub>

### LINKED ISSUE #1053: [SL-1811] [Bug] Offset metric error when querying with one grain & filtering with another

### Current Behavior

If you query an offset metric error with `metric_time` with one grain, but then filter by `metric_time` with another grain, you will get a SQL error saying the filtered column does not exist. The filtered grain is being removed from the subquery before the filter is applied.

### Expected Behavior

The filtered column needs to be kept in subqueries until the filter is applied so that the SQL can execute successfully.

### Steps To Reproduce

Define a derived metric that uses `offset_window` or `offset_to_grain`. Query that metric by `metric_time__day` with filter on `metric_time__month`.

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
- metricflow:
```

_No response_

### Additional Context

This is related to and potentially blocked by https://github.com/dbt-labs/metricflow/issues/1052.

<sub>[SL-1811](https://linear.app/dbt-labs/issue/SL-1811/[bug]-offset-metric-error-when-querying-with-one-grain-and-filtering)</sub>

courtneyholcomb:
Resolves SL-1777
Resolves SL-1810
Resolves #1053 
Resolves #1052 

### Description
Resolves 2 related issues:

1. If you query a time offset metric with multiple instances of `metric_time`/`agg_time_dimension`, only one of the instances will be offset. All of them should be.
2. If you query a time offset metric with one instance of `metric_time`/`agg_time_dimension` but filter by a different one, the query will fail

These issues were caused by issues in the `JoinToTimeSpineNode` dataflow to SQL code. The node was only designed to offset one time dimension. It was also filtering out columns prematurely, when that filtering should instead be happening in the `FilterElementsNode`. This PR fixes both issues.

I would recommend reviewing this PR by commit.
- The first commit (after changelog) adds tests that show what the SQL previously looked like, as well as the scenarios that trigger an error before SQL can be generated.
- Next commit updates the dataflow to SQL logic.
- Third commit updates the snapshots so you can clearly see how the logic change impacts the SQL.


metricflow/test/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/DuckDB/test_offset_to_grain_metric_filter_and_query_have_different_granularities__plan0_optimized.sql
```diff
@@ -0,0 +1,36 @@
...
+      SELECT
+        DATE_TRUNC('day', ds) AS metric_time__day
+        , 1 AS bookings
+      FROM ***************************.fct_bookings bookings_source_src_28000
+    ) subq_11
+    ON
+      DATE_TRUNC('month', subq_13.ds) = subq_11.metric_time__day
+    WHERE DATE_TRUNC('month', subq_13.ds) = subq_13.ds
+  ) subq_15
+  WHERE metric_time__day = '2020-01-01'
```
tlento, line 33: This all feels very convoluted but maybe that's just how we have to do offset to grain calculations? Like if this grouped by metric_time__day I guess we'd drop the where filter and just have a repeated bookings_start_of_month value collected alongside each day.

I'm actually not sure how I'd write the SQL for this to make it less confusing. For this case I'd probably do the filter + aggregation on the month grains and then join to time spine to fill out other granularities, but that doesn't generalize.

Anyway, something for us to think about later - it really feels like there has to be a cleaner way to render this, but maybe there isn't and that's another reason why MetricFlow is useful.

metricflow/test/integration/query_output/test_offset_metrics.py
```diff
@@ -21,6 +21,7 @@ def test_offset_to_grain_with_single_granularity(  # noqa: D
         MetricFlowQueryRequest.create_with_random_request_id(
             metric_names=["bookings_at_start_of_month"],
             group_by_names=["metric_time__day"],
+            order_by_names=["metric_time__day"],
```
tlento, line 24: Yay! Thank you!

tlento (approved):
This seems like what we want to happen instead of whatever was happening before, but I do wonder about the invariant handling I mentioned inline - there's an implicit expectation of a granularity relationship between the time spine and the selected columns but right now if that expectation isn't met we'll render a date trunc to a coarser granularity, which doesn't seem like what we really want to do.

Should we just put an assertion in there? Or is there some other mechanism enforcing this?

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1331,35 +1341,48 @@ def visit_join_to_time_spine_node(self, node: JoinToTimeSpineNode) -> SqlDataSet
...
+        # Add requested granularities (if different from time_spine) and date_parts to time spine column.
+        for time_dimension_instance in time_dimensions_to_select_from_time_spine:
+            time_dimension_spec = time_dimension_instance.spec
+
+            # Apply grain to time spine column select expression, unless grain already matches time spine column.
+            select_expr: SqlExpressionNode = (
+                time_spine_column_select_expr
+                if time_dimension_spec.time_granularity == time_spine_dim_instance.spec.time_granularity
+                else SqlDateTruncExpression(
+                    time_granularity=time_dimension_spec.time_granularity, arg=time_spine_column_select_expr
```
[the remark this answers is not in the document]
courtneyholcomb: Good point! I feel like the time spine SHOULD be set with the smallest possible grain in order to fill all granularity periods we support.
In practice, though, it might not be - especially when we transition to supporting smaller grains. Users might need to update their time spines.
I'll add an assertion for now and a comment to explain!

metricflow/test/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/DuckDB/test_offset_to_grain_metric_filter_and_query_have_different_granularities__plan0_optimized.sql
courtneyholcomb: `offset_to_grain` ends up with the weirdest edge cases. We might just want to be more restrictive with what we allow for those metrics. It's already so weird when there are multiple instances of `metric_time` queried with an `offset_to_grain` metric because the where filter that we add by default doesn't totally do its job anymore. And I have no idea why someone would want to make that query. But alas, for now, we can just give the people what they asked for.

courtneyholcomb:
@tlento When I added that assertion, it revealed a sneaky bug related to a duplicate variable name. I added an additional commit to fix that bug so you might want to take a look at that before I merge.  It's [this commit](https://github.com/dbt-labs/metricflow/pull/1054/commits/dc60c11e00907c798c9b5c25865b3949184c9aff) which also updates basic snapshots. 
I'm updating the other engine snapshots now but don't think that should block your review.

metricflow/test/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/DuckDB/test_offset_to_grain_metric_filter_and_query_have_different_granularities__plan0_optimized.sql
tlento: > But alas, for now, we can just give the people what they asked for.

LOL, this is me every time I have to add a feature of any kind.

metricflow/test/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/DuckDB/test_cumulative_time_offset_metric_with_time_constraint__plan0.sql
```diff
@@ -30,7 +30,7 @@ FROM (
             , DATE_TRUNC('month', subq_5.metric_time__day) AS metric_time__month
             , DATE_TRUNC('quarter', subq_5.metric_time__day) AS metric_time__quarter
             , DATE_TRUNC('year', subq_5.metric_time__day) AS metric_time__year
-            , EXTRACT(year FROM DATE_TRUNC('day', subq_5.metric_time__day)) AS metric_time__extract_year
+            , EXTRACT(year FROM subq_5.metric_time__day) AS metric_time__extract_year
```
tlento, line 33: Oh wow I didn't even notice that.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1379,14 +1393,14 @@ def visit_join_to_time_spine_node(self, node: JoinToTimeSpineNode) -> SqlDataSet
...
+                element_name=original_time_spine_dim_instance.spec.element_name,
+                entity_links=original_time_spine_dim_instance.spec.entity_links,
                 time_granularity=time_dimension_spec.time_granularity,
                 date_part=time_dimension_spec.date_part,
-                aggregation_state=time_spine_dim_instance.spec.aggregation_state,
+                aggregation_state=original_time_spine_dim_instance.spec.aggregation_state,
             )
             time_spine_dim_instance = TimeDimensionInstance(
-                defined_from=time_spine_dim_instance.defined_from,
+                defined_from=original_time_spine_dim_instance.defined_from,
```
tlento, line 1403: Yikes, this was inside the loop!

tlento (approved):
Sneaky sneaky variable naming bugs. We probably need a refactor in there to make us less dependent on how we name things, but I don't think it's worth the time to do that today.

Looks good!

PR #1056, opened 2024-02-28
Bug fix: For measures using `join_to_timespine`, apply filters after time spine join

### LINKED ISSUE #1039: [SL-1732] [Bug] Where filter is applied to inner query when using `join_to_timespine`

### Current Behavior

If I use a where filter when querying metrics that are joined to a time spine the filter is applied to the inner query before the timespine join. This means data is filtered out, then joined to the time spine so I see dates in the final result set that I expect to be filtered out.

For example, I issued the following query in the jaffle-sl-template repo. 
`dbt sl query --metrics twice_orders_fill_nulls_with_0,mqls --group-by metric_time --where "{{ TimeDimension('metric_time')` > '2020-01-01'" --order-by metric_time desc --compile}}

The data return includes records from 2014, which is when my time spine model starts

In the compiled SQL you can see that the filter is gettings applied to the metric sub query.

```
SELECT
  COALESCE(subq_5131.metric_time**day, subq_5141.metric_time**day) AS metric_time__day
  , COALESCE(MAX(subq_5131.mqls), 0) AS mqls
  , MAX(subq_5141.twice_orders_fill_nulls_with_0) AS twice_orders_fill_nulls_with_0
FROM (
  SELECT
    metric_time__day
    , COALESCE(mqls, 0) AS mqls
  FROM (
    SELECT
      subq_5129.date_day AS metric_time__day
      , subq_5127.mqls AS mqls
    FROM dbt_jstein.metricflow_time_spine subq_5129
    LEFT OUTER JOIN (
      SELECT
        metric_time__day
        , SUM(mqls) AS mqls
      FROM (
        SELECT
          DATE_TRUNC('day', first_contact_date) AS metric_time__day
          , CASE WHEN mql_id IS NOT NULL THEN 1 ELSE 0 END AS mqls
        FROM ANALYTICS.dbt_jstein.olist_marketing_qualified_leads olist_mqls_src_678
      ) subq_5125
      WHERE metric_time__day > '2020-01-01'
      GROUP BY
        metric_time__day
    ) subq_5127
    ON
      subq_5129.date_day = subq_5127.metric_time__day
  ) subq_5130
) subq_5131
FULL OUTER JOIN (
  SELECT
    metric_time__day
    , 2 * orders_fill_nulls_with_zero AS twice_orders_fill_nulls_with_0
  FROM (
    SELECT
      metric_time__day
      , COALESCE(order_count, 0) AS orders_fill_nulls_with_zero
    FROM (
      SELECT
        subq_5138.date_day AS metric_time__day
        , subq_5136.order_count AS order_count
      FROM dbt_jstein.metricflow_time_spine subq_5138
      LEFT OUTER JOIN (
        SELECT
          metric_time__day
          , SUM(order_count) AS order_count
        FROM (
          SELECT
            DATE_TRUNC('day', ordered_at) AS metric_time__day
            , 1 AS order_count
          FROM ANALYTICS.dbt_jstein.orders orders_src_680
        ) subq_5134
        WHERE metric_time__day > '2020-01-01'
        GROUP BY
          metric_time__day
      ) subq_5136
      ON
        subq_5138.date_day = subq_5136.metric_time__day
    ) subq_5139
  ) subq_5140
) subq_5141
ON
  subq_5131.metric_time**day = subq_5141.metric_time**day
GROUP BY
  COALESCE(subq_5131.metric_time**day, subq_5141.metric_time**day)
ORDER BY metric_time__day
LIMIT 100
```

I would expect the filter to also be applied to the final sub query to only show the dates I requested. i.e

```
ON
  subq_5131.metric_time**day = subq_5141.metric_time**day
WHERE
  COALESCE(subq_5131.metric_time**day, subq_5141.metric_time**day) > '2020-01-01'
GROUP BY
  COALESCE(subq_5131.metric_time**day, subq_5141.metric_time**day)
ORDER BY metric_time__day
```

### Expected Behavior

I would expect the result set to start on `2020-01-01`

### Steps To Reproduce

1. Checkout main on the jaffle-sl-template repo
2. Run the following query  `dbt sl query --metrics twice_orders_fill_nulls_with_0,mqls --group-by metric_time --where "{{ TimeDimension('metric_time')` > '2020-01-01'" --order-by metric_time desc --compile}}

### Relevant log output

*No response*

### Environment

```markdown
- OS:
- Python:
- dbt:
- metricflow:
```

*No response*

### Additional Context

*No response*

<sub>From [SyncLinear.com](https://synclinear.com) | [SL-1732](https://linear.app/dbt-labs/issue/SL-1732/[bug]-where-filter-is-applied-to-inner-query-when-using-join-to)</sub>

courtneyholcomb:
Resolves #1039 
Completes SL-1732

### Description
For queries using measures that join to time spine, filters were being applied before the time spine join, but not after. This means that there might be additional rows added from the time spine join that should be removed by the filters, resulting in unexpected rows in the output data. Here, we update the DataFlow Plan to apply those filters again after the time spine join, ensuring there will be no unexpected rows in the result.
Note: this means the filters are applied multiple times, which is redundant for the final output but ideally results in a more efficient query.

### DIFF
--- a/metricflow/test/snapshots/test_fill_nulls_with_rendering.py/SqlQueryPlan/BigQuery/test_join_to_time_spine_with_filters__plan0_optimized.sql
+++ b/metricflow/test/snapshots/test_fill_nulls_with_rendering.py/SqlQueryPlan/BigQuery/test_join_to_time_spine_with_filters__plan0_optimized.sql
@@ -1,0 +1,52 @@
+-- Compute Metrics via Expressions
+SELECT
+  metric_time__day
+  , COALESCE(bookings, 0) AS bookings_fill_nulls_with_0
+FROM (
+  -- Constrain Output with WHERE
+  -- Constrain Time Range to [2020-01-03T00:00:00, 2020-01-05T00:00:00]
+  SELECT
+    metric_time__day
+    , bookings
+  FROM (
+    -- Join to Time Spine Dataset
+    SELECT
+      subq_19.metric_time__day AS metric_time__day
+      , subq_18.bookings AS bookings
+    FROM (
+      -- Time Spine
+      SELECT
+        ds AS metric_time__day
+      FROM ***************************.mf_time_spine subq_20
+      WHERE ds BETWEEN '2020-01-03' AND '2020-01-05'
+    ) subq_19
+    LEFT OUTER JOIN (
+      -- Constrain Output with WHERE
+      -- Aggregate Measures
+      SELECT
+        metric_time__day
+        , SUM(bookings) AS bookings
+      FROM (
+        -- Read Elements From Semantic Model 'bookings_source'
+        -- Metric Time Dimension 'ds'
+        -- Constrain Time Range to [2020-01-03T00:00:00, 2020-01-05T00:00:00]
+        -- Pass Only Elements: ['bookings', 'metric_time__day']
+        SELECT
+          DATE_TRUNC(ds, day) AS metric_time__day
+          , 1 AS bookings
+        FROM ***************************.fct_bookings bookings_source_src_28000
+        WHERE DATE_TRUNC(ds, day) BETWEEN '2020-01-03' AND '2020-01-05'
+      ) subq_16
+      WHERE metric_time__day > '2020-01-01'
+      GROUP BY
+        metric_time__day
+    ) subq_18
+    ON
+      subq_19.metric_time__day = subq_18.metric_time__day
+  ) subq_21
+  WHERE (
+    metric_time__day BETWEEN '2020-01-03' AND '2020-01-05'
+  ) AND (
+    metric_time__day > '2020-01-01'
+  )
+) subq_23
--- a/.changes/unreleased/Fixes-20240228-084335.yaml
+++ b/.changes/unreleased/Fixes-20240228-084335.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: 'Bug fix: if measure joins to time spine, apply filters again after that join.'
+time: 2024-02-28T08:43:35.044076-08:00
+custom:
+  Author: courtneyholcomb
+  Issue: "1039"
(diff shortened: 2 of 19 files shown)


metricflow/test/snapshots/test_fill_nulls_with_rendering.py/SqlQueryPlan/BigQuery/test_join_to_time_spine_with_filters__plan0_optimized.sql
```diff
@@ -0,0 +1,52 @@
...
+        metric_time__day
+    ) subq_18
+    ON
+      subq_19.metric_time__day = subq_18.metric_time__day
+  ) subq_21
+  WHERE (
+    metric_time__day BETWEEN '2020-01-03' AND '2020-01-05'
+  ) AND (
+    metric_time__day > '2020-01-01'
+  )
```
tlento, line 51: This is just an artifact of passing both a time constraint and a where filter on metric_time in the same query, right? Because if the filter in the query was `metric_time__day = '2020-01-01'` then this would return the empty set, but that's actually what the constraint combination would tell us to do.

tlento (approved):
This looks good, the inline is just me double-checking that we're rendering faithfully even if the input might be odd.

metricflow/test/snapshots/test_fill_nulls_with_rendering.py/SqlQueryPlan/BigQuery/test_join_to_time_spine_with_filters__plan0_optimized.sql
courtneyholcomb: Yes. I knew it was weird when I wrote this test case but I just wanted to make sure both filters got rendered 😆

courtneyholcomb: But also it won't return an empty dataset since `'2020-01-03'` through `'2020-01-05'` is `> '2020-01-01'`!

PR #1057, opened 2024-02-28
Tests for min/max with `metric_time`

courtneyholcomb:
Resolves SL-1808

### Description
Resolves a #TODO that I left in here a while back. Test cases combining min/max query with `metric_time` only. (These features were worked on in parallel, so I couldn't add tests until they both merged.)


tlento (approved):
Nice, thank you!

PR #1059, opened 2024-02-29
Improve error message for metrics with no parents

tlento:
Due to a gap in our upstream model validation it is possible for
a user to define a derived metric with no inputs. This causes a
fall-through error to trigger of the form `unbound method
set.intersection needs an argument`.

The root cause of this is our use of the `*` operator for expanding
an iterable into a set of values. If the iterable is empty, Python
will simply pass nothing - not None, but nothing at all. This means
any call like `set.intersection(*input)` will raise this strange error
if `input` is empty.

This change does two things to resolve this issue:

1. It updates the set.intersection(*input) call to a more robust
approach, which will return an empty set if the input is empty
2. It updates the invocation to check for empty input and include
a more refined error issue indicating the root cause.

The reason we do it this way is the offending call is inside of
a more generic helper, so we allow the helper to accept empty
input and return correspondingly empty output, but at the callsite
where we know the input should not be empty and that empty input
indicates a configuration issue we raise the appropriate error.

### DIFF
--- a/metricflow/query/group_by_item/candidate_push_down/group_by_item_candidate.py
+++ b/metricflow/query/group_by_item/candidate_push_down/group_by_item_candidate.py
@@ -49,7 +49,7 @@ class GroupByItemCandidateSet(PathPrefixable):
         where the new candidate set was created.
         """
         specs_as_sets = tuple(set(candidate_set.specs) for candidate_set in candidate_sets)
-        common_specs = set.intersection(*specs_as_sets)
+        common_specs = set.intersection(*specs_as_sets) if specs_as_sets else set()
         if len(common_specs) == 0:
             return GroupByItemCandidateSet.empty_instance()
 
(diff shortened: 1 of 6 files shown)


tlento:
* **#1059** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1059?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dbt-labs/metricflow/1059?utm_source=stack-comment).

metricflow/query/group_by_item/candidate_push_down/group_by_item_candidate.py
```diff
@@ -49,7 +49,7 @@ def intersection(
         where the new candidate set was created.
         """
         specs_as_sets = tuple(set(candidate_set.specs) for candidate_set in candidate_sets)
-        common_specs = set.intersection(*specs_as_sets)
+        common_specs = set.intersection(*specs_as_sets) if specs_as_sets else {}
```
plypaul, line 52: Might need to be `set()` as `{}` seems to evaluate to a dict when I tried it out.

plypaul (approved):

metricflow/query/group_by_item/candidate_push_down/group_by_item_candidate.py
tlento: le sigh

PR #1068, opened 2024-03-07
Add script for helping with MetricFlow releases

tlento:
The MetricFlow release process is a little annoying with
updates through FOSSA (which we can hopefully replace with
easier to manage SBOM automation), changelog management,
and so forth.

This automates some of the manual bits and makes the release
PR and tagging processes less error-prone.

This is not production quality deployment code by any means,
but it's better than hand-rolling everything.


courtneyholcomb (approved):
🚀

PR #1074, opened 2024-03-12
Separate method for recursive dataflow plans

courtneyholcomb:
### Description
We will need to build recursive dataflow plans for metrics as dimensions. This PR should not change any output behavior - just separates some DFP logic into a helper method that can be used to build source nodes.


tlento (approved):
Seems reasonable!

PR #1082, opened 2024-03-18
 Support `CREATE TABLE ... AS SELECT` in SQL Object Model

plypaul:
### Description

When generating the tasks for the execution plan, the `CREATE TABLE ... AS SELECT` statement was not generated through the SQL object model. Instead, it was generated when the associated task ran. For better consistency and to support potential SQL-engine differences, this PR adds the appropriate SQL node to model that statement. The execution task updates will be done in a separate PR.

### DIFF
--- a/metricflow/sql/sql_plan.py
+++ b/metricflow/sql/sql_plan.py
@@ -8,6 +8,8 @@ from dataclasses import dataclass
 from enum import Enum
 from typing import Generic, List, Optional, Sequence, Tuple
 
+from typing_extensions import override
+
 from metricflow.dag.id_prefix import IdPrefix, StaticIdPrefix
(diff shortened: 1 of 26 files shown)


metricflow/sql/sql_plan.py
```diff
@@ -289,6 +295,48 @@ def as_select_node(self) -> Optional[SqlSelectStatementNode]:  # noqa: D
...
+
+    @property
+    @override
+    def as_select_node(self) -> Optional[SqlSelectStatementNode]:
+        return None
+
+    @property
+    @override
+    def description(self) -> str:
+        return "Create table {self."
```
courtneyholcomb, line 323: looks like this didn't get finished!

courtneyholcomb (approved):

metricflow/sql/sql_plan.py
plypaul: Updated.

PR #1084, opened 2024-03-18
Update `MetricFlowTestSessionState` / `pytest` Flags for Displaying Graphs / Snapshots

plypaul:
### Description

`MetricFlowTestSessionState` was an odd object that contained both configuration and state for the test session. The reason for the state was that it needed to keep track of the number of displayed graphs so that in case the user tried to display graphs for all tests, the user wouldn't be flooded. To address this, this PR makes the following changes:

* Update the flag behavior so that you can only display graphs when running a single test. In retrospect, displaying graphs for multiple tests seems like an uncommon use case.
* Rename `MetricFlowTestSessionState` to `MetricFlowTestConfiguration`
* Use separate flags for displaying snapshots vs graphs.
* Other minor changes.


courtneyholcomb (approved):
LGTM!

PR #1085, opened 2024-03-18
Add `raw_strings` Argument to `preserve_raw_strings `

plypaul:
### Description

This PR adds an argument to `mf_pformat_many` so that for multi-line strings (e.g. rendered SQL), the log output looks cleaner.

### DIFF
--- a/metricflow/mf_logging/pretty_print.py
+++ b/metricflow/mf_logging/pretty_print.py
@@ -424,21 +424,32 @@ def mf_pformat_many(  # type: ignore
     include_object_field_names: bool = True,
     include_none_object_fields: bool = False,
     include_empty_object_fields: bool = False,
+    preserve_raw_strings: bool = False,
 ) -> str:
-    """Prints many objects in an indented form."""
+    """Prints many objects in an indented form.
+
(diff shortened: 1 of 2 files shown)


metricflow/mf_logging/pretty_print.py
```diff
@@ -424,21 +424,32 @@ def mf_pformat_many(  # type: ignore
     include_object_field_names: bool = True,
     include_none_object_fields: bool = False,
     include_empty_object_fields: bool = False,
+    raw_strings: bool = False,
```
tlento, line 427: nit: can we rename this to `preserve_raw_strings` or something? The `raw_strings` name isn't descriptive at the call site, and since the parameter value is a bool we're just requiring new devs to go read the docstring or function body to find out what `raw_strings` does.

tlento (approved):
As long as the newlines aren't messing up error log entries or something I'm on board.

metricflow/mf_logging/pretty_print.py
plypaul: Good point - updated.

PR #1099, opened 2024-03-22
Move MetricFlow CLI package to dbt-metricflow

tlento:
As we move towards making MetricFlow a stand-alone
library for building queries and rendering SQL, it makes
sense to remove any direct query execution components and
their attendant dependencies. For dependency management and package
deployment purposes, we are particularly interested in removing the
dbt-core dependencies from the production MetricFlow package.

The lone production component relying on dbt-core is the
MetricFlow CLI. In order to expedite our dependency refinement we
take the step of moving the CLI under dbt-metricflow here.

The packaging update here is as minimal as possible - there are a
load of problems with this layout, not least of which is we need to
do some pretty janky editable install operations in all of our test
environments and duplicate dependencies across the two packages just
to keep dev tools working.

In the long run, we should move to a more refined packaging model that
allows us to execute the MetricFlow test suite using the local
library with a stand-alone SqlClient built for testing AND to run the
same test suite using our supported adapter packages via dbt-metricflow.
Doing this requires us to:

1. Overhaul our test package and fixture setups
2. Refine MetricFlow's APIs so they can be simply invoked from whatever
components in dbt-metricflow require them (e.g., the CLI)
3. Re-organize our repo packaging setup to, at a minimum, move the CLI
tests under dbt-metricflow

While none of this is hard, exactly, it's a lot of effort that we don't
have the time to make at this time, so instead we have a janky package
layout and rely on a few questionable dev environment management practices.


courtneyholcomb (approved):
Awesome! All makes sense to me! 🚀

PR #1116, opened 2024-04-09
Bump dev version

courtneyholcomb:
### Description
Bump dev version to release https://github.com/dbt-labs/metricflow/pull/1115.


DevonFulcher:
@courtneyholcomb Considering the new PR workflow, do we need so many dev releases still? Can we just change MFS like this https://github.com/dbt-labs/metricflow-server/pull/529/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R15 (internal to dbt Labs)

DevonFulcher (approved):
Approving, but I don't think we need to do this practice anymore

courtneyholcomb:
> Approving, but I don't think we need to do this practice anymore

Ooo I'm behind! Will see if the action works when I merge the base PR!

PR #1118, opened 2024-04-09
Make recipe errors more user-friendly

courtneyholcomb:
Resolves SL-1954

### Description
[Users have been seeing](https://dbt-labs.slack.com/archives/C05D1MF1N4Q/p1712067795713559) "Recipe not found" errors, but they aren't likely to understand what that means. This can happen if the specs they requested are not all joinable, so we want to give them an error that makes that more clear.

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -253,11 +253,11 @@ class DataflowPlanBuilder:
         logger.info(f"Recipe for conversion measure aggregation:\n{mf_pformat(conversion_measure_recipe)}")
         if base_measure_recipe is None:
             raise UnableToSatisfyQueryError(
-                f"Recipe not found for measure spec: {base_measure_spec.measure_spec} and linkable specs: {base_required_linkable_specs}"
+                f"Unable to join all items in request. Measure: {base_measure_spec.measure_spec}; Specs to join: {base_required_linkable_specs}"
             )
         if conversion_measure_recipe is None:
             raise UnableToSatisfyQueryError(
-                f"Recipe not found for measure spec: {conversion_measure_spec.measure_spec}"
+                f"Unable to build dataflow plan for measure: {conversion_measure_spec.measure_spec}"
             )
 
         # Gets the aggregated opportunities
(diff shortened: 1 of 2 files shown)


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -653,7 +653,7 @@ def _build_plan_for_distinct_values(self, query_spec: MetricFlowQuerySpec) -> Da
             linkable_spec_set=required_linkable_specs, time_range_constraint=query_spec.time_range_constraint
         )
         if not dataflow_recipe:
-            raise UnableToSatisfyQueryError(f"Recipe not found for linkable specs: {required_linkable_specs}")
+            raise UnableToSatisfyQueryError(f"Unable to join all items in request: {required_linkable_specs}")
```
tlento, line 656: Is this user-facing, and do we have any idea know when it happens and why? If so we could maybe add some more details around possible causes, but at this point I don't think we have that info yet.

I'm kind of wondering why this happens.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -253,11 +253,11 @@ def _build_aggregated_conversion_node(
         logger.info(f"Recipe for conversion measure aggregation:\n{mf_pformat(conversion_measure_recipe)}")
         if base_measure_recipe is None:
             raise UnableToSatisfyQueryError(
-                f"Recipe not found for measure spec: {base_measure_spec.measure_spec} and linkable specs: {base_required_linkable_specs}"
+                f"Unable to join all items in request. Measure: {base_measure_spec.measure_spec}; Specs to join: {base_required_linkable_specs}"
             )
         if conversion_measure_recipe is None:
             raise UnableToSatisfyQueryError(
-                f"Recipe not found for measure spec: {conversion_measure_spec.measure_spec}"
+                f"Unable to build dataflow plan for measure: {conversion_measure_spec.measure_spec}"
```
tlento, line 260: Does the measure spec indicate that this is a conversion measure? If not let's do `for conversion measure` instead of `for measure` so we can see if conversion measures are more likely to cause this problem.

tlento (approved):
Thank you!

metricflow/dataflow/builder/dataflow_plan_builder.py
courtneyholcomb: Yeah, it's a good question! I did a quick DD search of the past week of these errors: https://dbtlabsmt.datadoghq.com/logs?query=recipe%20not%20found%20&cols=host%2Cservice&fromUser=true&index=%2A&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1712081485492&to_ts=1712686285492&live=true
There is some stuff I don't understand here, like how a non-additive dimension spec might impact the dataflow plan.

metricflow/dataflow/builder/dataflow_plan_builder.py
courtneyholcomb: Sounds good, will update!

PR #1121, opened 2024-04-10
Rename node visitor method to match typical pattern

courtneyholcomb:
### Description
Just a quick rename! This has been bothering me for a while because it's so hard to remember the name of this method. So I've updated it to match our typical pattern, which is just snake-casing the node class name.


tlento (approved):
I miss my `codecleanupraccoon` macro....

raccoon with a broom!

Yay thank you!

PR #1125, opened 2024-04-11
Remove pointless loop over DataflowPlan optimizer sets

tlento:
The DataflowPlan optimizers are, at least in theory, applied as part
of a loop of optimizer sets to try, with each iteration of the loop
theoretically applying a different set of optimizers, and concluding
with an empty set. The idea behind this was quite clearly to ensure
that the optimizers don't cause valid queries to fail due to some
incorrect optimization path. Why we deemed this necessary is completely
unclear, as no documentation has been provided to explain this bit
of complexity.

What was happening in practice was that on every iteration we were
applying the same hard-coded set consisting of the SourceScanOptimizer,
and the loop itself would incorrectly report error logs suggesting
that other things were being attempted.

Since this logic has never been applied correctly, the optimizer
parameterization is buried so deeply into an implementation
method that there's no way to access it via any reasonable mechanism,
and the built-in loop iteration makes everything here look more
complex than it really is, we remove the loop here.

Note this was added fairly recently, but since the reasoning for
the change was never provided it appears that what amounts to a
revert of the salient commit from PR https://github.com/dbt-labs/metricflow/pull/976 should be fine.

### DIFF
--- a/metricflow/engine/metricflow_engine.py
+++ b/metricflow/engine/metricflow_engine.py
@@ -497,49 +497,23 @@ class MetricFlowEngine(AbstractMetricFlowEngine):
                 time_dimension_specs=query_spec.time_dimension_specs,
             )
 
-        dataflow_plan_optimizer_sets_to_try = (
-            (SourceScanOptimizer(),),
-            (),
-        )
-        dataflow_plan: Optional[DataflowPlan] = None
-        convert_to_execution_plan_result: Optional[ConvertToExecutionPlanResult] = None
+        if query_spec.metric_specs:
+            dataflow_plan = self._dataflow_plan_builder.build_plan(
+                query_spec=query_spec,
+                output_sql_table=output_table,
+                output_selection_specs=output_selection_specs,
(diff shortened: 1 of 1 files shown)


tlento:
* **#1128** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1128?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1127** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1127?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1126** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1126?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1125** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1125?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

courtneyholcomb (approved):
🙏 🙏 🙏

PR #1127, opened 2024-04-11
Add missing aliases to integration test check queries

tlento:
Most engines don't require this, but several do, including
all Postgres versions prior to Postgres 16, so we should
probably keep including the subquery aliases for now.

PR #1128, opened 2024-04-11
Tidy up semantic manifest accessors and helper classes

tlento:
This PR does a bit of tidying on our SemanticModelLookup and MetricLookup classes
and their associated interfaces. The changes include:

1. Adding tests for a core element in ValidLinkableSpecResolver, which feeds
into a critical method in our MetricLookup class
2. Cleaning up the initializers and documentation in the manifest accessors themselves
3. Removing some unused properties from the SemanticModelLookup class

Commit-wise review might be a bit faster.


tlento:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1128?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1128** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1128?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1127** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1127?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1126** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1126?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1125** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1125?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

metricflow/model/semantics/semantic_model_lookup.py
```diff
@@ -41,27 +41,31 @@
 
 
 class SemanticModelLookup(SemanticModelAccessor):
```
courtneyholcomb, line 43: Unrelated to this PR, but I'm wondering if we need these ABC classes (`SemanticModelAccessor` and `MetricAccessor`). Seems like we could get rid of them and just use private functions to restrict what callers use. Though nothing is truly private in python. Or maybe protocols would be better.

metricflow/model/semantics/semantic_model_lookup.py
```diff
@@ -41,27 +41,31 @@
...
     That interface prevents unwanted calls to methods for adding semantic models to the container.
     """
 
-    def __init__(  # noqa: D107
+    def __init__(
         self,
         model: SemanticManifest,
     ) -> None:
-        self._model = model
+        """Initializer.
```
courtneyholcomb, line 54: Does this docstring do anything except to tell you that "model" actually means "semantic manifest"? (Probably a holdover from Transform.)
In which case should we just rename the param and delete the docstring?

courtneyholcomb (approved):
Thank you! 🧹  🧹 🧹

metricflow/model/semantics/semantic_model_lookup.py
tlento: Fun fact, these are Protocols, not ABCs, we just inherit from them because it makes it easier to find the relationship. This was from before the ProtocoloHint thing.

That aside, I had the exact same thought - the Protocols were built to ensure callers would never add models or whatever but that is no longer a meaningful concern. So I was actually thinking about doing this, but then I ran into the other bit about the protocol, which was how it allowed us to sidestep some circular import dependency problems, so I gave up for now.

But yeah, whenever we feel so inclined we should probably just ditch the interface classes.

metricflow/model/semantics/semantic_model_lookup.py
tlento: Not my finest work, I admit. 🤣  

I think I was trying to say that this is how we load all of the models into the lookup class, as in any test cases or whatever that want to adjust the model set should just initialize something with a custom semantic manifest. 

Also I totally meant to change that to semantic_manifest but I clearly forgot. I'll fix this in a follow-up, there's another tweak I wanted to make to MetricLookup around this as well, and I may take a shot at getting rid of the protocols as well.

tlento (commented):
Just filed #1134 for the follow-ups here!

metricflow/model/semantics/semantic_model_lookup.py
courtneyholcomb: Heck yeah 💥

PR #1130, opened 2024-04-12
Remove redundant entity names from valid linkable specs

courtneyholcomb:
### Description
I've been seeing suggestions for entities like `Entity("listing__listing")` show up when querying. This should resolve that.


tlento (approved):
Please regenerate snapshots before merge, because I think my PRs added a few more of these.

PR #1131, opened 2024-04-12
Pass join_path where missed in group by resolution

courtneyholcomb:
### Description
Fixes a small oversight. This didn't seem to cause any issues so it might not be used anywhere, but maybe will be eventually.


tlento (approved):
I don't know if we use that join_path attribute for anything, but yeah, if we do, it sure seems like we needed it there. Thank you!

PR #1136, opened 2024-04-15
Update `get_semantic_models_for_measure` to return just one semantic model

courtneyholcomb:
### Description
Measure names are required to be unique across semantic models, so we can safely assume that `get_semantic_models_for_measure()` should return one semantic model per measure.

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -715,15 +715,6 @@ class DataflowPlanBuilder:
         """Returns true if any of the linkable specs requires a multi-hop join to realize."""
         return any(len(x.entity_links) > 1 for x in linkable_specs)
 
-    def _get_semantic_model_names_for_measures(self, measures: Sequence[MeasureSpec]) -> Set[str]:
-        """Return the names of the semantic models needed to compute the input measures.
-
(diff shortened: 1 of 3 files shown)


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -720,12 +720,9 @@ def _get_semantic_model_names_for_measures(self, measures: Sequence[MeasureSpec]
...
         """
-        semantic_model_names: Set[str] = set()
-        for measure in measures:
-            semantic_model_names = semantic_model_names.union(
-                {d.name for d in self._semantic_model_lookup.get_semantic_models_for_measure(measure.reference)}
-            )
-        return semantic_model_names
+        return {
+            self._semantic_model_lookup.get_semantic_model_for_measure(measure.reference).name for measure in measures
+        }
```
tlento, line 725: Do you mind deleting this function and moving this one-liner over to where we call it? I think the docstring is super duper outdated and your change here removed all of the complexity from getting this set.

tlento (approved):
Love it, thanks!

metricflow/dataflow/builder/dataflow_plan_builder.py
courtneyholcomb: Will do!

PR #1137, opened 2024-04-15
Remove unneeded protocols for lookup classes

courtneyholcomb:
### Description
Deletes protocols used for `SemanticModelLookup` and `MetricLookup` in an effort to reduce boilerplate code and improve developer experience when working with these classes. Discussed with @tlento [here](https://github.com/dbt-labs/metricflow/pull/1128#discussion_r1561942760).
Also copies over any missing docstrings from the base classes so that we won't lose that context.


metricflow/model/semantics/metric_lookup.py
```diff
@@ -17,13 +17,12 @@
...
 from metricflow.model.semantics.semantic_model_join_evaluator import MAX_JOIN_HOPS
 from metricflow.model.semantics.semantic_model_lookup import SemanticModelLookup
-from metricflow.protocols.semantics import MetricAccessor
 from metricflow.specs.specs import TimeDimensionSpec
 
 logger = logging.getLogger(__name__)
 
 
-class MetricLookup(MetricAccessor):
+class MetricLookup:
```
tlento, line 25: I think the `add_metric` method here is the only "write" method in this class, and it's not called anywhere. Mind just renaming it to `_add_metric` so we don't have any "write" methods in the public interface?

metricflow/model/semantics/semantic_model_lookup.py
```diff
@@ -40,10 +38,10 @@
 logger = logging.getLogger(__name__)
 
 
-class SemanticModelLookup(SemanticModelAccessor):
+class SemanticModelLookup:
     """Tracks semantic information for semantic models held in a set of SemanticModelContainers.
 
-    This implements the SemanticModelAccessors protocol, the interface type we use throughout the codebase.
+    This implements the SemanticModelLookups protocol, the interface type we use throughout the codebase.
```
tlento, line 44: Nit - remove this line and the next, they're outdated.

tlento (approved):
Huh. That was easier than I thought it would be. See inline about some added method cleanup. Fine to do a separate PR or just layer on a new commit for them, either way. Thanks!

PR #1138, opened 2024-04-15
Give enums singular names

courtneyholcomb:
### Description
Updates two enums to have singular names, as is conventional (and as we do with all our other enums).
`LinkableElementProperties` -> `LinkableElementProperty`
`RequiredDwEngineFeatures` -> `RequiredDwEngineFeature`


tlento (approved):
dancing and sweeping

PR #1139, opened 2024-04-15
Group by resolution for group by metrics

courtneyholcomb:
### Description
Refactor group by resolution for `LinkableMetrics`. This gets more explicit about what the join paths will should look like for `LinkableMetrics` and updates that logic accordingly. `LinkableMetrics` are different from other linkable elements for a couple of reasons because there is no such thing as a "local" `LinkableMetric`. There is always a join required to get to the metric subquery. To make this clear, here we reclassify `LinkableMetrics` as either single-hop or muti-hop, where previously they were considered either local or single-hop.

There was also a bug here related to how we resolved what linkable metrics are available. Previously, the code assumed that if a metric could be queried with one entity in a semantic model, it could be queried with any metric in that semantic model, but that's not true because (1) some entities might cause potential fan-out and (2) derived metrics need all input measures to be joinable to the entity. Here, we fix that bug by updating the logic to be based on entities, not semantic models.

Those are the main changes in this PR, but reviewing by commit would be helpful to get more granular context.

Coming after this: Query tests for metric filters with single hop and multi-hop joins + ensure DFP is using expected join paths.


metricflow/model/semantics/linkable_spec_resolver.py
```diff
@@ -862,6 +899,39 @@ def _get_joined_elements(self, measure_semantic_model: SemanticModel) -> Linkabl
...
+                            join_path and entity in join_path.entity_links
+                        ):
+                            continue
+                        join_path_elements = join_path.path_elements if join_path else ()
+                        new_join_path_element = SemanticModelJoinPathElement(
+                            semantic_model_reference=next_semantic_model.reference, join_on_entity=entity.reference
+                        )
+                        next_semantic_models_and_join_paths.append(
+                            (next_semantic_model, SemanticModelJoinPath(join_path_elements + (new_join_path_element,)))
+                        )
```
tlento, line 930: Does this open the door to fanout on the intermediate step? Like I have a measure and that joins to the foreign `listing` entity via `account` in some other semantic model, and then we end up with `account__listing__some_listing_metric` but there's fanout between account and listing so our metrics aren't necessarily correct.

I don't see any checks for join validity within the entity block here.

tlento (commented):
Overall this change seems reasonable but I don't understand why the snapshot outputs are changing so drastically. It looks like we were cutting off a lot of multi-hop joins as well.

What worries me about this, apart from my general driving need to understand how things work, is that the multi-hop case might be overly broad. For example, do we have any way of asserting that we are, in fact, excluding intermediate fanout join steps?

tests/snapshots/test_semantic_model_container.py/tuple/test_linkable_elements_for_no_metrics_query__result0.txt
```diff
@@ -152,9 +151,6 @@
   'guest__bookings_offset_twice',
   'guest__bookings_per_booker',
   'guest__bookings_per_dollar',
-  'guest__bookings_per_listing',
```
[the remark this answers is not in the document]
courtneyholcomb: They shouldn't have been allowed. The previous implementation allowed some joins that weren't valid. In these cases, they're all derived metrics from multiple input metrics where only one of the input metrics is actually able to join to the prefixed entity. We need all input metrics to be joinable to all group bys.

metricflow/model/semantics/linkable_spec_resolver.py
courtneyholcomb: My brain realized this problem just when I was trying to fall asleep last night 🙃

metricflow-semantics/metricflow_semantics/model/semantics/linkable_spec_resolver.py
```diff
@@ -521,17 +500,11 @@ def _get_linkable_element_set_for_measure(
         measure_semantic_model = self._get_semantic_model_for_measure(measure_reference)
 
         elements_in_semantic_model = self._get_elements_in_semantic_model(measure_semantic_model)
-        metrics_linked_to_semantic_model = self._get_joinable_metrics_for_semantic_model(measure_semantic_model)
```
tlento, line 524: Oh ok this has been moved into get_joined_elements. That makes sense.

tlento (commented):
Ok this seems reasonable but I know you've got a monster update coming.

It'd be good to have the final PR with the change explain what we expect to see in the snapshot updates based on what's changing in the logic. It's really hard for me to visualize what's happening inside of the LinkableSpecResolver in general, so it'll be good to have some documentation in the PR description around what we're expecting to see and why this change is happening.

PR #1142, opened 2024-04-22
List dimensions

courtneyholcomb:
Resolves #SL-2079

### Description
Expose `list_dimensions()` on `MFEngine`. Different from our existing implementation because it does not require any metric inputs, instead just getting the entire list of dimensions for the semantic manifest. This is needed for the GSheets integration.


DevonFulcher (approved):
Awesome, thank you! You may want Paul or Tom's approval too.

PR #1148, opened 2024-04-23
Types cleanup: `defined_from` -> `semantic_model_element_reference`

courtneyholcomb:
### Description
Removes `SemanticModelElementInstance.defined_from` property in favor of `SemanticModelElementInstance.semantic_model_element_reference` made these changes:
1. There could never be multiple items in the tuple. It either contained one `SemanticModelElementReference` or was empty. This resulted in a lot of unnecessary type checking boilerplate code scattered throughout the codebase. Instead, I updated the type to be one optional `SemanticModelElementReference`.
2. This still left a lot of type assertions to make sure it wasn't `None`, so I added the `checked_semantic_model_element_reference` property to use wherever that assertion would be needed.
3. There was a TODO on the `origin_semantic_model_reference` property stating: `TODO: make this a required input value, rather than a derived property on these objects` so i went ahead and did that, replacing `defined_from` with `origin_semantic_model_reference` as a dataclass attribute.
4. Then I realized that `origin_semantic_model_reference` is not actually a `SemanticModelReference` as the name suggests. It's a `SemanticModelElementReference`. So I changed the name to be `origin_semantic_model_element_reference`. That was really long. So I removed the `origin_` prefix, shortening it to `semantic_model_element_reference`.

### DIFF
--- a/metricflow/dataflow/builder/node_evaluator.py
+++ b/metricflow/dataflow/builder/node_evaluator.py
@@ -233,17 +233,10 @@ class NodeEvaluatorForLinkableInstances:
                 if entity_instance_in_right_node is None:
                     raise RuntimeError(f"Could not find entity instance with name ({entity_spec_in_right_node})")
 
-                assert (
-                    len(entity_instance_in_right_node.defined_from) == 1
-                ), f"Did not get exactly 1 defined_from in {entity_instance_in_right_node}"
-
-                entity_in_right_node = self._semantic_model_lookup.get_entity_in_semantic_model(
-                    entity_instance_in_right_node.defined_from[0]
-                )
+                element_reference = entity_instance_in_right_node.checked_origin_semantic_model_reference
+                entity_in_right_node = self._semantic_model_lookup.get_entity_in_semantic_model(element_reference)
                 if entity_in_right_node is None:
-                    raise RuntimeError(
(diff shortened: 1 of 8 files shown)


plypaul:
For context, it was intended to be a tuple as the `*Instance` classes describe an instance of a model element in the dataflow. With that in mind, it was supposed to be possible that it's a tuple with multiple items when a join occurs. After the join, the instance is would show that it's effectively related to multiple semantic models. It may be partially implemented though - taking a look.

plypaul:
Seems like a bug here: 

https://github.com/dbt-labs/metricflow/blob/5ad161b3c7a891632550c976ed9785cbc90dd481/metricflow/plan_conversion/instance_converters.py#L453

I vaguely recall that it was related to support composite entities, but not sure.

tlento:
> For context, it was intended to be a tuple as the *Instance classes describe an instance of a model element in the dataflow. With that in mind, it was supposed to be possible that it's a tuple with multiple items when a join occurs. After the join, the instance would show that it's effectively related to multiple semantic models. It may be partially implemented though - taking a look.

This was never implemented at all. The only scenario where this was ever anything but a single-element tuple was due to an override from the old caching implementation, in which case it would be empty.

The only times we ever read this property we immediately assert that it's of length 1.

PR #1152, opened 2024-04-24
Add an Interface to Resolve Saved-Query Dependencies

plypaul:
### Description

This PR adds the initial interface for a resolver to get the dependencies of a saved query.


metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
```diff
@@ -105,9 +121,20 @@ def path_key(self) -> ElementPathKey:  # noqa: D102
...
+            semantic_model_references.add(self.semantic_model_origin)
+        for join_path_item in self.join_path:
+            semantic_model_references.add(join_path_item.semantic_model_reference)
+
+        return sorted(semantic_model_references, key=lambda reference: reference.semantic_model_name)
+
 
 @dataclass(frozen=True)
-class LinkableEntity:
+class LinkableEntity(LinkableElement, SemanticModelDerivation):
```
tlento, line 137: You don't need both of these, right, just LinkableElement?

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
```diff
@@ -127,9 +154,18 @@ def path_key(self) -> ElementPathKey:  # noqa: D102
...
+        semantic_model_references = {self.semantic_model_origin}
+        for join_path_item in self.join_path:
+            semantic_model_references.add(join_path_item.semantic_model_reference)
+
+        return sorted(semantic_model_references, key=lambda reference: reference.semantic_model_name)
+
 
 @dataclass(frozen=True)
-class LinkableMetric:
+class LinkableMetric(LinkableElement, SemanticModelDerivation):
```
tlento, line 168: Same thing here, just LinkableElement?

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element_set.py
```diff
@@ -290,3 +288,104 @@ def only_unique_path_keys(self) -> LinkableElementSet:
...
+        return specs
+
+    def _path_key_to_spec(self, path_key: ElementPathKey) -> LinkableInstanceSpec:
+        if path_key.element_type is LinkableElementType.DIMENSION:
+            return DimensionSpec(
+                element_name=path_key.element_name,
+                entity_links=path_key.entity_links,
+            )
+        elif path_key.element_type is LinkableElementType.TIME_DIMENSION:
+            assert path_key.time_granularity is not None
```
tlento, line 338: nit: error message, maybe something like "type refinement, should have been checked in dataclass validation"

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element_set.py
```diff
@@ -290,3 +288,104 @@ def only_unique_path_keys(self) -> LinkableElementSet:
                 if len(set(linkable_metrics)) <= 1
             },
         )
+
+    @property
+    @override
+    def derived_from_semantic_models(self) -> Sequence[SemanticModelReference]:
```
tlento, line 294: Could we add some test cases for this and the filter by spec pattern operations? They're quite a bit more complex than the direct to spec conversion.

metricflow-semantics/metricflow_semantics/api/v0_1/saved_query_dependency_resolver.py
```diff
@@ -0,0 +1,75 @@
...
+                )
+            ),
+        )
+
+    def resolve_dependencies(self, saved_query_name: str) -> SavedQueryDependencySet:
+        """Return the dependencies of the given saved query in the manifest."""
+        try:
+            return self._resolve_dependencies(saved_query_name)
+        except Exception:
+            logger.exception(
```
tlento, line 64: Ok. We might need to do something about this log level for deployment - if it turns out to be common we'll want it to show up with level WARN or INFO instead of EXCEPTION - but I think we can do that via datadog instead of here.

tlento (approved):
Seems reasonable! I can make use of the LinkableELementSet return value - I was going to make a similar set of changes before all this started anyway.

Lots of small things inline. I think the main things are some added test coverage and the usual "I don't like visitors" thing. I'd like the tests done either separately as a follow-up or prior to merge, it's up to you.

metricflow-semantics/metricflow_semantics/query/query_parser.py
```diff
@@ -333,9 +338,9 @@ def parse_and_validate_query(
             order_by_names=order_by_names,
             order_by=order_by,
             min_max_only=min_max_only,
-        )
+        ).query_spec
 
-    @log_runtime()
+    # @log_runtime()
```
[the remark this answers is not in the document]
plypaul: That's a typo - updated.

metricflow-semantics/metricflow_semantics/specs/spec_set.py
```diff
@@ -0,0 +1,213 @@
...
+    def visit_group_by_metric_spec(self, group_by_metric_spec: GroupByMetricSpec) -> None:
+        self.group_by_metric_specs.append(group_by_metric_spec)
+
+    @override
+    def visit_metric_spec(self, metric_spec: MetricSpec) -> None:
+        self.metric_specs.append(metric_spec)
+
+    @override
+    def visit_metadata_spec(self, metadata_spec: MetadataSpec) -> None:
+        self.metadata_specs.append(metadata_spec)
```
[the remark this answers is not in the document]
plypaul: The rule I've been following is: no `isinstance` calls unless it's in a equals or comparison method. It's a code-smell, and when it's in the code base, it tends to spread.

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
plypaul: Right - updated.

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
plypaul: Updated.

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element_set.py
plypaul: Updated.

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element_set.py
plypaul: Add some test cases.

plypaul:
```
Note: this approval assumes the large blocks of green in cases where things existed before are copy/paste. I did not read those sections carefully.
```

That's correct. There shouldn't be significant logic changes in those sections.

metricflow-semantics/metricflow_semantics/collection_helpers/dedupe.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def ordered_dedupe(*iterables: Iterable[IterableT]) -> Tuple[IterableT, ...]:
+    """De-duplicates the items in the iterables while preserving the order."""
+    ordered_results: Dict[IterableT, None] = {}
+    for iterable in iterables:
+        for item in iterable:
+            ordered_results[item] = None
+
+    return tuple(ordered_results.keys())
```
tlento, line 15: This is a less robust custom implementation of the (admittedly poorly named) unique_everseen function in more-itertools.

Since dbt-semantic-interfaces depends on more-itertools we should add the dependency with the same version range and use it here as well.

tlento (commented):
The updates look good, thank you, let's get this merged as soon as it's ready. The only update you might want to make before merge is the more_itertools change but even that I can do later as I grind my way through this thing I'm trying to get done. The useless visitor chafes but I'm just going to start removing stuff like that so one extra diff in that stack won't matter much to me one way or the other.

metricflow-semantics/metricflow_semantics/collection_helpers/dedupe.py
tlento: Actually I just noticed this takes `*iterables`, so it's a less robust implementation of `unique_everseen(itertools.chain(iterables))`. The point still stands, we should use the libraries, especially where iterator collection helpers are concerned.

plypaul: Sure, we can make that change, but let me check about the "minimize dependencies" request and check how that method works with types.

metricflow-semantics/metricflow_semantics/specs/spec_set.py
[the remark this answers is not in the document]
plypaul: Since we had had this discussion a few times and to avoid having this thread be the place where we debate, let me just say that I see your points and I'm up for continued discussion on the issue.

metricflow-semantics/metricflow_semantics/collection_helpers/dedupe.py
tlento: I already looked at the dependencies, that's not an issue. The typing is another matter, more-itertools is an old library and these methods date back a long way.

PR #1159, opened 2024-04-30
Fix bug in snapshot generation

courtneyholcomb:
### Description
Snapshot bug fix. Updates the broken snapshot. I don't think it would be useful to look super closely at the snapshot that's changed, since I have some incoming changes that will change that dramatically. This just gets us to an accurate base so that we can compare the diff in coming PRs.


tlento (approved):
Easy peasy mac n cheesy.....

PR #1160, opened 2024-04-30
Map joinable metrics to entities instead of semantic models

courtneyholcomb:
### Description
This fixes a bug in the logic deciding which `LinkableMetrics` show up as options for a given query. Previously, we mapped semantic models to the metrics they could join to, but that wasn't specific enough. Sometimes there are entities in a semantic model that can join to a metric, while there are others that can't. Here we update the logic to map `LinkableMetrics` to entities instead.
Also changes the way we categorize `LinkableMetrics` in terms of `LinkableElementProperties`. Previously, we would assign properties based on joins like this: 
`1 join = (METRIC), 2 joins = (METRIC, JOINED), 3+ joins = (METRIC, JOINED, MULTI_HOP)`
Here we update the logic to assign properties like this:
`1 join = (METRIC, JOINED), 2+ joins = (METRIC, JOINED, MULTI_HOP)`
We've discussed this already as a team, and it just seems to make more sense since all `LinkableMetrics` require a join.

Note that at the top of this stack will be a test that loops through all allowed linkable metrics and runs them through the query parser + dataflow plan builder to ensure we ACTUALLY allow all the options we say we do.


metricflow-semantics/metricflow_semantics/model/semantics/linkable_spec_resolver.py
```diff
@@ -244,25 +217,34 @@ def _get_semantic_model_for_measure(self, measure_reference: MeasureReference) -
...
+        properties = frozenset({LinkableElementProperty.METRIC, LinkableElementProperty.JOINED})
+        if using_join_path:
+            assert (
+                semantic_model.reference == using_join_path.last_semantic_model_reference
+            ), "Last join path element should match semantic model when building LinkableMetrics."
+            properties = properties.union(frozenset({LinkableElementProperty.MULTI_HOP}))
+
+        path_key_to_linkable_metrics: Dict[ElementPathKey, Tuple[LinkableMetric, ...]] = {}
+        for entity_reference in [entity.reference for entity in semantic_model.entities]:
+            if using_join_path and entity_reference in using_join_path.entity_links:
```
tlento, line 233: I'm still not clear on why this is here. Is this to avoid entity link cycles?

I put this in the bigger PR but maybe a local variable to explain the purpose would help clear thing sup

metricflow-semantics/metricflow_semantics/model/semantics/linkable_spec_resolver.py
```diff
@@ -244,25 +217,34 @@ def _get_semantic_model_for_measure(self, measure_reference: MeasureReference) -
...
-                    )
+                        semantic_model_join_path=using_join_path,
+                    ),
                 )
-        return LinkableElementSet(
-            path_key_to_linkable_metrics={
-                linkable_metric.path_key: (linkable_metric,) for linkable_metric in linkable_metrics
-            },
-        )
+                path_key_to_linkable_metrics[linkable_metric.path_key] = (linkable_metric,)
```
tlento, line 245: This is overwriting rather than extending. Do we want to verify that they're all the same `linkable_metric`?

tlento (commented):
Seems reasonable. I left a couple of questions inline.

metricflow-semantics/metricflow_semantics/model/semantics/linkable_spec_resolver.py
courtneyholcomb: Correct, avoiding a cycle!
Oops I should go back and make sure I addressed all your previous comments...

metricflow-semantics/metricflow_semantics/model/semantics/linkable_spec_resolver.py
courtneyholcomb: Ok yeah I fixed this in a later PR (why is this stack such a mess) but I can add a fix here. Notable though that that is the same behavior for all linkable elements in functions where we build `LinkableDimensions` and `LinkableEntities`, so I can fix this for those too.

courtneyholcomb: Updated!

metricflow-semantics/metricflow_semantics/model/semantics/linkable_spec_resolver.py
```diff
@@ -658,13 +666,19 @@ def create_linkable_element_set_from_join_path(
                     )
                 )
 
+        path_key_to_linkable_dimensions: Dict[ElementPathKey, Tuple[LinkableDimension, ...]] = {}
```
tlento, line 670: Same thing down here. Sigh. Sorry about that.

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
```diff
@@ -312,21 +312,10 @@ class MetricSubqueryJoinPath:
...
-            )
-
     @property
     def entity_links(self) -> Tuple[EntityReference, ...]:  # noqa: D102
-        return (
-            self.semantic_model_join_path.entity_links
-            if self.semantic_model_join_path
-            else (self.metric_subquery_join_path_element.join_on_entity,)
+        return (self.semantic_model_join_path.entity_links if self.semantic_model_join_path else ()) + (
+            self.metric_subquery_join_path_element.join_on_entity,
```
tlento, line 318: lol-lollerskates

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
```diff
@@ -312,21 +312,10 @@ class MetricSubqueryJoinPath:
...
     semantic_model_join_path: Optional[SemanticModelJoinPath] = None
 
-    def __post_init__(self) -> None:  # noqa: D105
-        if self.semantic_model_join_path:
-            assert (
-                self.metric_subquery_join_path_element.join_on_entity == self.semantic_model_join_path.last_entity_link
-            ), (
-                "Last entity link for `semantic_model_join_path` must match `semantic_model_join_path.join_on_entity`."
-                "This is the entity used to join the metric subquery to the outer query.",
-            )
```
tlento, line 322: Yeah, sorry about that.

tlento (approved):

PR #1162, opened 2024-04-30
`LinkableMetric` tests

courtneyholcomb:
### Description
Write some tests:
- one that shows what linkable elements are allowed in a multi-hop model
- one that tests filtering a metric by itself


courtneyholcomb:
@tlento wondering if you missed this one? Asking because it's fairly uncontroversial, and it's at the bottom of the stack!

tlento:
I sure did. I think it was still marked as draft mode when I looked at it the first time. Or maybe I just got confused.

tests_metricflow/snapshots/test_metric_filter_rendering.py/SqlQueryPlan/BigQuery/test_metric_filtered_by_itself__plan0_optimized.sql
```diff
@@ -0,0 +1,39 @@
...
+      guest_id AS guest
+      , COUNT(DISTINCT guest_id) AS bookers
+    FROM ***************************.fct_bookings bookings_source_src_28000
+    GROUP BY
+      guest
+  ) subq_26
+  ON
+    subq_20.guest = subq_26.guest
+) subq_28
+WHERE guest__bookers > 1.00
```
tlento, line 39: LOL. This is maybe not the best example query since the value of `bookers` will always be 1 here, but it's fine.

tests_metricflow/query_rendering/test_metric_filter_rendering.py
```diff
@@ -229,4 +229,28 @@ def test_distinct_values_query_with_metric_filter(
...
+    dataflow_plan_builder: DataflowPlanBuilder,
+    sql_client: SqlClient,
+    dataflow_to_sql_converter: DataflowToSqlQueryPlanConverter,
+    query_parser: MetricFlowQueryParser,
+) -> None:
+    """Tests a query for a metric that filters by the same metric."""
+    query_spec = query_parser.parse_and_validate_query(
+        metric_names=("bookers",),
+        where_constraint=PydanticWhereFilter(
+            where_sql_template="{{ Metric('bookers', ['guest']) }} > 1.00",
```
tlento, line 245: Just to have a slightly less awkward query (the current one will return 0 rows), shall we do `bookings` instead of `bookers` or something?

It is nice to see the distinct counts applied in the right place, so there is that. Another option is to key on `listing` instead of `guest`.

tlento (approved):
Sorry about that, I didn't notice this one until the end. It's fine as-is but maybe a less silly query would be nice.

tests_metricflow/query_rendering/test_metric_filter_rendering.py
courtneyholcomb: Yeah fair 😂 if you don't mind I'm going to add that on top of the stack, because I spent half of my day rebasing these PRs and I don't wanna do it again 😅

tlento: I'm a big fan of adding fixup changes to the top of the stack.

PR #1163, opened 2024-04-30
Track 2 join paths for `LinkableMetrics`

courtneyholcomb:
### Description
`LinkableMetrics` are extra complicated because they involve multiple join paths: 1) the join path needed to join the metric subquery to the outer query, and 2) the join path used in the subquery to join the metric to the group by entity. So far we have only been tracking join path 1. This commit adds tracking for join path 2.
This is necessary because some metrics are accessible via multiple join paths, and the user might need to specify which one they're looking for since we error if the join path is ambiguous.
**NOTE:** There are now duplicates in the linkable element snapshots - this is because the snapshots are only showing one of the two join paths. This also means that we could have duplicate SQL column names. I have a fix coming for this soon.


tlento (approved):
Ok this seems reasonable. The naming is super hard - every time I thought I had a name that would be easier to understand I realized it was not any better (and often was even more confusing), but we can think about that and see if we can improve things later.

PR #1164, opened 2024-04-30
Use query parser to build group by metric source node

courtneyholcomb:
Needed to resolve the appropriate join path for multi-hop joins. Moved source node-building logic to query parser class to avoid cirular import errors. This change also required adding the query parser as a dependency for the dataflow plan builder.

### Description
This is needed to resolve the appropriate join path for multi-hop joins. Currently users won't be able to pass a join path for those, so we need to build it ourselves.
Moved source node-building logic to query parser class to avoid circular import errors. This change also required adding the query parser as a dependency for the dataflow plan builder.
This doesn't change the output of any existing tests, but I have some relevant tests coming in future PRs.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/query/query_parser.py
+++ b/metricflow-semantics/metricflow_semantics/query/query_parser.py
@@ -33,6 +33,7 @@ from metricflow_semantics.protocols.query_parameter import (
 )
 from metricflow_semantics.query.group_by_item.filter_spec_resolution.filter_pattern_factory import (
     DefaultWhereFilterPatternFactory,
+    WhereFilterPatternFactory,
 )
 from metricflow_semantics.query.group_by_item.group_by_item_resolver import GroupByItemResolver
 from metricflow_semantics.query.group_by_item.resolution_dag.dag import GroupByItemResolutionDag
@@ -82,14 +83,18 @@ class MetricFlowQueryParser:
     TODO: Add fuzzy match results.
     """
 
-    def __init__(self, semantic_manifest_lookup: SemanticManifestLookup) -> None:  # noqa: D107
+    def __init__(  # noqa: D107
+        self,
+        semantic_manifest_lookup: SemanticManifestLookup,
+        where_filter_pattern_factory: WhereFilterPatternFactory = DefaultWhereFilterPatternFactory(),
+    ) -> None:
         self._manifest_lookup = semantic_manifest_lookup
         self._metric_naming_schemes = (MetricNamingScheme(),)
         self._group_by_item_naming_schemes = (
             ObjectBuilderNamingScheme(),
             DunderNamingScheme(),
         )
-        self._where_filter_pattern_factory = DefaultWhereFilterPatternFactory()
+        self._where_filter_pattern_factory = where_filter_pattern_factory
 
     def parse_and_validate_saved_query(
         self,
@@ -524,7 +529,7 @@ class MetricFlowQueryParser:
     def build_query_spec_for_group_by_metric_source_node(
         self, group_by_metric_spec: GroupByMetricSpec
     ) -> MetricFlowQuerySpec:
-        """Query spec that can be used to build a source node for this spec in the DFP."""
+        """Query spec that can be used to build a source node for this spec in the DataflowPlanBuilder."""
         return self.parse_and_validate_query(
             metrics=(MetricParameter(group_by_metric_spec.reference.element_name),),
             group_by=(DimensionOrEntityParameter(group_by_metric_spec.entity_spec.qualified_name),),
--- a/metricflow-semantics/metricflow_semantics/query/query_parser.py
+++ b/metricflow-semantics/metricflow_semantics/query/query_parser.py
@@ -55,8 +55,10 @@ from metricflow_semantics.query.resolver_inputs.query_resolver_inputs import (
 from metricflow_semantics.specs.patterns.base_time_grain import BaseTimeGrainPattern
 from metricflow_semantics.specs.patterns.metric_time_pattern import MetricTimePattern
 from metricflow_semantics.specs.patterns.none_date_part import NoneDatePartPattern
+from metricflow_semantics.specs.query_param_implementations import DimensionOrEntityParameter, MetricParameter
 from metricflow_semantics.specs.query_spec import MetricFlowQuerySpec
 from metricflow_semantics.specs.spec_classes import (
+    GroupByMetricSpec,
     InstanceSpec,
     TimeDimensionSpec,
 )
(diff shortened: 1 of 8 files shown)


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -33,14 +33,14 @@
 from metricflow_semantics.query.group_by_item.filter_spec_resolution.filter_spec_lookup import (
     FilterSpecResolutionLookUp,
 )
+from metricflow_semantics.query.query_parser import MetricFlowQueryParser
```
tlento, line 36: Oof. That isn't great. Not sure what our alternatives are here.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -836,7 +830,8 @@ def _find_dataflow_recipe(
         # MetricGroupBy source nodes could be extremely large (and potentially slow).
         candidate_nodes_for_right_side_of_join += [
             self._build_query_output_node(
-                query_spec=self._query_spec_for_source_node(group_by_metric_spec), for_group_by_source_node=True
+                query_spec=self._query_parser.build_query_spec_for_group_by_metric_source_node(group_by_metric_spec),
```
tlento, line 833: Ok, I did a fair bit of digging and in order to avoid this query parser -> dataflow plan builder -> query parser -> dataflow plan builder loop we'd have to re-order most of the meaningful operations between engine initialization and here, so I don't think there's any other reasonable way to do this.

Another options we might consider is moving this into the SourceNodeBuilder and making the SourceNodeBuilder a property of the DataflowPlanBuilder. Then we could access these as a set of "metric source nodes" and avoid direct access to the query parser here.

tlento (commented):
I'll comment here and leave it open for one round of discussion - what do you think about encapsulating the group by metric node generation in the SourceNodeBuilder? We can have a separate method that just returns the set of metric source nodes associated with the linkable specs we pass in. That way we aren't exposing the entire query parser to the DataflowPlanBuilder, but we will need to include the SourceNodeBuilder (which shouldn't be an issue, really).

I think that might be a little cleaner than wiring the query parser through here.

metricflow/dataflow/builder/dataflow_plan_builder.py
courtneyholcomb: I have no problem doing that - but curious why that seems better to you?

courtneyholcomb: i.e., what is so bad about wiring the query parser through here?

tlento: The downside of directly importing the query parser is it opens the door to weird cycles where we could, at any point in dataflow plan construction, decide to parse another query and build another dataflow plan. Super powerful, hard to reason about, so I'd rather avoid making the query parser available here until we really have no other choice. If everything is behind the source node builder we make it more obvious that these special processing directives are scope-limited to source nodes, rather than something we can apply to arbitrary mid-DAG nodes.

There's also an upside to moving this call to the SourceNodeBuilder - these GroupByMetric objects are a special type of source node in the DAG. Even though they'd be typed differently from the existing source nodes, bundling their accessor in the provider we created for source nodes feels more natural to me than doing an inlined query parser call.

So on balance I have a preference for moving this out if that doesn't cause annoying dependency cycles somewhere else.

courtneyholcomb:
@tlento - rewrote this PR to put that method on the source node builder, as discussed!
Commit history is new so you can start over here.

metricflow-semantics/metricflow_semantics/query/query_parser.py
```diff
@@ -81,18 +82,14 @@ class MetricFlowQueryParser:
     TODO: Add fuzzy match results.
     """
 
-    def __init__(  # noqa: D107
-        self,
-        semantic_manifest_lookup: SemanticManifestLookup,
-        where_filter_pattern_factory: WhereFilterPatternFactory = DefaultWhereFilterPatternFactory(),
```
tlento, line 87: Wait, we still use this in MFS. I think all of us want to remove the WhereFilterPatternFactory altogether, but we can't do it just yet unless we move the MFS compatibility shims into MetricFlow itself. In theory we won't need it after end of May so I'm inclined to just wait.

metricflow-semantics/metricflow_semantics/query/query_parser.py
```diff
@@ -524,6 +521,15 @@ def _parse_and_validate_query(
             queried_semantic_models=query_resolution.queried_semantic_models,
         )
 
+    def build_query_spec_for_group_by_metric_source_node(
+        self, group_by_metric_spec: GroupByMetricSpec
+    ) -> MetricFlowQuerySpec:
+        """Query spec that can be used to build a source node for this spec in the DFP."""
```
tlento, line 527: micro-nit, I don't like acronyms, it's a personal problem, so if this makes the line too long you can ignore it:
```suggestion
        """Query spec that can be used to build a source node for this spec in the DataflowPlanBuilder."""
```

metricflow/dataflow/builder/source_node.py
```diff
@@ -95,3 +97,11 @@ def create_from_data_sets(self, data_sets: Sequence[SemanticModelDataSet]) -> So
             source_nodes_for_group_by_item_queries=tuple(group_by_item_source_nodes) + (self._time_spine_source_node,),
             source_nodes_for_metric_queries=tuple(source_nodes_for_metric_queries),
         )
+
+    def build_source_node_for_group_by_metric(self, group_by_metric_spec: GroupByMetricSpec) -> MetricFlowQuerySpec:
+        """Build source node used to satisfy requested group by metrics.
+
+        This is just a wrapper around the query parser method, stored here to limit the scope of the DataFlowPlanBuilder's
+        dependency on the query parser to only source nodes.
+        """
```
tlento, line 106: Documentation and naming nit - this is not building the source node itself, it's actually providing the query spec so the DataflowPlanBuilder can do that. Maybe this instead?
```suggestion
    def build_source_node_inputs_for_group_by_metric(self, group_by_metric_spec: GroupByMetricSpec) -> MetricFlowQuerySpec:
        """Build source node inputs used to satisfy requested group by metrics.

        Group by metrics are essentially metric subqueries that operate as source nodes in
    the DataflowPlanBuilder. We provide the inputs here because they require an entire
    DataFlowPlan generation step.

    This is just a wrapper around the query parser method, stored here to limit the scope of the DataFlowPlanBuilder's
        dependency on the query parser to only source nodes.
        """
```

tlento (commented):
Oops, meant to leave a note - looks good apart from some minor nits and the part where it breaks MFS.

metricflow-semantics/metricflow_semantics/query/query_parser.py
courtneyholcomb: Ahh got it! Will update

courtneyholcomb:
Addressed your comments! This one is now at the bottom of the stack.

metricflow-semantics/tests_metricflow_semantics/query/test_ambiguous_entity_path.py
```diff
@@ -20,10 +17,7 @@
 def multi_hop_query_parser(  # noqa: D103
     simple_multi_hop_join_manifest_lookup: SemanticManifestLookup,
 ) -> MetricFlowQueryParser:
-    return MetricFlowQueryParser(
-        semantic_manifest_lookup=simple_multi_hop_join_manifest_lookup,
-        where_filter_pattern_factory=DefaultWhereFilterPatternFactory(),
```
tlento, line 25: Oh yeah, nice.

tlento (approved):
Let's go!

PR #1166, opened 2024-04-30
Handle both join paths in `GroupByMetricPattern`

courtneyholcomb:
### Description
The main purpose of this PR is to ensure we parse both join paths from the `MetricCallParameterSet` and handle them in the `GroupByMetricPattern`. This piece is a bit hacky because we currently don't support a couple things that we eventually will: 1) multiple dimensions & entities in the `group_by` param for metric filters, and 2) an `entity_links` param on metric filters. That will get cleaned up in the next iterations of this feature.
The second commit here updates the metric subquery entity path (on various classes) to include the last entity link, where previously it didn't. This is due to a bug I found in the case when you have no entity links left for the outer query - which happens when you have already completed all joins needed to get to the metric subquery.

### DIFF
--- a/metricflow/plan_conversion/dataflow_to_sql.py
+++ b/metricflow/plan_conversion/dataflow_to_sql.py
@@ -604,6 +604,11 @@ class DataflowToSqlQueryPlanConverter(DataflowPlanNodeVisitor[SqlDataSet]):
         output_instance_set = output_instance_set.transform(ChangeAssociatedColumns(self._column_association_resolver))
         output_instance_set = output_instance_set.transform(RemoveMetrics())
 
+        if node.for_group_by_source_node:
+            assert (
+                len(node.metric_specs) == 1 and len(output_instance_set.entity_instances) == 1
+            ), "Group by metrics currently only support exactly one metric grouped by exactly one entity."
+
         non_metric_select_column_set: SelectColumnSet = output_instance_set.transform(
             CreateSelectColumnsForInstances(
                 table_alias=from_data_set_alias,
@@ -614,6 +619,7 @@ class DataflowToSqlQueryPlanConverter(DataflowPlanNodeVisitor[SqlDataSet]):
         # Add select columns that would compute the metrics to the select columns.
         metric_select_columns = []
         metric_instances = []
+        group_by_metric_instance: Optional[GroupByMetricInstance] = None
         for metric_spec in node.metric_specs:
             metric = self._metric_lookup.get_metric(metric_spec.reference)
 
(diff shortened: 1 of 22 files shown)


metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -727,20 +727,26 @@ def visit_compute_metrics_node(self, node: ComputeMetricsNode) -> SqlDataSet:
...
+
+        transform_func: InstanceSetTransform = AddMetrics(metric_instances)
+        if node.for_group_by_source_node:
+            assert (
+                len(metric_instances) == 1 and len(output_instance_set.entity_instances) == 1
+            ), "Group by metrics currently only support exactly one metric grouped by exactly one entity."
+            metric_instance = metric_instances[0]
+            entity_instance = output_instance_set.entity_instances[0]
+            group_by_metric_instance = GroupByMetricInstance(
+                associated_columns=metric_instance.associated_columns,
```
tlento, line 739: Oh wow, that seems important...

tests_metricflow/snapshots/test_metric_filter_rendering.py/SqlQueryPlan/BigQuery/test_distinct_values_query_with_metric_filter__plan0_optimized.sql
```diff
@@ -11,7 +11,7 @@ FROM (
   FULL OUTER JOIN (
     -- Aggregate Measures
     -- Compute Metrics via Expressions
-    -- Pass Only Elements: ['listing', 'bookings']
+    -- Pass Only Elements: ['listing', 'listing__bookings']
```
tlento, line 14: Wait, this is weird, the `listing__bookings` alias isn't applied in the rendered SQL internally, so it seems like this comment shouldn't be updated like this.

I don't think it matters but it indicates some potential for edge case mismatches, especially since this is only happening in the optimized plans. Is there some odd interaction with optimizers we need to look out for here?

tlento (commented):
Code changes seem reasonable to me given the temporary nature of some of the internal, but I'm a little concerned about the weirdness with the optimizers. Is that something you fix later, or something you already know about?

tests_metricflow/snapshots/test_metric_filter_rendering.py/SqlQueryPlan/BigQuery/test_distinct_values_query_with_metric_filter__plan0_optimized.sql
courtneyholcomb: Good question - I had not noticed this so will look into it!

courtneyholcomb: THIS WAS JUST DUMB. It's only on the BigQuery and Databricks snapshots, I forgot to rerun them after changing some stuff 🤦‍♀️  adding SQL engine tests to this PR to double check all engines

courtneyholcomb: But also prompted me to add a little cleanup - I was building the output column name from the metric spec instead of the group by metric spec. They were the same so it didn't matter, but just in case that changes at some point I updated the logic to pull from the correct spec.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -713,43 +719,38 @@ def visit_compute_metrics_node(self, node: ComputeMetricsNode) -> SqlDataSet:
...
+                output_column_association = self._column_association_resolver.resolve_spec(metric_spec)
+                metric_instances.append(
+                    MetricInstance(
+                        associated_columns=(output_column_association,),
+                        defined_from=defined_from,
+                        spec=metric_spec,
+                    )
                 )
+            metric_select_columns.append(
+                SqlSelectColumn(expr=metric_expr, column_alias=output_column_association.column_name)
```
tlento, line 747: This is super awkward with the assertion on line 607 serving as a guard for single-metric-special-case logic that mostly manifests on line 751 (!!!).

I don't think we should fix it here at all - you need to merge this and move on - but this whole function has clearly gotten completely out of control.

tlento (approved):
Ok let's approve this time instead of forgetting to approve this time. Sorry about that.

PR #1167, opened 2024-04-30
Store entity links on `MetricSubqueryJoinPathElement`

courtneyholcomb:
### Description
Store entity links on `MetricSubqueryJoinPathElement`.
Originally I thought this wasn't necessary since the entity links should be a duplicate of the entities in the join path. Later I realized that there is one scenario where they don't match the join path. We allow querying local entities both alone and prefixed with any of the other entities in the same semantic model. If you query a local entity with another entity prefix, the entity links won't match the join path. To account for that, we need to store both entity links and join path on `MetricSubqueryJoinPathElement`.


tlento (approved):
Makes sense, I wondered about this earlier but it all seemed fine at the time. We're adding more duplicates here (local plus linked entities, I guess?) but it looks like that'll be addressed up-stack, so I'll look at that in a bit.

nit: your PR description is duplicated.

PR #1168, opened 2024-04-30
Remove metrics that must be queried with `metric_time` from linkable metric options

courtneyholcomb:
### Description
Remove metrics that must be queried with `metric_time` from linkable metric options. That means cumulative metrics with a window or grain to date, as well as any time offset metrics. This restriction can be removed when we add support for using multiple entities/dimensions in the group by for metric filters.


tlento (approved):
Seems reasonable!

PR #1169, opened 2024-04-30
Update DFP join logic to account for group by metric source nodes

courtneyholcomb:
### Description
There was a lot of logic in the dataflow plan builder that assumed all source nodes and joinable nodes would be `ReadSqlSourceNode` or `MetricTimeDimensionTransformNode` (as was previously true). This is no longer true since adding `LinkableMetrics`, for which you might have a `ComputeMetricsNode` as a source node. This comes with a lot of changes, like the fact that the entity links might look different for this source node than for others. Here we update the DFP building logic to account for those nodes.

### DIFF
--- a/metricflow/dataflow/builder/node_evaluator.py
+++ b/metricflow/dataflow/builder/node_evaluator.py
@@ -223,11 +223,6 @@ class NodeEvaluatorForLinkableInstances:
             # then produce the linkable spec. See comments further below for more details.
 
             for entity_spec_in_right_node in entity_specs_in_right_node:
-                # If an entity has links, what that means and whether it can be used is unclear at the moment,
-                # so skip it.
-                if len(entity_spec_in_right_node.entity_links) > 0:
-                    continue
-
                 entity_instance_in_right_node = None
                 for instance in data_set_in_right_node.instance_set.entity_instances:
                     if instance.spec == entity_spec_in_right_node:
--- a/metricflow/dataflow/builder/node_evaluator.py
+++ b/metricflow/dataflow/builder/node_evaluator.py
@@ -40,7 +40,6 @@ from metricflow.dataflow.builder.partitions import (
     PartitionTimeDimensionJoinDescription,
 )
 from metricflow.dataflow.dataflow_plan import BaseOutput
-from metricflow.dataflow.nodes.compute_metrics import ComputeMetricsNode
 from metricflow.dataflow.nodes.filter_elements import FilterElementsNode
 from metricflow.dataflow.nodes.join_to_base import JoinDescription, ValidityWindowJoinDescription
 from metricflow.dataflow.nodes.metric_time_transform import MetricTimeDimensionTransformNode
@@ -257,17 +256,19 @@ class NodeEvaluatorForLinkableInstances:
                 assert len(entity_instance_in_left_node.defined_from) == 1
                 assert len(entity_instance_in_right_node.defined_from) == 1
 
-                if not self._join_evaluator.is_valid_semantic_model_join(
-                    left_semantic_model_reference=entity_instance_in_left_node.defined_from[0].semantic_model_reference,
-                    right_semantic_model_reference=entity_instance_in_right_node.defined_from[
-                        0
-                    ].semantic_model_reference,
-                    on_entity_reference=entity_spec_in_right_node.reference,
+                if not (
+                    self._join_evaluator.is_valid_semantic_model_join(
+                        left_semantic_model_reference=entity_instance_in_left_node.defined_from[
+                            0
+                        ].semantic_model_reference,
+                        right_semantic_model_reference=entity_instance_in_right_node.defined_from[
+                            0
+                        ].semantic_model_reference,
+                        on_entity_reference=entity_spec_in_right_node.reference,
+                    )
+                    or right_node.is_aggregated_to_elements == {entity_spec_in_right_node.reference}
                 ):
-                    # If joining to ComputeMetricsNode, the right node is pre-aggregated.
-                    # Since we currently only allow one entity on GroupByMetric, this won't fan out.
-                    if not isinstance(right_node, ComputeMetricsNode):
-                        continue
+                    continue
 
                 linkless_entity_spec_in_node = LinklessEntitySpec.from_element_name(
                     entity_spec_in_right_node.element_name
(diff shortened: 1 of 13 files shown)


metricflow/validation/dataflow_join_validator.py
```diff
@@ -43,15 +40,36 @@ def is_valid_instance_set_join(
         left_instance_set: InstanceSet,
         right_instance_set: InstanceSet,
         on_entity_reference: EntityReference,
+        right_node_is_subquery: bool = False,
```
tlento, line 43: Let's rename this parameter to `right_node_is_aggregated`

tlento (approved):
I think we should replace the isinstance checks here and in the node evaluator with a node-level property if you haven't already done that, that seems less fragile and easier for a reader to understand to me.

I also think that change is probably best done in a separate PR, so I'm approving this and you can add one more to the top of the stack.

metricflow/dataflow/builder/node_evaluator.py
```diff
@@ -223,11 +223,6 @@ def _find_joinable_candidate_nodes_that_can_satisfy_linkable_specs(
             # then produce the linkable spec. See comments further below for more details.
 
             for entity_spec_in_right_node in entity_specs_in_right_node:
-                # If an entity has links, what that means and whether it can be used is unclear at the moment,
-                # so skip it.
-                if len(entity_spec_in_right_node.entity_links) > 0:
```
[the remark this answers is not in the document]
courtneyholcomb: Correct!

metricflow/plan_conversion/node_processor.py
```diff
@@ -217,6 +215,8 @@ def _get_candidates_nodes_for_multi_hop(
                     left_instance_set=data_set_of_first_node_that_could_be_joined.instance_set,
                     right_instance_set=data_set_of_second_node_that_can_be_joined.instance_set,
                     on_entity_reference=entity_reference_to_join_first_and_second_nodes,
+                    # TODO: make this check more substantial in V2
+                    right_node_is_subquery=isinstance(second_node_that_could_be_joined, ComputeMetricsNode),
```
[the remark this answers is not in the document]
courtneyholcomb: Yeah, I thought this might be controversial 😅 
I guess I don't love this solution, so I'm trying to think of a better one. What if there is another node that technically is aggregated? We probably do have other nodes that are aggregated - like in `FilterElementsNode` we sometimes aggregate to the distinct values.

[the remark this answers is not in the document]
courtneyholcomb: Ok I'm thinking instead of `BaseOutput.is_aggregated -> bool`, I'll add something like `BaseOutput.is_aggregated_to_elements -> Sequence[LinkableElementSpec]`. That way we can store the list of entities & dimensions it's aggregated to, if possible. Should be useful for V2 of this feature, too.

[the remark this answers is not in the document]
courtneyholcomb: Ok I added that property! Took a few new commits so you can start with https://github.com/dbt-labs/metricflow/pull/1169/commits/8951ea06c64918ccfff060ad42345b45b9ded249. I'm pretty happy with this change, makes the join checks much more secure 👍

courtneyholcomb:
Oops, I missed your note about making that change in a separate PR, so I'm guessing you'll want to re-review this! @tlento

metricflow/dataflow/nodes/compute_metrics.py
```diff
@@ -19,7 +20,11 @@ class ComputeMetricsNode(ComputedMetricsOutput):
     """A node that computes metrics from input measures. Dimensions / entities are passed through."""
 
     def __init__(
-        self, parent_node: BaseOutput, metric_specs: Sequence[MetricSpec], for_group_by_source_node: bool = False
+        self,
+        parent_node: BaseOutput,
+        metric_specs: Sequence[MetricSpec],
+        is_aggregated_to_elements: Set[LinkableElementReference],
```
tlento, line 26: nit again: `aggregated_to_elements` - this is no longer a boolean.

metricflow/dataflow/builder/node_evaluator.py
```diff
@@ -257,17 +256,19 @@ def _find_joinable_candidate_nodes_that_can_satisfy_linkable_specs(
...
+                    self._join_evaluator.is_valid_semantic_model_join(
+                        left_semantic_model_reference=entity_instance_in_left_node.defined_from[
+                            0
+                        ].semantic_model_reference,
+                        right_semantic_model_reference=entity_instance_in_right_node.defined_from[
+                            0
+                        ].semantic_model_reference,
+                        on_entity_reference=entity_spec_in_right_node.reference,
+                    )
+                    or right_node.is_aggregated_to_elements == {entity_spec_in_right_node.reference}
```
tlento, line 269: nit: the formatting makes this very hard to read, and kind of hides the function of the `or` in here. Can we store these as locals, like:

```
is_valid_semantic_model_join = self._join_evaluator.is_valid_semantic_model_join(....)
has_matching_aggregation_elements = right_node.is_aggregated_to_elements == {entity_spec_in_right_node.reference}
if not (is_valid_semantic_model_join or has_matching_aggregation_elements):
    continue
```

metricflow/validation/dataflow_join_validator.py
```diff
@@ -40,13 +40,13 @@ def is_valid_instance_set_join(
         left_instance_set: InstanceSet,
         right_instance_set: InstanceSet,
         on_entity_reference: EntityReference,
-        right_node_is_subquery: bool = False,
+        right_node_is_aggregated_to_entity: bool = False,
```
tlento, line 43: Oh neat, so we can (probably) swap this to the elements and do a straight comparison later.

tlento (approved):
YES! I've been thinking that keeping the aggregation element set for comparison was a thing we'd need to do, so I'm glad you skipped ahead to making it happen. There are a couple of ways we could approach this but this looks very promising, and much better than a boolean flag.

metricflow/dataflow/nodes/compute_metrics.py
courtneyholcomb: DANGIT ok I'm going to add that to the top of the stack instead of fixing here

metricflow/dataflow/builder/node_evaluator.py
courtneyholcomb: Yes! Will also put on top of stack

PR #1170, opened 2024-04-30
Update `AddGroupByMetrics` to be singular `AddGroupByMetric`

courtneyholcomb:
…y ever accept one group by metric

### Description
Just a cleanup item. This should only accept one group by metric, so let's name it and type it appropriately.


tlento (approved):
Sure, why not?

PR #1172, opened 2024-04-30
Dedupe `LinkableMetrics` in snapshots

courtneyholcomb:
### Description
Show both the outer and inner query join paths for linkable metrics when generating test snapshots. This fixes the issue of duplicates showing up in these snapshots.


metricflow-semantics/metricflow_semantics/test_helpers/snapshot_helpers.py
```diff
@@ -301,7 +301,14 @@ def assert_linkable_element_set_snapshot_equal(  # noqa: D103
                         if linkable_metric.join_by_semantic_model
                         else ""
                     ),
-                    tuple(entity_link.element_name for entity_link in linkable_metric.join_path.entity_links),
+                    (
+                        str(tuple(entity_link.element_name for entity_link in linkable_metric.join_path.entity_links)),
```
tlento, line 305: This fixes the symptom but why are they showing up multiple times, anyway? Where's the ambiguity coming from?

tlento (commented):
Is it a code issue that these duplicates exist, rather than a test file layout one?

If not, do we still want to dedupe in the test file? It seems like we'd want this to reflect what the code is doing, duplicates and all.

metricflow-semantics/tests_metricflow_semantics/snapshots/test_linkable_spec_resolver.py/str/test_create_linkable_element_set_from_join_path__result0.txt
```diff
@@ -1,126 +1,126 @@
...
+listings_latest   ("('listing', 'user')", "('user',)")                     visit_buy_conversion_rate_7days_fill_nulls_with_0                                     ['JOINED', 'METRIC', 'MULTI_HOP']
+listings_latest   ("('listing', 'user')", "('user',)")                     visit_buy_conversion_rate_by_session                                                  ['JOINED', 'METRIC', 'MULTI_HOP']
+listings_latest   ("('listing', 'user')", "('user',)")                     visit_buy_conversions                                                                 ['JOINED', 'METRIC', 'MULTI_HOP']
+listings_latest   ("('listing', 'user')", "('verification', 'user')")      identity_verifications                                                                ['JOINED', 'METRIC', 'MULTI_HOP']
+listings_latest   ("('listing', 'user')", "('view', 'user')")              views                                                                                 ['JOINED', 'METRIC', 'MULTI_HOP']
+listings_latest   ("('listing', 'user')", "('visit', 'user')")             visit_buy_conversion_rate                                                             ['JOINED', 'METRIC', 'MULTI_HOP']
```
tlento, line 78: I made my browser bigger so this is a little more obvious to me, but I still don't understand what's happening here. Is the second tuple a different internal join path for resolving the visit_buy_conversions metric? But then isn't that an ambiguous join? How do we know which one to use?

courtneyholcomb: The first tuple is the entity links in the OUTER query (how we join to the metric subquery).
The second tuple is the entity links in the INNER query (how the metric joins to the entity). There can be multiple options for this path, even when the outer path is the same.
That does make the second one ambiguous, but can be deduped by what they pass into the filter! Currently they would need to specify the entity links via dundered syntax, but I have a task on my list to support the `Entity(...)` so that they can specify the entity links without the dundered syntax.

Not sure if this actually addresses your confusion so let me know if not!

tlento: OH I GET IT. This is the set of all metrics and their (internal) join paths, because from the outside those look like different metrics, and the user would have to indicate precisely which of these targets they want. So this is what we want the snapshot to represent. 

I wonder if we should make a separate column for this so we can label it, like internal_join_links or something, but I think this will do for now. No sense rewriting all the things right now, especially since that column can't be populated for anything other than a group by metric.

tlento (approved):
Ok I get it now, thank you for your patience in explaining this to me on nearly every PR about this feature.

PR #1173, opened 2024-04-30
Test that all available single-hop metrics don't error

courtneyholcomb:
### Description
Test that these things actually work! This test loops through the allowed linkable metrics and runs them through the query parser & dataflow plan builder. This ensures that they will work for both group by resolution logic and dataflow plan logic.

More tests to come:
- multi-hop linkable metrics (multi-hop for both inner and outer join paths)
- e2e query tests: query rendering + check query tests


tlento:
> Test that these things actually work! 

Best PR description opening ever. 🤣

PR #1180, opened 2024-05-06
Track `derived_from_semantic_models` properly for `LinkableMetrics`

courtneyholcomb:
### Description
Update to `derived_from_semantic_models()` for `LinkableMetrics`. Previously, we were only tracking the semantic models used in the outer join path. Here, we update to add tracking for semantic models used in the inner join path and the metric itself, as well. That should fully cover all semantic models touched for these elements.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
+++ b/metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
@@ -21,6 +21,7 @@ from typing_extensions import override
 
 from metricflow_semantics.model.linkable_element_property import LinkableElementProperty
 from metricflow_semantics.model.semantic_model_derivation import SemanticModelDerivation
+from metricflow_semantics.specs.spec_classes import GroupByMetricReference
 
 logger = logging.getLogger(__name__)
 
@@ -199,8 +200,12 @@ class LinkableMetric(LinkableElement, SerializableDataclass):
         )
 
     @property
-    def reference(self) -> MetricReference:  # noqa: D102
+    def metric_reference(self) -> MetricReference:  # noqa: D102
         return self.join_path.metric_subquery_join_path_element.metric_reference
+
+    @property
+    def reference(self) -> GroupByMetricReference:  # noqa: D102
+        return GroupByMetricReference(self.metric_reference.element_name)
 
     @property
     def join_by_semantic_model(self) -> Optional[SemanticModelReference]:  # noqa: D102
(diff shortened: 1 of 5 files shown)


metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
```diff
@@ -324,15 +344,22 @@ class MetricSubqueryJoinPathElement:
         metric_reference: The metric that's aggregated in the subquery.
         join_on_entity: The entity that the metric is grouped by in the subquery. This will be updated in V2 to allow a list
             of entitites & dimensions.
+        entity_links: Sequence of entities joined to get from a metric source to the `join_on_entity`.
         metric_to_entity_join_path: Describes the join path used in the subquery to join the metric to the `join_on_entity`.
             Can be none if all required elements are defined in the same semantic model.
     """
 
     metric_reference: MetricReference
+    metric_derived_from_semantic_models: Tuple[SemanticModelReference, ...]
```
tlento, line 353: Huh. This name is awkward. Can we just call this `derived_from_semantic_models`, too? Since this is a MetricSubqueryJoinPathElement it seems like it'll be ok. That also makes it a little easier for us to make this part of a common interface with the SemanticModelJoinPath if we decide to do so in the future.

metricflow-semantics/metricflow_semantics/model/semantics/linkable_spec_resolver.py
```diff
@@ -192,10 +192,15 @@ def __init__(
                 continue
             metric_reference = MetricReference(metric.name)
             linkable_element_set_for_metric = self.get_linkable_elements_for_metrics([metric_reference])
+            defined_from_semantic_models = tuple(
+                self._semantic_model_lookup.get_semantic_model_for_measure(input_measure.measure_reference).reference
+                for input_measure in metric.input_measures
+            )
```
tlento, line 198: Not so much this specific block, but how do we handle this type of resolution for query-level metrics? It feels like this has to be duplicating that work somehow, or that this, at least, could be done from an extension of the API that provides the saved query semantic model inputs. Or is this what that logic relies on for getting those semantic model references?

Anyway, I don't think that changes anything for this PR.

tlento (approved):
Left a naming nit inline. 

Also, outside the scope of this PR, it feels like there's a common solution to a bunch of this hiding in here somewhere. The problem is I can't remember how the query resolver works, so it's possible this IS the common solution already.

PR #1182, opened 2024-05-06
Fix pretty print bug for sets containing enums

courtneyholcomb:
### Description
The recent change to allow & sort sets in MF pretty printing was hiding a sneaky error when pretty printing objects containing enums. Enums can't be sorted without a key. This would only show up in logs that call `mf_pformat()`, since none of our test snapshots appear to contain enums - hence why it wasn't caught in CI.
Here's an example of a logging error triggered by query parsing that was being buried by the try/except in `mf_pformat()`:
`TypeError: '<' not supported between instances of 'LinkableElementProperty' and 'LinkableElementProperty'`

### DIFF
--- a/metricflow-semantics/metricflow_semantics/mf_logging/pretty_print.py
+++ b/metricflow-semantics/metricflow_semantics/mf_logging/pretty_print.py
@@ -45,7 +45,7 @@ class MetricFlowPrettyFormatter:
         """Pretty prints a sequence object i.e. list or tuple.
 
         Args:
-            list_like_obj: A list or a tuple.
+            list_like_obj: A list, tuple, set, or frozenset.
             remaining_line_width: If specified, try to make the string representation <= this many columns wide.
 
         Returns:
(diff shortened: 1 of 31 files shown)


metricflow-semantics/metricflow_semantics/mf_logging/pretty_print.py
```diff
@@ -60,7 +60,10 @@ def _handle_sequence_obj(
         elif isinstance(list_like_obj, set) or isinstance(list_like_obj, frozenset):
             left_enclose_str = f"{type(list_like_obj).__name__}("
             right_enclose_str = ")"
-            list_like_obj = sorted(list_like_obj)
+            list_like_obj = sorted(
+                list_like_obj,
+                key=(lambda x: x.name) if (list_like_obj and isinstance(list(list_like_obj)[0], Enum)) else None,
+            )
```
tlento, line 66: This will be more robust - albeit slower - if we do this instead:

```suggestion
            list_like_obj = sorted(self._handle_any_obj(obj) for obj in list_like_obj)
```

tlento (approved):
Oof, yeah, I thought about this but we DO have snapshots with enum values in sets (or at least we do at the top of my stack) so I assumed this worked. I suspect they all just had length 1 so the comparison was never triggered.

I left a suggestion inline for a more robust approach. It's slower but we don't do a ton of set printing so I think it'll be fine.

PR #1184, opened 2024-05-07
Finalize V1 LinkableMetric options

courtneyholcomb:
### Description
The good news here is that I realized the reason multi-hop `LinkableMetrics` weren't working was mostly a bug in the test, not the actual code. After a quick tweak to the dataflow plan builder (second commit), Multi-hop for inner join paths is working. Multi-hop for outer join paths is not yet working due to the lack of interface for specifying that join path. This PR restrict LinkableMetric options to only to the ones that are working thus far (single-hop outer join path, all inner join paths), and tests that they all pass through the dataflow plan builder successfully.


metricflow-semantics/tests_metricflow_semantics/snapshots/test_linkable_spec_resolver.py/str/test_create_linkable_element_set_from_join_path__result0.txt
```diff
@@ -1,126 +1,46 @@
...
-listings_latest   ('listing',)                                             ds                                                   WEEK                YEAR         ['DERIVED_TIME_GRANULARITY', 'JOINED']
-listings_latest   ('listing',)                                             ds                                                   YEAR                             ['DERIVED_TIME_GRANULARITY', 'JOINED']
-listings_latest   ('listing',)                                             ds                                                   YEAR                YEAR         ['DERIVED_TIME_GRANULARITY', 'JOINED']
-listings_latest   ('listing',)                                             is_lux_latest                                                                         ['JOINED']
-listings_latest   ('listing',)                                             user                                                                                  ['ENTITY', 'JOINED']
+Semantic Model    Entity Links    Name             Time Granularity    Date Part    Properties
```
courtneyholcomb, line 1: No metrics in this file or the next because join path implies multi-hop.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -897,6 +886,19 @@ def _find_dataflow_recipe(
                 f"nodes for the right side of the join:\n"
                 f"{mf_pformat(candidate_nodes_for_right_side_of_join)}"
             )
+
+        # If there are MetricGroupBys in the requested linkable specs, build source nodes to satisfy them.
+        # We do this at query time instead of during usual source node generation because the number of potential
+        # MetricGroupBy source nodes could be extremely large (and potentially slow).
+        logger.info(f"Building source nodes for group by metrics: {linkable_spec_set.group_by_metric_specs}")
+        candidate_nodes_for_right_side_of_join += [
```
tlento, line 894: Interesting. I think this is still fine for predicate pushdown, since they aren't really source nodes in the original sense, so I probably need to handle them separately anyway.

tlento (approved):
Seems reasonable!

Just to make sure I'm clear on this, the updated test actually does nothing with multi-hop paths since those always return an empty set, but once you add support it'll naturally execute them now, is that right?

courtneyholcomb:
> Seems reasonable!
> 
> 
> 
> Just to make sure I'm clear on this, the updated test actually does nothing with multi-hop paths since those always return an empty set, but once you add support it'll naturally execute them now, is that right?

Let me clarify!
When I put this up, I did so with the belief that multi-hop for the INNER query was working. (That's what I meant in the PR description.) Then, though, I realized that none of the LinkableMetrics for listings or bookings in the simple manifest actually have multi-hop inner query paths. The ones that appear to be multi-hop are actually just using the local entity prefix syntax.
But the place where we return an empty set only actually gets hit if the OUTER join path is multi-hop.
So alas, we still don't have multi-hop tested for either join path.

PR #1187, opened 2024-05-08
Fix path key conversion bug for `LinkableMetrics`

courtneyholcomb:
### Description
Turns out there was a bug where we were using the wrong set of entity links in path key conversion. This fixes that!

### DIFF
--- a/metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
+++ b/metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
@@ -80,6 +80,10 @@ class ElementPathKey:
         else:
             assert_values_exhausted(element_type)
 
+        assert len(set(self.entity_links)) == len(
+            self.entity_links
+        ), f"Duplicate found in `entity_links`: {self.entity_links}."
+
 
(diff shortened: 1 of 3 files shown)


tlento (approved):
Nice, thanks for the quick fix! This makes the relationship between these different elements a lot more clear, too.

Vitaliy3i:
Hi, I've using metricflow for a while, and this commit brakes our SL due to error:
- "AssertionError: Entity links have been misconfigured. They should not include the join_on_entity."

After some manual troubleshooting and debugging, I've hound the prohibited entity joins, but it require some effort to find it.
Overall, it great that we currently have more validations and better quality of our SL, but it will be great to put some more hints into error message - i.e. print the entity name that produce prohibited join. 

Thanks for your work!

courtneyholcomb:
@Vitaliy3i PR here: https://github.com/dbt-labs/metricflow/pull/1196

courtneyholcomb:
Realized this actually impacts users who aren't using this feature yet - so I'm removing this error and we can add a validation instead. Sorry about that!

PR #1189, opened 2024-05-08
Remove `pandas` / `numpy` Dependency From `metricflow-semantics`

plypaul:
### Description

`metricflow-semantics` currently is dependent on `pandas` and transitively, `numpy`. It would be helpful to remove those dependencies as:

* They often have hard version requirements due to backward-incompatible changes.
* They pull in many dependent packages.
* They are large packages that slow down building of the virtual environment and consume significant memory when loaded.

Removing those dependencies would make the package easier to use. Migrading away from `pandas` requires changing the implementation of time period calculations to use `dateutil.`.


plypaul:
Closing this out as this was significantly revised, and related changes will be in a another PR.

PR #1190, opened 2024-05-08
Remove `pandas` / `numpy` Dependency From `metricflow`

plypaul:
### Description

This PR removes `pandas` / `numpy` from `metricflow` for similar reasons as those listed in https://github.com/dbt-labs/metricflow/pull/1189.

The main challenge with doing so is that the `DataFrame` implementation needs to be swapped. The operations that are done with `DataFrames` are limited to access, comparison, and display operations, so an internal representation is reasonable.

Cleanup is still remaining:

* Docstrings need more elaboration.
* Some commits need to be split.
* `MetricFlowDataTable` needs refinement e.g. type-casting / inference of data types / logic consolidation.


plypaul:
Closing this out as this was significantly revised, and related changes will be in a another PR.

PR #1193, opened 2024-05-08
Use correct join node specs for `GroupByMetrics`

courtneyholcomb:
### Description
In the process of writing rendering tests, I found an issue with the logic used to determine which specs to include in a joined node. This PR primarily updates that logic to handle `GroupByMetricSpecs`, plus with some related adjustments along the way.


metricflow/dataflow/builder/node_evaluator.py
```diff
@@ -92,13 +98,13 @@ def join_description(self) -> JoinDescription:
...
-        include_specs.extend(
-            [
-                LinklessEntitySpec.from_reference(spec.entity_links[0])
-                for spec in self.satisfiable_linkable_specs
-                if len(spec.entity_links) > 0
-            ]
-        )
+        # Get the specs needed to join onto this node.
+        if self.node_to_join.aggregated_to_elements:
+            include_specs.extend(self.node_to_join.aggregated_to_elements)
```
tlento, line 103: Did we want these as-is or did we want to convert them to the LinklessEntitySpec types at this point?

metricflow/plan_conversion/instance_converters.py
```diff
@@ -558,19 +558,25 @@ def transform(self, instance_set: InstanceSet) -> InstanceSet:  # noqa: D102
...
                 if include_spec not in instance_set.spec_set.all_specs:
-                    raise RuntimeError(
-                        f"Include spec {include_spec} is not in the spec set {instance_set.spec_set} - "
-                        f"check if this node was constructed correctly."
-                    )
+                    include_specs_not_found.append(include_spec)
+            if include_specs_not_found:
+                raise RuntimeError(
+                    f"Include specs {include_specs_not_found} are not in the spec set {instance_set.spec_set} - "
+                    f"check if this node was constructed correctly."
```
tlento, line 568: nit: this is an odd follow up. If I think it was constructed correctly and I hit this error, what does that indicate?

metricflow/plan_conversion/instance_converters.py
```diff
@@ -558,19 +558,25 @@ def transform(self, instance_set: InstanceSet) -> InstanceSet:  # noqa: D102
...
         elif self._exclude_specs:
+            exclude_specs_not_found = []
             for exclude_spec in self._exclude_specs.all_specs:
                 if exclude_spec not in instance_set.spec_set.all_specs:
-                    raise RuntimeError(
-                        f"Exclude spec {exclude_spec} is not in the spec set {instance_set.spec_set} - "
-                        f"check if this node was constructed correctly."
-                    )
+                    exclude_specs_not_found.append(exclude_spec)
+            if exclude_specs_not_found:
```
tlento, line 575: It is very nice to get all of these at once.  😁

tests_metricflow/snapshots/test_metric_filter_rendering.py/SqlQueryPlan/DuckDB/test_multi_hop_inner_join_path__plan0_optimized.sql
```diff
@@ -0,0 +1,42 @@
+-- Constrain Output with WHERE
```
tlento, line 1: I'm a little confused about where these snapshots are coming from. Some other PR?

tlento (approved):
There are a few oddities in here but unless one of the inline comments makes you say "oh no!" I think the changes themselves are reasonable.

metricflow/dataflow/builder/node_evaluator.py
```diff
@@ -266,7 +266,8 @@ def _find_joinable_candidate_nodes_that_can_satisfy_linkable_specs(
                         ].semantic_model_reference,
                         on_entity_reference=entity_spec_in_right_node.reference,
                     )
-                    or right_node.aggregated_to_elements == {entity_spec_in_right_node.reference}
+                    or {spec.reference for spec in right_node.aggregated_to_elements}
+                    == {entity_spec_in_right_node.reference}
```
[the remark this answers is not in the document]
courtneyholcomb: Yeah sorry I had a top-of-stack TODO on my list to fix this, but hadn't gotten to it yet 😅

metricflow/plan_conversion/instance_converters.py
courtneyholcomb: It indicates that you either 1) constructed the instance set correctly or 2) constructed the node incorrectly, since they don't match!

tests_metricflow/snapshots/test_metric_filter_rendering.py/SqlQueryPlan/DuckDB/test_multi_hop_inner_join_path__plan0_optimized.sql
courtneyholcomb: Dang it, I had a test I was running in this PR and ended up deleting it (but forgot to delete the snapshots). Will remove!

metricflow/dataflow/builder/node_evaluator.py
courtneyholcomb: We want them as-is. The reason they were always `LinklessEntitySpecs` previously is that we were always joining to source nodes (which would not have joins contained WITHIN them). In this case, we are joining to a node that has joins within it, so its instance specs might have entity links.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -386,7 +386,7 @@ def _build_conversion_metric_output_node(
             metric_spec=metric_spec,
             aggregated_measures_node=aggregated_measures_node,
             for_group_by_source_node=for_group_by_source_node,
-            aggregated_to_elements=queried_linkable_specs.as_reference_set,
+            aggregated_to_elements=set(queried_linkable_specs.as_tuple),
```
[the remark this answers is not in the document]
courtneyholcomb: Yeah, previously we had no need for the full spec but now we do!

metricflow/dataflow/builder/node_evaluator.py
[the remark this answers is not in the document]
courtneyholcomb: Updated!

metricflow/plan_conversion/instance_converters.py
courtneyholcomb: This was related to this PR because I ran into this error and that's why I needed to change the instance specs built in the `JoinDescription`. Definitely could have been split into another PR though.

PR #1194, opened 2024-05-08
Add tests for metric filters using local entity prefix

courtneyholcomb:
### Description
Tests for metric filters where the group by entity uses local entity prefix syntax. By "local entity syntax" I mean an entity that is local to the metric's source table, but prefixed with the semantic model's primary entity.


tests_metricflow/query_rendering/test_metric_filter_rendering.py
```diff
@@ -254,3 +254,55 @@ def test_metric_filtered_by_itself(
...
+    )
+    dataflow_plan = dataflow_plan_builder.build_plan(query_spec)
+
+    convert_and_check(
+        request=request,
+        mf_test_configuration=mf_test_configuration,
+        dataflow_to_sql_converter=dataflow_to_sql_converter,
+        sql_client=sql_client,
+        node=dataflow_plan.sink_output_nodes[0].parent_node,
+    )
```
tlento, line 308: Do we need this? If so, can we give it a better name and move the snapshot files accordingly?

tlento (approved):
Please do something about the `prefix2` test before merge. Thank you!

tests_metricflow/query_rendering/test_metric_filter_rendering.py
courtneyholcomb: Yeah, ok maybe we don't need it. Will remove.

PR #1195, opened 2024-05-09
Test rendering for conversion metric filter

courtneyholcomb:
### Description
Rendering test for a conversion metric filter.
I intended to add a check query test, too, but the SQL result turned out so complicated that it didn't seem particularly useful.


tlento:
> I intended to add a check query test, too, but the SQL result turned out so complicated that it didn't seem particularly useful.

Yeah, for conversion metrics I think we need to do output checks - it's often easier, with a simple example, to do the math than it is to figure out how to write the SQL.

tlento (approved):
Nice spot. Conversion metrics are so incredibly complicated.....

PR #1196, opened 2024-05-10
Improve entity links error message on `MetricSubqueryJoinPathElement`

courtneyholcomb:
### Description
Improve error message, as [requested](https://github.com/dbt-labs/metricflow/pull/1187#issuecomment-2105026807).


courtneyholcomb:
Closing - removing this error instead. A few customers have this misconfiguration so we should roll out a validation for it instead.

PR #1198, opened 2024-05-10
Remove error that users are running into

courtneyholcomb:
### Description
Some customers have a misconfiguration (I think) that is triggering this error. We should roll out a validation to get that fixed instead of blocking all their queries with this error.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
+++ b/metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
@@ -336,10 +336,6 @@ class MetricSubqueryJoinPathElement:
             self.derived_from_semantic_models
         ), "There must be at least one semantic model from which the metric is derived."
 
-        assert (
-            self.join_on_entity not in self.entity_links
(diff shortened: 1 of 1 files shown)


metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
```diff
@@ -336,10 +336,6 @@ def __post_init__(self) -> None:  # noqa: D105
             self.derived_from_semantic_models
         ), "There must be at least one semantic model from which the metric is derived."
 
-        assert (
```
tlento, line 339: Do we want to consider a logger.warning or something (debug is fine, anything that isn't an error) for tracking while we roll out other validation and see if this issue is what we think it is?

tlento (approved):
🤷‍♂️ :shipit:

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
tlento: Update - based on what we just learned, no, I don't think we want logging here, at least not at this time.

PR #1200, opened 2024-05-10
Don't allow LinkableMetrics for case when entity reference is in entity links

courtneyholcomb:
### Description
Some users have run into a scenario where the entity reference can be found in the entity links for some `LinkableEntities`. I haven't yet been able to reproduce that error with our test configs. Releasing metrics as dimensions is blocked on this edge case, so I'm skipping that scenario here to ensure that customers won't run into it until when using `LinkableMetrics` until we can debug further. This will allow us to release the feature and handle this case in a follow up version.


tlento (approved):
Ok, this looks good, per lengthy discussion about what this means:

1. Our current hypothesis is this is a multi-hop join of something like listing -> account and then there are multiple datasets with a primary account entity
2. This logging will show us what these paths look like and how common they are
3. This is most likely correct behavior given how metric joins are meant to work, certainly for v1

PR #1201, opened 2024-05-11
Update `SemanticModelJoinPath` to Include Left-Most Semantic Model

plypaul:
### Description

`SemanticModelJoinPath` represents the joins between the model that contains the measure and the models that contain the group-by-items. However, it currently does not include a reference to the semantic model containing the measure, so this PR adds a field `left_semantic_model_reference` to include that context. It's named left as the model containing the measure is on the left-most side of the join.


metricflow-semantics/metricflow_semantics/test_helpers/snapshot_helpers.py
```diff
@@ -259,41 +259,43 @@ def assert_linkable_element_set_snapshot_equal(  # noqa: D103
...
+                tuple(entity_link.element_name for entity_link in linkable_entity.entity_links),
+                linkable_entity.element_name,
+                "",
+                "",
+                sorted(linkable_element_property.name for linkable_element_property in linkable_entity.properties),
             )
+            if row_to_add not in rows:
+                rows.append(row_to_add)
 
     for linkable_metric_iterable in linkable_element_set.path_key_to_linkable_metrics.values():
```
courtneyholcomb, line 300: The same logic should apply for LinkableMetrics, too. You can access the outer join path via `LinkableMetric.join_path.semantic_model_join_path`.

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
```diff
@@ -267,7 +265,8 @@ class SemanticModelJoinPath:
     would be represented by 2 path elements [(semantic_model0, A), (dimension_source1, B)]
     """
 
-    path_elements: Tuple[SemanticModelJoinPathElement, ...]
+    left_semantic_model_reference: SemanticModelReference
```
courtneyholcomb, line 268: With this change, `LinkableMetric.join_path.semantic_model_join_path` should no longer be an optional field (otherwise we won't know the `left_semantic_model_reference`). In the cases where it was previously None, it should now be a `SemanticModelJoinPath` that includes a `left_semantic_model_reference` and no path elements.

courtneyholcomb (changes requested):
This should also apply to `LinkableMetrics` so I left some comments related to that! Let me know if you want to talk through those, I know you haven't seen much of the logic related to `LinkableMetrics`.

plypaul:
@courtneyholcomb In the following snippet:

```
    def get_joinable_metrics_for_semantic_model(
        self, semantic_model: SemanticModel, using_join_path: Optional[SemanticModelJoinPath] = None
    ) -> LinkableElementSet:
        """Get the set of linkable metrics that can be joined to this semantic model."""
```

Would you be able to clarify what `using_join_path` is supposed to mean?

plypaul:
@courtneyholcomb Updated.

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
```diff
@@ -222,7 +222,10 @@ def reference(self) -> GroupByMetricReference:  # noqa: D102
 
     @property
     def join_by_semantic_model(self) -> Optional[SemanticModelReference]:  # noqa: D102
-        return self.join_path.last_semantic_model_reference
+        if len(self.join_path.semantic_model_join_path.path_elements) == 0:
```
courtneyholcomb, line 225: Should this return `self.join_path.semantic_model_join_path.left_semantic_model_reference` instead of `None`? Depends how it's used so I'll leave that up to you.

metricflow-semantics/metricflow_semantics/model/semantics/linkable_spec_resolver.py
```diff
@@ -264,24 +266,32 @@ def _get_semantic_model_for_measure(self, measure_reference: MeasureReference) -
...
-        if using_join_path:
+
+        join_path_has_path_links = len(using_join_path.path_elements) > 0
+        if join_path_has_path_links:
             assert semantic_model.reference == using_join_path.last_semantic_model_reference, (
                 "Last join path element should match semantic model when building LinkableMetrics. "
                 f"Got semantic model: {semantic_model.reference.semantic_model_name}; "
                 f"last join path element: {using_join_path.last_semantic_model_reference.semantic_model_name}",
             )
-            properties = properties.union(frozenset({LinkableElementProperty.MULTI_HOP}))
```
courtneyholcomb, line 278: We do want to keep this in the case when `join_path_has_path_links=True`. Currently you won't see this property in any of the snapshots because multi-hop options are temporarily disabled until we can test them more thoroughly. But we'll need this back when they're enabled again.

courtneyholcomb (approved):
One comment about keeping the `MULTI_HOP` property, but otherwise looks great! Thank you!!

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
plypaul: Actually, this is no longer used, so I've removed it.

metricflow-semantics/metricflow_semantics/model/semantics/linkable_spec_resolver.py
plypaul: Moved to a comment if you need it as a reminder to get rid of the dead code flag.

PR #1205, opened 2024-05-11
Remove Dataflow Plan Node Types

plypaul:
### Description

The original class hierarchy for the `DataflowPlanNodes` included types that described the data that was output by the node. However, those turned out to not be useful in practice (e.g. `BaseOutput` was the majority of use cases), so this PR removes them.

### DIFF
--- a/metricflow/dataflow/dataflow_plan.py
+++ b/metricflow/dataflow/dataflow_plan.py
@@ -9,11 +9,10 @@ from typing import Generic, Optional, Sequence, Set, Type, TypeVar
 
 from metricflow_semantics.dag.id_prefix import StaticIdPrefix
 from metricflow_semantics.dag.mf_dag import DagId, DagNode, MetricFlowDag, NodeId
+from metricflow_semantics.specs.spec_classes import LinkableInstanceSpec
 from metricflow_semantics.visitor import Visitable, VisitorOutputT
 
 if typing.TYPE_CHECKING:
-    from metricflow_semantics.specs.spec_classes import LinkableInstanceSpec
-
     from metricflow.dataflow.nodes.add_generated_uuid import AddGeneratedUuidColumnNode
     from metricflow.dataflow.nodes.aggregate_measures import AggregateMeasuresNode
     from metricflow.dataflow.nodes.combine_aggregated_outputs import CombineAggregatedOutputsNode
@@ -75,7 +74,7 @@ class DataflowPlanNode(DagNode, Visitable, ABC):
         raise NotImplementedError
 
     @abstractmethod
-    def with_new_parents(self: NodeSelfT, new_parent_nodes: Sequence[BaseOutput]) -> NodeSelfT:
+    def with_new_parents(self: NodeSelfT, new_parent_nodes: Sequence[DataflowPlanNode]) -> NodeSelfT:
         """Creates a node with the same behavior as this node, but with a different set of parents.
 
         typing.Self would be useful here, but not available in Python 3.8.
@@ -86,6 +85,11 @@ class DataflowPlanNode(DagNode, Visitable, ABC):
     def node_type(self) -> Type:  # noqa: D102
         # TODO: Remove.
         return self.__class__
+
+    @property
+    def aggregated_to_elements(self) -> Set[LinkableInstanceSpec]:
+        """Indicates that the node has been aggregated to these specs, guaranteeing uniqueness in all combinations."""
+        return set()
 
 
 class DataflowPlanNodeVisitor(Generic[VisitorOutputT], ABC):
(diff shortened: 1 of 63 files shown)


metricflow/execution/dataflow_to_execution.py
```diff
@@ -99,4 +115,70 @@ def visit_write_to_result_table_node(self, node: WriteToResultTableNode) -> Conv
...
+    def visit_min_max_node(self, node: MinMaxNode) -> ConvertToExecutionPlanResult:
+        raise NotImplementedError
+
+    @override
+    def visit_add_generated_uuid_column_node(self, node: AddGeneratedUuidColumnNode) -> ConvertToExecutionPlanResult:
+        raise NotImplementedError
+
+    @override
+    def visit_join_conversion_events_node(self, node: JoinConversionEventsNode) -> ConvertToExecutionPlanResult:
+        raise NotImplementedError
```
tlento, line 184: Speaking of things looking weird, big blocks of green inside mechanical change diffs always make an impression. Now if anybody attempts to convert a dataflow plan to an execution plan using the wrong node type the runtime will blow up with a NotImplementedError. This seems undesirable. While I was never a fan of the existence of the SinkNodeVisitor interface, its one redeeming feature was preventing this from happening. Indeed, I originally suggested it as a "if you must use a visitor to do this graph level property access then please at least make it a different type"

Do you have a follow-up planned where you get rid of this one way or another? My planned stack was going to involve replacing the visitor itself with a property on the DataflowPlan and pausing on removing the execution plan stuff until later just because I don't want to deal with the MFS changes right now, but removing execution plans (so much for our earlier execution plan aspirations....) altogether would be welcome.

tlento (commented):
I love this PR. It is third in the set of cleanup changes I was planning to make after predicate pushdown, but that ordering had more to do with personal preferences than anything else.

I have not really read it because it's, well, hard to read, and I'm trying to get another feature out the door. 

The main problem I ran into here is this PR represents, conceptually, three or four changes in one:
1. removing useless type marker nodes
2. removing SinkNodes
3. simplifying the optimizer return types (which is probably fine to consolidate into the SinkNode removal, and is a separate commit here)
4. doing the natural consolidation of BaseOutput into DataflowPlanNode.

If this were split along these - or other equally readable - lines it'd be a "scroll through and make sure nothing looks weird" review. As it is now it's quite a bit more daunting. 

Similarly, if this causes some weirdness then viewing via git diff or whatever is going to be a headache since we squash and merge, and it'd be more accessible in every piece of tooling if it was broken into more readily addressable chunks.

Anyway, you can do what you will with this. If you're going to break this up let me know and I won't review it until you're done, otherwise I'll take a pass through when I have more time.

metricflow/execution/dataflow_to_execution.py
plypaul: This is what I got now: https://github.com/dbt-labs/metricflow/blob/cdfdce9681f1cbd28b387efe2dbce697a70671c4/metricflow/execution/dataflow_to_execution.py

plypaul:
> Anyway, you can do what you will with this. If you're going to break this up let me know and I won't review it until you're done, otherwise I'll take a pass through when I have more time.

Unfortunately, breaking out that change might be a bit tough, so let's go with the assumption that the commits will be as is. If that changes, I'll let you know.

metricflow/execution/dataflow_to_execution.py
tlento: Consolidating to the one execution plan type we actually use looks like a win to me. Implementing a DagWalker that does not - and in fact absolutely SHOULD NOT - walk the DAG is confusing. Maybe just update that to be a regular class?

metricflow/dataflow/dataflow_plan.py
```diff
@@ -178,22 +178,15 @@ def visit_join_conversion_events_node(self, node: JoinConversionEventsNode) -> V
...
     """Describes the flow of metric data as it goes from source nodes to sink nodes in the graph."""
 
-    def __init__(  # noqa: D107
-        self, sink_output_nodes: Sequence[DataflowPlanNode], plan_id: Optional[DagId] = None
-    ) -> None:
-        if len(sink_output_nodes) == 0:
-            raise RuntimeError("Can't create a dataflow plan without sink node(s).")
-        self._sink_output_nodes = tuple(sink_output_nodes)
+    def __init__(self, sink_nodes: Sequence[DataflowPlanNode], plan_id: Optional[DagId] = None) -> None:  # noqa: D107
+        assert len(sink_nodes) == 1, f"Exactly 1 sink node is supported. Got: {sink_nodes}"
```
tlento, line 182: Out of curiosity, are we going to formalize this via the type system?

metricflow/dataflow/dataflow_plan.py
```diff
@@ -178,22 +178,15 @@ def visit_join_conversion_events_node(self, node: JoinConversionEventsNode) -> V
...
 
     @property
-    def sink_output_nodes(self) -> Sequence[DataflowPlanNode]:  # noqa: D102
-        return self._sink_output_nodes
-
-    @property
-    def sink_output_node(self) -> DataflowPlanNode:  # noqa: D102
-        assert len(self._sink_output_nodes) == 1, f"Only 1 sink node supported. Got: {self._sink_output_nodes}"
-        return self._sink_output_nodes[0]
+    def checked_sink_node(self) -> DataflowPlanNode:
```
tlento, line 189: Can this just be sink_node? We already have the assertion in the initializer.

tlento (approved):
Thanks for splitting out the rename, that made a surprisingly big difference!

At some point we should really reconsider the multiple sink nodes with the runtime errors. I don't think we're likely to need multiple sinks anymore, since all we do is render queries to a single output stream (i.e., the outer SELECT statement), and any forking of that output data stream is probably best handled outside of MetricFlow. Not sure if you're doing that upstack or not, but it's a thing to consider.

metricflow/dataflow/dataflow_plan.py
plypaul: I haven't had a chance to think about how to simplify this, but yeah, that sounds like a good idea.

metricflow/dataflow/dataflow_plan.py
plypaul: Yeah, updated.

PR #1207, opened 2024-05-11
Add Check to Verify `queried_semantic_models`

plypaul:
### Description

This PR adds a check to the integration test cases to verify that the queried semantic models returned by the parser match the semantic models queried in the dataflow plan.

Tagging @courtneyholcomb as there's currently a case that is skipped - do you have context offhand why that's the case?

### DIFF
--- a/tests_metricflow/integration/test_configured_cases.py
+++ b/tests_metricflow/integration/test_configured_cases.py
@@ -20,7 +20,7 @@ from metricflow_semantics.specs.query_param_implementations import DimensionOrEn
 from metricflow_semantics.test_helpers.config_helpers import MetricFlowTestConfiguration
 from metricflow_semantics.test_helpers.time_helpers import ConfigurableTimeSource
 
-from metricflow.engine.metricflow_engine import MetricFlowEngine, MetricFlowQueryRequest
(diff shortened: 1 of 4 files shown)


tests_metricflow/integration/test_configured_cases.py
```diff
@@ -361,3 +362,16 @@ def test_case(
     )
     # If we sort, it's effectively not checking the order whatever order that the output was would be overwritten.
     assert_dataframes_equal(actual, expected, sort_columns=not case.check_order, allow_empty=case.allow_empty)
+
+    # Check that the parse result and the dataflow plan show the same semantic models read.
+    if name in {"itest_dimensions.yaml/distinct_values_query_with_metric_filter"}:
```
courtneyholcomb, line 367: @plypaul did some digging and it turns out this bug is not in the metric filters feature, it's in queries without metrics in general. You can see [this test case](https://github.com/dbt-labs/metricflow/pull/1209/files) gets the same error. I'll put up an issue to fix that, but you might want to update the message below to reflect that.

courtneyholcomb (approved):
Great test to add 🙏 Thank you!

PR #1208, opened 2024-05-11
Tests for metric filters with inner-query joins

courtneyholcomb:
### Description
Add tests for metric filters where the metric subquery has one or two joins.


tests_metricflow/query_rendering/test_metric_filter_rendering.py
```diff
@@ -306,3 +306,57 @@ def test_filter_with_conversion_metric(  # noqa: D103
...
+    sql_client: SqlClient,
+    multihop_dataflow_to_sql_converter: DataflowToSqlQueryPlanConverter,
+    multihop_query_parser: MetricFlowQueryParser,
+) -> None:
+    """Tests rendering for a metric filter using a two-hop join in the inner query."""
+    query_spec = multihop_query_parser.parse_and_validate_query(
+        metric_names=("third_hop_count",),
+        where_constraint=PydanticWhereFilter(
+            where_sql_template="{{ Metric('txn_count', ['account_id__customer_id__customer_third_hop_id']) }} > 2",
+        ),
```
tlento, line 352: 🤯

tlento (approved):
The merge conflicts are going to cause other lint errors. Sorry about the delays on that, totally missed this PR.

Please make sure all engine tests are passing before merge! If you add the persist and run tests labels at the same time I think the second one clobbers the first and then everything will fail, so it might be a little annoying.

PR #1212, opened 2024-05-14
Appropriate handling for group by source nodes in `ComputeMetricsBranchCombiner`

courtneyholcomb:
### Description
Jordan found a bug when using metric filters on ratio metrics in prod. The [error](https://dbtlabsmt.datadoghq.com/logs?query=Include%20specs%20GroupByMetricSpec%20&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&event=AgAAAY90J9TAPXKJvQAAAAAAAAAYAAAAAEFZOTBKLWpfQUFBdjR1VVdjbGg5WEFCaQAAACQAAAAAMDE4Zjc0MjgtMmY1NC00OTM1LTg2Y2QtYTgxZjYxZDdhN2Rl&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1715638086706&to_ts=1715641686706&live=true) indicates that a `ComputeMetricsNode` is being unexpectedly created with `MetricSpecs` instead of `GroupByMetricSpecs`. 
I was able to reproduce the error in check query tests. You can see the test case & the CI failures in the first commit here. To fix that bug, I updated the `ComputeMetricsBranchCombiner` to use the `ComputeMetricsNode.for_group_by_source_node` param where missed. I also added a step to ensure we don't combine nodes if only one is a group by source node (since that needs to be aggregated separately before joining to the outer query).


metricflow/dataflow/optimizer/source_scan/cm_branch_combiner.py
```diff
@@ -293,6 +293,14 @@ def visit_compute_metrics_node(self, node: ComputeMetricsNode) -> ComputeMetrics
             )
             return ComputeMetricsBranchCombinerResult()
 
+        if self._current_left_node.for_group_by_source_node != current_right_node.for_group_by_source_node:
```
plypaul, line 296: Can you move this check (and the above `aggregated_to_elements` check) to a method similar to:

```
def combine(other: ComputeMetricsNode) -> CombinationResult
    ...
```

That should make it easier update that node in the future.

plypaul (approved):
LGTM if the `ComputeMetricsNode.combine` works.

metricflow/dataflow/optimizer/source_scan/cm_branch_combiner.py
plypaul: Might be easier to do `can_combine` instead for now.

PR #1215, opened 2024-05-15
Encapsulate time range constraints for predicate pushdown

tlento:
MetricFlow currently allows for a limited scope form of filter
predicate pushdown that is particular to time range constraints where
the querying user has provided an explicit time window for us to
query against. The application of the pushdown operation is managed
by threading the time range constraint through the entire dataflow
plan builder and applying the filter operation as appropriate.

This is exactly what we need to do for robust predicate pushdown
evaluation for our expanded set of pushdown operations. Rather than
wire a whole new set of parameters through, we simply encapsulate
the time range constraints inside of a new object that is more
readily extensible to other predicate pushdown handling.

This specific change is as mechanical as possible in order to minimize
confusion. Places that stood out for improvement via the encapsulating
object have been marked for later updates, which will follow shortly.

PR #1214, opened 2024-05-15
Add conversion metric rendering tests

tlento:
Add conversion metric rendering tests

We do not have rendering tests for conversion metrics, nor do we
have any tests with query-time filters added.

This change addresses both of these gaps in our coverage, which will
give us more confidence as we move ahead with predicate pushdown.

Add snapshots for all engines

PR #1216, opened 2024-05-15
Use pushdown params to disable time range constraint pushdown

tlento:
The predicate pushdown operations for time range constriants currently
rely on None/not-None state to determine whether or not to push the
ConstrainTimeRange filter node to the source. With the switch to
pushdown params this None/not-None state gets tenuous, as future
updates will need to do things like manage time range constraints,
other time dimension filters, and categorical dimension (and entity)
filters, and relying on externalizing None/not-None combinations for
these various filter types will be challenging.

This change encapsulates the enabling and disabling of time range
constraints inside of PredicatePushdownParams. At the moment, in
order to maintain the existing behavior, we simply internalize the
None/not-None behavior for time range constraints. This will also
allow us to easily retain pushdown processing for categorical dimensions
even when time constraint filters should not be applied, and
gradually centralize those controls as we streamline the callsites.

There is added complexity to this change because of two things.

1. Time constraint updating is scattered around in the DataflowPlanBuilder
2. Conversion metrics currently do predicate pushdown for time constraints

The first of these will be addressed later.
Since we cannot reliably support general predicate pushdown for conversion
metrics, we need to allow for a time-range-only pushdown operation. Today
this is equivalent to pushdown enabled, but that will change shortly.

### DIFF
--- a/metricflow/plan_conversion/node_processor.py
+++ b/metricflow/plan_conversion/node_processor.py
@@ -140,20 +140,29 @@ class PredicatePushdownParameters:
 
     @staticmethod
     def with_time_range_constraint(
-        original_pushdown_params: PredicatePushdownParameters, time_range_constraint: Optional[TimeRangeConstraint]
+        original_pushdown_params: PredicatePushdownParameters, time_range_constraint: TimeRangeConstraint
     ) -> PredicatePushdownParameters:
-        """Factory method for overriding the time range constraint value in a given set of pushdown parameters.
+        """Factory method for adding or updating a time range constraint input to a set of pushdown parameters.
 
-        This allows for crude updates to the core time range constraint, including selectively disabling time range
-        predicate pushdown in certain sub-branches of the dataflow plan, such as in complex cases involving time spine
-        joins and cumulative metrics.
+        This allows for temporarily overriding a time range constraint with an adjusted one, or enabling a time
+        range constraint filter if one becomes available mid-stream during dataflow plan construction.
         """
-        if original_pushdown_params.is_pushdown_enabled_for_time_range_constraint:
-            return PredicatePushdownParameters(
-                time_range_constraint=time_range_constraint,
-            )
-        else:
-            return original_pushdown_params
+        pushdown_enabled_types = original_pushdown_params.pushdown_enabled_types.union(
+            {PushdownPredicateInputType.TIME_RANGE_CONSTRAINT}
+        )
+        return PredicatePushdownParameters(
+            time_range_constraint=time_range_constraint, pushdown_enabled_types=pushdown_enabled_types
+        )
+
+    @staticmethod
+    def without_time_range_constraint(
+        original_pushdown_params: PredicatePushdownParameters,
+    ) -> PredicatePushdownParameters:
+        """Factory method for removing the time range constraint, if any, from the given set of pushdown parameters."""
+        pushdown_enabled_types = original_pushdown_params.pushdown_enabled_types.difference(
+            {PushdownPredicateInputType.TIME_RANGE_CONSTRAINT}
+        )
+        return PredicatePushdownParameters(time_range_constraint=None, pushdown_enabled_types=pushdown_enabled_types)
 
     @staticmethod
     def with_pushdown_disabled() -> PredicatePushdownParameters:
--- a/metricflow/plan_conversion/node_processor.py
+++ b/metricflow/plan_conversion/node_processor.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import dataclasses
 import logging
 from enum import Enum
-from typing import List, Optional, Sequence, Set
+from typing import FrozenSet, List, Optional, Sequence, Set
 
 from dbt_semantic_interfaces.enum_extension import assert_values_exhausted
 from dbt_semantic_interfaces.references import EntityReference, TimeDimensionReference
@@ -61,7 +61,7 @@ class MultiHopJoinCandidate:
     lineage: MultiHopJoinCandidateLineage
 
 
-class PredicatePushdownState(Enum):
+class PushdownPredicateInputType(Enum):
     """Enumeration of constraint states describing when predicate pushdown may operate in a dataflow plan.
 
     This is necessary for holistic checks against the set of potentially enabled pushdown operations, because the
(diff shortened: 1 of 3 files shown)


tlento:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1216?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1217** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1217?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1216** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1216?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1215** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1215?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1214** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1214?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

metricflow/plan_conversion/node_processor.py
```diff
@@ -59,14 +61,111 @@ class MultiHopJoinCandidate:
...
+    """Container class for managing information about whether and how to do filter predicate pushdown.
 
-    Stores time constraint information for applying pre-join time filters.
+    The time_range_constraint property holds the time window for setting up a time range filter expression.
     """
 
-    time_range_constraint: Optional[TimeRangeConstraint]
+    _PREDICATE_METADATA_KEY = "is_predicate_property"
+
+    time_range_constraint: Optional[TimeRangeConstraint] = dataclasses.field(metadata={_PREDICATE_METADATA_KEY: True})
```
courtneyholcomb, line 89: This last was pretty hard to understand without diving into the dataclasses docs a bit & running the code to see what the value actually is. Might be helpful to leave a comment to explain that this is not a typical default arg (i.e., you still need to pass a value in for `time_range_constraint`)

metricflow/plan_conversion/node_processor.py
```diff
@@ -59,14 +61,111 @@ class MultiHopJoinCandidate:
...
+        """Validation to ensure pushdown properties are configured correctly.
+
+        In particular, this asserts that cases where pushdown is disabled cannot leak pushdown operations via
+        outside property access - if pushdown is disabled, no further pushdown operations of any kind are allowed
+        on that particular code branch.
+        """
+        if self.pushdown_state is PredicatePushdownState.FULLY_ENABLED:
+            return
+
+        invalid_predicate_property_names = {
```
courtneyholcomb, line 102: So is this essentially just checking that `time_range_constraint` is always `None` when `pushdown_state` is `DISABLED`?
If so, this seems pretty complicated & hard to read for just that functionality, and it might be helpful to simplify by just checking that one scenario. But maybe I'm missing something or this gets used more generically down the stack?

metricflow/plan_conversion/node_processor.py
```diff
@@ -59,14 +61,111 @@ class MultiHopJoinCandidate:
...
+            "Invalid pushdown parameter configuration! Disabled pushdown parameters cannot have properties "
+            "set that may lead to improper access and use in other contexts, as that can lead to unintended "
+            "filtering operations in cases where these properties are accessed without appropriate checks against "
+            "pushdown configuration. This indicates that pushdown is configured for limited scope operations, but "
+            f"other predicate properties are set to non-None values.\nFull configuration: {instance_configuration}\n"
+            f"Invalid predicate properties: {invalid_disabled_properties}"
+        )
+
+    @property
+    def is_pushdown_enabled(self) -> bool:
```
courtneyholcomb, line 130: Might be helpful to call this `is_pushdown_enabled_for_time_range` or something more specific so that other devs don't mistake this as an indication that pushdown is FULLY enabled!

courtneyholcomb (approved):
Code looks good! Left some in-line suggestions to improve readability, but feel free to address those at the top of the stack.

metricflow/plan_conversion/node_processor.py
tlento: Yeah, to be honest I can probably get rid of this and just keep a list of predicate key names. That's what I had originally and I think it's a little less magical overall, but it's still a lot of magic string handling.

metricflow/plan_conversion/node_processor.py
tlento: Oh yeah, I definitely should've explained that in the commit message. There will be other parameters added which will also need to be nulled out for disabled. Will update things accordingly.

I'm essentially trying to enforce that when the state is disabled there's no possibility of somebody doing something like `if pushdown_params.time_range_constraint or pushdown_params.where_filter_specs` and then triggering a pushdown operation. If the disabled state was binary I could just make it a derived property but because of our conversion metric and time window handling I have to deal with things like time range only and categorical only pushdown states.

metricflow/plan_conversion/node_processor.py
tlento: That isn't exactly what this does, but to your point inverting it to is_pushdown_disabled might be better.

tlento (commented):
Thanks @courtneyholcomb ! I'm still playing around with the expanded pushdown mechanics so I'll come back to this before merge and either put up another PR or revise this one and request a follow up, depending on what I want to do.

metricflow/plan_conversion/node_processor.py
courtneyholcomb: Sounds good!

courtneyholcomb: Sounds good!

metricflow/plan_conversion/node_processor.py
tlento: I've made a ton of updates and now this is EXACTLY what that does. I also put in the disabled thing but I'll probably remove it. I think I should stop tinkering here and move ahead, though.

metricflow/plan_conversion/node_processor.py
```diff
@@ -61,7 +61,7 @@ class MultiHopJoinCandidate:
     lineage: MultiHopJoinCandidateLineage
 
 
-class PredicatePushdownState(Enum):
+class PushdownPredicateInputType(Enum):
```
courtneyholcomb: I like this change!!

courtneyholcomb (approved):
LGTM!

courtneyholcomb (approved):
All looks good!

tlento:
### Merge activity

* **May 23, 5:33 PM PDT**: @tlento started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/dbt-labs/metricflow/1216).

PR #1223, opened 2024-05-21
Add logging to track latency for metrics as dimensions in query parsing

courtneyholcomb:
### Description
Logging to see if metrics as dimensions is causing major slowdowns.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/model/semantics/linkable_spec_resolver.py
+++ b/metricflow-semantics/metricflow_semantics/model/semantics/linkable_spec_resolver.py
@@ -186,6 +186,7 @@ class ValidLinkableSpecResolver:
 
         # Populate storage dicts with linkable metrics. This loop must happen after the one above so that
         # _metric_to_linkable_element_sets is populated with entities and dimensions.
+        linkable_metrics_start_time = time.time()
         for metric in self._semantic_manifest.metrics:
             # Cumulative metrics and time offset metrics require grouping by metric_time, which is not yet available for
             # linkable metrics. So skip those.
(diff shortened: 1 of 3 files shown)


DevonFulcher (approved):
Thanks! Although, does it make more sense to use a span for this?

courtneyholcomb:
> Thanks! Although, does it make more sense to use a span for this?

@DevonFulcher not exactly my area of expertise, but it's my understanding that we can't do the same type of tracing in MF that we do in MFS since it's a third party package 🤔

metricflow-semantics/metricflow_semantics/model/semantics/linkable_spec_resolver.py
```diff
@@ -218,13 +219,17 @@ def __init__(
...
         linkable_element_sets_for_no_metrics_queries: List[LinkableElementSet] = []
+        # Temp: separate loop to track latency for linkable metrics. TODO: combine with loop below.
         for semantic_model in semantic_manifest.semantic_models:
-            linkable_element_sets_for_no_metrics_queries.append(self._get_elements_in_semantic_model(semantic_model))
             linkable_element_sets_for_no_metrics_queries.append(
                 self.get_joinable_metrics_for_semantic_model(
                     semantic_model, SemanticModelJoinPath(left_semantic_model_reference=semantic_model.reference)
                 )
             )
+        logger.info(f"Building valid linkable metrics took: {time.time() - linkable_metrics_start_time:.2f}s")
```
aiguofer, line 229: :( we need to figure out how to add stuff to `extra` in MF logs.... having it as a field in DD would allow us to easily calculate all sorts of stats on it

PR #1227, opened 2024-05-22
Enable predicate pushdown for categorical dimension filters

tlento:
Enable predicate pushdown for categorical dimension filters

We now have the ability to push down filter predicates within
the DataflowPlan. We start with categorical dimension filters,
as they are the simplest.

This change simply tracks the where filters applied at the measure
node and pushes all of them down to the construction of the source
node for evaluation. At this time a filter is eligible to be applied
to the source node if it only contains references to categorical dimensions
that originate from the same, singular semantic model definition that
feeds into the source node in question.

We do not support time dimensions at this time, as they can cause strange
interactions with things like cumulative metrics, which could result in
inappropriate input filtering that produces non-obviously censored metric
results.

We also do not support entities at this time, as entities may be defined
in multiple semantic models and as such filters must be applied with more
care to ensure we are correctly accounting for the entity link paths to
the relevant source node, if any, when we apply the filter.

Finally, we are not able to safely push predicates down for the "null value"
side of an outer join, which, in practice, restricts us to only doing predicate
pushdown to the measure source nodes.

The snapshot test changes for existing snapshots highlight the new behavior,
while the added test snapshots demonstrate specific circumstances of
interest.


tlento (commented):
Quick note - the intervening commits show where I got a little too aggressive with pushdown, and the snapshots show one of the issues with the "during construction" approach - it's exceedingly difficult to properly push the filter down without duplicating it in the query, because the recursive tracking of which filters got pushed and which ones didn't is super complicated. Some of these pushdown operations are effectively undoing the subquery reducing optimizer, but have no other practical effect.

Future updates will:

1. Add some dedicated join + filter test cases to ensure things keep returning the same results in the face of predicate pushdown. We incidentally seem to be covering the currently operating edge cases, but having explicit tests for all of the boundary conditions will be helpful as we restructure.
2. Finish consolidating our pushdown logic - including time window adjustments - into the pushdown state object
3. Convert our process to an optimizer pass and remove most predicate management from the DataflowPlanBuilder
4. Fix up the pushdown rendering so we don't double-render filters without justification

Whether those come before or after basic support for time dimensions remains to be seen, but I think it likely we'll do them first.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -680,7 +691,9 @@ def _build_plan_for_distinct_values(self, query_spec: MetricFlowQuerySpec) -> Da
         required_linkable_specs, _ = self.__get_required_and_extraneous_linkable_specs(
             queried_linkable_specs=query_spec.linkable_specs, filter_specs=query_level_filter_specs
         )
-        predicate_pushdown_state = PredicatePushdownState(time_range_constraint=query_spec.time_range_constraint)
+        predicate_pushdown_state = PredicatePushdownState(
+            time_range_constraint=query_spec.time_range_constraint, where_filter_specs=query_level_filter_specs
```
courtneyholcomb, line 695: Where does this get narrowed down to only categorical dimensions? 🤔

tests_metricflow/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/BigQuery/test_nested_filters__plan0_optimized.sql
```diff
@@ -22,17 +22,34 @@ FROM (
         -- Join Standard Outputs
         -- Pass Only Elements: ['average_booking_value', 'listing__is_lux_latest', 'booking__is_instant']
         SELECT
```
courtneyholcomb, line 24: Unrelated to this PR, but I've been wondering if we can optimize away these interim subqueries that do nothing but select columns 🤔

courtneyholcomb (commented):
All the snapshots and tests look good.
After reading this, I'm still confused as to how we ensure only categorical dimensions get pushed down 🤔 So trying to dig in to understand that before I approve!

metricflow/plan_conversion/node_processor.py
```diff
@@ -272,6 +345,49 @@ def _add_time_range_constraint(
...
+        enabled_element_types: FrozenSet[LinkableElementType],
+    ) -> Sequence[DataflowPlanNode]:
+        """Processes where filter specs and evaluates their fitness for pushdown against the provided node set."""
+        eligible_filter_specs_by_model: Dict[SemanticModelReference, Sequence[WhereFilterSpec]] = {}
+        for spec in where_filter_specs:
+            semantic_models = set(element.semantic_model_origin for element in spec.linkable_elements)
+            invalid_element_types = [
+                element for element in spec.linkable_elements if element.element_type not in enabled_element_types
+            ]
+            if len(semantic_models) == 1 and len(invalid_element_types) == 0:
```
courtneyholcomb, line 361: Ok, so we only push down filters if ALL the filtered elements are eligible element types. Is that because we don't know if this is an `AND` or an `OR` filter at this point?

metricflow/plan_conversion/node_processor.py
```diff
@@ -272,6 +345,49 @@ def _add_time_range_constraint(
...
+        self,
+        source_nodes: Sequence[DataflowPlanNode],
+        where_filter_specs: Sequence[WhereFilterSpec],
+        enabled_element_types: FrozenSet[LinkableElementType],
+    ) -> Sequence[DataflowPlanNode]:
+        """Processes where filter specs and evaluates their fitness for pushdown against the provided node set."""
+        eligible_filter_specs_by_model: Dict[SemanticModelReference, Sequence[WhereFilterSpec]] = {}
+        for spec in where_filter_specs:
+            semantic_models = set(element.semantic_model_origin for element in spec.linkable_elements)
+            invalid_element_types = [
```
courtneyholcomb, line 358: Ok this is the logic I was looking for 👍

metricflow/plan_conversion/node_processor.py
```diff
@@ -272,6 +345,49 @@ def _add_time_range_constraint(
...
+                eligible_filter_specs = eligible_filter_specs_by_model[node_semantic_models[0]]
+                source_node_specs = self._node_data_set_resolver.get_output_data_set(source_node).instance_set.spec_set
+                matching_filter_specs = [
+                    filter_spec
+                    for filter_spec in eligible_filter_specs
+                    if all([spec in source_node_specs.linkable_specs for spec in filter_spec.linkable_specs])
+                ]
+                if len(matching_filter_specs) == 0:
+                    filtered_nodes.append(source_node)
+                    continue
```
courtneyholcomb, line 380: Does this `continue` actually do anything?

metricflow/plan_conversion/node_processor.py
```diff
@@ -272,6 +345,49 @@ def _add_time_range_constraint(
...
+        filtered_nodes: List[DataflowPlanNode] = []
+        for source_node in source_nodes:
+            node_semantic_models = tuple(source_node.as_plan().source_semantic_models)
+            if len(node_semantic_models) == 1 and node_semantic_models[0] in eligible_filter_specs_by_model:
+                eligible_filter_specs = eligible_filter_specs_by_model[node_semantic_models[0]]
+                source_node_specs = self._node_data_set_resolver.get_output_data_set(source_node).instance_set.spec_set
+                matching_filter_specs = [
+                    filter_spec
+                    for filter_spec in eligible_filter_specs
+                    if all([spec in source_node_specs.linkable_specs for spec in filter_spec.linkable_specs])
```
courtneyholcomb, line 376: Is there ever a time where this won't be true, since we get the semantic model from the linkable element above? Or is this just an extra safety check in case something gets misconfigured?

courtneyholcomb (approved):
Looooks goood!!!
Left a couple of inline nits, nothing blocking!

tests_metricflow/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/BigQuery/test_nested_filters__plan0_optimized.sql
tlento: They're supposed to get optimized away, and I'm not entirely sure why they don't. There may be some column or subquery alias thing that's getting in the way.

tlento:
### Merge activity

* **Jun 24, 8:29 PM PDT**: @tlento started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/dbt-labs/metricflow/1227).

PR #1229, opened 2024-05-23
Cache linkable elements

courtneyholcomb:
Completes SL-2262

### Description
Recently we've seen an increase in latency for compile time on queries. The theory is that this might be caused by the addition of `LinkableMetrics` to group by options, which may have dramatically increased the number of group by options available. Looping through them could add a lot of slowness to the query.

I added some logs to see how long those loops were taking, and looking at data from the past 24 hours, the call to `linkable_elements_for_measure` took about .1 seconds on each call for long-running queries. This could be called quite a few times for the same measure within a given query, resulting in a lot of duplicate work & unnecessary latency. To fix that, I decided to cache that method. I also cached a couple of similar methods along the way.

I tested this primarily by running queries in pytest and showing INFO logs. There was a major reduction in the logs for these functions after adding the cache (especially in scenarios where a given measure is used several times in one query).

The other changes in this PR are making params hashable, which is required for the cache decorator.
I did not add an expiration for the cache - since these classes are scoped to a given semantic manifest, the valid linkable elements for a given query should never change.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/model/semantics/linkable_element_set.py
+++ b/metricflow-semantics/metricflow_semantics/model/semantics/linkable_element_set.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import logging
+import time
 from collections import defaultdict
 from dataclasses import dataclass, field
 from typing import Dict, FrozenSet, List, Sequence, Set, Tuple
@@ -27,6 +29,8 @@ from metricflow_semantics.specs.spec_classes import (
     LinkableInstanceSpec,
     TimeDimensionSpec,
 )
+
+logger = logging.getLogger(__name__)
 
 
 @dataclass(frozen=True)
@@ -379,6 +383,8 @@ class LinkableElementSet(SemanticModelDerivation):
         Returns a new set consisting of the elements in the `LinkableElementSet` that have a corresponding spec that
         match all the given spec patterns.
         """
+        start_time = time.time()
+
         # Spec patterns need all specs to match properly e.g. `BaseTimeGrainPattern`.
         matching_specs: Sequence[InstanceSpec] = self.specs
 
@@ -402,8 +408,10 @@ class LinkableElementSet(SemanticModelDerivation):
             if LinkableElementSet._path_key_to_spec(path_key) in specs_to_include:
                 path_key_to_linkable_metrics[path_key] = linkable_metrics
 
-        return LinkableElementSet(
+        filtered_elements = LinkableElementSet(
             path_key_to_linkable_dimensions=path_key_to_linkable_dimensions,
             path_key_to_linkable_entities=path_key_to_linkable_entities,
             path_key_to_linkable_metrics=path_key_to_linkable_metrics,
         )
+        logger.info(f"Filtering valid linkable elements took: {time.time() - start_time:.2f}s")
+        return filtered_elements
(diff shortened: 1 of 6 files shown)


metricflow-semantics/metricflow_semantics/model/semantics/linkable_element_set.py
```diff
@@ -402,8 +408,10 @@ def filter_by_spec_patterns(self, spec_patterns: Sequence[SpecPattern]) -> Linka
             if LinkableElementSet._path_key_to_spec(path_key) in specs_to_include:
                 path_key_to_linkable_metrics[path_key] = linkable_metrics
 
-        return LinkableElementSet(
+        filtered_elements = LinkableElementSet(
             path_key_to_linkable_dimensions=path_key_to_linkable_dimensions,
             path_key_to_linkable_entities=path_key_to_linkable_entities,
             path_key_to_linkable_metrics=path_key_to_linkable_metrics,
         )
+        logger.info(f"Filtering valid linkable elements took: {time.time() - start_time:.2f}s")
```
courtneyholcomb, line 416: I would love to cache this function too, but that change would be a bit more involved since base class (`LinkableElementSet`) relies heavily on dictionaries, which are not hashable.

tlento: Can we just store `Sequence[Tuple[ElementPathKey, Sequence[LinkableDimension]]]` or whatever instead of a dict? We go to some lengths to avoid mutation of these things anyway, might as well make it official. 

We shouldn't do it in this PR, anyway, that sounds like a fairly involved change given the need to update callsites and all of those merge functions and whatnot.

metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
```diff
@@ -44,11 +45,12 @@ def __init__(self, semantic_manifest: SemanticManifest, semantic_model_lookup: S
...
         )
 
+    @functools.cache
     def linkable_elements_for_measure(
         self,
         measure_reference: MeasureReference,
-        with_any_of: Optional[Set[LinkableElementProperty]] = None,
-        without_any_of: Optional[Set[LinkableElementProperty]] = None,
+        with_any_of: Optional[FrozenSet[LinkableElementProperty]] = None,
+        without_any_of: Optional[FrozenSet[LinkableElementProperty]] = None,
```
tlento, line 53: I kind of wondered why these weren't frozenset to begin with.

metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
```diff
@@ -80,6 +83,7 @@ def linkable_elements_for_no_metrics_query(
             without_any_of=frozen_without_any_of,
         )
 
+    @functools.cache
     def linkable_elements_for_metrics(
         self,
         metric_references: Sequence[MetricReference],
```
tlento, line 89: Oof Sequence isn't counted as immutable. Should we update this to Tuple or does the typechecker catch it if we pass in a non-hashable type?

tlento (approved):
Sure, why not?

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element_set.py
courtneyholcomb: Yeah that could work!

metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
courtneyholcomb: The type checker catches it! It fails if the type isn't hashable.

PR #1232, opened 2024-05-30
Move `pandas` time-period calls to an interface

plypaul:
### Description

This PR moves time-adjustment operations into a interface / class so that
uses of `pandas` for this purpose can be encapsulated. This will facilitate
the switch from `pandas` to `dateutil` for these time-related operations.


tlento (approved):
Seems reasonable given the dateutil-based implementation upstack.

PR #1234, opened 2024-05-30
Add replacement class for dataframes

plypaul:
### Description

This PR adds a replacement class for `pandas.Dataframe` that has the minimum functionality as required by MF. This replacement will allow removal of the `pandas` dependency later on.

### DIFF
--- a/metricflow/data_table/mf_column.py
+++ b/metricflow/data_table/mf_column.py
@@ -1,0 +1,39 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from functools import cached_property
+from typing import Iterator, Sequence, Tuple, Type
+
+from metricflow.data_table.column_types import CellValue
+
+
+@dataclass(frozen=True)
+class ColumnDescription:
+    """Describes a single column in a data table."""
+
+    column_name: str
+    column_type: Type[CellValue]
+
+    def with_lower_case_column_name(self) -> ColumnDescription:  # noqa: D102
+        return ColumnDescription(
+            column_name=self.column_name.lower(),
+            column_type=self.column_type,
+        )
+
+
+@dataclass(frozen=True)
+class ColumnDescriptionSet:
+    """Describes a collection of columns in a data table."""
+
+    column_descriptions: Tuple[ColumnDescription, ...]
+
+    def __iter__(self) -> Iterator[ColumnDescription]:  # noqa: D105
+        return iter(self.column_descriptions)
+
+    @cached_property
+    def column_names(self) -> Sequence[str]:  # noqa: D102
+        return tuple(column_description.column_name for column_description in self.column_descriptions)
+
+    @cached_property
+    def column_types(self) -> Sequence[Type]:  # noqa: D102
+        return tuple(column_description.column_type for column_description in self.column_descriptions)
--- a/metricflow/data_table/mf_table.py
+++ b/metricflow/data_table/mf_table.py
@@ -1,0 +1,279 @@
+from __future__ import annotations
+
+import datetime
+import itertools
+import logging
+from dataclasses import dataclass
+from decimal import Decimal
+from typing import Iterable, Iterator, List, Optional, Sequence, Tuple, Type
+
+import tabulate
+from metricflow_semantics.mf_logging.formatting import indent
+from metricflow_semantics.mf_logging.pretty_print import mf_pformat, mf_pformat_many
+from typing_extensions import Self
+
+from metricflow.data_table.column_types import CellValue, InputCellValue, row_cell_types
+from metricflow.data_table.mf_column import ColumnDescription
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True, eq=False)
+class MetricFlowDataTable:
+    """Container for tabular data stored in memory.
+
+    This is feature-limited and is used to pass tabular data for tests and the CLI. The only data types that are
+    supported are described by `CellValue`.
+
+    When constructing the table, additional input types (as described by `InputCellValue`) can be used, but those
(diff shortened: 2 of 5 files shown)


metricflow/data_table/mf_column.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+@dataclass(frozen=True)
+class ColumnDescription:
+    """Describes a single column in a data table."""
+
+    column_name: str
+    column_type: Type[CellValue]
+
+    def lower_case_column_name(self) -> ColumnDescription:  # noqa: D102
```
tlento, line 17: nit: given that method name I'd expect this to be a string. Given how it's currently used I think this is better:

```
def with_lower_case_column_name(self) -> ColumnDescription:
```

metricflow/data_table/mf_column.py
```diff
@@ -0,0 +1,39 @@
...
+@dataclass(frozen=True)
+class ColumnDescriptionSet:
+    """Describes a collection of columns in a data table."""
+
+    column_descriptions: Tuple[ColumnDescription, ...]
+
+    def __iter__(self) -> Iterator[ColumnDescription]:  # noqa: D105
+        return iter(self.column_descriptions)
+
+    @cached_property
```
tlento, line 33: Oh neat!

metricflow/data_table/mf_table.py
```diff
@@ -0,0 +1,279 @@
...
+from typing_extensions import Self
+
+from metricflow.data_table.column_types import CellValue, InputCellValue, row_cell_types
+from metricflow.data_table.mf_column import ColumnDescription
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True, eq=False)
+class MetricFlowDataTable:
```
tlento, line 22: This is way more than I was expecting. I was thinking we could define a protocol that would let us implement wrappers around Agate/Pandas interfaces, cut over, and remove Pandas that way, not that we'd have a full bore data container class, but if this is what it takes let's just proceed.

metricflow/data_table/mf_table.py
```diff
@@ -0,0 +1,279 @@
...
+            rows=tuple(
+                tuple(row_dict[column_index] for column_index in range(self.column_count)) for row_dict in new_rows
+            ),
+        )
+
+    def _sorted_by_row(self) -> MetricFlowDataTable:  # noqa: D102
+        def _cell_sort_key(cell: CellValue) -> str:
+            if isinstance(cell, datetime.datetime):
+                return cell.isoformat()
+            return str(cell)
```
tlento, line 120: Is this potentially non-deterministic for floats?

tests_metricflow/sql/compare_data_table.py
```diff
@@ -0,0 +1,159 @@
...
+                        column_index=column_index,
+                        expected_value=expected_value,
+                        actual_value=actual_value,
+                    )
+            # It should be safe to remove this once `MetricFlowDataTable` is validated as it doesn't allow timezones.
+            elif (
+                isinstance(expected_value, datetime.datetime)
+                and isinstance(actual_value, datetime.datetime)
+                # If expected has no tz but actual is UTC, remove timezone. Some engines add UTC by default.
+                and (actual_value.tzinfo == "UTC" and expected_value.tzinfo is None)
```
tlento, line 64: oof, timezone support is coming, this may get gnarly.....

Those are tomorrow problems, though, for now copying what we've got is good.

tests_metricflow/sql/test_data_table.py
```diff
@@ -0,0 +1,124 @@
...
+def test_get_cell_value(example_table: MetricFlowDataTable) -> None:  # noqa: D103
+    assert example_table.get_cell_value(0, 0) == 0
+    assert example_table.get_cell_value(0, 1) == "a"
+    assert example_table.get_cell_value(1, 0) == 1
+    assert example_table.get_cell_value(1, 1) == "b"
+
+
+def test_column_values_iterator(example_table: MetricFlowDataTable) -> None:  # noqa: D103
+    assert tuple(example_table.column_values_iterator(0)) == (0, 1)
+    assert tuple(example_table.column_values_iterator(0)) == (0, 1)
```
tlento, line 124: Was this supposed to be: 
```suggestion
    assert tuple(example_table.column_values_iterator(1)) == ('a', 'b')
```

tlento (approved):
Some nits inline. If we don't know about the float -> string key thing let's just proceed and deal with it if it comes up. In terms of the overall approach, I'm assuming there isn't an easier way that works reasonable well across the Pandas -> Agate boundary.

I hope for a future where we can move to some kind of standard. Seems like things are converging on Arrow, but we aren't there yet.

metricflow/data_table/mf_column.py
plypaul: Updated.

metricflow/data_table/mf_table.py
plypaul: I thought this would be small going in, but the type-handling threw a wrench in there. We could still move to a protocol and this can help to come up with that definition since it has a limited interface.

metricflow/data_table/mf_table.py
plypaul: This could be tricky - did you have thoughts on an approach?

tests_metricflow/sql/test_data_table.py
plypaul: Good catch, updated.

plypaul:
> I hope for a future where we can move to some kind of standard. Seems like things are converging on Arrow, but we aren't there yet.

I took a look at `pyarrow`, but that had a `numpy` dependency and versioning requirements.

PR #1235, opened 2024-05-31
Replace uses of `DataFrame` with `MetricflowDataTable`

plypaul:
### Description

This replacement is needed for the later removal of the `pandas` dependency.


tests_metricflow/sql_clients/test_sql_client.py
```diff
@@ -25,17 +26,14 @@ def _select_x_as_y(x: int = 1, y: str = "y") -> str:
...
     We lower-case the names due to snowflake's tendency to capitalize things. This isn't ideal but it'll do for now.
     """
-    df.columns = df.columns.str.lower()
-    col = col.lower()
-    assert isinstance(df, pd.DataFrame)
-    assert df.shape == (len(vals), 1)
-    assert df.columns.tolist() == [col]
-    assert set(df[col]) == vals
+    assert df.column_count == 1
+    assert df.column_names == (col,)
```
tlento, line 35: This is case sensitive, which doesn't work with Snowflake.

tests_metricflow/snapshots/test_cli.py/str/BigQuery/test_saved_query__cli_output.txt
```diff
@@ -1,11 +1,11 @@
...
-| 2019-12-19 00:00:00 |                          4 |          6 |                  6 |
-| 2019-12-19 00:00:00 |                          5 |          2 |                  0 |
-| 2019-12-20 00:00:00 |                          5 |          2 |                  0 |
-| 2020-01-01 00:00:00 |                          4 |          2 |                  1 |
-| 2020-01-02 00:00:00 |                          4 |          3 |                  3 |
-| 2020-01-02 00:00:00 |                          5 |          1 |                  0 |
-| 2020-01-03 00:00:00 |                          5 |          1 |                  0 |
+metric_time__day      listing__capacity_latest    bookings    instant_bookings
+------------------  --------------------------  ----------  ------------------
+2019-12-01                                   5           1                   0
```
tlento, line 3: Oh interesting. Is that because the datetime/dateutil objects don't format in 0 values for time?

tlento (commented):
Thanks for splitting this up into reviewable chunks. Note the change doesn't work on Snowflake due to their default behavior of SHOUTING ALL THE COLUMN NAMES.

metricflow/engine/metricflow_engine.py
```diff
@@ -712,22 +712,10 @@ def get_dimension_values(  # noqa: D102
...
-        result_dataframe.columns = result_dataframe.columns.str.lower()
-
-        # Get dimension values regardless of input name -> output dimension mapping. This is necessary befcause
-        # granularity adjustments on time dimensions produce different output names for dimension values.
-        # Note: this only works as long as we have exactly one column of group by values
-        # and no other extraneous output columns
-        dim_vals = result_dataframe[result_dataframe.columns[~result_dataframe.columns.isin(metric_names)]].iloc[:, 0]
-
-        return sorted([str(val) for val in dim_vals])
+        return sorted([str(val) for val in query_result.result_df.column_values_iterator(0)])
```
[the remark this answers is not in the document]
plypaul: Yeah, there's a specified order when the SQL is rendered to have the dimension values first

https://github.com/dbt-labs/metricflow/blob/bbd2901522e3b8926de4e733571af13d88eba606/metricflow/plan_conversion/select_column_gen.py#L36

It would be better to do a lookup, but mapping the name to the output column is not as straightforward as it should be since the output column name can be different from the input.

tests_metricflow/snapshots/test_cli.py/str/BigQuery/test_saved_query__cli_output.txt
plypaul: This is actually due to a case in the text formatting logic. In retrospect, let me remove that.

tlento (approved):
:shipit:

PR #1245, opened 2024-06-04
Remove `inference` module

plypaul:
### Description

This removes the `inference` module as it is no longer relevant since `metricflow` uses dbt models.


pyproject.toml
```diff
@@ -34,7 +34,6 @@ dependencies = [
   "pydantic>=1.10.0, <1.11.0",
   "python-dateutil>=2.8.2, <2.9.0",
   "rapidfuzz>=3.0, <4.0",
-  "ruamel.yaml>=0.17.21, <0.18.0",
```
tlento, line 37: Wait, that's all we were using this for? Nice!

I'm pretty sure there's some special casing around this package in our mypy.ini, but I don't know if we can actually remove it.

tlento (approved):
raccoon_sweeping

pyproject.toml
plypaul: Good point - updated `mypy.ini`.

PR #1246, opened 2024-06-04
Update `metricflow` dependencies

plypaul:
### Description

This PR moves dependencies for `metricflow` so that they are shared with `metricflow-semantics` and makes some updates to put them in the proper sections.

### DIFF
--- a/extra-hatch-configuration/requirements-dev-packages.txt
+++ b/extra-hatch-configuration/requirements-dev-packages.txt
@@ -4,6 +4,8 @@ pre-commit>=3.2.2, <3.3.0
 pytest-mock>=3.7.0, <3.8.0
 pytest-xdist>=3.2.1, <3.3.0
 pytest>=7.1.1, < 7.2.0
+PyYAML>=6.0, <7.0.0
+graphviz>=0.18.2, <0.21
 types-PyYAML
 types-python-dateutil
 types-tabulate
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,17 +25,16 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = [
-  "Jinja2>=3.1.3",
-  "dbt-semantic-interfaces>=0.5.1, <0.6.0",
-  "more-itertools>=8.10.0, <10.2.0",
-  "pydantic>=1.10.0, <1.11.0",
-  "tabulate>=0.8.9",
-  "typing_extensions>=4.4, <5.0",
+
+# Dependencies are specified through the `hatch-requirements-txt` plug-in.
+dynamic = ["dependencies", "optional-dependencies"]
+
+[tool.hatch.metadata.hooks.requirements_txt]
+files = [
+  "metricflow-semantics/extra-hatch-configuration/requirements.txt",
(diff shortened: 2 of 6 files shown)


extra-hatch-configuration/requirements-dev-packages.txt
```diff
@@ -4,6 +4,8 @@ pre-commit>=3.2.2, <3.3.0
 pytest-mock>=3.7.0, <3.8.0
 pytest-xdist>=3.2.1, <3.3.0
 pytest>=7.1.1, < 7.2.0
+PyYAML>=6.0, <7.0.0
+graphviz>=0.18.2, <0.21
```
tlento, line 8: These aren't dev dependencies, are they? I'm not sure why we need pyYAML, but graphviz is referenced in our production package. As such, it should probably be in the metricflow-semantics dependency set, or at least listed as an optional dependency there.

pyproject.toml
```diff
@@ -113,11 +111,11 @@ MF_TEST_ADAPTER_TYPE="duckdb"
 MF_SQL_ENGINE_URL="duckdb://"
 # This allows us to use the classes in the `dbt-metricflow` package for tests without installing the package.
 # `dbt-metricflow` can't be installed as it has `metricflow` as a dependency.
-PYTHONPATH="./dbt-metricflow"
+PYTHONPATH="metricflow-semantics:dbt-metricflow"
```
tlento, line 114: Ah, nice, I see, so this should maybe make the install from the last PR editable? Let's give it a try!

tlento (commented):
I'm not convinced this will run unless the user happens to have graphviz installed, since metricflow_semantics [uses it](https://github.com/dbt-labs/metricflow/blob/main/metricflow-semantics/metricflow_semantics/dag/dag_visualization.py#L47).

We only access this module in the CLI package so maybe we move it over there along with the graphviz dependency. Or else we can remove the graphviz rendering altogether, I don't think anybody's using it. It's cool and all, but eh.

extra-hatch-configuration/requirements-dev-packages.txt
plypaul: `pyYAML` is needed to parse YAML configuration files for tests. `graphviz` is used in the tests and the CLI. Let me put it back as a regular dependency until I figure out what to do with it.

plypaul:
Can you clarify what you mean by `I'm not convinced this will run`?

tlento:
I think if you try to use the CLI without a separate installation of graphviz the import will fail and metricflow won't load.

plypaul:
Yeah, prior to the latest commit, that would have been the case. Now that `graphviz` is a dependency of `metricflow-semantics` (as that's where the usage is), it should be good.

tlento (approved):
Nice, thanks!

Graphviz appears to have basically no dependencies so maybe we just keep it there until somebody hits a version conflict, but since they haven't had a minor version release since 2022 I don't think this is an issue. 

We can also make it optional, condition the import appropriately, and then make the CLI package pull in the optional dep from metricflow-semantics, but I'm more inclined to move the relevant module and maybe rename that one `graphviz_whatever` property we have floating around - I looked at the logic and it's portable, the only unpleasant thing is it adds a typing dependency on the DagNode class, but when we harden the API we can replace that with some kind of MFDagNode protocol.

PR #1247, opened 2024-06-04
Fix warnings that appear when running `pytest`

plypaul:
### Description

With the update to Python 3.12, there a few deprecation warnings that have popped up. This PR updates packages and adds appropriate filters so that those warnings don't show up in test sessions.


tlento (approved):
Seems reasonable!

PR #1248, opened 2024-06-04
Mark slow tests

plypaul:
### Description

This PR marks slow tests and updates the `Makefile` so that they are only run in CI.


courtneyholcomb (approved):
Love this solution! Thank you!

PR #1249, opened 2024-06-05
Update funky test case

courtneyholcomb:
### Description
Very belatedly addressing some feedback about this test case! The existing case would always end up with zero rows, so this changes it to a more realistic scenario.
Context: https://github.com/dbt-labs/metricflow/pull/1162#discussion_r1588434046


tlento:
Looks like this is missing the new snapshots, maybe?

courtneyholcomb:
> Looks like this is missing the new snapshots, maybe?

@tlento whoops! Snapshots are there - I didn't intent to include the entity object syntax test case so that's what's failing. (Started working on a new test case while snapshots were running... always a mistake lol.) Removed that so this is good to go now!

tlento (approved):
LGTM!

PR #1253, opened 2024-06-06
Fix label removal logic for SQL engine tests

plypaul:
### Description

This PR fixes an issue where `Run Tests With Other SQL Engines` label isn't getting removed when tests fail.


.github/workflows/cd-sql-engine-tests.yaml
```diff
@@ -15,7 +15,7 @@ env:
 jobs:
   snowflake-tests:
     environment: DW_INTEGRATION_TESTS
-    if: ${{ github.event.action != 'labeled' || github.event.label.name == 'Run Tests With Other SQL Engines' }}
```
tlento, line 18: Removing the first part of this check breaks all other triggers, because it requires the event to have this label name.

.github/workflows/cd-sql-engine-tests.yaml
```diff
@@ -15,7 +15,7 @@ env:
 jobs:
```
tlento, line 15: <i class='graphite__hidden'>[Re: line 6]<br/><br/></i>

With this change this workflow_dispatch trigger will succeed in doing nothing.<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1253?utm_source=unchanged-line-comment">Graphite</a>.</span>

tlento (dismissed):
This breaks our workflow_dispatch

plypaul:
I'm missing some context - what were the use cases for `workflow_dispatch`?

plypaul:
Got context from @tlento that the workflow dispatched used to check if `main` is broken. Added a comment about it.

.github/workflows/cd-sql-engine-tests.yaml
```diff
@@ -115,7 +118,8 @@ jobs:
     name: Remove Label After Running Tests
     runs-on: ubuntu-latest
     needs: [ snowflake-tests, redshift-tests, bigquery-tests, databricks-tests ]
-    if: github.event.action == 'labeled' && github.event.label.name == 'Run Tests With Other SQL Engines'
+    # Default behavior for `needs` is that it requires success, so a success / failure needs to be specifically checked.
+    if: ${{ (success() || failure()) && github.event.action == 'labeled' }}
```
tlento, line 122: Oh nice.

tlento (approved):
LGTM, thanks!

PR #1257, opened 2024-06-07
Upgrade DSI to acquire sub-daily granularity options

courtneyholcomb:
Primarily, this PR upgrades DSI. This is needed to get sub-daily granularity options into MF. Along the way:
- Points to the latest commit of dbt-core, which is the first commit that does not have a dependency conflict with the version of DSI we need.
- Updates various places in code affected by the upgrade.

### DIFF
--- a/dbt-metricflow/extra-hatch-configuration/requirements-cli.txt
+++ b/dbt-metricflow/extra-hatch-configuration/requirements-cli.txt
@@ -1,5 +1,5 @@
 # Internal dependencies
-dbt-core>=1.8.0, <1.9.0
+dbt-core @ git+https://github.com/dbt-labs/dbt-core@e53420c1d073dc81609ae7aa84cef6ee09650576#subdirectory=core
 
 # dsi version should be fixed by MetricFlow/dbt-core, not set here
 dbt-semantic-interfaces
--- a/metricflow-semantics/metricflow_semantics/time/dateutil_adjuster.py
+++ b/metricflow-semantics/metricflow_semantics/time/dateutil_adjuster.py
@@ -22,7 +22,20 @@ class DateutilTimePeriodAdjuster(TimePeriodAdjuster):
 
     def _relative_delta_for_window(self, time_granularity: TimeGranularity, count: int) -> relativedelta:
         """Relative-delta to cover time windows specified at different grains."""
-        if time_granularity is TimeGranularity.DAY:
+        if time_granularity is TimeGranularity.NANOSECOND:
+            # TODO: figure out a workaround when enabling time constraints
+            raise ValueError("`relativedelta` does not support nanoseconds.")
+        elif time_granularity is TimeGranularity.MICROSECOND:
+            return relativedelta(microseconds=count)
+        elif time_granularity is TimeGranularity.MILLISECOND:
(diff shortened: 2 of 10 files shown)


dbt-metricflow/extra-hatch-configuration/requirements-cli.txt
```diff
@@ -1,5 +1,5 @@
 # Internal dependencies
-dbt-core>=1.8.0, <1.9.0
+dbt-core @ git+https://github.com/dbt-labs/dbt-core@e53420c1d073dc81609ae7aa84cef6ee09650576#subdirectory=core
```
tlento, line 2: Argh, this means we cannot release dbt-metricflow until 1.9 rolls out. I'm so frustrated.

metricflow-semantics/metricflow_semantics/time/dateutil_adjuster.py
```diff
@@ -22,7 +22,20 @@ class DateutilTimePeriodAdjuster(TimePeriodAdjuster):
 
     def _relative_delta_for_window(self, time_granularity: TimeGranularity, count: int) -> relativedelta:
         """Relative-delta to cover time windows specified at different grains."""
-        if time_granularity is TimeGranularity.DAY:
+        if time_granularity is TimeGranularity.NANOSECOND:
+            # TODO: figure out a workaround when enabling time constraints
```
tlento, line 26: Eh, maybe we just drop support for nanosecond for now, although I guess for cumulative metrics this isn't a big deal.

Note: BigQuery, Redshift, and Postgres only support microsecond. Trino and DuckDB support nanosecond although DuckDB uses a different type for it (and Trino requires a more explicit precision for anything smaller than millisecond).

metricflow-semantics/metricflow_semantics/time/time_constants.py
```diff
@@ -1,15 +1,5 @@
 from __future__ import annotations
 
-from dbt_semantic_interfaces.type_enums.time_granularity import TimeGranularity
-
 # Python formatting string to use for converting datetime to ISO8601
 ISO8601_PYTHON_FORMAT = "%Y-%m-%d"
 ISO8601_PYTHON_TS_FORMAT = "%Y-%m-%d %H:%M:%S"
```
tlento, line 5: Oh we're so close to removing this file....

tlento (commented):
This whole dependency situation is a huge bummer.

Options:

1. We downgrade DSI to 0.5.2dev or whatever
2. We release MetricFlow and dbt-metricflow before merging it, and acknowledge that it's the only release of MetricFlow we'll deploy for public use prior to dbt-core 1.9
3. We figure something else out

I'm inclined towards option 2.

metricflow-semantics/metricflow_semantics/time/dateutil_adjuster.py
courtneyholcomb: As discussed offline, we'll keep as is.

courtneyholcomb:
I believe this PR is only blocked due to waiting on `metricflow` / `dbt-metricflow` release, which will then allow us to point to a commit for `dbt-core`.

tlento (approved):

PR #1258, opened 2024-06-07
Engine-specific rendering for sub-daily granularity options

courtneyholcomb:
Support for rendering `DATE_TRUNC` statements with sub-daily granularity options.
Includes validation for engines that don't support certain granularities.
Note about validation: warehouse validations in YAML would be ideal, but we don't have those currently. Query parsing validations would also be ideal, but the query parse has no awareness of the engine being queried. This seemed like the next best option, and it's such an rare edge case that I'm not too concerned about optimizing this error.


metricflow/protocols/sql_client.py
```diff
@@ -24,6 +26,26 @@ class SqlEngine(Enum):
     DATABRICKS = "Databricks"
     TRINO = "Trino"
 
+    @property
+    def unsupported_granularities(self) -> Set[TimeGranularity]:
+        """Granularities that can't be used with this SqlEngine."""
```
tlento, line 31: Per discussion, all looks good with this list of limitations. Let's just update the documentation to explain why these are in place. Maybe something like:
```suggestion
        """Granularities that can't be used with this SqlEngine.

        We allow the smallest granularity the SQLEngine supports for its base TIMESTAMP type and all our required
        operations (e.g., DATE_TRUNC). For example, when we added support for these granularities
        Trino supported more precise types for storage and access, but Trino's base TIMESTAMP type and
        date_trunc function only supported millisecond precision.
        """
```

metricflow/sql/render/expr_renderer.py
```diff
@@ -41,6 +41,10 @@
 )
 from metricflow.sql.sql_plan import SqlSelectColumn
 
+if TYPE_CHECKING:
+    from metricflow.protocols.sql_client import SqlEngine
```
tlento, line 45: Huh. I see the circular dependency, it's just a little funny.

We should probably move this enum into its own module at some point. Up to you whether or not you want to deal with it now.

metricflow/sql/render/big_query.py
```diff
@@ -120,14 +123,18 @@ def visit_cast_to_timestamp_expr(self, node: SqlCastToTimestampExpression) -> Sq
...
+        self._validate_granularity_for_engine(node.time_granularity)
+
         arg_rendered = self.render_sql_expr(node.arg)
 
         prefix = ""
         if node.time_granularity == TimeGranularity.WEEK:
             prefix = "iso"
 
+        trunc_expr = "DATE_TRUNC" if node.time_granularity.to_int() >= TimeGranularity.DAY.to_int() else "TIME_TRUNC"
+
```
tlento, line 135: We should use DATETIME_TRUNC for now, and eventually we'll need to shift to either DATETIME_TRUNC or TIMESTAMP_TRUNC.

The good news is we don't need this conditional at the moment, since DATETIME_TRUNC  supports time-level granularities.

Worth noting, BigQuery's TIME type is independent of date. https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#time_type
```suggestion
```

metricflow/sql/render/big_query.py
```diff
@@ -120,14 +123,18 @@ def visit_cast_to_timestamp_expr(self, node: SqlCastToTimestampExpression) -> Sq
...
         return SqlExpressionRenderResult(
-            sql=f"DATE_TRUNC({arg_rendered.sql}, {prefix}{node.time_granularity.value})",
+            sql=f"{trunc_expr}({arg_rendered.sql}, {prefix}{node.time_granularity.value})",
```
tlento, line 137:
```suggestion
            sql=f"DATETIME_TRUNC({arg_rendered.sql}, {prefix}{node.time_granularity.value})",
```

tlento (commented):
The BigQuery update will not work as expected for sub-daily granularities. To be honest I don't know what'll happen when you put a DATETIME in there, probably a type mismatch error.

tlento (commented):
Engine failures abound. Looks like a snapshot update and some way of installing databricks to make them happy (editable install in the databricks env, I think) should do it.

courtneyholcomb:
@tlento SORRY! Should be good now!

tlento (approved):
Let's goooooo! 🚀

PR #1259, opened 2024-06-11
DFP & SQL rendering for cumulative metrics with non-default grain

courtneyholcomb:
Completes SL-2368


metricflow-semantics/metricflow_semantics/instances.py
```diff
@@ -216,3 +218,18 @@ def spec_set(self) -> InstanceSpecSet:  # noqa: D102
             metric_specs=tuple(x.spec for x in self.metric_instances),
             metadata_specs=tuple(x.spec for x in self.metadata_instances),
         )
+
+    def agg_time_dimension_instances(
```
courtneyholcomb, line 222: still needed? TBD

PR #1264, opened 2024-06-12
Use default arg for `SingleColumnCorrelationKey`

courtneyholcomb:
For the love of code reduction!!

### DIFF
--- a/metricflow-semantics/metricflow_semantics/specs/column_assoc.py
+++ b/metricflow-semantics/metricflow_semantics/specs/column_assoc.py
@@ -37,7 +37,7 @@ class ColumnAssociation(SerializableDataclass):
     # When an instance is passed from one dataflow node to another, we need to know how the columns from the input
     # corresponds to the columns from the output. Equality of this key is used to determine that relationship.
     # This could be made to be in a dictionary instead, but having it here means that it doesn't need to be hashable.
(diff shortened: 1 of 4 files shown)


metricflow-semantics/metricflow_semantics/specs/column_assoc.py
```diff
@@ -37,7 +37,7 @@ class ColumnAssociation(SerializableDataclass):
     # When an instance is passed from one dataflow node to another, we need to know how the columns from the input
     # corresponds to the columns from the output. Equality of this key is used to determine that relationship.
     # This could be made to be in a dictionary instead, but having it here means that it doesn't need to be hashable.
-    single_column_correlation_key: SingleColumnCorrelationKey
+    single_column_correlation_key: SingleColumnCorrelationKey = SingleColumnCorrelationKey()
```
tlento, line 40: Funny thing, we only use this in one place and because we only ever use instances of this class, and because the `__eq__` is defined based on an isinstance() check, the place we use this effectively does `if True`.

I think this was in place to support some other operation, perhaps having to do with our original caching implementation, and it either never got built out or was removed in the post-acquisition integration process.

tlento (approved):
Nice, thanks! TODO: Look into removing column correlation keys, as they do not appear to do anything.

PR #1265, opened 2024-06-12
Use default args for unused `SqlSelectStatementNode` params

courtneyholcomb:
For the love of code reduction!! Part 2


tlento (approved):
Huh, maybe we should all disable our IDE auto-complete.... nah, that's bananas.

PR #1266, opened 2024-06-12
Update `SqlWindowFunction` enum

courtneyholcomb:
Update `SqlWindowFunction` enum in a few ways:
- Add `AVERAGE` & `LAST_VALUE` for cumulative metrics with non-default granularities.
- Remove `ROW_NUMBER` since it isn't used anywhere.
- Update values to be uppercase for consistency with our other SQL rendering practices.

All snapshot updates are just upper-casing from `first_value` to `FIRST_VALUE`.

### DIFF
--- a/metricflow/sql/sql_exprs.py
+++ b/metricflow/sql/sql_exprs.py
@@ -944,8 +944,9 @@ class SqlWindowFunction(Enum):
     Values are the SQL string to be used in rendering.
     """
 
-    FIRST_VALUE = "first_value"
-    ROW_NUMBER = "row_number"
+    FIRST_VALUE = "FIRST_VALUE"
+    LAST_VALUE = "LAST_VALUE"
+    AVERAGE = "AVG"
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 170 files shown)


metricflow/sql/sql_exprs.py
```diff
@@ -944,8 +944,9 @@ class SqlWindowFunction(Enum):
     Values are the SQL string to be used in rendering.
     """
 
-    FIRST_VALUE = "first_value"
-    ROW_NUMBER = "row_number"
+    FIRST_VALUE = "FIRST_VALUE"
+    LAST_VALUE = "LAST_VALUE"
```
WilliamDee, line 948: do we need to include this?

WilliamDee (approved):

metricflow/sql/sql_exprs.py
WilliamDee: the last_value enum

courtneyholcomb: Oh yes! Sorry forgot to include in the PR description! Will also be used for non-default granularities with cumulative metrics. We could use `FIRST_VALUE` and order `desc`, but I decided this was more readable because it aligns with the YAML argument names (which are `first`/`last`).

PR #1270, opened 2024-06-12
Refactor `JoinOverTimeRangeNode` SQL rendering logic

courtneyholcomb:
There should be no behavior changes associated with this PR, only some code refactoring. This refactor was in service of:
1. Simplicity & readability. Removed some code that is no longer needed, made other pieces more DRY.
2. An upcoming change that will support multiple `agg_time_dimensions` in these functions, which is needed for enabling cumulative metrics with non-default granularities.


tlento (approved):
raccoon_sweeping

PR #1271, opened 2024-06-12
Bug fix: querying multiple agg time dimensions with cumulative metrics

courtneyholcomb:
First commit here demonstrates the bug. If you query multiple instances of metric time or the agg time dimension with a cumulative metric, only the one with the smallest requested grain will be selected from the time spine. This results in a weird result where `metric_time__day` might have values for every row, but `metric_time__month` might not (and similar issues). This update ensures that all agg time dimensions queried with cumulative metrics will be selected from the time spine table.

### DIFF
--- a/metricflow/plan_conversion/dataflow_to_sql.py
+++ b/metricflow/plan_conversion/dataflow_to_sql.py
@@ -310,7 +310,7 @@ class DataflowToSqlQueryPlanConverter(DataflowPlanNodeVisitor[SqlDataSet]):
 
         time_spine_data_set_alias = self._next_unique_table_alias()
 
-        # Assemble time_spine dataset with agg_time_dimension_instance_for_join selected.
+        # Assemble time_spine dataset with requested agg time dimension instances selected.
         time_spine_data_set = self._make_time_spine_data_set(
             agg_time_dimension_instances=requested_agg_time_dimension_instances,
             time_spine_source=self._time_spine_source,
@@ -1052,9 +1052,9 @@ class DataflowToSqlQueryPlanConverter(DataflowPlanNodeVisitor[SqlDataSet]):
                     spec=metric_time_dimension_spec,
                 )
             )
-            output_column_to_input_column[
-                metric_time_dimension_column_association.column_name
-            ] = matching_time_dimension_instance.associated_column.column_name
+            output_column_to_input_column[metric_time_dimension_column_association.column_name] = (
+                matching_time_dimension_instance.associated_column.column_name
+            )
 
         output_instance_set = InstanceSet(
             measure_instances=tuple(output_measure_instances),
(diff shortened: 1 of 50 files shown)


tests_metricflow/snapshots/test_cumulative_metric_rendering.py/SqlQueryPlan/DuckDB/test_cumulative_metric_with_multiple_metric_time_dimensions__plan0_optimized.sql
```diff
@@ -3,18 +3,27 @@
 -- Aggregate Measures
 -- Compute Metrics via Expressions
 SELECT
-  subq_10.ds AS metric_time__day
-  , DATE_TRUNC('month', revenue_src_28000.created_at) AS metric_time__month
+  subq_9.metric_time__day AS metric_time__day
+  , subq_9.metric_time__month AS metric_time__month
```
courtneyholcomb, line 7: This one still has a bug! Should be agg time dimension column name

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -290,20 +291,28 @@ def visit_join_over_time_range_node(self, node: JoinOverTimeRangeNode) -> SqlDat
         input_data_set = node.parent_node.accept(self)
         input_data_set_alias = self._next_unique_table_alias()
 
-        agg_time_dimension_instance: Optional[TimeDimensionInstance] = None
+        # Find requested agg_time_dimensions in parent instance set.
+        # For now, will use instance with smallest granularity in time spine join.
+        # TODO: use metric's default_grain once that property is available.
```
tlento, line 296: Is that what we want? I think we need to always join on the smallest selected. If default grain is DAILY and someone selects `metric_time__hourly, metric_time__daily` I think we still need to join on hourly. But of course that's only true if hourly is requested. Or maybe it should be whatever the window grain is set to? I'm confused.

Anyway, that's a tomorrow problem, doesn't affect this PR.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -290,20 +291,28 @@ def visit_join_over_time_range_node(self, node: JoinOverTimeRangeNode) -> SqlDat
...
+                    agg_time_dimension_instance_for_join = instance
         assert (
-            agg_time_dimension_instance
+            agg_time_dimension_instance_for_join
         ), "Specified metric time spec not found in parent data set. This should have been caught by validations."
 
         time_spine_data_set_alias = self._next_unique_table_alias()
 
-        # Assemble time_spine dataset with agg_time_dimension_instance selected.
+        # Assemble time_spine dataset with agg_time_dimension_instance_for_join selected.
```
tlento, line 313: nit: `with requested agg time dimension instances selected`

tests_metricflow/snapshots/test_cumulative_metric_rendering.py/SqlQueryPlan/DuckDB/test_cumulative_metric_with_agg_time_and_metric_time__plan0.sql
```diff
@@ -18,7 +18,8 @@ FROM (
     FROM (
       -- Join Self Over Time Range
       SELECT
-        subq_2.metric_time__day AS metric_time__day
+        subq_2.revenue_instance__ds__month AS revenue_instance__ds__month
```
tlento, line 21: I find it hilarious that we were just not including these values. Guess it's not a common query pattern.

tlento (approved):
Nice catch, fix seems reasonable!

metricflow/plan_conversion/dataflow_to_sql.py
courtneyholcomb: I think what it will end up being is: use the default unless that's too big based on the selected grains, in which case use the largest grain allowed for the selected grains.

PR #1272, opened 2024-06-13
Update DWH validator to include filters + saved query support

WilliamDee:
## Context
We want to add support for filtering to certain semantic models when validating measures/dimensions/entities/semantic_models, certain metrics when validating metrics, certain saved queries when validating saved queries.

Additionally, we want to add support to generating SQL for validating saved queries

Resolves SL-2375

### DIFF
--- a/metricflow/validation/data_warehouse_model_validator.py
+++ b/metricflow/validation/data_warehouse_model_validator.py
@@ -135,10 +135,14 @@ class DataWarehouseTaskBuilder:
         return (rendered_plan.sql, rendered_plan.bind_parameters)
 
     @classmethod
-    def gen_semantic_model_tasks(cls, manifest: SemanticManifest) -> List[DataWarehouseValidationTask]:
+    def gen_semantic_model_tasks(
+        cls, manifest: SemanticManifest, filter_by_semantic_models: Optional[Sequence[str]] = None
+    ) -> List[DataWarehouseValidationTask]:
         """Generates a list of tasks for validating the semantic models of the manifest."""
         tasks: List[DataWarehouseValidationTask] = []
         for semantic_model in manifest.semantic_models:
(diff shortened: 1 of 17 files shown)


tests_metricflow/snapshots/test_data_warehouse_tasks.py/str/BigQuery/test_build_metric_tasks__query0.sql
```diff
@@ -1,12 +1,7 @@
--- Aggregate Measures
```
aliceliu, line 1: Are the comment removal changes intentional?

metricflow/validation/data_warehouse_model_validator.py
```diff
@@ -135,10 +136,14 @@ def renderize(
         return (rendered_plan.sql, rendered_plan.bind_parameters)
 
     @classmethod
-    def gen_semantic_model_tasks(cls, manifest: SemanticManifest) -> List[DataWarehouseValidationTask]:
+    def gen_semantic_model_tasks(
+        cls, manifest: SemanticManifest, filter_by_semantic_models: Optional[Sequence[str]] = None
```
aliceliu, line 140: nit: filter_by_semantic_models sounds like a boolean option, maybe semantic_models_filter is better?

aliceliu (approved):

tests_metricflow/snapshots/test_data_warehouse_tasks.py/str/BigQuery/test_build_metric_tasks__query0.sql
WilliamDee: Yea, realize this was unnecessary for this use case, so I changed to return the sql without descriptions

PR #1274, opened 2024-06-13
`WindowAggregationNode`: SQL rendering for cumulative metrics with non-default granularity

courtneyholcomb:
Add dataflow plan node `WindowAggregationNode`. This node will be used to re-aggregate cumulative metrics when they're queried with non-default granularities. This PR primarily includes the SQL rendering logic and the boilerplate code needed for new `DataflowPlanNodes`. Tests coming in an upcoming PR.
Note that one piece of this is a placeholder waiting for the DSI change that sets the `SqlWindowFunction` for a given cumulative metric.


metricflow/dataflow/nodes/window_reaggregation_node.py
```diff
@@ -0,0 +1,87 @@
...
+
+    Currently used for calculating cumulative metrics at various granularities.
+    """
+
+    def __init__(  # noqa: D107
+        self,
+        parent_node: ComputeMetricsNode,
+        metric_spec: MetricSpec,
+        order_by_spec: TimeDimensionSpec,
+        partition_by_specs: Sequence[TimeDimensionSpec],
```
courtneyholcomb, line 28: Realized the type for this should be `Sequence[InstanceSpec]` instead of `Sequence[TimeDimensionSpec]`. Fixed that higher up in the stack to avoid the rebasing, but will fix it here if I need to address feedback in this PR anyway.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1610,3 +1613,97 @@ def visit_join_conversion_events_node(self, node: JoinConversionEventsNode) -> S
...
+        order_by_instance = None
+        partition_by_instances: Tuple[MdoInstance, ...] = ()
+        for instance in parent_instance_set.as_tuple:
+            if instance.spec == node.metric_spec:
+                metric_instance = instance
+                continue
+            if instance.spec == node.order_by_spec:
+                order_by_instance = instance
+            if instance.spec in node.partition_by_specs:
+                partition_by_instances += (instance,)
```
tlento, line 1632: These are mutually exclusive, right? The metric can't be a partition or order by, and the node doesn't allow the order by to be a partition spec.

Should we make this if/elif/elif?

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1610,3 +1613,97 @@ def visit_join_conversion_events_node(self, node: JoinConversionEventsNode) -> S
...
+                        column_name=partition_by_instance.associated_column.column_name,
+                    )
+                    for partition_by_instance in partition_by_instances
+                ],
+                order_by_args=order_by_args,
+            ),
+            column_alias=metric_instance.associated_column.column_name,
+        )
+
+        # Order by instance should not be included in the output dataset because it was not requested in the query.
```
tlento, line 1676: Oh right, it never was, otherwise we wouldn't need the window function, right?

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1610,3 +1613,97 @@ def visit_join_conversion_events_node(self, node: JoinConversionEventsNode) -> S
...
+        # Can't include window function in a group by, so we use a subquery and apply group by in the outer query.
+        subquery_select_columns = output_instance_set.transform(
+            FilterElements(exclude_specs=InstanceSpecSet(metric_specs=(metric_instance.spec,)))
+        ).transform(
+            CreateSelectColumnsForInstances(parent_data_set_alias, self._column_association_resolver)
+        ).as_tuple() + (
+            metric_select_column,
+        )
+        subquery = SqlSelectStatementNode(
+            description="",  # description included in outer query
```
tlento, line 1690: Huh. This may look odd, though, every subquery is typically labeled. Maybe we just put something like "Window function for cumulative re-aggregation" or something?

tlento (approved):
Seems reasonable, kind of hard to envision but it looks like it should work.

metricflow/plan_conversion/dataflow_to_sql.py
courtneyholcomb: Good catch. I wrote this originally thinking that the order by could also be in the partition by, and then refactored later but missed this!

metricflow/dataflow/optimizer/source_scan/cm_branch_combiner.py
```diff
@@ -327,13 +328,19 @@ def _handle_unsupported_node(self, current_right_node: DataflowPlanNode) -> Comp
         )
         return ComputeMetricsBranchCombinerResult()
 
+    def visit_window_reaggregation_node(  # noqa: D102
+        self, node: WindowReaggregationNode
+    ) -> ComputeMetricsBranchCombinerResult:
+        self._log_visit_node_type(node)
+        return self._handle_unsupported_node(node)
```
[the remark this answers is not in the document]
courtneyholcomb: Needed to take another look here, but it looks like `_handle_unsupported_node()` is the right call here. `WindowAggregationNode` should always be a child of `ComputeMetricsNode`, and all other nodes that can only have `ComputeMetricsNode` as a child use `_handle_unsupported_node()`. Shouldn't impede the combining of branches up to the point of `ComputeMetricsNode`.

PR #1275, opened 2024-06-13
Update optimizer to avoid putting window functions in group by

courtneyholcomb:
When using the `WindowAggregationNode`, you will end up with a subquery that our current `SqlRewritingSubQueryReducer` thinks can be collapsed into the outer query. However, this would put window functions in the group by clause, which triggers a SQL error. This PR updates the `SqlRewritingSubQueryReducer` logic to avoid reducing in that circumstance.

### DIFF
--- a/metricflow/sql/optimizer/rewriting_sub_query_reducer.py
+++ b/metricflow/sql/optimizer/rewriting_sub_query_reducer.py
@@ -139,6 +139,10 @@ class SqlRewritingSubQueryReducerVisitor(SqlQueryPlanNodeVisitor[SqlQueryPlanNod
         return True
 
     @staticmethod
+    def _select_columns_with_window_functions(select_columns: Tuple[SqlSelectColumn, ...]) -> List[SqlSelectColumn]:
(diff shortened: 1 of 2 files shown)


metricflow/sql/optimizer/rewriting_sub_query_reducer.py
```diff
@@ -276,6 +280,32 @@ def _current_node_can_be_reduced(self, node: SqlSelectStatementNode) -> bool:
...
+                group_by.column_alias in parent_column_aliases_with_window_functions
+                or (
+                    group_by.expr.as_column_reference_expression
+                    and group_by.expr.as_column_reference_expression.col_ref.column_name
+                    in parent_column_aliases_with_window_functions
+                )
+                or (
+                    group_by.expr.as_string_expression
+                    and group_by.expr.as_string_expression.sql_expr in parent_column_aliases_with_window_functions
+                )
```
tlento, line 304: Oof, why are there so many ways to reference a group by? Oh well. I think I'll blame BigQuery and Trino on general principles.

tlento (approved):
LGTM!

PR #1276, opened 2024-06-13
Include group by metrics in `MatchingLinkableSpecsTransform`

courtneyholcomb:
Tiny cleanup item that was previously missed.


tlento (approved):
beard thumbs up

PR #1281, opened 2024-06-14
`DataflowPlan` for cumulative metrics queried with non-default granularity

courtneyholcomb:
When querying cumulative metrics without default granularity, build a dataflow plan that includes `WindowReaggregationNode`.

E2E tests coming soon for this feature!


tlento (approved):
Seems reasonable!

PR #1282, opened 2024-06-14
Remove granularity restriction for cumulative metrics

courtneyholcomb:
Remove the `BaseTimeGrainPattern` filter that was in place for cumulative metrics. This restricted users from querying cumulative metrics with `metric_time` using non-default granularities, which will be supported shortly.
Only thing left in the stack after this is tests!


tlento (approved):
ted lasso boom

I've just discovered the Ted Lasso giphy page and I might never communicate in any other way ever again.....

PR #1284, opened 2024-06-14
Tests for cumulative metrics queried with non-default granularities

courtneyholcomb:
Tests!!

### DIFF
--- a/tests_metricflow/integration/test_cases/itest_cumulative_metric.yaml
+++ b/tests_metricflow/integration/test_cases/itest_cumulative_metric.yaml
@@ -412,3 +412,183 @@ integration_test:
     WHERE {{ render_time_constraint("a.ds", "2020-03-05", "2021-01-04") }}
     GROUP BY a.ds
     ORDER BY a.ds
+---
+integration_test:
+  name: all_time_metric_with_non_default_grains
+  description: Query a cumulative all-time metric with non-default grains
+  model: SIMPLE_MODEL
+  metrics: ["revenue_all_time"]
+  group_bys: ["metric_time__week", "metric_time__quarter"]
+  check_query: |
+    SELECT
+      metric_time__week
+      , metric_time__quarter
+      , revenue_all_time
+    FROM (
+      SELECT
+        metric_time__week
+        , metric_time__quarter
+        , FIRST_VALUE(revenue_all_time) OVER (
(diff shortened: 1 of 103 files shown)


tests_metricflow/integration/test_cases/itest_cumulative_metric.yaml
```diff
@@ -412,3 +412,183 @@ integration_test:
...
+      , revenue_instance__ds__year
+      , revenue_mtd
+    FROM (
+      SELECT
+        revenue_instance__ds__quarter
+        , revenue_instance__ds__year
+        , FIRST_VALUE(revenue_mtd) OVER (
+          PARTITION BY revenue_instance__ds__quarter, revenue_instance__ds__year
+          ORDER BY metric_time__day
+          ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING
```
tlento, line 518: Whoa, all of this works on all engines?

tests_metricflow/integration/test_cases/itest_cumulative_metric.yaml
```diff
@@ -412,3 +412,183 @@ integration_test:
...
+  check_query: |
+    SELECT
+      revenue_instance__ds__quarter
+      , revenue_instance__ds__year
+      , revenue_mtd
+    FROM (
+      SELECT
+        revenue_instance__ds__quarter
+        , revenue_instance__ds__year
+        , FIRST_VALUE(revenue_mtd) OVER (
```
tlento, line 515: Oh this is interesting. If it's month to date and people truncate to year that's super strange, right? It's just always month to date for January or whatever? I guess the answer is "don't do that" - but I do wonder if we should have a different default from FIRST_VALUE - I guess people can eventually override it to LAST_VALUE (which might be more sensible).

tlento (approved):
ted lasso woohoo

tests_metricflow/integration/test_cases/itest_cumulative_metric.yaml
courtneyholcomb: This SQL rendering logic already existed for conversion metrics, so it should!

courtneyholcomb: And also SQL engine & check query tests pass, so yes.

PR #1288, opened 2024-06-17
Use mantle-friendly DSI deps in `metricflow-semantics`

courtneyholcomb:
We can't rely on a dev version of DSI in mantle. Since mantle depends on `metricflow-semantics`, we can't point to a dev version there.

### DIFF
--- a/metricflow-semantics/extra-hatch-configuration/requirements.txt
+++ b/metricflow-semantics/extra-hatch-configuration/requirements.txt
@@ -1,4 +1,5 @@
-dbt-semantic-interfaces>=0.5.0, <0.7.0 # Should match what is supported in dbt-mantle.
+
+dbt-semantic-interfaces>=0.5.0, <2.0.0 # dbt-mantle depends on metricflow-semantics, so DSI must always point to a production version here.
 graphviz>=0.18.2, <0.21
 python-dateutil>=2.9.0, <2.10.0
 rapidfuzz>=3.0, <4.0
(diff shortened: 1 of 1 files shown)


metricflow-semantics/extra-hatch-configuration/requirements.txt
```diff
@@ -1,4 +1,5 @@
-dbt-semantic-interfaces==0.6.0.dev2
+
+dbt-semantic-interfaces>=0.5.0, <2.0.0 # dbt-mantle depends on metricflow-semantics, so DSI must always point to a production version here.
```
tlento, line 2: We might consider cutting back a bit here, but in theory the pip resolver should work correctly and return 0.6.0 or whatever for the various dbt builds.

Also, just for readability, let's reference the product rather than the repo in our commentary.

```suggestion
dbt-semantic-interfaces>=0.5.0, <2.0.0 # dbt cloud depends on metricflow-semantics, so DSI must always point to a production version here.
```

tlento (approved):
I suggested it so I guess it must be reasonable, right? We can always lower the cap if we need to deploy breaking changes in DSI, but I imagine we wouldn't bother with that until we update this repo to handle them first.

metricflow-semantics/extra-hatch-configuration/requirements.txt
courtneyholcomb: I actually find it more readable the other way. So I'll update it to include both 🙃

PR #1293, opened 2024-06-18
Consume new `cumulative_type_params` fields

courtneyholcomb:
Completes SL-2413

Consume new `cumulative_type_params` fields (`window` and `grain_to_date`) instead of `type_params` cumulative fields. Eventually we will retire the `type_params` fields altogether. For now, we have a transformation that copies the fields from `type_params` to `cumulative_type_params`, if not yet set. This ensures that we only need to consume the new fields here.

This is the last step needed to enable non-default granularities for cumulative metrics in MF 🥳

### DIFF
--- a/metricflow-semantics/extra-hatch-configuration/requirements.txt
+++ b/metricflow-semantics/extra-hatch-configuration/requirements.txt
@@ -1,5 +1,5 @@
 # dbt Cloud depends on metricflow-semantics (dependency set in dbt-mantle), so DSI must always point to a production version here.
-dbt-semantic-interfaces==0.6.1.dev0 # Temp: pin to dev release to support cumulative_type_params in MF before they get released in prod.
+dbt-semantic-interfaces>=0.6.1, <2.0.0 # Temp: pin to dev release to support cumulative_type_params in MF before they get released in prod.
 graphviz>=0.18.2, <0.21
 python-dateutil>=2.9.0, <2.10.0
 rapidfuzz>=3.0, <4.0
--- a/metricflow-semantics/extra-hatch-configuration/requirements.txt
+++ b/metricflow-semantics/extra-hatch-configuration/requirements.txt
@@ -1,5 +1,5 @@
 # dbt Cloud depends on metricflow-semantics (dependency set in dbt-mantle), so DSI must always point to a production version here.
-dbt-semantic-interfaces>=0.6.1, <2.0.0 # Temp: pin to dev release to support cumulative_type_params in MF before they get released in prod.
+dbt-semantic-interfaces>=0.6.1, <2.0.0
 graphviz>=0.18.2, <0.21
 python-dateutil>=2.9.0, <2.10.0
 rapidfuzz>=3.0, <4.0
--- a/metricflow-semantics/extra-hatch-configuration/requirements.txt
+++ b/metricflow-semantics/extra-hatch-configuration/requirements.txt
@@ -1,5 +1,5 @@
 # dbt Cloud depends on metricflow-semantics (dependency set in dbt-mantle), so DSI must always point to a production version here.
-dbt-semantic-interfaces>=0.5.0, <2.0.0
+dbt-semantic-interfaces==0.6.1.dev0 # Temp: pin to dev release to support cumulative_type_params in MF before they get released in prod.
 graphviz>=0.18.2, <0.21
 python-dateutil>=2.9.0, <2.10.0
 rapidfuzz>=3.0, <4.0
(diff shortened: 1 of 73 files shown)


tests_metricflow/snapshots/test_cumulative_metric_rendering.py/SqlQueryPlan/BigQuery/test_derived_cumulative_metric_with_non_default_grains__plan0_optimized.sql
```diff
@@ -11,19 +11,14 @@ FROM (
...
-        ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING
-      ) AS t2mr
+      , AVG(t2mr) OVER (PARTITION BY metric_time__week) AS t2mr
     FROM (
       -- Join Self Over Time Range
       -- Pass Only Elements: ['txn_revenue', 'metric_time__week', 'metric_time__day']
       -- Aggregate Measures
       -- Compute Metrics via Expressions
       SELECT
-        subq_12.metric_time__day AS metric_time__day
```
tlento, line 25: Wait, this doesn't look right. We're now doing the initial aggregation against WEEK instead of DAY, and then taking the average of those grouped by week, which is effectively a no-op. Is that expected?

tlento (commented):
Oops, wrong button.

The changes are reasonable but the SQL for the AVG window function type doesn't seem correct to me. I don't see how the window function ever gets more than one input row per window now.

tests_metricflow/snapshots/test_cumulative_metric_rendering.py/SqlQueryPlan/BigQuery/test_derived_cumulative_metric_with_non_default_grains__plan0_optimized.sql
courtneyholcomb: Shoot, you're right. I wonder if the column pruner got to it - investigating 🧐

courtneyholcomb: Definitely seems like an optimizer issue since the non-optimized SQL looks right.

courtneyholcomb: @tlento fixed this bug in the latest 2 commits.

metricflow-semantics/extra-hatch-configuration/requirements.txt
```diff
@@ -1,5 +1,5 @@
 # dbt Cloud depends on metricflow-semantics (dependency set in dbt-mantle), so DSI must always point to a production version here.
-dbt-semantic-interfaces>=0.5.0, <2.0.0
+dbt-semantic-interfaces>=0.6.1, <2.0.0 # Temp: pin to dev release to support cumulative_type_params in MF before they get released in prod.
```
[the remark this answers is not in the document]
courtneyholcomb: Oops I just should have removed this comment!

tests_metricflow/snapshots/test_cumulative_metric_rendering.py/SqlQueryPlan/BigQuery/test_derived_cumulative_metric_with_non_default_grains__plan0_optimized.sql
```diff
@@ -8,23 +8,19 @@ FROM (
     metric_time__week
     , t2mr
   FROM (
+    -- Compute Metrics via Expressions
```
tlento, line 11: Oh that's weird, but I guess it makes sense?

tlento (approved):
ted lasso let's give it a shot

tests_metricflow/snapshots/test_cumulative_metric_rendering.py/SqlQueryPlan/BigQuery/test_derived_cumulative_metric_with_non_default_grains__plan0_optimized.sql
courtneyholcomb: Yeah I think they just get optimized in a different way with this change. But not an incorrect way...

PR #1299, opened 2024-06-25
Relax pydantic dependency range to allow for 2.x installs

tlento:
MetricFlow has a single dependency point on pydantic but all we
use it for is our pretty printer. Since all of the objects we
pass through there are, in practice, Pydantic 1.x objects loaded
via the dsi_pydantic_shim, we simply re-use that package element
from the dbt-semantic-interfaces installation and thereby ensure
we're loading the same base model.

This build state was tested via linting (which ran on a pydantic
1.10.x install) and unit tests (which ran on a pydantic 2.7.x install).

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -27,7 +27,7 @@ ignore_missing_imports = True
 [mypy-databricks]
 ignore_missing_imports = True
 
-[mypy-dbt_metadata_client.*]
+[mypy-dsi_pydantic_shim]
 ignore_missing_imports = True
 
 [mypy-google.*]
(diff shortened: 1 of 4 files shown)


mypy.ini
```diff
@@ -27,7 +27,7 @@ ignore_missing_imports = True
 [mypy-databricks]
 ignore_missing_imports = True
 
-[mypy-dbt_metadata_client.*]
```
tlento, line 30: Note - we haven't had this dependency for like a year

PR #1310, opened 2024-06-28
Use `default_granularity` to resolve `metric_time`

courtneyholcomb:
Completes SL-2433

TODO:
- Add thorough tests
- Make a new PR with clean commits


metricflow-semantics/tests_metricflow_semantics/snapshots/test_spec_lookup.py/str/test_filter_spec_resolution__metrics_with_different_time_grains__result.txt
```diff
@@ -25,111 +25,53 @@ FilterSpecResolutionLookUp(
...
-      resolved_linkable_element_set=LinkableElementSet(),
-      spec_pattern=TimeDimensionPattern(
-        parameter_set=EntityLinkPatternParameterSet(
-          fields_to_compare=(DATE_PART, ELEMENT_NAME, ENTITY_LINKS),
-          element_name='metric_time',
-        ),
-      ),
-      issue_set=MetricFlowQueryResolutionIssueSet(
-        issues=(
-          NoCommonItemsInParents(
```
courtneyholcomb, line 37: Why did all these NoCommonItemsInParents issues change?

courtneyholcomb: Because we now have a default rule for what to use if the metrics have different default granularities. Confirm this decision with product:
- before: if you query a metric with 2 agg_time_dimensions of different granularities, and you don't specify granularity, query will fail (forcing you to specify a granularity, theoretically to avoid unintended granularity assumption
- now: will default to the larger of the granularities (since that one is definitely allowed for both dimensions)

PR #1316, opened 2024-07-09
Allow passing granularity in where filter `TimeDimension` name

courtneyholcomb:
Allow users to use dunder syntax to pass granularity in where filter `TimeDimension`. Ex:
`{{ TimeDimension('metric_time__month') }} > 2020-01-01`
This behavior is currently implemented everywhere else that similar Jinja parsing happens, making this one site inconsistent. Fix that to avoid user confusion.


DevonFulcher (approved):
Thank you!

PR #1319, opened 2024-07-10
Add an interface to override pretty-printing behavior

plypaul:
There are some cases when pretty-printing or pretty-formatting an object using MF's pretty-formatter where a different representation is desired e.g. only show specific fields. This PR adds an interface that allows for such control.

### DIFF
--- a/metricflow-semantics/tests_metricflow_semantics/collection_helpers/test_pretty_print.py
+++ b/metricflow-semantics/tests_metricflow_semantics/collection_helpers/test_pretty_print.py
@@ -2,12 +2,16 @@ from __future__ import annotations
 
 import logging
 import textwrap
+from dataclasses import dataclass
+from typing import Optional
 
 from dbt_semantic_interfaces.implementations.elements.dimension import PydanticDimension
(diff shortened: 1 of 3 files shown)


metricflow-semantics/tests_metricflow_semantics/collection_helpers/test_pretty_print.py
```diff
@@ -179,3 +183,19 @@ def test_pformat_many_with_strings() -> None:  # noqa: D103
         ).rstrip()
         == result
     )
+
+
+def test_custom_pretty_print() -> None:
+    """Check that `MetricFlowPrettyFormatable` can be used to override the result when using MF's pretty-printer."""
```
courtneyholcomb, line 189: nit: `Formattable`*

courtneyholcomb (approved):

metricflow-semantics/tests_metricflow_semantics/collection_helpers/test_pretty_print.py
plypaul: Updated.

PR #1320, opened 2024-07-10
Convert DAG node classes to dataclasses

plypaul:
* Previously, the DAG-node classes in MF were not dataclasses due to issues with class inheritance. These have since been resolved via hierarchy changes, so this PR updates those classes to be dataclasses. In general, dataclasses make these easier to use, and there are upcoming use cases where dataclasses will simplify implementation (e.g. graph component comparison, serialization).
* A `create()` method was added to simplify many initialization use cases while not overriding the one generated by `dataclasses`.
* There is an update to how the `node_id` field is set - please see `mf_dag.py`.
* Otherwise, this should be a mechanical update with no substantive logic changes.
* There are no snapshot changes, so that should simplify review.
* Please view by commit.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/dag/mf_dag.py
+++ b/metricflow-semantics/metricflow_semantics/dag/mf_dag.py
@@ -7,13 +7,15 @@ import logging
 import textwrap
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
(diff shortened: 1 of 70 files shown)


courtneyholcomb (approved):
LGTM!

PR #1321, opened 2024-07-11
Default granularity prep

courtneyholcomb:
A few changes preparing to use the `default_granularity` field to resolve `metric_time` granularity.
- [Use `SetDefaultGranularityRule` in CLI rule set](https://github.com/dbt-labs/metricflow/commit/66dd77c0ff6e4aaeeaf1f6a718ac81de51710faf)
- [Add `default_granularity` to `MetricSpec`](https://github.com/dbt-labs/metricflow/commit/6e68ad6214616092c5cd1deef95dbbedb9694dec)
- [Add `MetricLookup` method to `get_default_granularity_for_metrics`](https://github.com/dbt-labs/metricflow/commit/fd57e01957e664c594625d8d15796fc3f6c1f8cf)

### DIFF
--- a/metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
+++ b/metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
@@ -210,3 +210,18 @@ class MetricLookup:
                     minimum_queryable_granularity = defined_time_granularity
 
         return minimum_queryable_granularity
+
+    def get_default_granularity_for_metrics(
+        self, metric_references: Sequence[MetricReference]
+    ) -> Optional[TimeGranularity]:
+        """When querying a group of metrics, the default granularity will be the largest of the metrics' default granularities."""
+        max_default_granularity: Optional[TimeGranularity] = None
+        for metric_reference in metric_references:
+            default_granularity = self.get_metric(metric_reference).default_granularity
+            assert (
+                default_granularity
+            ), f"No default_granularity set for {metric_reference}. Something has been misconfigured."
+            if not max_default_granularity or default_granularity.to_int() > max_default_granularity.to_int():
+                max_default_granularity = default_granularity
+
+        return max_default_granularity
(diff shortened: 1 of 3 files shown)


metricflow-semantics/metricflow_semantics/specs/spec_classes.py
```diff
@@ -570,6 +570,7 @@ class MetricSpec(InstanceSpec):  # noqa: D101
     alias: Optional[str] = None
     offset_window: Optional[PydanticMetricTimeWindow] = None
     offset_to_grain: Optional[TimeGranularity] = None
+    default_granularity: Optional[TimeGranularity] = None
```
plypaul, line 573: The fields in the spec classes should be values that can't be inferred from the semantic manifest. From `get_default_granularity_for_metrics`, it seems like it is?

metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
```diff
@@ -210,3 +210,18 @@ def get_min_queryable_time_granularity(self, metric_reference: MetricReference)
                     minimum_queryable_granularity = defined_time_granularity
 
         return minimum_queryable_granularity
+
+    def get_default_granularity_for_metrics(
```
plypaul, line 214: To confirm - what's the expected behavior when you have a derived metric that depends on two other metrics that have different default time granularities? Based on `SetDefaultGranularityRule`, it seems that the default granularity for parent metrics are not intended to be considered.

courtneyholcomb: That's correct. But now that you bring it up, I'm thinking I should confirm that behavior with product, so I'll post about it in #dev-metricflow and see what people think.
Thinking I can probably change this behavior on top of the stack if we decide to go that route.

metricflow-semantics/metricflow_semantics/specs/spec_classes.py
courtneyholcomb: Updated!

PR #1325, opened 2024-07-11
Tests for `Metric.time_granularity`

courtneyholcomb:
Adds `default_granularity` tests for query parsing, SQL rendering, query output, and check query tests.
Also cleans up a couple of comments due to product decisions that changed the expected implementation of this feature.

I will add more tests once we finalize the expected behavior for more specific scenarios.


plypaul:
Following what you noted in the description - are you planning to add more items to this PR or has that already been done?

courtneyholcomb:
> Following what you noted in the description - are you planning to add more items to this PR or has that already been done?

@plypaul planning to add them in a separate PR!

plypaul (commented):
Gotcha. In that case:

* Is there a specific reason why the default granularity was updated for those metrics? It has resulted in a number of snapshot changes, so I am wondering if there's an interaction between those metrics and default granularity that's important to check.
* For testing resolution, tests of the form in (https://github.com/dbt-labs/metricflow/blob/a55c4020829834701ef268241b5c872b6a17c8d3/metricflow-semantics/tests_metricflow_semantics/query/test_ambiguous_entity_path.py#L23) might be more helpful instead of the style of tests in `test_query_parser.py`.
* Does the DSI version bump need to go in this PR or can it be done in a separate PR?

courtneyholcomb:
@plypaul 

> * Is there a specific reason why the default granularity was updated for those metrics? It has resulted in a number of snapshot changes, so I am wondering if there's an interaction between those metrics and default granularity that's important to check.

I chose to add `time_granularity` to those metrics to that we could see the behavior for a few different types of metrics that I thought might behave differently, like a simple metric vs. a metric with input metrics (the ratio metric).
The snapshot changes were intentional. I thought it would be helpful to include this feature in some tests that were not just unit tests. I went through the snapshot changes before putting up the PR before putting up the PR and thought they made sense based on the query inputs. What snapshot changes are you seeing that concern you?

> * For testing resolution, tests of the form in (https://github.com/dbt-labs/metricflow/blob/a55c4020829834701ef268241b5c872b6a17c8d3/metricflow-semantics/tests_metricflow_semantics/query/test_ambiguous_entity_path.py#L23
>   ) might be more helpful instead of the style of tests in `test_query_parser.py`.

Sure, I can add some tests there!

courtneyholcomb:
@plypaul moved those query parser tests, and separated out the version bump into a new PR.

plypaul:
> What snapshot changes are you seeing that concern you?

It was more that I was wondering whether there as any particular interaction between the default granularity and the fill configuration option.

metricflow-semantics/metricflow_semantics/test_helpers/semantic_manifest_yamls/ambiguous_resolution_manifest/metrics.yaml
```diff
@@ -92,3 +92,31 @@ metric:
     metrics:
       - name: monthly_metric_0
         filter: "{{ TimeDimension('metric_time') }} = '2020-01-01'"
+---
+metric:
+  name: simple_metric_with_time_granularity
```
plypaul, line 97: More like `...with_default_time_granularity`?

metricflow-semantics/tests_metricflow_semantics/query/test_query_parser.py
```diff
@@ -51,6 +51,10 @@
           expr: "1"
           agg: sum
           create_metric: true
+        - name: monthly_bookings
```
plypaul, line 54: This can be removed now?

plypaul (approved):

courtneyholcomb:
> > What snapshot changes are you seeing that concern you?
> 
> It was more that I was wondering whether there as any particular interaction between the default granularity and the fill configuration option.

Nope, there shouldn't be. Just happened to be that the metrics I added a time granularity to were used in those tests.

PR #1328, opened 2024-07-11
Streamline `LinkableElementProperty.all_properties()`

plypaul:
This PR updates `LinkableElementProperty.all_properties()` to iterate over the enum to make it less error prone when new properties are added.


courtneyholcomb (approved):
Good call 🙏

PR #1331, opened 2024-07-12
Set `max_metric_default_time_granularity` property on `PushDownResult`

courtneyholcomb:
Per title. Set this property when visiting the metric node and narrow down to the max when visiting a query node.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/query/group_by_item/candidate_push_down/push_down_visitor.py
+++ b/metricflow-semantics/metricflow_semantics/query/group_by_item/candidate_push_down/push_down_visitor.py
@@ -7,6 +7,7 @@ from typing import Dict, FrozenSet, Iterator, List, Optional, Sequence, Tuple
 
 from dbt_semantic_interfaces.enum_extension import assert_values_exhausted
 from dbt_semantic_interfaces.type_enums import MetricType
+from dbt_semantic_interfaces.type_enums.time_granularity import TimeGranularity
 from typing_extensions import override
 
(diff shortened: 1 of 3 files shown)


plypaul (approved):

metricflow-semantics/metricflow_semantics/query/group_by_item/candidate_push_down/push_down_visitor.py
```diff
@@ -353,10 +369,18 @@ def visit_metric_node(self, node: MetricGroupByItemResolutionNode) -> PushDownRe
                     )
                 )
 
+            # If time granularity is not set for the metric, defaults to DAY if available, else the smallest available granularity.
+            # Note: ignores any granularity set on input metrics.
+            metric_default_time_granularity = metric.time_granularity or max(
+                TimeGranularity.DAY,
+                self._semantic_manifest_lookup.metric_lookup.get_min_queryable_time_granularity(node.metric_reference),
```
plypaul, line 376: Double checking that we want the default to be the min. queryable time granularity if it's not set in the configs?

courtneyholcomb: The product decision was to default to:
1. `DAY`, if available (i.e., for dimensions with `DAY` granularity or smaller)
2. Min queryable granularity if `DAY` is not available (for dimensions with `WEEK` granularity or larger).
And this should implement that behavior! So this won't change the behavior for existing dimensions - will mainly be a change once people start adding sub-daily dimensions.

PR #1334, opened 2024-07-15
Expose MetricConfig in MFEngine

WilliamDee:
## Context
We need to expose `MetricConfig` in the MFEngine so that we can grab the `meta` field now that versionless customers are on DSI 0.6.x which contains that field


tlento (approved):
Seems reasonable

PR #1335, opened 2024-07-15
Use `LinkableSpecSet` in `WhereFilterSpec`

plypaul:
In `WhereFilterSpec`, the field `linkable_specs` can't be used with `SerializableDataclass` since it's an interface. This PR changes that field to a `LinkableSpecSet`.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/specs/where_filter/where_filter_spec.py
+++ b/metricflow-semantics/metricflow_semantics/specs/where_filter/where_filter_spec.py
@@ -10,6 +10,7 @@ from metricflow_semantics.collection_helpers.dedupe import ordered_dedupe
 from metricflow_semantics.collection_helpers.merger import Mergeable
 from metricflow_semantics.model.semantics.linkable_element import LinkableElement
 from metricflow_semantics.specs.instance_spec import LinkableInstanceSpec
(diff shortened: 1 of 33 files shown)


metricflow-semantics/metricflow_semantics/specs/where_filter/where_filter_spec.py
```diff
@@ -61,7 +66,7 @@ def merge(self, other: WhereFilterSpec) -> WhereFilterSpec:  # noqa: D102
         return WhereFilterSpec(
             where_sql=f"({self.where_sql}) AND ({other.where_sql})",
             bind_parameters=self.bind_parameters.combine(other.bind_parameters),
-            linkable_specs=ordered_dedupe(self.linkable_specs, other.linkable_specs),
```
courtneyholcomb, line 64: Does the ordering matter from the previous code? Just checking since it looks like it was removed in the new code

courtneyholcomb (approved):
LGTM!

metricflow-semantics/metricflow_semantics/specs/where_filter/where_filter_spec.py
plypaul: The `.dedupe()` call is actually ordered, so it should be the same.

PR #1336, opened 2024-07-15
Update fields in `LinkableElement`-related classes to use tuples

plypaul:
Currently, serialization of sets is not supported so this PR changes the `properties` field of a few `LinkableElement` related classes to a tuple-based implementation.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
+++ b/metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import collections
 import logging
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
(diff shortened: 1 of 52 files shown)


metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
```diff
@@ -98,6 +99,16 @@ class SemanticModelJoinPathElement(SerializableDataclass):
 class LinkableElement(SemanticModelDerivation, SerializableDataclass, ABC):
     """An entity / dimension that may have been joined by entities."""
 
+    properties: Tuple[LinkableElementProperty, ...]
+
+    def __post_init__(self) -> None:  # noqa: D105
+        assert len(self.property_set) == len(self.properties)
```
courtneyholcomb, line 105: Can we add error messages to these asserts to help devs if they run into this in development?

courtneyholcomb (approved):
🚢

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element.py
plypaul: Added.

PR #1338, opened 2024-07-16
Updated error message for validation tasks

WilliamDee:
## Context
Currently, in dwh validations for dimensions, we query all the dimensions in 1 query and if that fails we attempt to query each dimension one by one. However, with time dimension, we end up querying multiple variants of a singular dimension (ie., `ordered_at` expands to querying `ordered_at__week`, `ordered_at__month`, etc...), but the error message only used the element_ment. This meant in the error, we ended up seeing "duplicates", but it was actually just querying each one of those variants. With this, we should use the qualfiied name to see what we're actually querying for

#### Example:
Screenshot 2024-07-15 at 10 33 32 PM

### DIFF
--- a/metricflow/validation/data_warehouse_model_validator.py
+++ b/metricflow/validation/data_warehouse_model_validator.py
@@ -238,7 +238,7 @@ class DataWarehouseTaskBuilder:
                             ),
                             element_type=SemanticModelElementType.DIMENSION,
                         ),
-                        error_message=f"Unable to query dimension `{spec.element_name}` on semantic model `{semantic_model.name}` in data warehouse",
+                        error_message=f"Unable to query dimension `{spec.qualified_name}` on semantic model `{semantic_model.name}` in data warehouse",
                     )
                 )
 


metricflow/validation/data_warehouse_model_validator.py
```diff
@@ -238,7 +238,7 @@ def gen_dimension_tasks(
                             ),
                             element_type=SemanticModelElementType.DIMENSION,
                         ),
-                        error_message=f"Unable to query dimension `{spec.element_name}` on semantic model `{semantic_model.name}` in data warehouse",
+                        error_message=f"Unable to query dimension `{spec.qualified_name}` on semantic model `{semantic_model.name}` in data warehouse",
```
WilliamDee, line 241: Q: This was logic already here, but do we need to query against every grain for a time dimension? Could we just consolidate this to only query for the time dimension column or are there cases where the grain/dow casting fails

courtneyholcomb: Hmmmm cases I can think of:
- the grain is too small for what the time dimension supports
- date part is not supported for cumulative metrics

courtneyholcomb (approved):

metricflow/validation/data_warehouse_model_validator.py
WilliamDee: So since these arent metric queries and just straight `SELECT <dim column> FROM sem_model` do those still apply?

courtneyholcomb: If the user can request a grain then the grain one still applies! Like is there a date trunc happening here? If so it could fail if the time column can't be trunced to a certain level

courtneyholcomb: Actually same for date part since they could request a date part that's too small for the column

WilliamDee: Yea the date trunc happens here

PR #1344, opened 2024-07-19
Update DAG formatting to not use Jinja

plypaul:
This PR changes the method to format a DAG to text to not use Jinja as during performance profiling, Jinja was found to be quite slow.


tlento (approved):
Seems reasonable! Yay for less jinja!

PR #1345, opened 2024-07-19
Remove unnecessary `deepcopy` call

plypaul:
This PR removes a `deepcopy` call that is used when fetching dimensions. The call seems unnecessary as the manifest should not be modified during a query. This call was also found to have a significant impact in performance during profiling.


tlento (approved):
Huh. Yeah, I definitely saw that a long time back and I thought I took it out, must've forgot to do it.

PR #1348, opened 2024-07-25
Support multiple time spines using new time spine configs

courtneyholcomb:
Completes SL-2313

Consume new time spine configs and add support for using multiple time spines at different granularities.

### DIFF
--- a/metricflow/plan_conversion/dataflow_to_sql.py
+++ b/metricflow/plan_conversion/dataflow_to_sql.py
@@ -243,7 +243,9 @@ class DataflowToSqlQueryPlanConverter(DataflowPlanNodeVisitor[SqlDataSet]):
         if not compatible_time_spine_grains:
             raise RuntimeError(
                 # TODO: update docs link when new docs are available
-                f"No compatible time spine found. This query requires a time spine with granularity {smallest_agg_time_grain} or smaller. See docs to configure a new time spine: https://docs.getdbt.com/docs/build/metricflow-time-spine"
+                f"This query requires a time spine with granularity {smallest_agg_time_grain.name} or smaller, which is not configured. "
+                f"The smallest available time spine granularity is {min(self._time_spine_sources.keys()).name}, which is too large."
+                "See documentation for how to configure a new time spine: https://docs.getdbt.com/docs/build/metricflow-time-spine"
             )
         return self._time_spine_sources[max(compatible_time_spine_grains)]
 
--- a/metricflow/plan_conversion/dataflow_to_sql.py
+++ b/metricflow/plan_conversion/dataflow_to_sql.py
@@ -183,7 +183,7 @@ class DataflowToSqlQueryPlanConverter(DataflowPlanNodeVisitor[SqlDataSet]):
         self._semantic_manifest_lookup = semantic_manifest_lookup
         self._metric_lookup = semantic_manifest_lookup.metric_lookup
         self._semantic_model_lookup = semantic_manifest_lookup.semantic_model_lookup
-        self._time_spine_source = TimeSpineSource.create_from_manifest(semantic_manifest_lookup.semantic_manifest)
+        self._time_spine_sources = TimeSpineSource.create_from_manifest(semantic_manifest_lookup.semantic_manifest)
 
     @property
     def column_association_resolver(self) -> ColumnAssociationResolver:  # noqa: D102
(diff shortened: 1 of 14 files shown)


metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -222,24 +222,47 @@ def _next_unique_table_alias(self) -> str:
...
+        Will choose the time spine with the largest granularity that can be used to get the smallest granularity requested
+        in the agg time dimension instances. Example:
+        - Time spines available: SECOND, MINUTE, DAY
+        - Agg time dimension granularity needed for request: HOUR, DAY
+        --> Selected time spine: MINUTE
+        """
+        assert (
+            agg_time_dimension_instances
+        ), "Building time spine dataset requires agg_time_dimension_instances, but none were found."
+        smallest_agg_time_grain = min(dim.spec.time_granularity for dim in agg_time_dimension_instances)
```
tlento, line 239: Ok, this works but it'll start failing with custom granularities. Good thing for me to note here - we'll need to be able to go from the agg time dimension instance to the time spine index granularity and check up on that.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -222,24 +222,47 @@ def _next_unique_table_alias(self) -> str:
...
+        compatible_time_spine_grains = [
+            grain for grain in self._time_spine_sources.keys() if grain.to_int() <= smallest_agg_time_grain.to_int()
+        ]
+        if not compatible_time_spine_grains:
+            raise RuntimeError(
+                # TODO: update docs link when new docs are available
+                f"No compatible time spine found. This query requires a time spine with granularity {smallest_agg_time_grain} or smaller. See docs to configure a new time spine: https://docs.getdbt.com/docs/build/metricflow-time-spine"
```
tlento, line 246: Let's include the smallest time spine granularity as well. Something like:

`This query requires a time spine with granularity {smallest_agg_time_grain}, but the smallest available time spine granularity is {min(self._time_spine_sources.keys())}, which is too large.`

metricflow/plan_conversion/time_spine.py
```diff
@@ -34,26 +33,42 @@ def spine_table(self) -> SqlTable:
...
-            and time_spine_table_configurations[0].grain == DEFAULT_TIME_GRANULARITY
-        ):
-            raise NotImplementedError(
-                f"Only a single time spine table configuration with {DEFAULT_TIME_GRANULARITY} is currently "
-                f"supported. Got:\n"
-                f"{mf_pformat(time_spine_table_configurations)}"
+        time_spine_sources = {
+            time_spine.primary_column.time_granularity: TimeSpineSource(
+                schema_name=time_spine.node_relation.schema_name,
+                table_name=time_spine.node_relation.relation_name,  # is relation name the table name? double check
```
tlento, line 41: relation_name should be the table-or-view name.

metricflow/plan_conversion/time_spine.py
```diff
@@ -34,26 +33,42 @@ def spine_table(self) -> SqlTable:
...
-            table_name=time_spine_table.table_name,
-            db_name=time_spine_table.db_name,
-            time_column_name=time_spine_table_configuration.column_name,
-            time_column_granularity=time_spine_table_configuration.grain,
-        )
+        # For backward compatibility: if legacy time spine config exists in the manifest, add that time spine here for
+        # backward compatibility. Ignore it if there is a new time spine config with the same granularity.
+        legacy_time_spines = semantic_manifest.project_configuration.time_spine_table_configurations
+        for legacy_time_spine in legacy_time_spines:
+            if not time_spine_sources.get(legacy_time_spine.grain):
```
tlento, line 53: We should use `in` instead of `.get()` here:

`if legacy_time_spine.grain not in time_spine_sources:`

That way you don't get weird bugs with falsey hashable values.

tests_metricflow/plan_conversion/test_time_spine.py
```diff
@@ -14,7 +15,9 @@ def test_date_spine_date_range(  # noqa: D103
     simple_semantic_manifest_lookup: SemanticManifestLookup,
     create_source_tables: None,
 ) -> None:
-    time_spine_source = TimeSpineSource.create_from_manifest(simple_semantic_manifest_lookup.semantic_manifest)
+    time_spine_source = TimeSpineSource.create_from_manifest(simple_semantic_manifest_lookup.semantic_manifest)[
+        TimeGranularity.DAY
```
tlento, line 19: Oh ok. Technically this is the legacy time spine fallback.

tlento (approved):
Seems reasonable so far. I assume we will have more robust tests added as we get some models with custom time spines in place.

I left a bunch of inlines. None of them are merge blocking, but may as well clean up whatever you think needs cleaning up before hitting the button.

metricflow/plan_conversion/time_spine.py
courtneyholcomb: Whoops I forgot about this comment 😅 thank you!

tests_metricflow/plan_conversion/test_time_spine.py
courtneyholcomb: Yeah, honestly, I'm a bit confused about why this test is here. I guess it's essentially testing that the legacy time spine model we have documented somewhere has the values we expect. But this feels like it should live in core or something, not here, since it's not actually related to MF functionality.

courtneyholcomb: Actually it looks like its just testing the test data we have in `tests_metricflow/fixtures/source_table_snapshots/time_spine_table/mf_time_spine.yaml`! I guess that might be useful to keep people from changing it? ¯\_(ツ)_/¯

PR #1356, opened 2024-07-27
Enable Trusted Publishing for PyPI actions

tlento:
In order to use Trusted Publishing we have to do some special things.

Most of these things have been abstracted away through the official
action plugin provided by the Python Packaging Authority.

This change simply conforms our actions configs with the requirements
for the plugin. Specifically, providing the `id-token: write` permission
and the target URL for publication in the relevant environment config.


tlento:
Note - the deployment failed because I was publishing the same package. All of the authentication appeared to work, and we got a duplicate file detected response from pypi.

plypaul (approved):

PR #1358, opened 2024-07-27
Add test time spines for sub-daily granularity

courtneyholcomb:
Add test time spines for sub-daily granularity
I added as much data as seemed reasonable to include manually, and figure we can focus tests to those time ranges. And we always have the flexibility to add more data later.
I included one time spine per sub-daily grain so that we could be sure to test the syntax for each grain against each engine that supports it.
Snapshot changes are separated into isolated commits for easier review. They only include the following changes:
- Node ID changes due to new time spine source nodes
- Time spine table alias changes due to new time spine source nodes


tests_metricflow/fixtures/dataflow_fixtures.py
```diff
@@ -90,7 +91,26 @@ def scd_query_parser(  # noqa: D103
...
+) -> Mapping[TimeGranularity, TimeSpineSource]:
+    legacy_time_spine_grain = TimeGranularity.DAY
+    time_spine_base_table_name = "mf_time_spine"
+    print("expected schema name:", mf_test_configuration.mf_source_schema)
+    # Legacy time spine
+    time_spine_sources = {
+        legacy_time_spine_grain: TimeSpineSource(
+            schema_name=mf_test_configuration.mf_source_schema, table_name=time_spine_base_table_name
+        )
+    }
```
tlento, line 105: Wait, does this mean that we're no longer triggering the fallback behavior in the runtime, because we're overriding it with an explicit time spine input?

What happens if I define an HOUR grain time spine but leave the DAY grain one in the original configuration? Does that fail unceremoniously, do we raise an informative error, or do we just use the HOUR spine for everything?

I'm actually fine with raising an informative error or using the HOUR spine, especially for now, I just realized I'm just not clear on what happens.

metricflow-semantics/metricflow_semantics/test_helpers/semantic_manifest_yamls/simple_manifest/metrics.yaml
```diff
@@ -783,22 +783,3 @@ metric:
...
   filter: "{{ Metric('views', ['listing']) }} > 10"
   time_granularity: week
----
-metric:
-  name: bookings_before_dec_20_2019
-  description: Bookings up to the end of 2022
-  type: simple
-  type_params:
-    measure: bookings
-  filter: "{{ TimeDimension('metric_time') }} < '2012-12-20'"
```
tlento, line 793: Oh, I think I know why this is. Timestamp literals in Trino have to take the form `TIMESTAMP <literal>` so we'd need to do some substitution somewhere.

If the error is due to the filter it means we have to do custom rendering against the filter expr.

tlento (commented):
Quick question about the fallback behavior:

What happens when a user adds an HOUR grain time spine but not a DAY? What if they add a MONTH grain time spine but not a DAY?

metricflow-semantics/metricflow_semantics/test_helpers/semantic_manifest_yamls/simple_manifest/metrics.yaml
courtneyholcomb: Yep. I couldn't come up with an easy way to fix it here and wasn't sure it was worth doing the hard fix for an engine we barely use.

tests_metricflow/fixtures/source_table_snapshots/time_spine_table/mf_time_spine_hour.yaml
```diff
@@ -0,0 +1,42 @@
+table_snapshot:
+  table_name: mf_time_spine_hour
+  column_definitions:
+    - name: ts
+      type: TIME
+  rows:
+    - ["2020-01-01 01:00:00"]
```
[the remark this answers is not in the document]
courtneyholcomb: We don't apply that `DATE_TRUNC` in `JoinToTimeSpineNode` or `JoinOverTimeRangeNode`, But we do in `ReadSqlSourceNode` & `MetricTimeDimensionTransformNode`. This feature doesn't change that behavior so far, but we could discuss if we want to change it.
I think the warehouse validations will be a good idea so we can have more efficient queries.

tests_metricflow/fixtures/dataflow_fixtures.py
courtneyholcomb: The legacy time spine config will only be included in your manifest if you have the `metricflow_time_spine` model in your project. It will be overridden if you add configs pointing to a time spine with `DAY`. The legacy time spine will be treated like any other time spine you have configured.
If you have a new time spine with `HOUR` grain in addition to the legacy `DAY` time spine, we'll use the `DAY` time spine if you query something with `DAY` or higher. In that case we would only use the `HOUR` time spine if you query with `HOUR`. We'll choose the time spine with the largest grain available that can resolve the requested grain.

tests_metricflow/fixtures/source_table_snapshots/time_spine_table/mf_time_spine_microsecond.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - ["2020-01-01 00:00:00.000020"]
+    - ["2020-01-01 00:00:00.000021"]
+    - ["2020-01-01 00:00:00.000022"]
+    - ["2020-01-01 00:00:00.000023"]
+    - ["2020-01-01 00:00:00.000024"]
+    - ["2020-01-01 00:00:00.000025"]
+    - ["2020-01-01 00:00:00.000026"]
+    - ["2020-01-01 00:00:00.000027"]
+    - ["2020-01-01 00:00:00.000028"]
+    - ["2020-01-01 00:00:00.000029"]
```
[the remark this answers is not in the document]
courtneyholcomb: Fair enough - I can update that tomorrow! Shouldn't impact any of the tests, just will need to repopulate the source schemas.

courtneyholcomb:
> What happens when a user adds an HOUR grain time spine but not a DAY? What if they add a MONTH grain time spine but not a DAY?

@tlento If they configure `HOUR` but not `DAY` (and they don't have the legacy time spine model in their project), we'll use `HOUR` to resolve queries of all grains larger than or equal to `HOUR`.
If they configure `MONTH` only (with no legacy time spine), I actually set up the core parser to error so they won't be able to parse successfully. We're requiring at least one time spine using `DAY` or higher. (Previously, we would error during parsing if they didn't have the `metricflow_time_spine` model in their project specifically.)

metricflow-semantics/metricflow_semantics/test_helpers/semantic_manifest_yamls/simple_manifest/metrics.yaml
tlento: Agreed. I think the right way to deal with this is via a more holistic approach to filter expression inputs. In the meantime, having a small gap in Trino test coverage - particularly one where the main difference essentially boils down to how we go about pasting in the user-provided expression at render time - seems fine to me.

tlento (approved):
Ok, that all makes sense. We should follow up on:

1. granularity validations/date_trunc behaviors for time spine inputs (I favor validation, I think, as littering the codebase with date_trunc calls is kind of crappy)
2. longer term approach to default fallback interactions. For now this feels like a documentation problem but we may need to refine some of our internals a bit.

I think both of these will come up with custom calendar so we'll have an opportunity to re-visit these in the coming months.

courtneyholcomb:
> Ok, that all makes sense. We should follow up on:
> 
> 1. granularity validations/date_trunc behaviors for time spine inputs (I favor validation, I think, as littering the codebase with date_trunc calls is kind of crappy)
> 2. longer term approach to default fallback interactions. For now this feels like a documentation problem but we may need to refine some of our internals a bit.
> 
> I think both of these will come up with custom calendar so we'll have an opportunity to re-visit these in the coming months.

@tlento Sounds good! I'll put up tasks for those in the Wrap Up milestone for this project, and we can move them to the custom calendar project if we see fit later.
If you don't mind I'm going to change the date ranges of the time spine data in a follow up PR to avoid interfering with reviewability of the other PRs since this is at the bottom of the stack.

PR #1359, opened 2024-07-27
Handle no-metric queries with sub-daily `metric_time`

courtneyholcomb:
Makes some updates to enable no-metric queries with sub-daily time dimensions and adds integration tests for those queries.


metricflow-semantics/metricflow_semantics/model/semantics/linkable_spec_resolver.py
```diff
@@ -463,15 +466,20 @@ def _get_metric_time_elements(self, measure_reference: Optional[MeasureReference
...
-
-        # It's possible to aggregate measures to coarser time granularities (except with cumulative metrics).
-        possible_metric_time_granularities = tuple(
-            time_granularity
-            for time_granularity in TimeGranularity
-            if defined_granularity.is_smaller_than_or_equal(time_granularity)
-        )
+            # If querying metric_time without metrics, will query from time spines.
+            # Defaults to DAY granularity if available in time spines, else smallest available granularity.
+            min_time_spine_granularity = min(self._time_spine_sources.keys())
```
tlento, line 477: This default is because the time_spine_sources automatically include the DAY granularity (via `TimeSpineSource.create_from_manifest()`) if no custom time spines are found, correct?

metricflow-semantics/tests_metricflow_semantics/snapshots/test_semantic_model_container.py/tuple/test_linkable_elements_for_no_metrics_query__result0.txt
```diff
@@ -419,7 +419,6 @@
   'lux_listing__listing__lux_listing__twice_bookings_fill_nulls_with_0_without_time_spine',
   'lux_listing__listing__lux_listing__views',
   'lux_listing__listing__lux_listing__views_times_booking_value',
-  'metric_time__day',
```
tlento, line 422: Wait, is this expected?

tlento (commented):
There's a snapshot change I don't understand but if that's expected this seems ok.

I think we probably need to streamline how we handle all of these granularity locations. The TimeDimension granularity specification is quite confusing in this new world where it only fixes a minimum.

metricflow-semantics/metricflow_semantics/specs/time_dimension_spec.py
```diff
@@ -101,7 +101,12 @@ def without_first_entity_link(self) -> TimeDimensionSpec:  # noqa: D102
 
     @property
     def without_entity_links(self) -> TimeDimensionSpec:  # noqa: D102
-        return TimeDimensionSpec.from_name(self.element_name)
+        return TimeDimensionSpec(
+            element_name=self.element_name,
+            time_granularity=self.time_granularity,
+            date_part=self.date_part,
+            entity_links=(),
```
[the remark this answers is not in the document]
courtneyholcomb: Your PR comment gifs are always the best.

metricflow-semantics/metricflow_semantics/model/semantics/linkable_spec_resolver.py
courtneyholcomb: This default is set so that if you query ambiguous `metric_time` without metrics, we don't return nanoseconds if you have that time spine configured. Since `metric_time` defaults to `DAY` in all other scenarios (unless overridden by the metric's default granularity), this seemed like the expected behavior.

metricflow-semantics/tests_metricflow_semantics/snapshots/test_semantic_model_container.py/tuple/test_linkable_elements_for_no_metrics_query__result0.txt
courtneyholcomb: Yes! I had the same question and meant to leave a comment explaining but I forgot.
This test is specifically filtering out specs with the property `DERIVED_TIME_GRANULARITY`. I'm not sure why. But I made a change in this PR that makes all `metric_time` granularities derived for no-metric queries.
Debatable which behavior would be expected for that property here, but doesn't seem super important since it's only internal.

tlento (approved):
I could've sworn I approved this before lunch but apparently hunger was making me delusional....

PR #1361, opened 2024-07-29
Tests for metrics that require time spine with sub-daily granularity

courtneyholcomb:
More tests. Specifically covering metrics that require the time spine. Types covered:
- Cumulative metrics
- Offset metrics
- Join to time spine metrics


tlento (approved):
:shipit:

PR #1373, opened 2024-08-21
Add `custom_granularity_columns` to `TimeSpineSource`

courtneyholcomb:
Completes SL-2744

Add `custom_granularity_columns` to `TimeSpineSource`. Along the way:
- Bumped DSI version to get the new spec that includes custom granularities. Updated this in both MF and `metricflow-semantics`, since the spec was needed in both places. Now that mantle points to an older commit of MF, it should be fine to use a dev version here for now. We'll just need to update to a non-dev version if we ever need to update the commit that mantle is using.
- Updated the commit of `dbt-core` that we use for tests. This was needed to enable compatibility with DSI 0.7.0.
- Renamed old attributes that weren't specific enough to differentiate them from the new attributes.

### DIFF
--- a/extra-hatch-configuration/requirements.txt
+++ b/extra-hatch-configuration/requirements.txt
@@ -1,5 +1,5 @@
 Jinja2>=3.1.3
-dbt-semantic-interfaces==0.6.10
+dbt-semantic-interfaces==0.7.1.dev0
 more-itertools>=8.10.0, <10.2.0
 pydantic>=1.10.0, <3.0
 tabulate>=0.8.9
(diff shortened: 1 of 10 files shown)


tlento (commented):
This seems reasonable at a glance but I'd like to come back to it after I make a bit more progress with the container object for managing custom granularities.

courtneyholcomb:
> This seems reasonable at a glance but I'd like to come back to it after I make a bit more progress with the container object for managing custom granularities.

@tlento any thoughts on merging this yet? Seems easy enough to change later!

tlento (approved):
Seems reasonable!

PR #1375, opened 2024-08-21
Improve branch combining logic for `AggregateMeasuresNode`

courtneyholcomb:
There was a TODO in here to improve this logic, and it seemed pretty simple so I snagged it.
Previously, if there were aliases in any metric input measures, we would avoid trying to combine the `AggregateMeasuresNodes`. The comment clarified that the scenario we were trying to avoid was a case where different measures used the same alias. To make this more specific, I updated the logic to only skip combining nodes if duplicate aliases were found in the input measures.


tlento (approved):
lasso-whynot

PR #1381, opened 2024-08-22
Update aggregated conversion node to only have queried linkable specs

### LINKED ISSUE #1210: [SL-2194] Conversion metrics cannot render queries with filters unless the element is in the group by
In adding rendering tests for conversion metrics I noticed that a query like the following will raise an exception:

```
    query_spec = query_parser.parse_and_validate_query(
        metric_names=("visit_buy_conversion_rate",),
        group_by_names=("visit__referrer_id",),
        where_constraint=PydanticWhereFilter(
            where_sql_template=("{{ TimeDimension('metric_time', 'day') }} = '2020-01-01'")
        ),
    )
```

This fails with the following error:

` AssertionError: All parent nodes should have the same set of linkable instances since all values are coalesced.`

This is in the dataflow to sql query plan converter. It's possible this is due to removing filter specs from the conversion metric query as detailed in #1199 but as this may not be related I'm opening a separate issue for investigation.

<sub>[SL-2194](https://linear.app/dbt-labs/issue/SL-2194/conversion-metrics-cannot-render-queries-with-filters-unless-the)</sub>

WilliamDee:
## Context
When you try to query a conversion metric by supplying a filter with a dimension that exists in the base measure's semantic model without providing it in the group by, the rendering fails. This is because with the logic in the conversion metric builder, it ended up "requiring" all the linkable specs from the query (group by + filter), so the group by provided in the filter gets added to the resulting `SELECT` even though it was not supplied in the group by. This caused an error with the join as it didn't expect that additional element. This PR updates it so that instead of using all the required linkable specs, we should only use the queried linkable specs when building out the conversion aggregate node.

Resolves #1210 
Resolves SL-2777


courtneyholcomb:
Sweet! Can we add a test?

WilliamDee:
> Sweet! Can we add a test?

Oh ya i was just wondering if this looked good first!

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -361,7 +361,7 @@ def _build_aggregated_conversion_node(
         # Aggregate the conversion events with the JoinConversionEventsNode as the source node
         recipe_with_join_conversion_source_node = DataflowRecipe(
             source_node=join_conversion_node,
-            required_local_linkable_specs=base_measure_recipe.required_local_linkable_specs,
+            required_local_linkable_specs=queried_linkable_specs.as_tuple,
```
tlento, line 364: One of the things I love most about GitHub is how they don't allow me to comment on lines that haven't been modified. Seriously, have these people ever even reviewed code before?

Anyway, this does not work because of line 269 above. The queried_linkable_specs do not include any filter specs. So if you don't include required_local_linkable_specs you miss out on propagating any specs requested as filter elements that originate in the base measure dataset.

That's why the missing spec assertion is tripping in that conversion metric test. When we try to run a query that filters on visit__referrer_id, which is local to the base measure, that only gets added to the set of required local specs generated on line 269.

I do not know how to fix this while also fixing the other issue you're trying to resolve here. I don't even know if that's possible without a major overhaul to how we deal with filters. That said, it might be worth postponing a fix for this, since it's not clear to me what our filter behaviors against these nodes are meant to be. There are some fuzzy edge cases around applying a filter to two parts of a conversion calculation, and my sense is we do not correctly differentiate between input-measure-level filters and query-level filters, so whatever we do is going to produce unexpected results for somebody.

WilliamDee: Since the issue here was that the final aggregated conversion node had the extraneous group bys, could we just pass the final aggregated node through a `FilterElementNode` instead of fixing it like this?

tlento: Maybe? To be honest, I don't even know what's supposed to happen here, much less whether or not removing that column from the base output will solve this problem in a reasonable way. 

The basic question here is how should query-time filters get applied? If I run a query for conversions computed as visits / buys, and I filter to something like `WHERE visit__referrer_id = 1234 OR buy__coupon_id = 1234` in order to get all conversions from partner ID 1234, what is supposed to happen? If I add a filter to a conversion measure input to exclude `buyer__is_campaign_sourced` for our `organic_reach_conversion` metric, what should happen? What should happen if I apply the query-time filter for partner 1234 with the organic_reach_conversion metric?

I don't think we ever answered these questions, and while some of them are pretty obvious (the organice_reach_conversion filter should be properly applied) the others are less obvious to me, and I think there are a few where it really is just a product decision around how we want this to behave and what kind of traade-offs our users will have in the configuration experience.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -372,6 +377,16 @@ def _build_aggregated_conversion_node(
             predicate_pushdown_state=disabled_pushdown_state,
             measure_recipe=recipe_with_join_conversion_source_node,
         )
+        if not extraneous_linkable_specs.is_subset_of(queried_linkable_specs):
```
courtneyholcomb, line 380: Don't we need to this node BEFORE aggregating the measure?

courtneyholcomb: Otherwise we will aggregate with the filter column that we don't want included.
Can we write tests for the bug scenario (rendering & check query) to make sure this fixes it?

WilliamDee: removed this btw, and as for tests, i wrote them here and they include ones that had the bug https://github.com/dbt-labs/metricflow/pull/1381/commits/260148903cae5ace20539823092a24c74ae841d1

tests_metricflow/query_rendering/test_conversion_metric_rendering.py
```diff
@@ -164,3 +164,31 @@ def test_conversion_metric_with_window_and_time_constraint(
         dataflow_plan_builder=dataflow_plan_builder,
         query_spec=parsed_query.query_spec,
     )
+
+
+@pytest.mark.sql_engine_snapshot
+def test_conversion_metric_with_filter(
```
courtneyholcomb, line 170: Can we add a similar test for a windowed conversion metric? I think that requires a group by (right?) so would want a different group by than the one we're filtering on

courtneyholcomb (approved):
Requested one more test but otherwise looks great!!

PR #1385, opened 2024-08-29
Add ExpandedTimeGranularity struct to linkable elements

tlento:
With the impending introduction of custom granularities,
which are named granularities accessed via a lookup keyed
by a date/time value with a specified base grain, our reliance
on raw TimeGranularity enumeration elements will no longer
be sufficient for managing dataflow plans or queries.

In order to be able to reference custom granularities in
our TimeDimensionSpec and associated ElementPathKey and
LinkableElement subclasses we need to make space for both
the granularity name and a base grain value.

This change adds an ExpandedTimeGranularity struct to
encapsulate both of these pieces of information in a single
object value, and switches the LinkableDimension time granularity
property to use it in place of the bare TimeGranularity enum value.

This is just the first step. Later changes will migrate most
internal TimeGranularity-typed properties to the new ExpandedTimeGranularity
construct.

Once we have a proof of concept for this working we will decide
whether this construct belongs in dbt-semantic-interfaces or if we can
keep this internal to MetricFlow. The main reason to expand this
is for more natural support for time windows and offsets, but
we may be able to simply allow strings in the protocol spec and
convert them to the struct types inside of MetricFlow.

PR #1386, opened 2024-08-29
Remove `from_name` methods on spec classes

tlento:
A number of our spec classes have a `from_name` method which
wraps the StructuredLinkableSpecName.from_name method.

In the dominant majority of cases these calls are unnecessary,
and the behavior of our code would be far more clear with a targeted
set of assertions and standard dataclass initialization.

We also have a looming issue with custom granularities affecting
the way we manage granularity configurations, which will typically
require us to initialize spec classes in a context where we have
access to the properties of any custom granularity. This can't be
readily handled from within the spec classes themselves.

This change rmeoves the `from_name` methods to avoid proliferation of
unnecessary uses, and to streamline the process of moving to direct
initialization when custom granularities are involved.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/model/semantics/semantic_model_lookup.py
+++ b/metricflow-semantics/metricflow_semantics/model/semantics/semantic_model_lookup.py
@@ -24,6 +24,7 @@ from metricflow_semantics.errors.error_classes import InvalidSemanticModelError
 from metricflow_semantics.mf_logging.pretty_print import mf_pformat
 from metricflow_semantics.model.semantics.element_group import ElementGrouper
 from metricflow_semantics.model.spec_converters import MeasureConverter
+from metricflow_semantics.naming.linkable_spec_name import StructuredLinkableSpecName
 from metricflow_semantics.specs.dimension_spec import DimensionSpec
 from metricflow_semantics.specs.entity_spec import EntitySpec
 from metricflow_semantics.specs.instance_spec import LinkableInstanceSpec
@@ -241,17 +242,23 @@ class SemanticModelLookup:
             semantic_models_for_dimension = self._dimension_index.get(dim.reference, []) + [semantic_model]
             self._dimension_index[dim.reference] = semantic_models_for_dimension
 
+            assert StructuredLinkableSpecName.from_name(dim.name).is_element_name, (
+                f"Dimension name `{dim.name}` contains annotations, but this name should be the plain element name "
+                "from the original model. This should have been blocked by validation!"
+            )
+
+            # TODO: Construct these specs correctly. All of the time dimension specs have the default granularity
             self._dimension_ref_to_spec[dim.time_dimension_reference or dim.reference] = (
-                TimeDimensionSpec.from_name(dim.name)
+                TimeDimensionSpec(element_name=dim.name, entity_links=())
                 if dim.type is DimensionType.TIME
-                else DimensionSpec.from_name(dim.name)
+                else DimensionSpec(element_name=dim.name, entity_links=())
             )
 
         for entity in semantic_model.entities:
             semantic_models_for_entity = self._entity_index.get(entity.reference, []) + [semantic_model]
             self._entity_index[entity.reference] = semantic_models_for_entity
 
-            self._entity_ref_to_spec[entity.reference] = EntitySpec.from_name(entity.name)
+            self._entity_ref_to_spec[entity.reference] = EntitySpec(element_name=entity.name, entity_links=())
 
         self._semantic_model_reference_to_semantic_model[semantic_model.reference] = semantic_model
 
--- a/metricflow-semantics/metricflow_semantics/model/semantics/semantic_model_lookup.py
+++ b/metricflow-semantics/metricflow_semantics/model/semantics/semantic_model_lookup.py
@@ -242,10 +242,12 @@ class SemanticModelLookup:
             semantic_models_for_dimension = self._dimension_index.get(dim.reference, []) + [semantic_model]
             self._dimension_index[dim.reference] = semantic_models_for_dimension
 
-            assert StructuredLinkableSpecName.from_name(dim.name).is_element_name, (
-                f"Dimension name `{dim.name}` contains annotations, but this name should be the plain element name "
-                "from the original model. This should have been blocked by validation!"
-            )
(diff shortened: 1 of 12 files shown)


tlento:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1386?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1387** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1387?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1386** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1386?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1385** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1385?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

metricflow-semantics/metricflow_semantics/specs/non_additive_dimension_spec.py
```diff
@@ -33,6 +34,13 @@ class NonAdditiveDimensionSpec(SerializableDataclass):
     window_choice: AggregationType
     window_groupings: Tuple[str, ...] = ()
 
+    def __post_init__(self) -> None:
+        """Post init validator to ensure names with double-underscores are not allowed."""
+        assert self.name.find(DUNDER) == -1, (
```
courtneyholcomb, line 39: Did you already check that this is actually handled in validation? If not I would be worried this might be a braking change.

metricflow-semantics/metricflow_semantics/model/semantics/semantic_model_lookup.py
```diff
@@ -241,17 +242,23 @@ def _add_semantic_model(self, semantic_model: SemanticModel) -> None:
             semantic_models_for_dimension = self._dimension_index.get(dim.reference, []) + [semantic_model]
             self._dimension_index[dim.reference] = semantic_models_for_dimension
 
+            assert StructuredLinkableSpecName.from_name(dim.name).is_element_name, (
+                f"Dimension name `{dim.name}` contains annotations, but this name should be the plain element name "
+                "from the original model. This should have been blocked by validation!"
+            )
+
+            # TODO: Construct these specs correctly. All of the time dimension specs have the default granularity
             self._dimension_ref_to_spec[dim.time_dimension_reference or dim.reference] = (
```
courtneyholcomb, line 251: This dict seems pretty useless since the specs don't have any new info compared to the references 🤔 I wonder if we can remove it.

metricflow-semantics/metricflow_semantics/model/semantics/semantic_model_lookup.py
```diff
@@ -241,17 +242,23 @@ def _add_semantic_model(self, semantic_model: SemanticModel) -> None:
             semantic_models_for_dimension = self._dimension_index.get(dim.reference, []) + [semantic_model]
             self._dimension_index[dim.reference] = semantic_models_for_dimension
 
+            assert StructuredLinkableSpecName.from_name(dim.name).is_element_name, (
```
courtneyholcomb, line 245: Should we log this error instead of raising it? Worried it could be a breaking change, and this is triggered on init so it will get hit even if the incorrect object is never accessed.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -400,7 +401,11 @@ def _build_conversion_metric_output_node(
             descendent_filter_specs=metric_spec.filter_specs,
             queried_linkable_specs=queried_linkable_specs,
         )
-        entity_spec = EntitySpec.from_name(conversion_type_params.entity)
+        assert StructuredLinkableSpecName.from_name(conversion_type_params.entity).is_element_name, (
```
courtneyholcomb, line 404: Do we need to raise an error here? Similar to the last one, I don't want any unexpected blockers to show up when this released to MFS. Might be better if we could log it instead.

metricflow-semantics/metricflow_semantics/specs/non_additive_dimension_spec.py
tlento: I did, but it's kind of roundabout. This value has to be a name matching the name attribute of a Dimension object in the same semantic model as the measure. So it can't even be metric_time, no matter what our docs examples show.

metricflow-semantics/metricflow_semantics/model/semantics/semantic_model_lookup.py
tlento: This is unclear to me. It's used in conversion metrics somehow, so maybe we can replace it over there?

metricflow-semantics/metricflow_semantics/model/semantics/semantic_model_lookup.py
tlento: Good call, I'll make this change here and upstack for things which are in use.

tlento (commented):
Will update assertions to logger.warning level items for now

metricflow-semantics/metricflow_semantics/model/semantics/semantic_model_lookup.py
courtneyholcomb: SG! We might want to make sure we merge the whole stack at once in that case, otherwise this error could end up deploying to MFS unexpectedly.

metricflow-semantics/metricflow_semantics/specs/non_additive_dimension_spec.py
courtneyholcomb: So you're not worried it could be a breaking change? If you are at all worried, I'd recommend making this a logged error instead of raised like the other ones mentioned in this PR! But if not then ignore me

courtneyholcomb (approved):

tlento:
### Merge activity

* **Sep 3, 3:08 PM PDT**: @tlento started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/dbt-labs/metricflow/1386).

PR #1389, opened 2024-09-03
Remove unused DimensionSpecResolver class

tlento:
The DimensionSpecResolver did something or other with where
clause inputs, but it is not used anymore, if it ever was.

PR #1387, opened 2024-08-29
Allow custom granularity names in string name to struct conversion

tlento:
MetricFlow relies on a class called StructuredLinkableSpecName to manage
conversion between structured names and their string representations,
e.g., `listing__is_lux` or `listing__created_at__month`.

Up until now the time granularity component of this naming struct
required an enumeration value, which would then be converted to its
string representation at serialization time. With custom granularities
we need to explicitly allow for a string name for the input time
granularity. This implements that change, along with some added
enforcement to ensure consistency in how we conver to/from the string
representations of these names.

PR #1390, opened 2024-09-03
Use ExpandedTimeGranularity in TimeDimensionSpec

tlento:
Use ExpandedTimeGranularity in TimeDimensionSpec

In order to support custom granularities we need a way to propagate
them through the dataflow plan building and plan conversion processes.
We typically use the TimeDimensionSpec classes for carrying that
information, since the specs indicate the grain of the specific
variation of the TimeDimension element we're operating on. We
decided to use an ExpandedTimeGranularity construct to encapsulate
the custom granularity metadata in a manner interchangeable with the
default granularity enumeration values, with the idea that these
would make it more obvious when and how to do comparisons and things.

This change updates the TimeDimensionSpec itself to use the new construct.
The changes here are meant to be as mechanical as possible, so there are
places where restructuring or follow-up is needed. It was not possible to
do purely mechanical changes in all cases.

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -54,6 +54,7 @@ from metricflow_semantics.specs.where_filter.where_filter_transform import Where
 from metricflow_semantics.sql.sql_join_type import SqlJoinType
 from metricflow_semantics.sql.sql_table import SqlTable
 from metricflow_semantics.time.dateutil_adjuster import DateutilTimePeriodAdjuster
+from metricflow_semantics.time.granularity import ExpandedTimeGranularity
 
 from metricflow.dataflow.builder.node_data_set import DataflowPlanNodeOutputDataSetResolver
 from metricflow.dataflow.builder.node_evaluator import (
@@ -440,7 +441,10 @@ class DataflowPlanBuilder:
         predicate_pushdown_state: PredicatePushdownState,
         for_group_by_source_node: bool = False,
     ) -> DataflowPlanNode:
-        default_granularity = self._metric_lookup.get_min_queryable_time_granularity(metric_spec.reference)
+        # TODO: [custom granularity] Figure out how to support custom granularities as defaults
+        default_granularity = ExpandedTimeGranularity.from_time_granularity(
+            self._metric_lookup.get_min_queryable_time_granularity(metric_spec.reference)
+        )
 
         queried_agg_time_dimensions = queried_linkable_specs.included_agg_time_dimension_specs_for_metric(
             metric_reference=metric_spec.reference, metric_lookup=self._metric_lookup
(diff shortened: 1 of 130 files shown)


tlento:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1390?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1390** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1390?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1389** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1389?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1387** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1387?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1386** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1386?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1385** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1385?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -1556,18 +1560,18 @@ def _build_aggregated_measure_from_measure_source_node(
...
-            ] = self._find_non_additive_dimension_in_linkable_specs(
-                agg_time_dimension=agg_time_dimension,
-                linkable_specs=queried_linkable_specs.as_tuple,
-                non_additive_dimension_spec=non_additive_dimension_spec,
+            queried_time_dimension_spec: Optional[TimeDimensionSpec] = (
+                self._find_non_additive_dimension_in_linkable_specs(
+                    agg_time_dimension=agg_time_dimension,
+                    linkable_specs=queried_linkable_specs.as_tuple,
+                    non_additive_dimension_spec=non_additive_dimension_spec,
+                )
```
tlento, line 1568: Ignore, lint version mismatch, fixed in later commit.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -996,18 +1004,25 @@ def visit_constrain_time_range_node(self, node: ConstrainTimeRangeNode) -> SqlDa
             ds >= '2020-01-01' AND ds <= '2020-02-01'
 
         instead of this: DATE_TRUNC('month', ds) >= '2020-01-01' AND DATE_TRUNC('month', ds <= '2020-02-01')
+
+        Since time range constraints are always bound by a range of standard date/time values, this conversion
```
courtneyholcomb, line 1008: Maybe we should allow custom granularities here? Won't they be date/time values, too? We could enforce that if needed, but maybe that's too restrictive.

courtneyholcomb (approved):
Nothing too controversial in here, looks good! Left one comment inline.

metricflow/plan_conversion/dataflow_to_sql.py
tlento: The values can be arbitrary, there's no guarantee they look like date/time at all. I can have a retail calendar with entries like Y2024Q01M02 and Y2024Q02M02.

For something like that we'd have to do the range adjustment based on a query against the time spine, which we can't do at query planning time, or by doing a full join. Either way I don't think it's worth the effort.

There is an interesting issue here - a custom grain defined in a cumulative metric means we'll never be able to do time range constraint predicate push down for that metric unless we get some added input from the user.

courtneyholcomb: Well, can't we apply a time range constraint on the base grain?

PR #1394, opened 2024-09-05
Lowercase all names passed into `StructuredLinkableSpecName`

courtneyholcomb:
Per title - lowercase all inputs to avoid errors later.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/naming/linkable_spec_name.py
+++ b/metricflow-semantics/metricflow_semantics/naming/linkable_spec_name.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import logging
-from dataclasses import dataclass
 from typing import Optional, Tuple
 
 from dbt_semantic_interfaces.type_enums.date_part import DatePart
@@ -14,7 +13,6 @@ DUNDER = "__"
 logger = logging.getLogger(__name__)
 
 
-@dataclass(frozen=True)
 class StructuredLinkableSpecName:
     """Parse a qualified name into different parts.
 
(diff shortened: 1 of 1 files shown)


metricflow-semantics/metricflow_semantics/naming/linkable_spec_name.py
```diff
@@ -29,6 +27,18 @@ class StructuredLinkableSpecName:
...
+    def __init__(  # noqa: D107
+        self,
+        entity_link_names: Tuple[str, ...],
+        element_name: str,
+        time_granularity_name: Optional[str] = None,
+        date_part: Optional[DatePart] = None,
+    ) -> None:
+        self.entity_link_names = tuple([entity_link_name.lower() for entity_link_name in entity_link_names])
+        self.element_name = element_name.lower()
+        self.time_granularity_name = time_granularity_name.lower() if time_granularity_name else None
```
tlento, line 39: Careful, the time_granularity_name can be defined in an enum, if we alter it then calling TimeGranularity(time_granularity_name) won't work. More importantly, the checks we have elsewhere to see if a given granularity is custom or not won't work for those values either.

Let's move the assertion about is_standard_name on the TimeGranularity enumeration into this initializer (unless there's a way to do that assertion in the enum class itself). Then we can delete the other dataclass elements.

metricflow-semantics/metricflow_semantics/naming/linkable_spec_name.py
```diff
@@ -29,6 +27,18 @@ class StructuredLinkableSpecName:
...
+        self.date_part = date_part
+
     def __post_init__(self) -> None:
```
tlento, line 42: This is a dataclass construct and will no longer be called. Let's merge the remaining bits into `__init__` and then remove this method.

metricflow-semantics/metricflow_semantics/naming/linkable_spec_name.py
```diff
@@ -29,6 +27,18 @@ class StructuredLinkableSpecName:
     time_granularity_name: Optional[str] = None
     date_part: Optional[DatePart] = None
```
tlento, line 28: Let's remove these and define property accessor methods instead.

tlento (approved):
I'm approving to unblock the various cross-repo wrangling we're trying to get in place, but this change needs a follow-up as it introduces a bunch of subtle issues with this class that will crop up over time. It's up to you whether to do this before merge or separately.

metricflow-semantics/metricflow_semantics/naming/linkable_spec_name.py
courtneyholcomb: Oooh TIL!! I thought this applied to all Python classes!

tlento:
Actually ignore the enum validation stuff and just delete post init. I found out how to verify enum structure so we'll fix that in DSI.

metricflow-semantics/metricflow_semantics/naming/linkable_spec_name.py
courtneyholcomb: Oops that was just an oversight 🤦‍♀️ We shouldn't need to add property methods, having public attributes in the init should cover it.

PR #1395, opened 2024-09-05
Add `JoinToCustomGranularityNode`

courtneyholcomb:
Basics for adding the new `JoinToCustomGranularityNode`. We will add one of these nodes for each custom granularity requested in the query, since there will need to be a separate join for each one. 
This PR includes the node itself and the basic visitor methods. The SQL rendering node will be added in a follow-up commit since the logic is more complex there. This node is never used in the dataflow plan at this point.
@tlento I left a couple questions inline about these methods to make sure I'm taking the right approach - I'm not completely familiar with some of the logic for those visitors.


metricflow/dataflow/optimizer/predicate_pushdown_optimizer.py
```diff
@@ -458,6 +459,11 @@ def visit_join_conversion_events_node(self, node: JoinConversionEventsNode) -> O
             )
         )
 
+    def visit_join_to_custom_granularity_node(  # noqa: D102
```
courtneyholcomb, line 462: @tlento I might need some help to figure this one out correctly - I looked at the logic in `visit_join_to_time_spine_node()` to try to emulate that, but I didn't totally understand it, so it would be helpful for me if we could talk through it!

tests_metricflow/dataflow/optimizer/source_scan/test_source_scan_optimizer.py
```diff
@@ -110,6 +111,9 @@ def visit_add_generated_uuid_column_node(self, node: AddGeneratedUuidColumnNode)
     def visit_join_conversion_events_node(self, node: JoinConversionEventsNode) -> int:  # noqa: D102
         return self._sum_parents(node)
 
+    def visit_join_to_custom_granularity_node(self, node: JoinToCustomGranularityNode) -> int:  # noqa: D102
```
courtneyholcomb, line 114: @tlento Should the time spine be considered a source node? It isn't for other nodes, so I didn't treat it as such here. But we should update any nodes that pull in the time spine here if it should be

metricflow/dataflow/optimizer/predicate_pushdown_optimizer.py
tlento: Yeah, this is complicated. As discussed offline, raising here and disabling the optimizer seems like our best option.

tests_metricflow/dataflow/optimizer/source_scan/test_source_scan_optimizer.py
tlento: As discussed offline.... Maybe?

I think our best path forward is to move off of our current single-pass approach and instead build CTE source nodes in one pass and then layer the metric nodes on top of those. If we do a good job with that this optimizer will no longer be necessary.

tlento (approved):
Seems reasonable!

PR #1399, opened 2024-09-06
`TimeSpineSource` additions & rename `SqlTableFromClauseNode` -> `SqlTableNode`

courtneyholcomb:
More setup that will be used for rendering custom granularity SQL.

Previously, we were never rendering a join with only a table (no subquery), but that will change for custom granularities. Those joins will use this node, which essentially just renders a table name and alias. I've renamed the node to reflect that change.

Also adds some functionality to `TimeSpineSource` that will be needed soon.


tlento (approved):
:shipit:

PR #1401, opened 2024-09-06
Include custom granularities in source nodes

courtneyholcomb:
Add custom granularities to source node datasets and other source node handling. Reviewing by commit is recommended.

### DIFF
--- a/metricflow/dataset/convert_semantic_model.py
+++ b/metricflow/dataset/convert_semantic_model.py
@@ -516,20 +516,20 @@ class SemanticModelToDataSetConverter:
         """Build data set for time spine."""
         from_source_alias = SequentialIdGenerator.create_next_id(StaticIdPrefix.TIME_SPINE_SOURCE).str_value
         base_granularity = time_spine_source.base_granularity
-        time_column_name = time_spine_source.base_column
+        base_column_name = time_spine_source.base_column
 
         time_dimension_instances: List[TimeDimensionInstance] = []
         select_columns: List[SqlSelectColumn] = []
 
         # Build base time dimension instances & columns
         base_time_dimension_instance = self._create_time_dimension_instance(
-            element_name=time_column_name,
+            element_name=base_column_name,
             entity_links=(),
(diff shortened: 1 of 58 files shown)


metricflow/dataset/convert_semantic_model.py
```diff
@@ -515,35 +515,55 @@ def create_sql_source_data_set(self, semantic_model: SemanticModel) -> SemanticM
...
         )
-        time_dimension_instances.extend(new_instances)
-        select_columns.extend(new_columns)
+        time_dimension_instances.extend(new_base_instances)
+        select_columns.extend(new_base_columns)
+
+        # Build custom granularity time dimension instances & columns
+        for custom_granularity in time_spine_source.custom_granularities:
+            custom_time_dimension_instance = self._create_time_dimension_instance(
+                element_name=time_column_name,
```
tlento, line 552: Can we add a comment explaining why we use the time_column_name here? It feels odd as a reader, because it seems like we should be using the custom name, but from looking at the snapshots I think we need this so that the granularity is appended as it is elsewhere - i.e., we want `metric_time__martian_day` not `martian_day`, and that implies also `ds__martian_day` instead of `martian_day`.

tests_metricflow/snapshots/test_metric_time_without_metrics.py/SqlQueryPlan/DuckDB/test_dimensions_with_time_constraint__plan0.sql
```diff
@@ -204,6 +204,7 @@ FROM (
           , subq_1.ds__extract_day
           , subq_1.ds__extract_dow
           , subq_1.ds__extract_doy
+          , subq_1.ds__martian_day
```
tlento, line 207: Ah, this is why we use the same name, so we're setting the standard that the base grain is encoded in the name in all cases.

tlento (approved):
LGTM, assuming I understood everything correctly. If so please do add a little explanation.

PR #1403, opened 2024-09-09
Make SemanticManifestLookup available for spec pattern creation

tlento:
MetricFlow's query input parameter processing runs through a
pattern matching/parsing step that then produces a parameter spec
for subsequent matching during group by resolution. In practice,
every query parameter either directly references or is derived
directly from one or more elements in a semantic manifest.

Converting a parameter into a corresponsing spec pattern for a
given query input can, therefore, benefit from access to
the semantic manifest lookup objects we provide. This will
allow us to do things like map custom granularities to their
corresponding base granularities when doing group by resolution.

This change simply makes the semantic manifest available to the
method that converts between the query input parameter and the
spec pattern. It isolates the availability of the semantic
manifest lookup to this method because the other methods in the
relevant classes should generally not be using the lookup to
perform their operations, and the change as implemented
here also has a more limited blast radius.

PR #1402, opened 2024-09-09
Remove unused parse_object_builder_naming_scheme helper function

tlento:
This module, and the lone function it contains, is impeding progress.
Happily, it is no longer used, so I'm deleting it.

PR #1404, opened 2024-09-09
Switch query parameter classes onto ExpandedTimeGranularity

tlento:
In preparation for supporting user-provided custom granularity names
at query time we need to update our query parameter processing
layer to use ExpandedTimeGranularities internally while accepting
string names from the original user input.

This change makes that update, although it is necessarily incomplete.
The TimeDimensionCallParameterSet, in particular, still uses an
enumeration-typed property for representing the user-requested grain,
but this interface is imported from dbt-semantic-interfaces and will
be updated in a later series of changes.

Follow-ups inside MetricFlow will enable lookups against custom granularity
values as we solidify our test layouts.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/naming/object_builder_scheme.py
+++ b/metricflow-semantics/metricflow_semantics/naming/object_builder_scheme.py
@@ -81,7 +81,7 @@ class ObjectBuilderNamingScheme(QueryItemNamingScheme):
                 ParameterSetField.DATE_PART,
             ]
 
-            time_granularity = None
+            time_granularity: Optional[ExpandedTimeGranularity] = None
             if time_dimension_call_parameter_set.time_granularity is not None:
                 fields_to_compare.append(ParameterSetField.TIME_GRANULARITY)
                 time_granularity = ExpandedTimeGranularity.from_time_granularity(
--- a/metricflow-semantics/metricflow_semantics/naming/object_builder_scheme.py
+++ b/metricflow-semantics/metricflow_semantics/naming/object_builder_scheme.py
@@ -23,6 +23,7 @@ from metricflow_semantics.specs.patterns.entity_link_pattern import (
 )
 from metricflow_semantics.specs.patterns.spec_pattern import SpecPattern
 from metricflow_semantics.specs.patterns.typed_patterns import DimensionPattern, TimeDimensionPattern
+from metricflow_semantics.time.granularity import ExpandedTimeGranularity
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


tlento:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1404?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1404** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1404?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1403** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1403?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1402** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1402?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

metricflow-semantics/metricflow_semantics/naming/object_builder_scheme.py
```diff
@@ -80,14 +81,18 @@ def spec_pattern(self, input_str: str, semantic_manifest_lookup: SemanticManifes
                 ParameterSetField.DATE_PART,
             ]
 
+            time_granularity = None
```
courtneyholcomb, line 84: surprised this didn't require a type annotation 🤔

metricflow-semantics/metricflow_semantics/specs/patterns/entity_link_pattern.py
```diff
@@ -136,13 +120,10 @@ def match(self, candidate_specs: Sequence[InstanceSpec]) -> Sequence[LinkableIns
         # Entity links could be a partial match, so it's handled separately.
         if ParameterSetField.ENTITY_LINKS in self.parameter_set.fields_to_compare:
             filtered_candidate_specs = self._match_entity_links(filtered_candidate_specs)
-        # Time granularities are special, so they are also handled separately.
```
courtneyholcomb, line 139: So we just won't check this field anymore ?

tests_metricflow/integration/test_configured_cases.py
```diff
@@ -261,8 +261,6 @@ def test_case(
         if date_part or grain:
             if date_part:
                 kwargs["date_part"] = DatePart(date_part)
-            if grain:
```
courtneyholcomb, line 264: how did this not break anything 🤔

metricflow-semantics/metricflow_semantics/naming/object_builder_scheme.py
tlento: Good call, I'll add one just to make future errors less confusing.

As for why it's not required, MyPy works in mysterious ways...... but in this case the type assignment flow is simple. In the end it can only be `None | ExpandedTimeGranularity` because there's just that one conditional branch below. If someone inserts another branch with a new potential type assignment that'll alter the type to `None | ExpandedTimeGranularity | whatever` and we'll get a confusing type error at the point where we try to pass this variable into `EntityLinkPatternParameterSet.from_parameters`.

metricflow-semantics/metricflow_semantics/specs/patterns/entity_link_pattern.py
tlento: We check it via the standard comparison, see the change below to the set.difference() call.

tests_metricflow/integration/test_configured_cases.py
tlento: This is the fix for a change to TimeDimensionParameter above. 

We no longer need to convert the string input to a TimeGranularity in the TimeDimensionParameter initializer - it accepts string values - so without this change we were passing an incorrectly typed value in for that argument and weird runtime failures were cropping up.

PR #1406, opened 2024-09-16
wip - allow grain in where filter

courtneyholcomb:


courtneyholcomb:
Not related to metric filters - this is super out of date! This PR was working on adding support for grain in where filters on time dimensions, which has been supported for a while now.

PR #1407, opened 2024-09-18
Select appropriate time spine node for `metric_time` queries without metrics

courtneyholcomb:
There are very few cases where multiple source nodes can satisfy the same linkable spec, but one instance in which that is the case is when `metric_time` is queried without metrics. In that case, we will use a time spine node to satisfy `metric_time`. Previously, we were choosing the time spine nodes that could support the selected grain, and then narrowing down to the one with the fewest satisfiable linkable specs. This logic was working because, before custom granularities, the time spine with the fewest satisfiable linkable specs was always the time spine with the largest compatible base granularity. With the addition of custom granularities, though, the time spine with the largest compatible base granularity might have any number of satisfiable linkable specs. This changes updates the source node selection logic to choose the time spine with the largest compatible base granularity, regardless of the number of satisfiable linkable specs.


tlento (approved):
:shipit:

PR #1408, opened 2024-09-18
Add SQL rendering logic for custom granularities

courtneyholcomb:
Render SQL for custom granularities.
The `JoinToCustomGranularityNode` will build a time spine dataset and `LEFT OUTER JOIN` it to the parent dataset, without a subquery. This is done by joining the time spine table directly, instead of creating a subquery that selects from the time spine table and then joins. That makes this the first node that does not create its own subquery. This join should be safe to do without a subquery since it is always a `LEFT OUTER JOIN`, ensuring the number of rows will not be changed by the join, making it relatively safe for optimizers to handle.
This behavior was discussed with @tlento and agreed upon in order to ensure optimization and readability.
You can see examples of what this SQL will look like in a [testing PR](https://github.com/dbt-labs/metricflow/pull/1410) higher up the stack.

### DIFF
--- a/metricflow/plan_conversion/dataflow_to_sql.py
+++ b/metricflow/plan_conversion/dataflow_to_sql.py
@@ -1451,11 +1451,6 @@ class DataflowToSqlQueryPlanConverter(DataflowPlanNodeVisitor[SqlDataSet]):
         left_expr_for_join: SqlExpressionNode = SqlColumnReferenceExpression.from_table_and_column_names(
             table_alias=parent_alias, column_name=parent_time_dimension_instance.associated_column.column_name
         )
-        if parent_time_dimension_instance.spec.time_granularity.base_granularity != time_spine_source.base_granularity:
-            # If needed, apply DATE_TRUNC to parent column match the time spine spine that's column being joined to.
-            left_expr_for_join = SqlDateTruncExpression.create(
-                time_granularity=time_spine_source.base_granularity, arg=left_expr_for_join
-            )
         join_description = SqlJoinDescription(
             right_source=SqlTableNode.create(sql_table=time_spine_source.spine_table),
             right_source_alias=time_spine_alias,
--- a/metricflow/plan_conversion/dataflow_to_sql.py
+++ b/metricflow/plan_conversion/dataflow_to_sql.py
@@ -199,6 +199,9 @@ class DataflowToSqlQueryPlanConverter(DataflowPlanNodeVisitor[SqlDataSet]):
         self._time_spine_sources = TimeSpineSource.build_standard_time_spine_sources(
             semantic_manifest_lookup.semantic_manifest
         )
+        self._custom_granularity_time_spine_sources = TimeSpineSource.build_custom_time_spine_sources(
+            tuple(self._time_spine_sources.values())
+        )
 
     @property
     def column_association_resolver(self) -> ColumnAssociationResolver:  # noqa: D102
@@ -262,9 +265,10 @@ class DataflowToSqlQueryPlanConverter(DataflowPlanNodeVisitor[SqlDataSet]):
             column_alias = self.column_association_resolver.resolve_spec(agg_time_dimension_instance.spec).column_name
             # If the requested granularity is the same as the granularity of the spine, do a direct select.
             # TODO: also handle date part.
-            # TODO: [custom granularity] add support for custom granularities to make_time_spine_data_set
             agg_time_grain = agg_time_dimension_instance.spec.time_granularity
-            assert not agg_time_grain.is_custom_granularity, "Custom time granularities are not yet supported!"
+            assert (
+                not agg_time_grain.is_custom_granularity
+            ), "Custom time granularities are not yet supported for all queries."
             if agg_time_grain.base_granularity == time_spine_source.base_granularity:
                 select_columns += (SqlSelectColumn(expr=column_expr, column_alias=column_alias),)
             # If any columns have a different granularity, apply a DATE_TRUNC() and aggregate via group_by.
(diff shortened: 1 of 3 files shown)


metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -262,9 +265,10 @@ def _make_time_spine_data_set(
             column_alias = self.column_association_resolver.resolve_spec(agg_time_dimension_instance.spec).column_name
             # If the requested granularity is the same as the granularity of the spine, do a direct select.
             # TODO: also handle date part.
-            # TODO: [custom granularity] add support for custom granularities to make_time_spine_data_set
             agg_time_grain = agg_time_dimension_instance.spec.time_granularity
-            assert not agg_time_grain.is_custom_granularity, "Custom time granularities are not yet supported!"
+            assert (
+                not agg_time_grain.is_custom_granularity
+            ), "Custom time granularities are not yet supported for all queries."
```
tlento, line 271: Is this because this block is for direct references to agg_time_dimension_instances, and we are currently only supporting this for metric_time? Or is this just not hooked up yet?

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1408,8 +1410,108 @@ def visit_join_to_time_spine_node(self, node: JoinToTimeSpineNode) -> SqlDataSet
...
+        assert parent_time_dimension_instance, (
+            "JoinToCustomGranularityNode's expected time_dimension_spec not found in parent dataset instances. "
+            "This indicates internal misconfiguration."
+        )
+
+        # Build join expression.
+        time_spine_alias = self._next_unique_table_alias()
+        custom_granularity_name = node.time_dimension_spec.time_granularity.name
+        time_spine_source = self._get_time_spine_for_custom_granularity(custom_granularity_name)
+        left_expr_for_join: SqlExpressionNode = SqlColumnReferenceExpression.from_table_and_column_names(
```
tlento, line 1451: Was chatting with @theyostalservice about why we merged nodes instead of keeping them separate and I realized it's partly because this always has to be a left outer join, right? Given this, it's always safe for things like predicate pushdown to push any predicate past this node, since no rows will be filtered out by the internal join here.

More generally, it's probably worth it for us to document the reasoning for deviating from our previous patterns in the PR description before merge.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1408,8 +1410,108 @@ def visit_join_to_time_spine_node(self, node: JoinToTimeSpineNode) -> SqlDataSet
...
+            table_alias=parent_alias, column_name=parent_time_dimension_instance.associated_column.column_name
+        )
+        if parent_time_dimension_instance.spec.time_granularity.base_granularity != time_spine_source.base_granularity:
+            # If needed, apply DATE_TRUNC to parent column match the time spine spine that's column being joined to.
+            left_expr_for_join = SqlDateTruncExpression.create(
+                time_granularity=time_spine_source.base_granularity, arg=left_expr_for_join
+            )
```
tlento, line 1458: Are we certain that the parent grain is always finer than the time spine source grain here? Do we need to switch on the `.to_int()` cases so we date_trunc to whatever is smaller?

Or does it matter? Maybe we only need to date_trunc parent grain since it'll always resolve to the 00 moment, like 2024-01-01T00:00:00? If so then this is fine but maybe we add a comment or something?

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1408,8 +1410,108 @@ def visit_join_to_time_spine_node(self, node: JoinToTimeSpineNode) -> SqlDataSet
...
+
+        # Build output time spine instances and columns.
+        time_spine_instance = TimeDimensionInstance(
+            defined_from=parent_time_dimension_instance.defined_from,
+            associated_columns=(self._column_association_resolver.resolve_spec(node.time_dimension_spec),),
+            spec=node.time_dimension_spec,
+        )
+        time_spine_instance_set = InstanceSet(time_dimension_instances=(time_spine_instance,))
+        time_spine_select_columns = (
```
tlento, line 1496: This only does one, so if I request `metric_time__martian_year` and `metric_time__martian_quarter` we'll do multiple joins. I think that's ok for now, we can optimize it later if we feel like it.

tlento (commented):
I have a few questions inline. My main worries are the assertions, which maybe are addressed upstack, and the thing about the date_trunc. I *think* this works as-is, but I'm always skittish about relying on date/time operations behaving sensibly. Let's sync up about it really quickly and if we need changes we can figure out if it's easier to merge this as-is and apply them on top of the stack or if we can fix up here and rebase.

metricflow/plan_conversion/dataflow_to_sql.py
courtneyholcomb: Just not hooked up yet!! (`agg_time_dimension_instances` here might include `metric_time`.) This was more just an error for me to encounter intentionally when working on custom granularity queries that encounter the `JoinToTimeSpineNode` or `JoinOverTimeRangeNode`, which I wanted to test more thoroughly. Also didn't want users to encounter a more obscure error if they were playing around with this before it was ready.

metricflow/plan_conversion/dataflow_to_sql.py
courtneyholcomb: Yes, good call!

metricflow/plan_conversion/dataflow_to_sql.py
courtneyholcomb: Looking back at this, I don't think this `DATE_TRUNC` is ever actually necessary. I think there was just a bug somewhere in my code when I added it that made me think it was necessary.
The reason it's not necessary is that on line 1439, we choose this parent instance BECAUSE it is the instance that matches the base grain.
I tested removing this logic at the top of the stack and nothing change, so seems safe to remove. Will remove it here since I need to add some comments to this PR anyway!

metricflow/plan_conversion/dataflow_to_sql.py
courtneyholcomb: Also - I haven't encountered this in my basic testing thus far, but I have a task up to make sure this gets handled/removed before we finish the production-ready feature.

tlento (approved):
I sure did forget to approve this one....

ted lasso oops

PR #1409, opened 2024-09-18
Dataflow plan for custom granularities

courtneyholcomb:
Update the `DataflowPlanBuilder` to support custom granularities. Steps included:
- For each custom granularity requested, add the appropriate `JoinToCustomGranularityNode` to the `DataflowPlan`.
- When looking for nodes that can satisfy a given linkable spec, check for the ability to satisfy the spec's base granularity, not the custom granularity requested. That will be joined in later.

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -812,13 +812,8 @@ class DataflowPlanBuilder:
 
         for time_dimension_spec in required_linkable_specs.time_dimension_specs:
             if time_dimension_spec.time_granularity.is_custom_granularity:
-                include_base_grain = (
-                    time_dimension_spec.with_base_grain() in required_linkable_specs.time_dimension_specs
-                )
                 output_node = JoinToCustomGranularityNode.create(
-                    parent_node=output_node,
-                    time_dimension_spec=time_dimension_spec,
-                    include_base_grain=include_base_grain,
+                    parent_node=output_node, time_dimension_spec=time_dimension_spec
                 )
 
         if len(query_level_filter_specs) > 0:
@@ -1449,6 +1444,12 @@ class DataflowPlanBuilder:
         extraneous_linkable_specs = LinkableSpecSet.merge_iterable(linkable_spec_sets_to_merge).dedupe()
         required_linkable_specs = queried_linkable_specs.merge(extraneous_linkable_specs).dedupe()
 
+        # Custom grains require joining to their base grain, so add base grain to extraneous specs.
+        base_grain_set = LinkableSpecSet.create_from_specs(
+            [spec.with_base_grain() for spec in required_linkable_specs.time_dimension_specs_with_custom_grain]
+        )
+        extraneous_linkable_specs = extraneous_linkable_specs.merge(base_grain_set).dedupe()
+
         return required_linkable_specs, extraneous_linkable_specs
 
     def _build_aggregated_measure_from_measure_source_node(
@@ -1611,7 +1612,12 @@ class DataflowPlanBuilder:
             )
 
             specs_to_keep_after_join = InstanceSpecSet(measure_specs=(measure_spec,)).merge(
-                InstanceSpecSet.create_from_specs(required_linkable_specs.as_tuple),
+                InstanceSpecSet.create_from_specs(
+                    [
+                        spec.with_base_grain() if isinstance(spec, TimeDimensionSpec) else spec
+                        for spec in required_linkable_specs.as_tuple
+                    ]
+                ),
             )
 
             after_join_filtered_node = FilterElementsNode.create(
@@ -1621,15 +1627,10 @@ class DataflowPlanBuilder:
         else:
             unaggregated_measure_node = filtered_measure_source_node
 
-        for time_dimension_spec in queried_linkable_specs.time_dimension_specs:
+        for time_dimension_spec in required_linkable_specs.time_dimension_specs:
             if time_dimension_spec.time_granularity.is_custom_granularity:
-                include_base_grain = (
-                    time_dimension_spec.with_base_grain() in required_linkable_specs.time_dimension_specs
-                )
                 unaggregated_measure_node = JoinToCustomGranularityNode.create(
-                    parent_node=unaggregated_measure_node,
-                    time_dimension_spec=time_dimension_spec,
-                    include_base_grain=include_base_grain,
+                    parent_node=unaggregated_measure_node, time_dimension_spec=time_dimension_spec
                 )
 
         # If time constraint was previously adjusted for cumulative window or grain, apply original time constraint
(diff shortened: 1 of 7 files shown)


metricflow/dataflow/builder/node_evaluator.py
```diff
@@ -406,6 +407,10 @@ def evaluate_node(
         logger.debug(f"Candidate spec set is:\n{mf_pformat(candidate_spec_set)}")
 
         data_set_linkable_specs = candidate_spec_set.linkable_specs
+        # Look for which nodes can satisfy the linkable specs at their base grains. Custom grains will be joined later.
+        required_linkable_specs_with_base_grains = [
+            spec.with_base_grain if isinstance(spec, TimeDimensionSpec) else spec for spec in required_linkable_specs
```
tlento, line 412: I just realized this with_base_grain thing isn't a function. Is this a property? If so, what is it doing? It looks like it's creating a new spec class with some property set, but if that's the case can't we have this be a property accessor defined directly? It doesn't seem like it saves us from any further instance checking/type refinement operations down the line.

Or is this switching the dimension grain from a custom one to the base grain? If so, do we need to do that, or should we just point to the `granularity.base_grain` or whatever the properties are called where needed?

tlento: Update - we just looked at the definition in the branch and I think it'd be a good idea to make with_base_grain a function, it's more obvious that we're converting the spec into a different spec rather than accessing some granularity property or something.

Separately it'd be nice to not have to do this type refinement but I think it's fine here for now.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -885,11 +895,25 @@ def _select_source_nodes_with_linkable_specs(
...
+
+        # Find the source node that will satisfy the base granularity. Custom granularities will be joined in later.
+        linkable_specs_set_with_base_granularities: Set[LinkableInstanceSpec] = set()
+        # TODO: Add support for no-metrics queries for custom grains without a join (i.e., select directly from time spine).
+        for linkable_spec in linkable_specs.as_tuple:
+            if isinstance(linkable_spec, TimeDimensionSpec) and linkable_spec.time_granularity.is_custom_granularity:
+                linkable_spec_with_base_grain = linkable_spec.with_grain(
+                    ExpandedTimeGranularity.from_time_granularity(linkable_spec.time_granularity.base_granularity)
+                )
+                linkable_specs_set_with_base_granularities.add(linkable_spec_with_base_grain)
```
tlento, line 907: It's not clear to me why we're doing these granularity transformations here and below (and I think maybe in another place in the overall stack).

tlento (commented):
@theyostalservice and I are both a bit confused by the conversion operations we're doing on the TimeDimensionSpecs (see inlines for where the confusion arises).

It feels like the custom granularity specs should be checkable via simple equality in their own right, making the conversion to base grain unnecessary. Then when we encounter one we should be able to get the base grain we need to construct the join just at that moment in the DataflowToSQLQueryPlanConverter logic. 

If that isn't the case it seems like there's some construct we're missing someplace. It's possible I'm the one missing something, though, so please do let us know. If it's easier to explain in real time let's jump on a call.

courtneyholcomb:
@tlento @theyostalservice we should have a huddle about this one today, if possible!
Essentially, source nodes can't satisfy custom grains, because those require a join. I agree that this mid-DataflowPlan grain conversion logic is a bit flimsy, but I haven't been able to come up with a better solution thus far and would love to hear if you have an idea.

metricflow/dataflow/builder/node_evaluator.py
courtneyholcomb: Updated the property to be a function!
I would also love to not do the type refinement so maybe if we come up with a better solution today that can be removed.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -885,11 +895,25 @@ def _select_source_nodes_with_linkable_specs(
         """Find source nodes with requested linkable specs and no measures."""
         # Use a dictionary to dedupe for consistent ordering.
         selected_nodes: Dict[DataflowPlanNode, None] = {}
-        requested_linkable_specs_set = set(linkable_specs.as_tuple)
+
+        # Find the source node that will satisfy the base granularity. Custom granularities will be joined in later.
```
courtneyholcomb, line 896: Make this comment more clear

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -885,11 +895,25 @@ def _select_source_nodes_with_linkable_specs(
...
+        linkable_specs_set_with_base_granularities: Set[LinkableInstanceSpec] = set()
+        # TODO: Add support for no-metrics queries for custom grains without a join (i.e., select directly from time spine).
+        for linkable_spec in linkable_specs.as_tuple:
+            if isinstance(linkable_spec, TimeDimensionSpec) and linkable_spec.time_granularity.is_custom_granularity:
```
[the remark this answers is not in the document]
courtneyholcomb: use helper to get all custom time dims from linkable spec set

metricflow/dataflow/builder/node_evaluator.py
courtneyholcomb: Update this function to take a LinkableSpecSet instead of a Sequence

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -1020,10 +1044,12 @@ def _find_dataflow_recipe(
             metric_time_dimension_reference=self._metric_time_dimension_reference,
```
courtneyholcomb, line 1041: Update this function to do the conversion from custom to base grain, keep that change contained inside it
and remove related logic from elsewhere

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -1562,7 +1563,12 @@ def _build_aggregated_measure_from_measure_source_node(
             )
 
             specs_to_keep_after_join = InstanceSpecSet(measure_specs=(measure_spec,)).merge(
-                InstanceSpecSet.create_from_specs(required_linkable_specs.as_tuple),
+                InstanceSpecSet.create_from_specs(
+                    [
+                        spec.with_base_grain() if isinstance(spec, TimeDimensionSpec) else spec
```
tlento, line 1568: I'm assuming this will also get consolidated like the other sections, so it'll be based on the LinkableSpecSet operation rather than this inline type refinement.

tlento (approved):
OK! Per earlier discussion:

1. We'll move the custom grain -> base grain conversion logic into _find_dataflow_recipe, since that's where it's needed, and remove the inlined bits from the node evaluator and other locations
2. We'll use the LinkableSpecSet for managing the relevant type refinement instead of inline isinstance checks
3. The calls to do these can have comments explaining why they're happening as appropriate. I think it's most important in the doc block for _find_dataflow_recipe, since we'll be doing these operations at the entry point to that method.

As discussed @courtneyholcomb will make these changes at the top of the stack.

metricflow/dataflow/builder/dataflow_plan_builder.py
theyostalservice: This is resolved in #1427

PR #1410, opened 2024-09-18
SQL rendering tests for metric queries with custom granularity

courtneyholcomb:
SQL rendering tests for querying metrics with custom granularities. Does not include tests with filters (coming later). If you have suggestions for additional metric scenarios that should be tested, please let me know!


tlento (approved):
LGTM! One note - we need to explicitly test all the time spine/fill nulls scenarios, and maybe some offset metrics, but I assume the former is not implemented yet and the latter might be more valuable as part of a test suite when we begin supporting custom granularity windows. The base case (like with cumulative) might be nice to have from the beginning, though.

PR #1413, opened 2024-09-20
Query parsing for custom granularities

courtneyholcomb:
Support custom granularities in the query parser and the group by resolution DAG.
Reviewing by commit is recommended.


metricflow-semantics/metricflow_semantics/specs/patterns/entity_link_pattern.py
```diff
@@ -112,6 +111,19 @@ def _match_entity_links(self, candidate_specs: Sequence[LinkableInstanceSpec]) -
         shortest_entity_link_length = min(len(matching_spec.entity_links) for matching_spec in matching_specs)
         return tuple(spec for spec in matching_specs if len(spec.entity_links) == shortest_entity_link_length)
 
+    def _match_time_granularities(
```
tlento, line 114: My original idea was to avoid all of this by converting to the ExpandedTimeGranularity as early as possible, but it looks like you're keeping them as strings until the resolver does its thing, which also seems reasonable to me.

metricflow-semantics/metricflow_semantics/naming/dunder_scheme.py
```diff
@@ -77,8 +77,9 @@ def spec_pattern(self, input_str: str, semantic_manifest_lookup: SemanticManifes
             )
 
         # At this point, len(input_str_parts) >= 2
-        custom_granularity_names: List[str] = []  # TODO
-        valid_granularity_names = [granularity.value for granularity in TimeGranularity] + custom_granularity_names
+        valid_granularity_names = [granularity.value for granularity in TimeGranularity] + list(
```
tlento, line 80: Shall we make this a set?

metricflow-semantics/metricflow_semantics/specs/patterns/minimum_time_grain.py
```diff
@@ -48,20 +47,23 @@ class MinimumTimeGrainPattern(SpecPattern):
...
-                spec_key_to_specs[spec_key][0].with_grain(
-                    ExpandedTimeGranularity.from_time_granularity(min(time_grains))
-                )
+            sorted_time_grains = sorted(
+                time_grains,
+                # Sort by smallest to largest standard granularity, with custom granularities last (sorted by their
+                # base granularity) since we don't know how large they are.
+                key=lambda grain: (grain.is_custom_granularity, grain.base_granularity.to_int()),
             )
+            assert sorted_time_grains, "Each time dimension spec should have at least one grain."
```
tlento, line 65: Is that always true? We never checked this before.

tlento (approved):
Left some inlines. The one potential blocking question is about that new assertion. Will there always be a match? If there isn't, what used to happen? Would that fall through to some validation that returns a nicer error than the assertion will throw?

tlento:
(Note - if the assertion is correct and that represents an internal problem feel free to merge.)

metricflow-semantics/metricflow_semantics/specs/patterns/minimum_time_grain.py
courtneyholcomb: Yes! If you read the loop above on line 52, if we ever add a spec key to the dict, we also add its time granularity. And time dimension specs are required to have a non-null time granularity.
Also, if that weren't true, we would have previously run into a `KeyError` due to the indexing on this list.

PR #1414, opened 2024-09-20
SQL rendering tests for custom granularity time dimensions requiring join

courtneyholcomb:
SQL rendering tests for querying time dimensions that require a join with custom granularity.


tlento (approved):
:shipit:

PR #1415, opened 2024-09-20
Support custom granularity in linkable spec resolution

courtneyholcomb:
Further updates to linkable spec resolution needed to support custom granularities.

This also fixes a bug related to cumulative metric restrictions: Cumulative metrics CAN be queried with non-default granularity now, so this removes that restriction. This was not an issue in actual queries because that logic was only hit when querying metadata, but it could result in a bug when listing dimensions for a cumulative metric. Instead of restricting derived time granularity, restrict cumulative metrics from being used with date part (this matches existing query behavior). This change required adding a `DATE_PART` `LinkableElementProperty` and updating `LinkableDimensions` accordingly, so that's what you'll see in the snapshot changes.


metricflow-semantics/metricflow_semantics/model/linkable_element_property.py
```diff
@@ -27,6 +27,8 @@ class LinkableElementProperty(Enum):
     METRIC_TIME = "metric_time"
     # Refers to a metric, not a dimension.
     METRIC = "metric"
+    # An time dimension with a DatePart.
```
theyostalservice, line 30: nit: I think you meant "A" not "An" here, right?  (Or is this a convention I'm not aware of?)

tlento (approved):
LGTM! Please fix typo @theyostalservice noticed

Apropos of nothing but I need some amusement...

roy kent growling at some kid

PR #1416, opened 2024-09-20
Basic check query tests for custom granularity

courtneyholcomb:
Adds some check query tests for various custom granularity scenarios. I have some tasks up for other scenarios to test, but also open to more suggestions!

### DIFF
--- a/tests_metricflow/integration/test_cases/itest_granularity.yaml
+++ b/tests_metricflow/integration/test_cases/itest_granularity.yaml
@@ -310,3 +310,142 @@ integration_test:
       GROUP BY 2
     ) b
     ON a.listing__ds__day = b.listing__ds__day
+---
+integration_test:
+  name: simple_metric_with_custom_granularity
+  description: A simple metric queried with a custom granularity
+  model: SIMPLE_MODEL
+  metrics: [ "bookings"]
+  group_bys: ["booking__ds__martian_day"]
+  check_query: |
+    SELECT
+      b.martian_day AS booking__ds__martian_day
+      , SUM(1) AS bookings
+    FROM {{ source_schema }}.fct_bookings a
+    LEFT OUTER JOIN {{ source_schema }}.mf_time_spine b
+    ON {{ render_date_trunc("a.ds", TimeGranularity.DAY) }} = b.ds
+    GROUP BY b.martian_day
+---
+integration_test:
+  name: simple_metric_with_custom_granularity_and_join
+  description: A simple metric queried with a custom granularity
+  model: SIMPLE_MODEL
+  metrics: [ "bookings"]
+  group_bys: ["listing__ds__martian_day"]
+  check_query: |
+    SELECT
+      c.martian_day AS listing__ds__martian_day
+      , SUM(1) AS bookings
+    FROM {{ source_schema }}.fct_bookings a
+    LEFT OUTER JOIN {{ source_schema }}.dim_listings_latest l
+      ON a.listing_id = l.listing_id
+    LEFT OUTER JOIN {{ source_schema }}.mf_time_spine c
+      ON {{ render_date_trunc("l.created_at", TimeGranularity.DAY) }} = c.ds
+    GROUP BY c.martian_day
+---
+integration_test:
+  name: cumulative_metric_with_custom_granularity
+  description: A cumulative metric queried with a custom granularity
+  model: SIMPLE_MODEL
+  metrics: [ "trailing_2_months_revenue"]
+  group_bys: ["metric_time__martian_day"]
+  check_query: |
+    SELECT
+      metric_time__martian_day
+      , trailing_2_months_revenue
+    FROM (
+      SELECT
+        metric_time__martian_day
+        , AVG(txn_revenue) OVER (PARTITION BY metric_time__martian_day) AS trailing_2_months_revenue
+      FROM (
+        SELECT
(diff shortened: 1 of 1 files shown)


tests_metricflow/integration/test_cases/itest_granularity.yaml
```diff
@@ -310,3 +310,142 @@ integration_test:
...
+      ON subq_8.metric_time__martian_day = subq_14.metric_time__martian_day
+      GROUP BY COALESCE(subq_8.metric_time__martian_day, subq_14.metric_time__martian_day)
+    ) subq_15
+---
+integration_test:
+  name: multiple_metrics_with_custom_granularity
+  description: A derived metric queried with a custom granularity
+  model: SIMPLE_MODEL
+  metrics: [ "bookings", "listings"]
+  group_bys: ["metric_time__martian_day", "listing__ds__month"]
```
theyostalservice, line 419: This is exercising the new datepart logic in the linker for custom granularities, right?  (I'm just trying to confirm the things I'm learning.)

theyostalservice (approved):
Would it be appropriate(*) to also include one or more no-metric queries?

(*) - I don't even know if you are able to pass '[]' for the metrics in this test set up, so this might be a very silly question.

Otherwise, I don't know the edge cases well enough to be sure we aren't skipping any, but the broad strokes of what the PR stack has been about seem to be covered - using custom granularities, joining on them, and connecting nodes based on date parts (plus some of the edge cases like cumulative metrics and offset metrics.). Accepting.

courtneyholcomb:
> Would it be appropriate(*) to also include one or more no-metric queries?
> 
> (*) - I don't even know if you are able to pass '[]' for the metrics in this test set up, so this might be a very silly question.
> 
> Otherwise, I don't know the edge cases well enough to be sure we aren't skipping any, but the broad strokes of what the PR stack has been about seem to be covered - using custom granularities, joining on them, and connecting nodes based on date parts (plus some of the edge cases like cumulative metrics and offset metrics.). Accepting.

🤔 You're right! I'm pretty sure I added them at some point at then I must have lost them in the bad rebase I did on this PR 😅  I'll add some to the top of the stack.

tests_metricflow/integration/test_cases/itest_granularity.yaml
courtneyholcomb: > the new datepart logic in the linker

Maybe! What do you mean by this?

tests_metricflow/integration/test_cases/itest_granularity.yaml
```diff
@@ -310,3 +310,142 @@ integration_test:
...
+        LEFT OUTER JOIN {{ source_schema }}.mf_time_spine ts3
+          ON {{ render_date_trunc("b.ds", TimeGranularity.DAY) }} = ts3.ds
+        GROUP BY ts3.martian_day
+      ) subq_14
+      ON subq_8.metric_time__martian_day = subq_14.metric_time__martian_day
+      GROUP BY COALESCE(subq_8.metric_time__martian_day, subq_14.metric_time__martian_day)
+    ) subq_15
+---
+integration_test:
+  name: multiple_metrics_with_custom_granularity
```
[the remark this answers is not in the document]
courtneyholcomb: Nice!! Yeah, these end to end tests are great because they actually check that the SQL can actually execute, too, unlike most of our other tests.

tests_metricflow/integration/test_cases/itest_granularity.yaml
theyostalservice: hahaha uh oh sounds like I'm learning less than I thought.

I *think* this query exercises the changes in #1415 - we're able to groupBy DatePart on a derived metric with custom granularity because because of that PR, right?

courtneyholcomb: Ohh ok I get what you meant by `the new datepart logic in the linker` now!

> we're able to groupBy DatePart on a derived metric with custom granularity because because of that PR


Not quite. Date part and custom granularity can't be used together. That PR does 2 things:
- Bug fix: updates the `LinkableSpecResolver` to allow non-default granularity for cumulative metrics. By non-default I don't mean custom. The _default_ granularity for a dimension is the one it's defined at in the YAML definition. For example, if a metric is defined with `DAY` granularity, the default is `DAY` and all other standard granularities (`MONTH`, `YEAR`, etc.) _and_ custom granularities are considered non-default. Cumulative metrics can be queried with all granularities, which is a change we made this year, and the `LinkableSpecResolver` was out of date for that logic. Cumulative metrics still cannot be queried with date part, though. This change was all in the first commit.
- Adds custom granularities to the `LinkableSpecResolver` for all metrics. This change is in all the later commits.
There are a lot of different metric types and group by options to grasp all at once here 😅 So let me know if this makes sense or if you have questions about any of it!

As for this test - it is testing that we can use a custom granularity with multiple simple metrics, and that does rely on the logic from that PR!

Also, I realized just now that I didn't update the descriptions for these tests properly (good old copy/paste) and that probably contributed to the confusion. I'll update those before merging.

theyostalservice: ok, i think it does make sense.  I may need a reminder or two about some of these details in the future, but I'm definitely putting things together a bit more as I go. :)

Thank you, Courtney!

PR #1419, opened 2024-09-23
Add `LazyFormat` to reduce logging overhead on performance

plypaul:
For larger semantic models / queries, the overhead of formatting objects for logging / writing log lines can be significant. This adds a `LazyFormat` class can be used with the existing logger so that log messages at the debug level don't get formatted nor written when the configured log level would exclude it. Using this, a quick way to migrate the code base would be a replacement like the following:

```
    Example:
        logger.debug(f"Result is: {expensive_function()}")
        ->
        logger.debug(LazyFormat(lambda: f"Result is: {expensive_function()}")
```


marcodamore (approved):
overall lgtm!

PR #1420, opened 2024-09-24
Change log calls to use `LazyFormat` and `DEBUG` level

plypaul:
As per https://github.com/dbt-labs/metricflow/pull/1419, this PR updates log calls to lazily evaluate the arguments, and changes most of them to the `DEBUG` level. This way, performance in production will be improved if the log level is set to `INFO` or higher.

### DIFF
--- a/dbt-metricflow/dbt_metricflow/cli/dbt_connectors/adapter_backed_client.py
+++ b/dbt-metricflow/dbt_metricflow/cli/dbt_connectors/adapter_backed_client.py
@@ -9,6 +9,7 @@ from dbt_common.exceptions.base import DbtDatabaseError
 from dbt_semantic_interfaces.enum_extension import assert_values_exhausted
 from metricflow_semantics.errors.error_classes import SqlBindParametersNotSupportedError
 from metricflow_semantics.mf_logging.formatting import indent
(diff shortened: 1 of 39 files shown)


dbt-metricflow/dbt_metricflow/cli/dbt_connectors/adapter_backed_client.py
```diff
@@ -177,14 +183,15 @@ def execute(
             )
         start = time.time()
         request_id = SqlRequestId(f"mf_rid__{random_id()}")
-        logger.info(AdapterBackedSqlClient._format_run_query_log_message(stmt, sql_bind_parameters))
+        logger.info(LazyFormat("Running query", statement=stmt, param_dict=sql_bind_parameters.param_dict))
```
courtneyholcomb, line 186: Can we differentiate the log message for `execute` from `query`? Makes it harder to figure out where a log is coming from if the message is logged in multiple places!

courtneyholcomb (approved):

dbt-metricflow/dbt_metricflow/cli/dbt_connectors/adapter_backed_client.py
plypaul: Updated.

PR #1425, opened 2024-09-24
Bug fix: custom granularity join column gets pruned

courtneyholcomb:
We have an issue with the `SqlColumnPruner` where it might prune a column that's used in the join condition, if it's not used anywhere else. There isn't a simple way to fix that issue in the optimizer itself, since the join condition can contain any `SqlExpressionNode`, and this appears to only be relevant for the new custom granularity node. Thus, I've fixed this in the custom granularity node instead of the optimizer by rendering the column's expression in the join condition instead of its alias.

### DIFF
--- a/metricflow/plan_conversion/dataflow_to_sql.py
+++ b/metricflow/plan_conversion/dataflow_to_sql.py
@@ -1441,23 +1441,32 @@ class DataflowToSqlQueryPlanConverter(DataflowPlanNodeVisitor[SqlDataSet]):
             if instance.spec == node.time_dimension_spec.with_base_grain():
                 parent_time_dimension_instance = instance
                 break
+        parent_column: Optional[SqlSelectColumn] = None
         assert parent_time_dimension_instance, (
             "JoinToCustomGranularityNode's expected time_dimension_spec not found in parent dataset instances. "
-            "This indicates internal misconfiguration."
+            f"This indicates internal misconfiguration. Expected: {node.time_dimension_spec.with_base_grain}; "
+            f"Got: {[instance.spec for instance in parent_data_set.instance_set.time_dimension_instances]}"
+        )
+        for select_column in parent_data_set.checked_sql_select_node.select_columns:
+            if select_column.column_alias == parent_time_dimension_instance.associated_column.column_name:
+                parent_column = select_column
+                break
+        assert parent_column, (
+            "JoinToCustomGranularityNode's expected time_dimension_spec not found in parent columns. "
+            f"This indicates internal misconfiguration. Expected: "
+            f"{parent_time_dimension_instance.associated_column.column_name}; Got: "
+            f"{[column.column_alias for column in parent_data_set.checked_sql_select_node.select_columns]}"
         )
 
         # Build join expression.
         time_spine_alias = self._next_unique_table_alias()
         custom_granularity_name = node.time_dimension_spec.time_granularity.name
         time_spine_source = self._get_time_spine_for_custom_granularity(custom_granularity_name)
-        left_expr_for_join: SqlExpressionNode = SqlColumnReferenceExpression.from_table_and_column_names(
-            table_alias=parent_alias, column_name=parent_time_dimension_instance.associated_column.column_name
-        )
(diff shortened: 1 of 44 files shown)


metricflow-semantics/metricflow_semantics/naming/dunder_scheme.py
```diff
@@ -77,8 +77,9 @@ def spec_pattern(self, input_str: str, semantic_manifest_lookup: SemanticManifes
             )
 
         # At this point, len(input_str_parts) >= 2
-        custom_granularity_names: List[str] = []  # TODO
-        valid_granularity_names = [granularity.value for granularity in TimeGranularity] + custom_granularity_names
+        valid_granularity_names = [granularity.value for granularity in TimeGranularity] + list(
+            semantic_manifest_lookup._custom_granularities.keys()
```
WilliamDee, line 81: nit: prefer if we could make `_custom_granularities` method not protected or add another accessor that is not a protected method

metricflow-semantics/metricflow_semantics/specs/patterns/minimum_time_grain.py
```diff
@@ -48,20 +47,23 @@ class MinimumTimeGrainPattern(SpecPattern):
...
         for spec_key, time_grains in spec_key_to_grains.items():
-            matched_time_dimension_specs.append(
-                spec_key_to_specs[spec_key][0].with_grain(
-                    ExpandedTimeGranularity.from_time_granularity(min(time_grains))
-                )
+            sorted_time_grains = sorted(
+                time_grains,
+                # Sort by smallest to largest standard granularity, with custom granularities last (sorted by their
+                # base granularity) since we don't know how large they are.
+                key=lambda grain: (grain.is_custom_granularity, grain.base_granularity.to_int()),
```
WilliamDee, line 63: Idea: Would it be nicer if we added the comparison methods to `ExpandedTimeGranularity` so you can just sort it as normal
```python
def __lt__(self, other):
        return self._cmp(other) < 0

    def __le__(self, other):
        return self._cmp(other) <= 0

    def __eq__(self, other):
        return self._cmp(other) == 0

    def __ne__(self, other):
        return self._cmp(other) != 0

    def __ge__(self, other):
        return self._cmp(other) >= 0

    def __gt__(self, other):
        return self._cmp(other) > 0
```
might be more forward thinking tho since i don't think it's used in multiple places

WilliamDee (approved):
Did a pass, this looks great to me 🥳 would love for another reviewer to just do a pass as well tho

metricflow/dataflow/nodes/join_to_custom_granularity.py
```diff
@@ -14,10 +14,21 @@
...
+    """Join parent dataset to time spine dataset to convert time dimension to a custom granularity.
+
+    Args:
+        time_dimension_spec: The time dimension spec with a custom granularity that will be satisfied by this node.
+    """
 
     time_dimension_spec: TimeDimensionSpec
 
+    def __post_init__(self) -> None:  # noqa: D105
+        super().__post_init__()
```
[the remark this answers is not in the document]
courtneyholcomb: Yes! There is a logging function that relies on the `__post_init__` logic from one of the parent classes, so I was seeing logging errors. I checked and this is the only `DagNode` subclass that has a `__post_init__` and didn't also have this call.

metricflow-semantics/metricflow_semantics/specs/patterns/minimum_time_grain.py
courtneyholcomb: That's a good idea 🙂‍↕️

courtneyholcomb: Actually, I changed my mind. Only because we can't actually compare the size of custom granularities - we have no idea which ones are bigger than the others. So we can sort by base granularity but then after that it would be random. Here it doesn't matter in practice, so this sorting method is fine, but we probably shouldn't add a partly-random sorting method that might be used elsewhere without knowing the context.

PR #1427, opened 2024-09-24
Improved source node logic for custom granularities

courtneyholcomb:
Based on feedback in https://github.com/dbt-labs/metricflow/pull/1409, this PR makes some updates to the dataflow plan builder. These are only refactoring changes, there should be no behavioral changes. Changes include renaming some things and simplifying the code related to converting custom granularities to base granularities in source node resolution. Reviewing by commit is recommended - full details in commit descriptions.

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -907,18 +907,8 @@ class DataflowPlanBuilder:
         # Use a dictionary to dedupe for consistent ordering.
         selected_nodes: Dict[DataflowPlanNode, None] = {}
 
-        # Find the source node that will satisfy the base granularity. Custom granularities will be joined in later.
-        linkable_specs_set_with_base_granularities: Set[LinkableInstanceSpec] = set()
         # TODO: Add support for no-metrics queries for custom grains without a join (i.e., select directly from time spine).
-        for linkable_spec in linkable_specs.as_tuple:
-            if isinstance(linkable_spec, TimeDimensionSpec) and linkable_spec.time_granularity.is_custom_granularity:
-                linkable_spec_with_base_grain = linkable_spec.with_grain(
-                    ExpandedTimeGranularity.from_time_granularity(linkable_spec.time_granularity.base_granularity)
-                )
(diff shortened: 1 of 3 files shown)


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -1614,10 +1610,7 @@ def _build_aggregated_measure_from_measure_source_node(
 
             specs_to_keep_after_join = InstanceSpecSet(measure_specs=(measure_spec,)).merge(
                 InstanceSpecSet.create_from_specs(
-                    [
-                        spec.with_base_grain() if isinstance(spec, TimeDimensionSpec) else spec
-                        for spec in required_linkable_specs.as_tuple
-                    ]
+                    required_linkable_specs.replace_custom_granularity_with_base_granularity().as_tuple
```
theyostalservice, line 1613: Ok, this class is fairly complex, so I'm going to give up and just ask - I thought that by moving one of the `replace_custom_granularity_with_base_granularity` calls to `_find_source_node_recipe`, we obviated the need to do this again anywhere else.  Was that impression wrong?

theyostalservice (approved):

theyostalservice:
Please take a look at https://github.com/dbt-labs/metricflow/pull/1427#discussion_r1776013843 before committing.

metricflow/dataflow/builder/dataflow_plan_builder.py
courtneyholcomb: You're right, that was the plan! I hadn't realized this use case was outside of that function.
I took a stab at seeing if we could add this spec set as a property to the `SourceNodeRecipe` to avoid calculating it again here, which would clean things up a bit. It works everywhere except for conversion metrics, which have a separate `SourceNodeRecipe` that they build. I tried to mess around with that code to make it work but I'm not familiar enough with conversion metrics to make any significant changes without consulting someone so I scrapped it.
Basically the blocker is that we haven't consolidated this code enough to make this simple. I would love to clean up the whole `DataflowPlanBuilder` to remove redundancies, and I we would certainly be able to clean this piece up then, but I think that's too big a task to block this feature.

PR #1431, opened 2024-09-27
Use list for where filters

courtneyholcomb:
## Context

There's a bug where time filters is not being re-applied after joining to timespine. Currently, we take all filter strings in a list of filters and mash them together. We need to keep them separate so that we can differentiate metric_time filters.

Resolves SL-2955


WilliamDee (approved):
image

PR #1435, opened 2024-10-01
Cache `ValidLinkableSpecResolver._get_joined_elements()`

plypaul:
This PR caches retrieval of the joined elements of a measure as it can be called repeatedly when resolving the group-by items for a derived metric (or a simple metric that is common to many saved queries).


courtneyholcomb (approved):
LGTM!

PR #1436, opened 2024-10-01
Cache methods in `ExpandedTimeGranularity`

plypaul:
The methods in `ExpandedTimeGranularity` benefits from caching as they are called in many cases where `TimeDimensionSpecs` are created. `*Spec` objects are one of the most frequently created / used objects.


courtneyholcomb (approved):
🚀

PR #1437, opened 2024-10-01
Cache `SemanticModelLookup.get_defined_time_granularity()`

plypaul:
The defined time granularity of a time dimension is frequently called during query resolution to figure out the time granularity for an ambiguous group-by item.


courtneyholcomb (approved):
👍 👍

PR #1438, opened 2024-10-02
Cache additional methods in `MetricLookup` / `SemanticModelLookup`

plypaul:
Similar to the previous PRs, this PR caches a few additional methods. I did a little bit of memory profiling using a large manifest, and total memory allocation did not change significantly. I believe this is because these additional caches reference objects that are already stored in `ValidLinkableSpecResolver` or the cached objects have low overhead.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/model/semantics/semantic_model_lookup.py
+++ b/metricflow-semantics/metricflow_semantics/model/semantics/semantic_model_lookup.py
@@ -67,6 +67,9 @@ class SemanticModelLookup:
         # Cache for defined time granularity.
         self._time_dimension_to_defined_time_granularity: Dict[TimeDimensionReference, TimeGranularity] = {}
 
+        # Cache for agg. time dimension for measure.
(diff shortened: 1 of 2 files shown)


metricflow-semantics/metricflow_semantics/model/semantics/semantic_model_lookup.py
```diff
@@ -67,6 +67,9 @@ def __init__(self, model: SemanticManifest, custom_granularities: Dict[str, Expa
         # Cache for defined time granularity.
         self._time_dimension_to_defined_time_granularity: Dict[TimeDimensionReference, TimeGranularity] = {}
 
+        # Cache for agg. time dimension for measure.
```
courtneyholcomb, line 70: did you mean to include a period after agg?

courtneyholcomb (approved):
LGTM!

metricflow-semantics/metricflow_semantics/model/semantics/semantic_model_lookup.py
plypaul: Yes, since it is an abbreviation.

PR #1441, opened 2024-10-02
Consolidate property filters into a class

plypaul:
The signature of a few functions in `ValidLinkableSpecResolver` take in a few arguments that describe how to filter the returned items based on properties. To simplify the signatures and enable improved caching in a later PR, this PR groups them into the `LinkableElementFilter` class.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/model/semantics/element_filter.py
+++ b/metricflow-semantics/metricflow_semantics/model/semantics/element_filter.py
@@ -1,0 +1,31 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import FrozenSet
+
+from typing_extensions import Self, override
+
+from metricflow_semantics.collection_helpers.merger import Mergeable
+from metricflow_semantics.model.linkable_element_property import LinkableElementProperty
+
+
+@dataclass(frozen=True)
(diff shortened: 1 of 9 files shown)


metricflow-semantics/metricflow_semantics/model/semantics/element_filter.py
```diff
@@ -0,0 +1,31 @@
...
+from typing import FrozenSet
+
+from typing_extensions import Self, override
+
+from metricflow_semantics.collection_helpers.merger import Mergeable
+from metricflow_semantics.model.linkable_element_property import LinkableElementProperty
+
+
+@dataclass(frozen=True)
+class LinkableElementFilter(Mergeable):
```
courtneyholcomb, line 13: Should we call this `LinkableElementPropertyFilter`? On first glance I would have assumed this is referring to a filter on a group by spec or something similar to that.

courtneyholcomb (approved):

metricflow-semantics/metricflow_semantics/model/semantics/element_filter.py
plypaul: This actually gets updated in a later PR to filter by element name, so I kept this more general.

PR #1443, opened 2024-10-03
Improve cache for `MetricLookup.linkable_elements_for_measure()`

plypaul:
This PR changes the cache for `MetricLookup.linkable_elements_for_measure()` to allow caching all measures. Memory use remained reasonable and because it's bound to `MetricLookup`, it will be GCed when a given manifest is no longer used.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
+++ b/metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import functools
 import logging
 import time
-from typing import Dict, Optional, Sequence, Set
+from typing import Dict, Optional, Sequence, Set, Tuple
 
 from dbt_semantic_interfaces.enum_extension import assert_values_exhausted
 from dbt_semantic_interfaces.protocols.metric import Metric, MetricInputMeasure, MetricType
@@ -13,6 +13,7 @@ from dbt_semantic_interfaces.type_enums.time_granularity import TimeGranularity
 
 from metricflow_semantics.errors.error_classes import DuplicateMetricError, MetricNotFoundError, NonExistentMeasureError
 from metricflow_semantics.mf_logging.lazy_formattable import LazyFormat
+from metricflow_semantics.model.linkable_element_property import LinkableElementProperty
 from metricflow_semantics.model.semantics.element_filter import LinkableElementFilter
 from metricflow_semantics.model.semantics.linkable_element_set import LinkableElementSet
 from metricflow_semantics.model.semantics.linkable_spec_resolver import (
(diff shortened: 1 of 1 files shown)


metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
```diff
@@ -62,25 +61,34 @@ def __init__(
...
-        linkable_element_set = self._linkable_spec_resolver.get_linkable_element_set_for_measure(
-            measure_reference=measure_reference,
-            element_filter=element_filter,
-        )
-        logger.debug(
-            LazyFormat(
-                lambda: f"Getting valid linkable elements for measure '{measure_reference.element_name}' took: {time.time() - start_time:.2f}s"
-            )
-        )
+        # Don't cache the result when group-by-metrics are selected as there can be many of them and may significantly
```
courtneyholcomb, line 75: If we remove the cache for group by metrics, doesn't that mean this will be super slow when we try to fetch group by metric options?

metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
```diff
@@ -62,25 +61,34 @@ def __init__(
...
         measure_reference: MeasureReference,
         element_filter: LinkableElementFilter = LinkableElementFilter(),
     ) -> LinkableElementSet:
         """Return the set of linkable elements reachable from a given measure."""
-        start_time = time.time()
-        linkable_element_set = self._linkable_spec_resolver.get_linkable_element_set_for_measure(
-            measure_reference=measure_reference,
-            element_filter=element_filter,
-        )
-        logger.debug(
```
courtneyholcomb, line 77: Curious why you removed this log? Seems helpful to see if there was improvement on performance for this method. Also seems like it would be a good one to make `INFO` instead of `DEBUG` so that we can see how it performs for customers with larger manifests.

metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
plypaul: I had some concerns about memory usage, so caching of group-by-metrics is handled later through a separate cache that's LRU and avoids GC issues.

metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
plypaul: This method can potentially be called a lot during query parsing, so it's easier to assess the runtime through DD tooling. e.g. a query with ~100 metrics, ~100 group-by items (including ones in filters) leads to a lot of log lines. Could leave a debug line in there though.

metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
courtneyholcomb: Got it 👍

courtneyholcomb (approved):

metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
plypaul: Added that debug line.

PR #1448, opened 2024-10-07
Cache building operations in `DataflowPlanBuilder`

plypaul:
This PR adds a few LRU caches to handle building operations within the `DataflowPlanBuilder`. Since the same metric may be used multiple times in a derived metric (or between queries), there can be significant performance improvements. Please view by commit as there were signature / type changes to making the changes cleaner.

### DIFF
--- a/metricflow/dataflow/builder/builder_cache.py
+++ b/metricflow/dataflow/builder/builder_cache.py
@@ -1,0 +1,78 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Optional
+
+from metricflow_semantics.collection_helpers.lru_cache import LruCache
+from metricflow_semantics.specs.linkable_spec_set import LinkableSpecSet
+from metricflow_semantics.specs.metric_spec import MetricSpec
+from metricflow_semantics.specs.where_filter.where_filter_transform import WhereSpecFactory
+
+from metricflow.dataflow.builder.measure_spec_properties import MeasureSpecProperties
+from metricflow.dataflow.builder.source_node_recipe import SourceNodeRecipe
+from metricflow.dataflow.dataflow_plan import DataflowPlanNode
+from metricflow.plan_conversion.node_processor import PredicatePushdownState
+
+
+@dataclass(frozen=True)
+class FindSourceNodeRecipeParameterSet:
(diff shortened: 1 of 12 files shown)


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -100,6 +106,10 @@
 
 logger = logging.getLogger(__name__)
 
+_MEASURE_RECIPE_CACHE_SIZE = 1000
```
courtneyholcomb, line 109: Curious if we've put much thought into what the size should be? I guess we'll just experiment and tune it accordingly?

metricflow/dataflow/builder/builder_cache.py
```diff
@@ -0,0 +1,78 @@
...
+    for_group_by_source_node: bool
+
+
+class DataflowPlanBuilderCache:
+    """Cache for internal methods in `DataflowPlanBuilder`."""
+
+    def __init__(  # noqa: D107
+        self, find_source_node_recipe_cache_size: int = 1000, build_any_metric_output_node_cache_size: int = 1000
+    ) -> None:
+        self._find_source_node_recipe_cache = LruCache[FindSourceNodeRecipeParameterSet, FindSourceNodeRecipeResult](
```
courtneyholcomb, line 50: It would be pretty cool if we could use the new `LruCache` while still using decorator syntax to maintain readability of the methods that are cached. But I guess it's also debatable if that's more readable (e.g., it obscures the cache code if you end up needing to debug something there). And probably not worth the effort it would take to make it work anyway...

courtneyholcomb (approved):
LGTM!

metricflow/dataflow/builder/builder_cache.py
plypaul: Yeah, the obscuring of the dependencies with a decorator would make it harder to follow.

metricflow/dataflow/builder/dataflow_plan_builder.py
plypaul: These are based on rough sizes of semantic manifests and performance tests on the larger ones we've seen. We do need to encapsulate all these caches into a single object and add instrumentation.

PR #1449, opened 2024-10-07
Update use cases of `fuctools.lru_cache` to use `LruCache`

plypaul:
To address potential memory leaks, PR updates use cases of `functools.lru_cache` to use `LruCache`.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
+++ b/metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import functools
 import logging
 import time
 from typing import Dict, Optional, Sequence, Set, Tuple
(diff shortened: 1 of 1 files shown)


metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
```diff
@@ -106,21 +124,33 @@ def linkable_elements_for_measure(
...
         self, metric_references: Sequence[MetricReference], element_set_filter: LinkableElementFilter
     ) -> LinkableElementSet:
         """Retrieve the matching set of linkable elements common to all metrics requested (intersection)."""
-        return self._linkable_spec_resolver.get_linkable_elements_for_metrics(
+        cache_key = (metric_references, element_set_filter)
+        result = self._linkable_elements_for_metrics_cache.get(cache_key)
+        if result is not None:
+            return result
+
+        result = self._linkable_spec_resolver.get_linkable_elements_for_metrics(
```
courtneyholcomb, line 149: There is an issue in the linkable spec resolver related to this, and I want to make sure we're not encountering it here. `get_linkable_elements_for_metrics()` returns a different result than `get_linkable_elements_for_measures()`. But the _measures_ method is the one that is actually used in queries to resolve specs for metrics. the metrics method is only used in metadata queries (i.e., list dimensions). There are minor differences between the two results, the main one I can think of being the date part restriction for cumulative metrics. I haven't had time to dig into that bug but we should make sure we're using the correct logic for this call.
It probably doesn't impact this logic, but would be good to double check.

courtneyholcomb (approved):

metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
plypaul: Yeah, this change should not be affected by that issue.

PR #1451, opened 2024-10-09
Add validation for missing `metric_time` when querying to/through SCD

serramatutu:
## Summary
This PR changes `MetricQueryValidationRule` to raise `ScdRequiresMetricTimeIssue` in case the user submits a query that uses an SCD but does not group by time.

## Implementation
To make this happen, I:
- refactored how `MetricQueryValidationRule` works to return a new `QueryAnalysisResult` and avoid creating a bunch of separate methods that return booleans
- changed `MetricQueryValidationRule` to make it get the specs for all the joinable SCDs for the given metric, match the group by input against it and raise an issue if there's no `metric_time` in the query
- created a new `LinkableElementProperty` called `SCD_HOP`, which indicates that you need to join to an SCD somewhere along the path to join to that element
- changed the `ValidLinkableSpecResolver` to add `SCD_HOP` to all the linkable elements which go to/through SCDs
- added tests for joins to an SCD and through an SCD


metricflow-semantics/metricflow_semantics/model/semantics/linkable_spec_resolver.py
```diff
@@ -188,6 +188,10 @@ def __init__(
 
         logger.debug(LazyFormat(lambda: f"Building valid group-by-item indexes took: {time.time() - start_time:.2f}s"))
 
+    def _semantic_model_is_scd(self, semantic_model: SemanticModel) -> bool:
```
serramatutu, line 191: Might wanna cache this in a dict to avoid iterating dimensions over and over and over again.

metricflow-semantics/metricflow_semantics/model/semantics/linkable_spec_resolver.py
```diff
@@ -711,12 +731,21 @@ def create_linkable_element_set_from_join_path(
         join_path: SemanticModelJoinPath,
     ) -> LinkableElementSet:
         """Given the current path, generate the respective linkable elements from the last semantic model in the path."""
+        semantic_model = self._semantic_model_lookup.get_by_reference(join_path.last_semantic_model_reference)
+        assert semantic_model
```
courtneyholcomb, line 735: Can you add an error message for this assertion? Something that tells us that something is internally misconfigured, not user error. We tend to hit these errors more than we expect, so trying to get in the habit of always including a message.

metricflow-semantics/metricflow_semantics/model/semantics/linkable_spec_resolver.py
```diff
@@ -711,12 +731,21 @@ def create_linkable_element_set_from_join_path(
...
         properties = frozenset({LinkableElementProperty.JOINED})
         if len(join_path.path_elements) > 1:
             properties = properties.union({LinkableElementProperty.MULTI_HOP})
 
-        semantic_model = self._semantic_model_lookup.get_by_reference(join_path.last_semantic_model_reference)
-        assert semantic_model
+        # If any of the semantic models in the join path is an SCD, add SCD_HOP
+        for reference_to_derived_model in join_path.derived_from_semantic_models:
+            derived_model = self._semantic_model_lookup.get_by_reference(reference_to_derived_model)
+            assert derived_model
```
courtneyholcomb, line 744: same here re: error message

metricflow-semantics/metricflow_semantics/query/validation_rules/metric_time_requirements.py
```diff
@@ -18,18 +23,33 @@
...
+)
 from metricflow_semantics.query.validation_rules.base_validation_rule import PostResolutionQueryValidationRule
 from metricflow_semantics.specs.time_dimension_spec import TimeDimensionSpec
 
 
+@dataclass(frozen=True)
+class QueryItemsAnalysis:
+    """Contains data about which items a query contains."""
+
+    scds: Sequence[str]
```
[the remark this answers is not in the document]
courtneyholcomb: Also - can we store the `LinkableInstanceSpecs` (`TimeDimensionSpec`/`DimensionSpec`/`EntitySpec`)  instead of the qualified names here? It seems fragile to just pass around the string names, especially since it's not always clear that they are qualified names.
Then you can pass the specs into `SCDRequiresMetricTimeIssue` and parse out the qualified names within that class.

courtneyholcomb (commented):
Looking good!! I left a bunch of small notes inline, and one bigger question.

metricflow-semantics/metricflow_semantics/query/validation_rules/metric_time_requirements.py
```diff
@@ -18,18 +23,33 @@
 from metricflow_semantics.query.issues.parsing.offset_metric_requires_metric_time import (
     OffsetMetricRequiresMetricTimeIssue,
 )
-from metricflow_semantics.query.resolver_inputs.query_resolver_inputs import ResolverInputForQuery
+from metricflow_semantics.query.issues.parsing.scd_requires_metric_time import (
```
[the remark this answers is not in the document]
serramatutu: Shouldn't black or flake8 keep this consistent tho? It passed pre-commit

metricflow-semantics/metricflow_semantics/query/validation_rules/metric_time_requirements.py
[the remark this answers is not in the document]
serramatutu: Will do!

metricflow-semantics/metricflow_semantics/query/validation_rules/metric_time_requirements.py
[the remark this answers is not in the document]
courtneyholcomb: If you add the comma then it will allow this! But if you remove the comma then it would collapse to one line

metricflow-semantics/tests_metricflow_semantics/snapshots/test_linkable_spec_resolver.py/str/test_create_linkable_element_set_from_join_path_multi_hop__result0.txt
```diff
@@ -1,48 +1,48 @@
-Model Join-Path                                  Entity Links          Name             Time Granularity    Date Part    Properties
```
serramatutu, line 1: When looping through models in the join path, I added a new assertion that the model exists in `SemanticModelLookup`. This assertion caused the test to break. This is expected since the model `bookings` doesn't exist. The file names it `bookings_source`, so I changed the test comparison to expect `bookings_source` instead.

metricflow-semantics/tests_metricflow_semantics/model/semantics/test_linkable_spec_resolver.py
```diff
@@ -165,7 +165,7 @@ def test_create_linkable_element_set_from_join_path_multi_hop(  # noqa: D103
                 left_semantic_model_reference=SemanticModelReference("views_source"),
                 path_elements=(
                     SemanticModelJoinPathElement(
-                        semantic_model_reference=SemanticModelReference("bookings"),
+                        semantic_model_reference=SemanticModelReference("bookings_source"),
```
serramatutu, line 168: I had to change this to `bookings_source` since `bookings` doesn't exist in `simple_manifest`, which was causing an assertion to break.

metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
```diff
@@ -221,3 +222,12 @@ def get_min_queryable_time_granularity(self, metric_reference: MetricReference)
                     minimum_queryable_granularity = defined_time_granularity
 
         return minimum_queryable_granularity
+
+    def get_joinable_scd_specs_for_metric(self, metric_reference: MetricReference) -> Sequence[LinkableInstanceSpec]:
```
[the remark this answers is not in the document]
serramatutu: We discussed this in DMs but I wanna document it here: the reason I'm using the `LinkableElements` is because a dimension being an `SCD_HOP` or not depends on the join path. 

Let's say we have 4 models: `bookings`, `agents`, `users` and `agencies`. Each `booking` is booked by an `user` with the help of an `agent` which is part of an `agency`. Each `user`'s bookings are always handled by the same `agency`, so there is a direct link between `user` and `agency` as well. Now suppose `agents` is an SCD.

- If we query for `bookings_revenue` grouped by `user__agency__name`, we don't need `metric_time` since neither `user` nor `agency` are SCDs. In this case, `agency__name` DOES NOT require `metric_time`.

- If we now query for `bookings_revenue` grouped by `agent__agency__name`, we do need `metric_time` since we join to `agency` via `agent`, and `agent` is an SCD. In this context, `agency__name` DOES require `metric_time`.

Since the same dimension might or might not be treated as an SCD depending on the join path, we need to analyze this in the path resolver based on the source metric, so it can't go in `SemanticModelLookup`.

serramatutu:
@plypaul I'm asking for your review since you're the one doing most of the profiling and I refactored the `metric_time_requirements.py` file. Is there a reason you only cached one of the methods and not the other? Do you think if I cache the whole of `_get_query_analysis` will that grow too much in memory or affect performance somehow?

I'll wait for this answer before I rebase from main :)

serramatutu:
Since the last review, I:
- Changed the 2 tests to use `assert_str_snapshot_equal` with the error message. I added 2 `.txt` files with the expected validation errors to compare against.
- Rebased the `fixup!` commits.
- Rebased against `main` to get @plypaul's caching changes.
- Added a new changelog entry with `changie new`.

metricflow-semantics/tests_metricflow_semantics/query/test_query_parser.py
```diff
@@ -459,6 +466,52 @@ def test_cumulative_metric_agg_time_dimension_name_validation(
...
+    )
+
+
+def test_join_through_scd_no_time_dimension_validation(
+    request: FixtureRequest,
+    mf_test_configuration: MetricFlowTestConfiguration,
+    scd_query_parser: MetricFlowQueryParser,
+) -> None:
+    """Test that queries that join through SCDs semantic models fail if no time dimensions are selected."""
+    with pytest.raises(InvalidQueryException) as exc_info:
```
serramatutu, line 497: @courtneyholcomb I change these tests from using the `match` regex to use `assert_str_snapshot_equal`, and I created the 2 new `.txt` files with the full error.

metricflow-semantics/tests_metricflow_semantics/snapshots/test_query_parser.py/str/test_join_through_scd_no_time_dimension_validation__error.txt
```diff
@@ -0,0 +1,17 @@
+Got errors while resolving the query.
+
+Error #1:
+  Message:
+
+    Your query contains the group bys which are SCDs or contain SCDs in the join path: [listing__user__home_state_latest].
```
courtneyholcomb, line 6: Nit on the message here - can we say (changes in code syntax):
Your query contains the `following` group bys`,` which are SCDs or contain SCDs in the join path: [`'`listing__user__home_state_latest`'`].

metricflow-semantics/metricflow_semantics/query/validation_rules/metric_time_requirements.py
```diff
@@ -46,77 +66,95 @@ def __init__(  # noqa: D107
...
-
-        return False
-
-    def _query_includes_metric_time_or_agg_time_dimension(self, metric_reference: MetricReference) -> bool:
-        return self._group_by_items_include_metric_time or self._group_by_items_include_agg_time_dimension(
-            query_resolver_input=self._resolver_input_for_query, metric_reference=metric_reference
-        )
-
-    def _group_by_items_include_agg_time_dimension(
+    @lru_cache
```
courtneyholcomb, line 69: Just checking - I think Paul said it was fine to use this `lru_cache` becasue the number of items is small enough?

courtneyholcomb (approved):
Left 2 small notes inline, but this looks great!! Awesome work!

PR #1452, opened 2024-10-09
Don't add measure-level filters to post-agg filters

WilliamDee:
## Context

We should respect filter hierarchy and have,
- measure-level filters should be applied only pre-aggregation
- metric-level filters and query-level filters should be applied pre AND post-aggregation

Meaning, measure-level filters should not be applied post-aggregation.

### Refactor
Changed `filter_specs` from being a tuple of filter specs all mashed together to a `WhereFilterSpecSet` where it stores which levels each of the filters were defined at. This was changed inside `MetricSpec` and `MetricInputMeasureSpec`.

Resolves SL-2956


tests_metricflow/snapshots/test_time_spine_join_rendering.py/SqlQueryPlan/DuckDB/test_join_to_time_spine_with_input_measure_constraint__plan0_optimized.sql
```diff
@@ -0,0 +1,39 @@
...
+    ) subq_11
+    WHERE (metric_time__day > '2020-01-01') AND (booking__is_instant)
+    GROUP BY
+      metric_time__day
+      , booking__is_instant
+  ) subq_13
+  ON
+    subq_15.ds = subq_13.metric_time__day
+) subq_16
+WHERE metric_time__day > '2020-01-01'
```
WilliamDee, line 39: `booking__is_instant` measure-level filter is not applied here

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -1339,18 +1348,30 @@ def _build_filter_specs_for_input_measure(
         metric: Metric,
         input_measure: MetricInputMeasure,
         descendent_filter_specs: Sequence[WhereFilterSpec],
-    ) -> Tuple[WhereFilterSpec, ...]:
+    ) -> MetricInputMeasureSpecFilterSet:
```
courtneyholcomb, line 1351: I would love if we could rename some stuff in this area to make it easier to read. This is not actually just building the filter set for the input measure, it's building filters for the aggregated metric. And since that is the only place filters get applied, it's really building filters for the whole metric query. (This was true before your change, too.) Can we rename it accordingly? Maybe something like:
- `_build_filter_specs_for_input_measure()` -> `_build_filter_specs_for_queried_metric()`

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -125,6 +125,14 @@ class MeasureSpecProperties:
     non_additive_dimension_spec: Optional[NonAdditiveDimensionSpec] = None
 
 
+@dataclass(frozen=True)
+class MetricInputMeasureSpecFilterSet:
```
courtneyholcomb, line 129: Similar to my other comment - can we rename this? Maybe: `MetricInputMeasureSpecFilterSet` -> `TieredFilterSpecSet`

I also would love to add some more detail to this class for readability. Can we have it take in `measure_filter_specs`, `metric_filter_specs`, and `query_filter_specs`? Then we can just add properties to resolve `all_filter_specs` and `after_aggregation_filter_specs`.
Also maybe call it `after_measure_aggregation_filter_specs` or something similar just to be extra clear, since there are other things that get aggregated in a query besides just measures.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -1327,7 +1335,8 @@ def _build_input_measure_spec(
             offset_window=child_metric_offset_window,
             offset_to_grain=child_metric_offset_to_grain,
             cumulative_description=cumulative_description,
-            filter_specs=filter_specs,
+            after_aggregation_filter_specs=filter_spec_set.after_aggregation_filter_specs if filter_spec_set else (),
```
courtneyholcomb, line 1338: Can we update this method to just take in the `filter_spec_set` instead of these two separate params?

metricflow/dataflow/builder/dataflow_plan_builder.py
courtneyholcomb: Also would be good to add a more descriptive docstring to explain what each tier of filter represents!

metricflow-semantics/metricflow_semantics/test_helpers/semantic_manifest_yamls/simple_manifest/metrics.yaml
```diff
@@ -11,7 +11,10 @@ metric:
   description: "average booking value metric"
   type: simple
   type_params:
-    measure: average_booking_value
+    measure:
+      name: average_booking_value
+      filter: "average_booking_value > 10"
```
courtneyholcomb, line 16: Thank you for adding! So silly that we didn't have this before

metricflow-semantics/metricflow_semantics/test_helpers/semantic_manifest_yamls/simple_manifest/metrics.yaml
```diff
@@ -629,6 +632,16 @@ metric:
       fill_nulls_with: 0
   time_granularity: week
 ---
+metric:
```
courtneyholcomb, line 635: Can we add a metric-level filter here too to see what happens in a query with both?

tests_metricflow/query_rendering/test_time_spine_join_rendering.py
```diff
@@ -152,3 +152,34 @@ def test_join_to_time_spine_with_queried_time_constraint(
         dataflow_plan_builder=dataflow_plan_builder,
         query_spec=query_spec,
     )
+
+
+@pytest.mark.sql_engine_snapshot
+def test_join_to_time_spine_with_input_measure_constraint(
```
courtneyholcomb, line 158: Will you add the same test or a similar one to the check query tests? i.e., the ones in `itest_***.yaml`. I always like to add both because those ones ensure that the query can also execute (in case we miss something when reading the SQL snapshots).

metricflow-semantics/metricflow_semantics/test_helpers/semantic_manifest_yamls/simple_manifest/metrics.yaml
WilliamDee: woops, i misadded this one actually. I added a new metric that has a measure filter below.

metricflow/dataflow/builder/dataflow_plan_builder.py
WilliamDee: Decided to keep this for now, until maybe we can rip it out of the input_measure potentially?

metricflow-semantics/metricflow_semantics/specs/where_filter/where_filter_spec_set.py
```diff
@@ -0,0 +1,41 @@
...
+    measure_level_filter_specs: Tuple[WhereFilterSpec, ...] = ()
+    metric_level_filter_specs: Tuple[WhereFilterSpec, ...] = ()
+    query_level_filter_specs: Tuple[WhereFilterSpec, ...] = ()
+
+    @property
+    def post_aggregation_filter_specs(self) -> Tuple[WhereFilterSpec, ...]:
+        """Returns filters relevant to post-measure aggregation."""
+        return self.metric_level_filter_specs + self.query_level_filter_specs
+
+    @property
```
courtneyholcomb, line 32: Can we specify in the docstring that all filters should get applied before measure aggregation?

courtneyholcomb (approved):
Looks great!! Should be good to go once you add those check query tests!

courtneyholcomb:
After rebasing and adding the check query tests, can you run the SQL engine tests on this PR too?

PR #1453, opened 2024-10-10
Remove unneeded group bys from time spine dataset

courtneyholcomb:
We've been applying an unnecessary group by to the time spine dataset. Previously, we were applying group bys whenever a `DATE_TRUNC` column was included in the select statement, _even if_ the base grain was included in the select statement. These group bys are unnecessary, because If there is no `DATE_TRUNC` applied to one select column, then none of the columns will be changed by the group by. For example:
```
SELECT
  ds
  , DATE_TRUNC(ds, month)
FROM time_spine
GROUP BY
  ds, DATE_TRUNC(ds, month)
```
will return the same result as:
```
SELECT
  ds
  , DATE_TRUNC(ds, month)
FROM time_spine
```
The results are the same, but the first query is inefficient because of the unnecessary group by. This PR removes the unnecessary group bys.
Note that if the base grain is not included, the group by is still necessary.

I recommend reviewing by commit because the number of snapshot changes is very large.

### DIFF
--- a/.changes/unreleased/Fixes-20241009-174346.yaml
+++ b/.changes/unreleased/Fixes-20241009-174346.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Remove unnecessary group bys that make queries less efficient.
+time: 2024-10-09T17:43:46.011252-07:00
+custom:
+  Author: courtneyholcomb
+  Issue: "1453"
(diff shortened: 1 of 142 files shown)


DevonFulcher (approved):
Nice! I'm curious, why two changelogs? Also, this is meant to improve data warehouse query-time performance, right? Not MF parse-time performance?

courtneyholcomb:
> Nice! I'm curious, why two changelogs? Also, this is meant to improve data warehouse query-time performance, right? Not MF parse-time performance?

@DevonFulcher whoops, I added a new changelog this morning after seeing the GitHub comment saying I needed one, not remembering that I already added one yesterday 😂 will remove the new one
And yes! The group by would slow down the query in the warehouse. This isn't part of the performance improvement project, it's just something I found while working on a different bug fix.

PR #1455, opened 2024-10-10
Court/ts filter3

courtneyholcomb:


metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -286,23 +296,81 @@ def _make_time_spine_data_set(
                 )
             # TODO: also handle date part.
 
+        output_instance_set = InstanceSet(
```
courtneyholcomb, line 299: This output set was incorrect before this PR, but was never used for any real logic so it didn't matter.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -286,23 +296,81 @@ def _make_time_spine_data_set(
...
+                column_alias=time_spine_source.base_column,
+                time_range_constraint=time_range_constraint,
+            )
+            if time_range_constraint
+            else None,
+        )
+
+        # Where constraints must be applied in an outer query since they are using an alias (e.g., 'metric_time__day'),
+        # and some engines do not support using an alias in the WHERE clause.
+        if not time_spine_where_constraints:
```
WilliamDee, line 332: nit: I always have a hard time with `not x` for these, it feels easier to read if we have like `if len(time_spine_where_constraints) == 0`

WilliamDee (approved):

PR #1456, opened 2024-10-15
Fix missing snapshots

WilliamDee:
## Context
After merging https://github.com/dbt-labs/metricflow/pull/1452, seems like snapshots in MF-semantics were missing (CI still passed - odd). But updated those and also updated the script to update snapshots in MF semantics as that seems to be missing.

### DIFF
--- a/tests_metricflow/generate_snapshots.py
+++ b/tests_metricflow/generate_snapshots.py
@@ -55,6 +55,7 @@ logger = logging.getLogger(__name__)
 
 
 TEST_DIRECTORY = "tests_metricflow"
+MF_SEMANTICS_TEST_DIRECTORY = "metricflow-semantics/tests_metricflow_semantics"
 
 
 class MetricFlowTestCredentialSet(FrozenBaseModel):  # noqa: D101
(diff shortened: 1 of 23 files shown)


tests_metricflow/generate_snapshots.py
```diff
@@ -142,6 +143,10 @@ def run_tests(test_configuration: MetricFlowTestConfiguration) -> None:  # noqa:
     if test_configuration.engine is SqlEngine.DUCKDB:
         # DuckDB is fast, so generate all snapshots, including the engine-agnostic ones
         run_command(f"pytest -x -vv -n 4 --overwrite-snapshots -k 'not itest' {TEST_DIRECTORY}")
+
+        # Run snapshots changes for metricflow-semantics
+        # these are not dialect specific, so only need to run once
+        run_command(f"pytest -x -vv -n 4 --overwrite-snapshots {MF_SEMANTICS_TEST_DIRECTORY}")
```
WilliamDee, line 149: put this to be ran under the `duckdb` engine only, as we only need to run these updates once since it's not dialect specific

courtneyholcomb: Did you test this to make sure it works as expected?

courtneyholcomb (approved):

tests_metricflow/generate_snapshots.py
WilliamDee: yup!

PR #1460, opened 2024-10-15
Implement `Mergeable` for `SqlBindParameterSet`

plypaul:
`SqlBindParameterSet` implements `combine()`, which is similar to the interface for `Mergeable` objects. This PR updates `SqlBindParameterSet` to be a subclass of `Mergeable` for consistency and the ability to use convenience methods for `Mergeable` objects in later PRs.


WilliamDee (approved):
TIL! I just added the `WhereFilterSpecSet` that had `merge` but was unaware of this `Mergeable` class

PR #1481, opened 2024-10-28
Improve performance of `MetricTimeQueryValidationRule`

plypaul:
This PR improves the performance of `MetricTimeQueryValidationRule`. One of the reasons for the slow performance is that it currently makes calls to `get_joinable_scd_specs_for_metric`, which can be slow with large manifests. Instead, this runs checks based on the resolved group-by-items, which is a much smaller set.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/query/validation_rules/metric_time_requirements.py
+++ b/metricflow-semantics/metricflow_semantics/query/validation_rules/metric_time_requirements.py
@@ -1,22 +1,21 @@
 from __future__ import annotations
 
 import typing
-from dataclasses import dataclass
-from typing import List, Sequence, Tuple
+from typing import List, Sequence
 
 from dbt_semantic_interfaces.enum_extension import assert_values_exhausted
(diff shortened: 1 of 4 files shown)


metricflow-semantics/metricflow_semantics/query/validation_rules/metric_time_requirements.py
```diff
@@ -71,73 +59,44 @@ def __init__(  # noqa: D107
...
-    def _uncached_query_items_analysis(
-        self, query_resolver_input: ResolverInputForQuery, metric_reference: MetricReference
-    ) -> QueryItemsAnalysis:
-        has_agg_time_dimension = False
-        has_metric_time = False
-        scds: List[InstanceSpec] = []
-
-        valid_agg_time_dimension_specs = self._manifest_lookup.metric_lookup.get_valid_agg_time_dimensions_for_metric(
-            metric_reference
+        self._query_includes_metric_time = (
```
serramatutu, line 62: lol this is so much better I didn't know you could do it like this. Next time I'll do it this way!

serramatutu (approved):

PR #1483, opened 2024-10-29
Remove unnecessary logic for retrieving agg time dimension for conversion metrics

WilliamDee:
## Context

For conversion metrics, we use the agg time dimesnion of each measure to do the conversion window calculation. Since we have `metric_time` which is already referencing the agg time dimension of a given semantic model, we don't need to use the actual agg time dimension reference. This removes the need for additional logic in the code as well as duplication in the rendered SQL.

**Note:** This should not affect any changes in the query result and should be just removing unnecessary duplication in the rendered SQL

### Example
Currently, the rendered SQL
```sql
...
SELECT
          , subq_25.ds__day AS ds__day <-- dupe
          , subq_25.ds__day AS metric_time__day  <-- dupe
          , subq_25.user AS user
          , subq_25.visits AS visits
        FROM (
          -- Read Elements From Semantic Model 'visits_source'
          SELECT
            1 AS visits
...
```
Now,
```sql
...
SELECT
          subq_25.ds__day AS metric_time__day
          , subq_25.user AS user
          , subq_25.visits AS visits
        FROM (
          -- Read Elements From Semantic Model 'visits_source'
          SELECT
            1 AS visits
...
```


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -302,19 +306,14 @@ def _build_aggregated_conversion_node(
...
-        # Get the agg time dimension for each measure used for matching conversion time windows
-        base_time_dimension_spec = TimeDimensionSpec.from_reference(
-            self._semantic_model_lookup.get_agg_time_dimension_for_measure(base_measure_spec.measure_spec.reference)
-        )
-        conversion_time_dimension_spec = TimeDimensionSpec.from_reference(
-            self._semantic_model_lookup.get_agg_time_dimension_for_measure(
-                conversion_measure_spec.measure_spec.reference
-            )
-        )
+        # Get the time dimension used to calculate the conversion window
```
courtneyholcomb, line 309: Some questions about this!
1. Why do we use the metric's default granularity here? It looks like previously we were using `TimeDimensionSpec.from_reference()`, which defaults to `DAY` grain, so this is an improvement. But what if the base and conversion measures have different agg time dimensions with different defined granularities? What would happen in that case? Maybe we should write tests for that scenario if we're not sure.
2. Do we currently require `metric_time` to be queried with conversion metrics? Or can users use `agg_time_dimension` instead? Just want to make sure this won't be a breaking change if we assume `metric_time` here.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -250,6 +251,9 @@ def _build_aggregated_conversion_node(
         # Due to other outstanding issues with conversion metric filters, we disable predicate
         # pushdown for any filter parameter set that is not part of the original time range constraint
         # implementation.
+        default_granularity = ExpandedTimeGranularity.from_time_granularity(
```
courtneyholcomb, line 254: nit: can we move this so that the comment above will still be next to the code it's referencing?

metricflow/dataflow/builder/dataflow_plan_builder.py
WilliamDee: 1. We're getting the min queryable granularity for the metric, so in that case, I'd expect it to take the finest grain possible amongst the 2 measure's agg time dimension and then it'll use that for both time dims when doing the conversion condition
2. We do not require `metric_time` to be in the group by, and it'll still work since we're adding that automatically to the conversion calculation join logic regardless. As for this change, it shouldn't be breaking anything as it doesn't change the result of anything, we're just cleaning up the rendered SQL.

Adding onto 2), we initially did this during the conversion calculation
```sql
SELECT ... FROM (
	...
)
INNER JOIN (
SELECT
  DATE_TRUNC('day', ds) AS metric_time__day,
  DATE_TRUNC('day', ds) AS ds__day
  , user_id AS user
  , 1 AS buys
FROM buy_source
) buys
ON
  (visits.ds__day <= buys.ds__day)
)
```
Now instead, we can just get rid of the redundant `ds__day` and just use `metric_time__day` directly, so it looks like,
```sql
SELECT ... FROM (
	...
)
INNER JOIN (
SELECT
  DATE_TRUNC('day', ds) AS metric_time__day,
  , user_id AS user
  , 1 AS buys
FROM buy_source
) buys
ON
  (visits.metric_time__day <= buys.metric_time__day)
)
```

courtneyholcomb: Can we sync on this on a call later? I want to clarify some stuff!

courtneyholcomb (approved):
Looks great!! 🚀

PR #1494, opened 2024-10-30
Fix `SourceScanOptimizer` generating duplicated columns

serramatutu:
## Summary

This PR fixes the issue with `SourceScanOptimizer` generating duplicated columns. The problem was happening because we didn't check for the same alias referring to different things before merging nodes.

The fix for it was pretty simple, and I added a few tests to catch this issue.

I recommend reviewing commit by commit :)


metricflow-semantics/metricflow_semantics/test_helpers/semantic_manifest_yamls/simple_manifest/metrics.yaml
```diff
@@ -805,3 +805,36 @@ metric:
       name: listings
   filter: "{{ Metric('views', ['listing']) }} > 10"
   time_granularity: week
+---
+# note that in both 1a and 1b "shared_alias" refer to "bookinks", whereas
+# in 2 it refers to "instant_bookings", hence the naming difference
+metric:
+  name: derived_sharedalias_1a
```
courtneyholcomb, line 812: can we add an underscore between `shared` and `alias` in these metric names? just to improve readability

metricflow-semantics/metricflow_semantics/test_helpers/semantic_manifest_yamls/simple_manifest/metrics.yaml
```diff
@@ -805,3 +805,36 @@ metric:
       name: listings
   filter: "{{ Metric('views', ['listing']) }} > 10"
   time_granularity: week
+---
+# note that in both 1a and 1b "shared_alias" refer to "bookinks", whereas
```
courtneyholcomb, line 809: nit: bookings is spelled wrong

tests_metricflow/query_rendering/test_derived_metric_rendering.py
```diff
@@ -795,3 +795,29 @@ def test_offset_to_grain_metric_filter_and_query_have_different_granularities(
...
+@pytest.mark.sql_engine_snapshot
+def test_derived_metric_that_defines_the_same_alias_in_different_components(
+    request: FixtureRequest,
+    mf_test_configuration: MetricFlowTestConfiguration,
+    dataflow_plan_builder: DataflowPlanBuilder,
+    query_parser: MetricFlowQueryParser,
+    sql_client: SqlClient,
+    dataflow_to_sql_converter: DataflowToSqlQueryPlanConverter,
+) -> None:
+    """Tests querying a cumulative metric which give the same alias to its components."""
```
courtneyholcomb, line 809: cumulative -> derived

tests_metricflow/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/DuckDB/test_derived_metric_that_defines_the_same_alias_in_different_components__plan0_optimized.sql
```diff
@@ -0,0 +1,59 @@
...
+-- Order By [] Limit 1
+SELECT
+  COALESCE(subq_18.booking__is_instant, subq_24.booking__is_instant) AS booking__is_instant
+  , MAX(subq_18.derived_sharedalias_1a) AS derived_sharedalias_1a
+  , MAX(subq_24.derived_sharedalias_2) AS derived_sharedalias_2
+FROM (
+  -- Compute Metrics via Expressions
+  SELECT
+    booking__is_instant
+    , shared_alias - 10 AS derived_sharedalias_1a
```
courtneyholcomb, line 11: Do these SQL queries run as expected even though we use the `bookings_source_src_28000` alias twice ?

tests_metricflow/integration/test_cases/itest_metrics.yaml
```diff
@@ -551,6 +551,21 @@ integration_test:
         ds
     ) a
 ---
+integration_test:
```
courtneyholcomb, line 554: Maybe add one more test here for querying 1a and 1b together? Just to make sure that SQL also executes as expected

.changes/unreleased/Fixes-20241030-201214.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Fixes
+body: Fix `SourceScanOptimizer` generated duplicated columns when querying for 2 derived metrics which define the same alias
```
courtneyholcomb, line 2: Would change this wording slightly since `SourceScanOptimizer` isn't actually _generating_ the columns. Something like:
> Prevent `SourceScanOptimizer` from combining nodes that use the same input metric alias in different derived metrics

metricflow/dataflow/nodes/compute_metrics.py
```diff
@@ -28,9 +28,11 @@ class ComputeMetricsNode(DataflowPlanNode):
     metric_specs: Tuple[MetricSpec, ...]
     for_group_by_source_node: bool
     _aggregated_to_elements: Tuple[LinkableInstanceSpec, ...]
+    _alias_to_metric_spec: Dict[str, MetricSpec] = field(hash=False)
```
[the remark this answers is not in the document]
serramatutu: I added this private thing as a way to cache the `alias -> MetricSpec` map, mostly because if we compare multiple times against the same node we don't need to iterate over all of its metric specs.

Not really sure if this optimization is necessary, though. We might be able to get away with just looping through things everytime.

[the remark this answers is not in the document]
serramatutu: In this case, there can't be multiple metrics associated with each alias, though. In one compute metrics node, there can only be one alias per metric since we're assuming metrics need to be unique within one derived metric. So keeping it as a dict is fine, I think.

Do you think we need this dict or should I just go back to a for loop and avoid this caching?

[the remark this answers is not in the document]
courtneyholcomb: Ohhh good point! I don't think we should keep it as a param that can be overridden. I'm not worried about the list of input metrics being long so yeah I'm fine with just putting this logic where it will be used.

tests_metricflow/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/DuckDB/test_derived_metric_that_defines_the_same_alias_in_different_components__plan0_optimized.sql
serramatutu: I think it's fine since the alias is defined separately in each subquery so the scope is different

metricflow/dataflow/nodes/compute_metrics.py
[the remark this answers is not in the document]
serramatutu: I removed this logic from the dataclass, and now we're building the dict inside `can_combine`.

tests_metricflow/integration/test_cases/itest_metrics.yaml
serramatutu: Done!

serramatutu:
Since the comments I added some fixup commits to fix the PR comments.

tests_metricflow/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/DuckDB/test_derived_metric_that_defines_the_same_alias_in_different_components__plan0_optimized.sql
courtneyholcomb: Yeah ok I think it should be fine!

courtneyholcomb (approved):
Looks great, thank you!! 🚢

PR #1495, opened 2024-10-30
Synthetic manifests for performance tests

plypaul:
This PR adds a few classes to generate a synthetic manifest. The synthetic manifest can then be used to profile performance and scaling behavior.

### DIFF
--- a/tests_metricflow/performance/synthetic_manifest_parameter_set.py
+++ b/tests_metricflow/performance/synthetic_manifest_parameter_set.py
@@ -1,0 +1,48 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class SyntheticManifestParameterSet:
+    """Describes how to generate a synthetic manifest for performance testing.
+
+    Goals are:
+    * Allow modeling of similar patterns seen in production manifests.
+    * Make generation straightforward.
+    * Minimize the number of parameters required.
+
+    Notes:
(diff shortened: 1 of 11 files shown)


tests_metricflow/performance/synthetic_manifest_parameter_set.py
```diff
@@ -0,0 +1,48 @@
...
+class SyntheticManifestParameterSet:
+    """Describes how to generate a synthetic manifest for performance testing.
+
+    Goals are:
+    * Allow modeling of similar patterns seen in production manifests.
+    * Make generation straightforward.
+    * Minimize the number of parameters required.
+
+    Notes:
+    * The synthetic manifest groups semantic models into two types - ones containing measures, and others containing dimensions.
```
serramatutu, line 16: I suppose this still captures the relevant behavior from production models, so it's fine? Is there anything extra we'd able to test by relaxing this and creating "mixed" models?

serramatutu (approved):

tests_metricflow/performance/synthetic_manifest_parameter_set.py
serramatutu: Same question for the other constraints: you can reach all measures via all dimensions etc.

plypaul: So those are simplifications to make it easier to generate semantic models and to make it easier to generate queries for metric. Mixed models would make it mimic production better, but I didn't think the added complexity would flesh out any performance cases that much better.

PR #1496, opened 2024-10-31
Update `LinkableSpecName.from_name` to parse custom grains

WilliamDee:
## Context

With the addition of custom granularities, when we parse the dundered names via `LinkableSpecName.from_name`, we need to be able to know whether the grain provided is a custom grain. This means we need to pass that information through to this classmethod.

Resolves SL-2971

### DIFF
--- a/metricflow-semantics/metricflow_semantics/model/semantics/semantic_model_lookup.py
+++ b/metricflow-semantics/metricflow_semantics/model/semantics/semantic_model_lookup.py
@@ -229,7 +229,9 @@ class SemanticModelLookup:
             semantic_models_for_dimension = self._dimension_index.get(dim.reference, []) + [semantic_model]
             self._dimension_index[dim.reference] = semantic_models_for_dimension
 
-            if not StructuredLinkableSpecName.from_name(dim.name).is_element_name:
+            if not StructuredLinkableSpecName.from_name(
+                qualified_name=dim.name, custom_granularity_names=self.custom_granularity_names
+            ).is_element_name:
                 # TODO: [custom granularity] change this to an assertion once we're sure there aren't exceptions
                 logger.warning(
                     LazyFormat(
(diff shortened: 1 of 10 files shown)


metricflow/engine/models.py
```diff
@@ -107,6 +111,7 @@ def from_pydantic(
             is_partition=pydantic_dimension.is_partition,
             expr=pydantic_dimension.expr,
             label=pydantic_dimension.label,
+            _possible_custom_granularity_names=custom_granularity_names,
```
courtneyholcomb, line 114: I think using this param is kinda hacky and there should be a better option. My thought is that we should add a new attribute to the `Dimension` object to store the `entity_links`. Then we wouldn't need custom granularity names here at all. Instead the `granularity_free_qualified_name` would just use something like this logic:
`StructuredLinkableSpecName(element_name=self.name, entity_links=(entity_prefix,)).granularity_free_qualified_name`
And we could also add a property to resolve `Dimension.qualified_name` instead of having to pass that in.
Seems cleaner and it definitely seems useful to have the entity prefix in general!

courtneyholcomb (commented):
Got a few suggestions, especially around performance! But otherwise looks great!

metricflow/engine/models.py
WilliamDee: ooo, ok that makes sense!

metricflow-semantics/metricflow_semantics/model/semantics/semantic_model_lookup.py
```diff
@@ -70,6 +70,11 @@ def __init__(self, model: SemanticManifest, custom_granularities: Dict[str, Expa
         # Cache for agg. time dimension for measure.
         self._measure_reference_to_agg_time_dimension_specs: Dict[MeasureReference, Sequence[TimeDimensionSpec]] = {}
 
+    @property
```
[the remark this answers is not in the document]
WilliamDee: I couldn't pass a set to `StructuredLinkableSpecName.from_name()` if we want to cache that, since sets aren't hashable so not cachable. I decided to use caching instead of changing this to a set.

metricflow/engine/models.py
courtneyholcomb: Sweet. Just a note here that we'll need to update MFS with this new property since I think we're using the `Dimension` object there!

courtneyholcomb (approved):
Looks great!! 🚀 🚀

PR #1500, opened 2024-11-04
Add `eq=False` to node classes

plypaul:
This sets `eq=False` for node dataclasses as nodes should be unique, and equivalence methods should be used for checking as the nodes represent recursive data structures. Without this, comparisons or using nodes as keys can be slow as the generated equals function will traverse the recursive structure.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/dag/mf_dag.py
+++ b/metricflow-semantics/metricflow_semantics/dag/mf_dag.py
@@ -57,9 +57,14 @@ class DagNodeVisitor(Generic[VisitorOutputT], ABC):
 DagNodeT = TypeVar("DagNodeT", bound="DagNode")
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True, eq=False)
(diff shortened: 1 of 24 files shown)


metricflow-semantics/metricflow_semantics/dag/mf_dag.py
```diff
@@ -57,7 +57,7 @@ def visit_node(self, node: DagNode) -> VisitorOutputT:  # noqa: D102
 DagNodeT = TypeVar("DagNodeT", bound="DagNode")
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True, eq=False)
```
courtneyholcomb, line 60: Can we add a comment here explaining essentially what you explained in the PR description? Will likely be helpful to future devs.

courtneyholcomb (approved):

metricflow-semantics/metricflow_semantics/dag/mf_dag.py
plypaul: Added.

PR #1502, opened 2024-11-04
Replace `is_table` with `as_sql_table_node`

plypaul:
This removes `SqlQueryPlanNode.is_table` and replaces it with a typed accessor for parallelism with `as_select_node`. This is helpful in handling some use cases for CTEs.

### DIFF
--- a/metricflow/sql/sql_plan.py
+++ b/metricflow/sql/sql_plan.py
@@ -41,14 +41,14 @@ class SqlQueryPlanNode(DagNode["SqlQueryPlanNode"], ABC):
 
     @property
     @abstractmethod
-    def is_table(self) -> bool:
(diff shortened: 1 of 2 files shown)


metricflow/sql/sql_plan.py
```diff
@@ -356,11 +358,11 @@ def render_node(self) -> SqlQueryPlanNode:  # noqa: D102
 class SqlCteNode(SqlQueryPlanNode):
     """Represents a single common table expression."""
 
-    select_statement: SqlSelectStatementNode
+    select_statement: SqlQueryPlanNode
```
plypaul, line 361: Notice this change got mixed in - this is to make recursive handling easier and it turns out the typed form is not necessary.

courtneyholcomb (approved):

PR #1503, opened 2024-11-04
Support CTEs in the column pruner

plypaul:
This updates the column pruner optimizer to support CTEs. To support CTEs, the required columns are mapped to the corresponding CTE and propagated like other parent nodes.

### DIFF
--- a/metricflow/sql/optimizer/tag_required_column_aliases.py
+++ b/metricflow/sql/optimizer/tag_required_column_aliases.py
@@ -106,13 +106,27 @@ class SqlMapRequiredColumnAliasesVisitor(SqlQueryPlanNodeVisitor[None]):
 
     @override
     def visit_cte_node(self, node: SqlCteNode) -> None:
-        raise NotImplementedError
+        select_statement = node.select_statement
+        # Copy the tagged aliases from the CTE to the SELECT since when visiting a SELECT, the CTE node (not the SELECT
+        # in the CTE) was tagged with the required aliases.
+        required_column_aliases_in_this_node = self._current_required_column_alias_mapping.get_aliases(node)
+        self._current_required_column_alias_mapping.add_aliases(select_statement, required_column_aliases_in_this_node)
+        # Visit parent nodes.
+        select_statement.accept(self)
 
     def _visit_parents(self, node: SqlQueryPlanNode) -> None:
         """Default recursive handler to visit the parents of the given node."""
         for parent_node in node.parent_nodes:
             parent_node.accept(self)
         return
+
+    def _tag_potential_cte_node(self, table_name: str, column_aliases: Set[str]) -> None:
+        """A reference to a SQL table might be a CTE. If so, tag the appropriate aliases in the CTEs."""
+        cte_node = self._cte_alias_to_cte_node.get(table_name)
+        if cte_node is not None:
+            self._current_required_column_alias_mapping.add_aliases(cte_node, column_aliases)
+            # `visit_cte_node` will handle propagating the required aliases to all CTEs that this CTE node depends on.
+            cte_node.accept(self)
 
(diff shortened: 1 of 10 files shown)


metricflow/sql/optimizer/tag_required_column_aliases.py
```diff
@@ -55,6 +55,7 @@ def __init__(self, tagged_column_alias_set: TaggedColumnAliasSet) -> None:
             traverses the SQL-query representation DAG.
         """
         self._column_alias_tagger = tagged_column_alias_set
+        self._cte_alias_to_cte_node: Dict[str, SqlCteNode] = {}
```
courtneyholcomb, line 58: Nit: can we rename this to something more clear like `_required_aliases_for_cte_nodes`?

metricflow/sql/optimizer/tag_required_column_aliases.py
```diff
@@ -86,17 +87,32 @@ def _search_for_expressions(
 
     @override
     def visit_cte_node(self, node: SqlCteNode) -> None:
-        raise NotImplementedError
+        select_statement = node.select_statement
+        # Copy the tagged aliases from the CTE to the SELECT since when visiting a SELECT, the CTE node (not the SELECT
```
courtneyholcomb, line 91: Why is it necessary to tag both the CTE node and CTE's select node with the same aliases?

metricflow/sql/optimizer/tag_required_column_aliases.py
```diff
@@ -86,17 +87,32 @@ def _search_for_expressions(
...
             parent_node.accept(self)
         return
 
+    def _tag_potential_cte_node(self, table_name: str, column_aliases: Set[str]) -> None:
+        """A reference to a SQL table might be a CTE. If so, tag the appropriate aliases in the CTEs."""
+        cte_node = self._cte_alias_to_cte_node.get(table_name)
+        if cte_node is not None:
+            self._column_alias_tagger.tag_aliases(cte_node, column_aliases)
+            # Propagate the required aliases to parents, which could be other CTEs.
+            cte_node.accept(self)
```
courtneyholcomb, line 110: Shouldn't CTEs be considered parent nodes? Why is this handled here instead of in `_visit_parents()`?

plypaul: This comment is not quite correct, so let me update. But CTE nodes are only the parents of the top level SELECT statement. So for something like:

```
-- node_id=cte_0
WITH some_table AS (
  SELECT 1
)

-- node_id=ss_0
SELECT * FROM (
  -- node_id=ss_1
  SELECT * 
  FROM 
   --node_id=st_0
    some_table
) a
```

The parents of `ss_0` are `[ss_1, cte_0]`, but the only parent of `ss_1` is `st_0`.

metricflow/sql/optimizer/tag_required_column_aliases.py
plypaul: Before visiting a node, it's assumed that we have already recorded the column aliases required at the node. So for a SELECT statement, the hierarchy looks like:

```
SELECT node -> CTE node -> SELECT node of the CTE
```

If we don't do this copy, this is what happens:

Let's say we start the process with having recorded that `col_0` is required in `SELECT node`. 

* We visit the `SELECT node` and record that `col_0` is required in  `CTE node`.
* Then we visit the CTE node. At the CTE node, we just go visit the `SELECT node of the CTE`
* At the `SELECT node of the CTE`, then there's no information about which columns were required.

courtneyholcomb (approved):

metricflow/sql/optimizer/tag_required_column_aliases.py
plypaul: This was resolved as a part of changes related to another comment.

PR #1504, opened 2024-11-04
Support CTEs in sub-query reducer

plypaul:
This updates the sub-query reducer to support CTEs. The `SELECT` statement in each CTE is treated like a `SELECT`, and all CTEs are retained. i.e. a CTE is not reduced with another CTE. The CTEs are retained to make mapping to dataflow plan nodes easier in later PRs.

### DIFF
--- a/metricflow/sql/optimizer/rewriting_sub_query_reducer.py
+++ b/metricflow/sql/optimizer/rewriting_sub_query_reducer.py
@@ -106,12 +106,12 @@ class SqlRewritingSubQueryReducerVisitor(SqlQueryPlanNodeVisitor[SqlQueryPlanNod
             ),
             join_descs=tuple(
                 SqlJoinDescription(
-                    right_source=x.right_source.accept(self),
-                    right_source_alias=x.right_source_alias,
-                    on_condition=x.on_condition,
-                    join_type=x.join_type,
+                    right_source=join_desc.right_source.accept(self),
+                    right_source_alias=join_desc.right_source_alias,
+                    on_condition=join_desc.on_condition,
+                    join_type=join_desc.join_type,
                 )
-                for x in node.join_descs
+                for join_desc in node.join_descs
             ),
             group_bys=node.group_bys,
             order_bys=node.order_bys,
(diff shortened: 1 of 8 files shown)


metricflow/sql/optimizer/rewriting_sub_query_reducer.py
```diff
@@ -196,7 +199,7 @@ def _is_simple_source(node: SqlSelectStatementNode) -> bool:
             if select_column.expr.lineage.contains_aggregate_exprs:
                 return False
         return (
-            len(node.parent_nodes) <= 1
+            len(node.join_descs) == 0
```
courtneyholcomb, line 202: This looks like a behavior change - were there changes to the snapshots from this change alone?

metricflow/sql/optimizer/rewriting_sub_query_reducer.py
```diff
@@ -612,10 +615,11 @@ def visit_select_statement_node(self, node: SqlSelectStatementNode) -> SqlQueryP
         # JOIN dim_listings c
         # ON a.listing_id = b.listing_id
 
-        assert len(node_with_reduced_parents.parent_nodes) == 1
```
courtneyholcomb, line 615: Was there a reason to remove this assertion?

metricflow/sql/optimizer/rewriting_sub_query_reducer.py
```diff
@@ -579,6 +578,13 @@ def _rewrite_node_with_join(node: SqlSelectStatementNode) -> SqlSelectStatementN
             select_columns=tuple(clauses_to_rewrite.select_columns),
             from_source=from_source,
             from_source_alias=from_source_alias,
+            cte_sources=tuple(
+                SqlCteNode.create(
```
courtneyholcomb, line 582: Should this use `cte_source.with_new_select`?

courtneyholcomb (approved):

metricflow/sql/optimizer/rewriting_sub_query_reducer.py
plypaul: There were no snapshot changes. This is not a behavior change as without CTEs, the number of parent nodes is 1 + the number of joins.

metricflow/sql/optimizer/rewriting_sub_query_reducer.py
plypaul: Good catch. `with_new_select` got added as a convenience method late in the stack.

metricflow/sql/optimizer/rewriting_sub_query_reducer.py
plypaul: Yes - since CTEs are modeled as a parent node of a SELECT node, this assertion does not hold.

PR #1511, opened 2024-11-06
Align dataflow plans

courtneyholcomb:
Dedupe logic in the `DataflowPlanBuilder`. Diverging paths in the dataflow plan have frequently been a source of bugs. To reduce similar future bugs, this PR unifies some of the logic across all paths. There are several steps we always take after source node generation and before aggregation. This PR moves all that logic into a helper function to ensure we always use the same logic.


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -328,34 +328,23 @@ def _build_aggregated_conversion_node(
             )
 
         # Build the unaggregated base measure node for computing conversions
-        unaggregated_base_measure_node = base_measure_recipe.source_node
-        if base_measure_recipe.join_targets:
```
courtneyholcomb, line 332: There are a few snapshot changes here, and this is the source of those changes. When we call `join_targets`, it creates a `FilterElementsNode` under the hood, even if we don't use it. On this line, we call `join_targets` without using the node that gets generated. The new code only calls `join_targets` once and uses the generated node, so there the source node IDs are incremented fewer times.
This is true for the other parts of the code referencing `join_targets` in this PR, too.

plypaul (approved):

PR #1516, opened 2024-11-10
Update `SqlTableAliasSimplifier` to handle CTEs

plypaul:
Similar to the updates for the other optimizers in https://github.com/dbt-labs/metricflow/pull/1503 and https://github.com/dbt-labs/metricflow/pull/1504, this PR implements the CTE case for the `SqlTableAliasSimplifier`

### DIFF
--- a/tests_metricflow/sql/optimizer/test_cte_table_alias_simplifier.py
+++ b/tests_metricflow/sql/optimizer/test_cte_table_alias_simplifier.py
@@ -34,7 +34,7 @@ def test_table_alias_simplification(
     mf_test_configuration: MetricFlowTestConfiguration,
     sql_plan_renderer: DefaultSqlQueryPlanRenderer,
 ) -> None:
-    """Tests that table aliases are removed when not needed in CTEs."""
+    """Tests that table aliases in the SELECT statement of a CTE are removed when not needed."""
     select_statement = SqlSelectStatementNode.create(
         description="Top-level SELECT",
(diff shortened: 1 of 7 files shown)


tests_metricflow/sql/optimizer/test_cte_table_alias_simplifier.py
```diff
@@ -0,0 +1,183 @@
...
+def sql_plan_renderer() -> DefaultSqlQueryPlanRenderer:  # noqa: D103
+    return DefaultSqlQueryPlanRenderer()
+
+
+def test_table_alias_simplification(
+    request: FixtureRequest,
+    mf_test_configuration: MetricFlowTestConfiguration,
+    sql_plan_renderer: DefaultSqlQueryPlanRenderer,
+) -> None:
+    """Tests that table aliases are removed when not needed in CTEs."""
```
courtneyholcomb, line 37: Should we also have a test to demonstrate that table aliases are not removed if they are needed in CTEs?
I'm assuming that would only happen if there was a join within the CTE statement. Not sure if that case will ever occur with the current CTE implementation.

courtneyholcomb (approved):

tests_metricflow/sql/optimizer/test_cte_table_alias_simplifier.py
plypaul: Can't hurt - added a test case.

PR #1517, opened 2024-11-10
Split traversal component from `DataflowToSqlQueryPlanConverter`

plypaul:
This PR splits off the part that does the dataflow traversal in `DataflowToSqlQueryPlanConverter` into a separate class to reduce the class size. Later, these two parts can be put into different files.

### DIFF
--- a/metricflow/plan_conversion/dataflow_to_sql.py
+++ b/metricflow/plan_conversion/dataflow_to_sql.py
@@ -210,39 +210,6 @@ class DataflowToSqlQueryPlanConverter:
         )
 
 
-def _make_time_range_comparison_expr(
-    table_alias: str, column_alias: str, time_range_constraint: TimeRangeConstraint
-) -> SqlExpressionNode:
-    """Build an expression like "ds BETWEEN CAST('2020-01-01' AS TIMESTAMP) AND CAST('2020-01-02' AS TIMESTAMP).
-
-    If the constraint uses day or larger grain, only render to the date level. Otherwise, render to the timestamp level.
-    """
-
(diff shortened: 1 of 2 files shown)


metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -237,6 +209,69 @@ def convert_to_sql_query_plan(
...
+        start_expr=SqlStringLiteralExpression.create(
+            literal_value=time_range_constraint.start_time.strftime(time_format_to_render),
+        ),
+        end_expr=SqlStringLiteralExpression.create(
+            literal_value=time_range_constraint.end_time.strftime(time_format_to_render),
+        ),
+    )
+
+
+class DataflowNodeToSqlSubqueryVisitor(DataflowPlanNodeVisitor[SqlDataSet]):
```
courtneyholcomb, line 213: I'm not sure about the name here - it feels like removing `Subquery` would be more accurate (i.e., `DataflowNodeToSqlVisitor`), since there isn't always a subquery involved. Technically, each node resolves to a full SQL query (and the parent node often resolves to a subquery).

courtneyholcomb (approved):
Left one recommendation about naming!

metricflow/plan_conversion/dataflow_to_sql.py
plypaul: Maybe `DataflowNodeToDefaultSqlVisitor`? Issue with `DataflowNodeToSqlVisitor` is differentiating this one from the new one that uses CTEs. I'll have to put renames at the end due to stack conflicts though.

courtneyholcomb: Works for me!

PR #1518, opened 2024-11-10
Assorted formatting / logging updates

plypaul:
This PR contains an assortment of updates related to logging / formatting. Although not really related, since they were small, they were put into a single PR. Please view by commit.

### DIFF
--- a/metricflow/dataset/sql_dataset.py
+++ b/metricflow/dataset/sql_dataset.py
@@ -5,6 +5,7 @@ from typing import List, Optional, Sequence
 from dbt_semantic_interfaces.references import SemanticModelReference
 from metricflow_semantics.assert_one_arg import assert_exactly_one_arg_set
 from metricflow_semantics.instances import EntityInstance, InstanceSet
+from metricflow_semantics.mf_logging.lazy_formattable import LazyFormat
 from metricflow_semantics.specs.column_assoc import ColumnAssociation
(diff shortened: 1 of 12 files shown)


metricflow/dataset/sql_dataset.py
```diff
@@ -52,7 +53,14 @@ def checked_sql_select_node(self) -> SqlSelectStatementNode:
         Otherwise, an exception is thrown.
         """
         if self._sql_select_node is None:
-            raise RuntimeError(f"{self} was created with a SQL node that is not a {SqlSelectStatementNode}")
+            raise RuntimeError(
+                str(
+                    LazyFormat(
+                        f"{self} was created with a SQL node that is not a {SqlSelectStatementNode}",
```
courtneyholcomb, line 59: Should we use the class name here instead of `self`? That way the logs will be more consistent/searchable and any changing attributes can go in the other args.
Also, wondering why `SqlSelectStatementNode` is in curly braces.

courtneyholcomb: I know this probably isn't an error you wrote - this doesn't need to block merge!

courtneyholcomb (approved):

metricflow/dataset/sql_dataset.py
plypaul: Updated.

PR #1520, opened 2024-11-10
Add a method to figure out common nodes in a dataflow plan

plypaul:
This PR adds a method to figure out nodes in a dataflow plan that appear more than once. i.e. a node that is the parent of multiple nodes. These common nodes indicate operations where a computation is reused, e.g. a metric that is used in the computation of multiple derived metrics in a query. These nodes will be later used to generate CTEs.

### DIFF
--- a/metricflow/dataflow/dataflow_plan.py
+++ b/metricflow/dataflow/dataflow_plan.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import functools
 import logging
 import typing
 from abc import ABC, abstractmethod
@@ -23,7 +24,11 @@ logger = logging.getLogger(__name__)
 
 NodeSelfT = TypeVar("NodeSelfT", bound="DataflowPlanNode")
 
+# Make it so that we only have to suppress errors here instead of both at the method and the class.
+ComparisonAnyType = typing.Any  # type: ignore[misc]
 
+
+@functools.total_ordering
 @dataclass(frozen=True, eq=False)
 class DataflowPlanNode(DagNode["DataflowPlanNode"], Visitable, ABC):
     """A node in the graph representation of the dataflow.
@@ -80,6 +85,12 @@ class DataflowPlanNode(DagNode["DataflowPlanNode"], Visitable, ABC):
     def aggregated_to_elements(self) -> Set[LinkableInstanceSpec]:
         """Indicates that the node has been aggregated to these specs, guaranteeing uniqueness in all combinations."""
         return set()
+
+    def __lt__(self, other: ComparisonAnyType) -> bool:  # noqa: D105
+        if not isinstance(other, DataflowPlanNode):
+            raise NotImplementedError
+
+        return self.node_id < other.node_id
 
 
 class DataflowPlan(MetricFlowDag[DataflowPlanNode]):
--- a/metricflow/dataflow/dataflow_plan_analyzer.py
+++ b/metricflow/dataflow/dataflow_plan_analyzer.py
@@ -1,0 +1,79 @@
+from __future__ import annotations
+
+from collections import defaultdict
+from typing import Dict, FrozenSet, Mapping, Sequence, Set
+
+from typing_extensions import override
+
+from metricflow.dataflow.dataflow_plan import DataflowPlan, DataflowPlanNode
+from metricflow.dataflow.dataflow_plan_visitor import DataflowPlanNodeVisitorWithDefaultHandler
+
+
+class DataflowPlanAnalyzer:
+    """Class to determine more complex properties of the dataflow plan.
+
+    These could also be made as member methods of the dataflow plan, but this requires resolving some circular
+    dependency issues to break out the functionality into separate files.
+    """
+
+    @staticmethod
+    def find_common_branches(dataflow_plan: DataflowPlan) -> Sequence[DataflowPlanNode]:
+        """Starting from the sink node, find the common branches that exist in the associated DAG.
+
+        Returns a sorted sequence for reproducibility.
+        """
+        counting_visitor = _CountDataflowNodeVisitor()
+        dataflow_plan.sink_node.accept(counting_visitor)
+
+        node_to_common_count = counting_visitor.get_node_counts()
+
+        common_nodes = []
+        for node, count in node_to_common_count.items():
+            if count > 1:
+                common_nodes.append(node)
+
+        common_branches_visitor = _FindLargestCommonBranchesVisitor(frozenset(common_nodes))
+
+        return tuple(sorted(dataflow_plan.sink_node.accept(common_branches_visitor)))
+
+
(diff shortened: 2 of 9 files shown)


metricflow/dataflow/dataflow_plan.py
```diff
@@ -81,6 +86,12 @@ def aggregated_to_elements(self) -> Set[LinkableInstanceSpec]:
         """Indicates that the node has been aggregated to these specs, guaranteeing uniqueness in all combinations."""
         return set()
 
+    def __lt__(self, other: ComparisonAnyType) -> bool:  # noqa: D105
+        if not isinstance(other, DataflowPlanNode):
+            raise NotImplementedError
```
courtneyholcomb, line 91: Should we add an error message here in case this somehow gets hit? Seems similar to a bare `AssertionError`.

metricflow/dataflow/dataflow_plan_analyzer.py
```diff
@@ -0,0 +1,77 @@
...
+from typing import Dict, FrozenSet, Mapping, Sequence, Set
+
+from typing_extensions import override
+
+from metricflow.dataflow.dataflow_plan import DataflowPlan, DataflowPlanNode
+from metricflow.dataflow.dataflow_plan_visitor import DataflowPlanNodeVisitorWithDefaultHandler
+
+
+class DataflowPlanAnalyzer:
+    """CLass to determine more complex properties of the dataflow plan.
```
courtneyholcomb, line 13: nit: Class*

metricflow/dataflow/dataflow_plan_analyzer.py
```diff
@@ -0,0 +1,77 @@
...
+        for node, count in node_to_common_count.items():
+            if count > 1:
+                common_nodes.append(node)
+
+        common_branches_visitor = _FindLargestCommonBranchesVisitor(frozenset(common_nodes))
+
+        return tuple(sorted(dataflow_plan.sink_node.accept(common_branches_visitor)))
+
+
+class _CountCommonDataflowNodeVisitor(DataflowPlanNodeVisitorWithDefaultHandler[None]):
```
courtneyholcomb, line 40: nit: This isn't counting just common nodes, it's counting all nodes (i.e., you'll get some nodes that are just used once, so they're not common). Could change the name of the class and related variables to reflect that - e.g., `_CountDataflowNodeVisitor`

metricflow/dataflow/dataflow_plan_analyzer.py
```diff
@@ -0,0 +1,77 @@
...
+    """Helper visitor to build a dict from a node in the plan to the number of times it appears in the plans."""
```
courtneyholcomb, line 41: This only traverses one dataflow plan at a time, right? If so, `plan*` instead of `plans`

metricflow/dataflow/dataflow_plan_visitor.py
```diff
@@ -39,82 +41,175 @@ class DataflowPlanNodeVisitor(Generic[VisitorOutputT], ABC):
...
-        pass
+        raise NotImplementedError
 
     @abstractmethod
     def visit_join_to_custom_granularity_node(self, node: JoinToCustomGranularityNode) -> VisitorOutputT:  # noqa: D102
-        pass
+        raise NotImplementedError
+
+
+class DataflowPlanNodeVisitorWithDefaultHandler(DataflowPlanNodeVisitor[VisitorOutputT], Generic[VisitorOutputT]):
```
courtneyholcomb, line 125: Appreciate this reduction in boilerplate! 🙏

metricflow/dataflow/dataflow_plan_analyzer.py
```diff
@@ -0,0 +1,77 @@
...
+    and `C -> D` can be considered common branches, and we want the largest one), this uses preorder traversal and
+    returns the first common node that is seen.
+    """
+
+    def __init__(self, common_nodes: FrozenSet[DataflowPlanNode]) -> None:
+        self._common_nodes = common_nodes
+
+    @override
+    def _default_handler(self, node: DataflowPlanNode) -> FrozenSet[DataflowPlanNode]:
+        if node in self._common_nodes:
```
courtneyholcomb, line 69: It took me a minute to understand why this gets the largest common nodes! Mind adding a comment or a docstring to explain that this early return works because we're traversing from largest to smallest?

tests_metricflow/snapshots/test_common_dataflow_branches.py/str/test_shared_metric_query__result.txt
```diff
@@ -0,0 +1,160 @@
+test_name: test_shared_metric_query
+test_filename: test_common_dataflow_branches.py
+docstring:
+  For a known case, test that a metric computation node is identified as a common branch.
+
+      A query for `bookings` and `bookings_per_booker` should have the computation for `bookings` as a common branch in
```
courtneyholcomb, line 6: This is an interesting case! The `ComputeMetricsNode` for `bookings` is common, but _within_ that node there is another common node that could also be used as a CTE - the `ReadSqlSourceNode` for the `bookings_source` table, which is also used in the `bookers` branch. Curious if you plan to add that optimization later?

courtneyholcomb (approved):
Left a few small comments inline, but overall looks great! 🚀 🚀 🚀

metricflow/dataflow/dataflow_plan.py
plypaul: This is actually a special value that is supposed to be returned for `__lt__` when you try to compare two objects that aren't comparable:

https://docs.python.org/3/library/constants.html

tests_metricflow/snapshots/test_common_dataflow_branches.py/str/test_shared_metric_query__result.txt
plypaul: That's correct, and the `ReadSqlSourceNode` is included in the result snapshot under `common_branch_1` below. I should probably add newlines because the different items are easy to miss.

courtneyholcomb: Ahh nice! Yep I missed that

metricflow/dataflow/dataflow_plan_analyzer.py
plypaul: Updated.

metricflow/dataflow/dataflow_plan_analyzer.py
plypaul: Updated.

metricflow/dataflow/dataflow_plan_analyzer.py
plypaul: Updated.

PR #1522, opened 2024-11-11
Update MF to support breaking changes in DSI for custom calendar

WilliamDee:
## Context

We merged 2 breaking changes in DSI https://github.com/dbt-labs/dbt-semantic-interfaces/pull/363 and https://github.com/dbt-labs/dbt-semantic-interfaces/pull/365 which changed most spec typing that used time granularity to be a `str` instead of `TimeGranularity` to enable support for custom granularity. Similarly, there were additional breaking changes to the objects that requires passing in `custom_granularity_names`. This PR updates all those callsites to be compatible with the new version of DSI (to be released)

Resolves SL-3097

### DIFF
--- a/metricflow-semantics/metricflow_semantics/specs/where_filter/where_filter_dimension.py
+++ b/metricflow-semantics/metricflow_semantics/specs/where_filter/where_filter_dimension.py
@@ -155,7 +155,5 @@ class WhereFilterDimensionFactory(ProtocolHint[QueryInterfaceDimensionFactory]):
             rendered_spec_tracker=self._rendered_spec_tracker,
             element_name=structured_name.element_name,
             entity_links=tuple(EntityReference(entity_link_name.lower()) for entity_link_name in entity_path)
-            + tuple(
-                EntityReference(entity_link_name.lower()) for entity_link_name in structured_name.entity_link_names
-            ),
+            + structured_name.entity_links,
         )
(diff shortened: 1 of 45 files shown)


metricflow-semantics/metricflow_semantics/specs/where_filter/where_filter_dimension.py
```diff
@@ -151,5 +155,7 @@ def create(self, name: str, entity_path: Sequence[str] = ()) -> WhereFilterDimen
             rendered_spec_tracker=self._rendered_spec_tracker,
             element_name=structured_name.element_name,
             entity_links=tuple(EntityReference(entity_link_name.lower()) for entity_link_name in entity_path)
-            + structured_name.entity_links,
+            + tuple(
```
courtneyholcomb, line 158: this change is just to support case insensitivity right?

courtneyholcomb (approved):
LGTM! Thanks for handling so quickly!!

metricflow-semantics/metricflow_semantics/specs/where_filter/where_filter_dimension.py
WilliamDee: nope, we switched from using `DunderedNameFormatter` to `StructuredLinkableSpecName` and that didn't have `entity_links: Tuple[EntityReference, ...]`, but only `entity_link_names: Tuple[str, ...]`. Now that I think about it, i'll just add a property to make `entity_links` work in `StructuredLinkableSpecName` so that we don't need this change

courtneyholcomb: Ooo ok sounds good!

PR #1526, opened 2024-11-13
Use CTEs for metrics in multi-metric cases

plypaul:
This PR updates the CTE generation logic to use CTEs for each metric in multi-metric queries. Although not necessary for performance, the generated SQL for multi-metric queries are easier to follow as the the top-level statement does not have any large subqueries.


courtneyholcomb (commented):
Let's discuss this one today?
I'm not sure that this does actually improve the readability of the SQL, especially in cases when it's actually adding a decent amount more SQL to the optimized query. I left several comments inline.

plypaul:
> I'm not sure that this does actually improve the readability of the SQL, especially in cases when it's actually adding a decent amount more SQL to the optimized query. I left several comments inline.

Because this change creates CTEs when CTEs aren't necessary, this will always produce more verbose SQL. This is more of a stylistic choice that provides a consistent form to the CTE with smaller parts. Analogous example is a long function that does a lot of things, or multiple smaller functions that each do one thing.

@Jstein77 might have some thoughts here.

metricflow/dataflow/dataflow_plan_analyzer.py
```diff
@@ -41,7 +41,7 @@ def find_common_branches(dataflow_plan: DataflowPlan) -> Sequence[DataflowPlanNo
 
     @staticmethod
     def group_nodes_by_type(dataflow_plan: DataflowPlan) -> DataflowPlanNodeSet:
-        """Grouops dataflow plan nodes by type."""
+        """Groups dataflow plan nodes by type."""
         grouping_visitor = _GroupNodesByTypeVisitor()
         return dataflow_plan.sink_node.accept(grouping_visitor)
```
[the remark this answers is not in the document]
plypaul: This should be fine since the number of nodes in a dataflow plan are generally small, e.g. < 100 nodes.

tests_metricflow/snapshots/test_custom_granularity.py/SqlQueryPlan/DuckDB/test_offset_metric_with_custom_granularity__plan0_optimized.sql
```diff
@@ -2,11 +2,8 @@ test_name: test_offset_metric_with_custom_granularity
 test_filename: test_custom_granularity.py
 sql_engine: DuckDB
 ---
--- Compute Metrics via Expressions
-SELECT
-  booking__ds__martian_day
-  , bookings_5_days_ago AS bookings_5_day_lag
-FROM (
+-- Read From CTE For node_id=cm_5
```
Jstein77, line 5: What does this comment mean? Does this mean the final select statment is from cm_5?

tests_metricflow/snapshots/test_custom_granularity.py/SqlQueryPlan/DuckDB/test_offset_metric_with_custom_granularity__plan0_optimized.sql
```diff
@@ -32,4 +29,23 @@ FROM (
...
+  FROM (
+    -- Read From CTE For node_id=cm_4
+    SELECT
+      booking__ds__martian_day
+      , bookings_5_days_ago
+    FROM cm_4_cte cm_4_cte
+  ) subq_17
+)
+
+SELECT
```
Jstein77, line 48: Do we need to rename these columns at the end? We already did that step in cm_5

tests_metricflow/snapshots/test_custom_granularity.py/SqlQueryPlan/DuckDB/test_offset_metric_with_custom_granularity__plan0_optimized.sql
```diff
@@ -2,11 +2,8 @@ test_name: test_offset_metric_with_custom_granularity
 test_filename: test_custom_granularity.py
 sql_engine: DuckDB
 ---
--- Compute Metrics via Expressions
-SELECT
-  booking__ds__martian_day
-  , bookings_5_days_ago AS bookings_5_day_lag
-FROM (
+-- Read From CTE For node_id=cm_5
+WITH cm_4_cte AS (
```
Jstein77, line 6: What does cm mean?

tests_metricflow/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/DuckDB/test_cumulative_time_offset_metric_with_time_constraint__plan0_optimized.sql
```diff
@@ -42,4 +39,23 @@ FROM (
   WHERE subq_17.metric_time__day BETWEEN '2019-12-19' AND '2020-01-02'
   GROUP BY
     subq_17.metric_time__day
-) subq_23
+)
+
+, cm_5_cte AS (
```
Jstein77, line 44: Is it possible to read from CTE cm_4 and compute metrics via expressions in one step? 

i.e if I we're to write this by hand I would write

cm_5_cte as (
select 
   metric_time__day
, every_2_days_bookers_2_days_ago as every_2_days_bookers_2_days_ago
from cm_4_cte
)

SELECT
 metric_time__day
 , every_2_days_bookers_2_days_ago
from cm_5_cte

tests_metricflow/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/DuckDB/test_cumulative_time_offset_metric_with_time_constraint__plan0_optimized.sql
```diff
@@ -42,4 +39,23 @@ FROM (
...
+  -- Compute Metrics via Expressions
+  SELECT
+    metric_time__day
+    , every_2_days_bookers_2_days_ago AS every_2_days_bookers_2_days_ago
```
Jstein77, line 48: if the metric and measure name are the same do we need to try and alias the measure? In this case the alias doesn't do anything.

tests_metricflow/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/DuckDB/test_cumulative_time_offset_metric_with_time_constraint__plan0_optimized.sql
```diff
@@ -42,4 +39,23 @@ FROM (
...
+    -- Read From CTE For node_id=cm_4
+    SELECT
+      metric_time__day
+      , every_2_days_bookers_2_days_ago
+    FROM cm_4_cte cm_4_cte
+  ) subq_23
+)
+
+SELECT
+  metric_time__day AS metric_time__day
```
Jstein77, line 59: Same comment as above where the alias is redudant

tests_metricflow/snapshots/test_custom_granularity.py/SqlQueryPlan/DuckDB/test_offset_metric_with_custom_granularity__plan0_optimized.sql
plypaul: Yes, that's the ID of the node that produces the final result.

tests_metricflow/snapshots/test_custom_granularity.py/SqlQueryPlan/DuckDB/test_offset_metric_with_custom_granularity__plan0_optimized.sql
plypaul: That's the internal ID prefix used for the dataflow node that computes metrics.

tests_metricflow/snapshots/test_custom_granularity.py/SqlQueryPlan/DuckDB/test_offset_metric_with_custom_granularity__plan0_optimized.sql
plypaul: This is the result of following the rule, if multiple metrics are computed in a query (parts of a derived query are each considered a metric computation), have each metric computation is out into a CTE. We can change the rule to get different behavior.

tests_metricflow/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/DuckDB/test_cumulative_time_offset_metric_with_time_constraint__plan0_optimized.sql
plypaul: Right now, the rule is that we always have column aliases. We could change this, but it's existing behavior.

tests_metricflow/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/DuckDB/test_cumulative_time_offset_metric_with_time_constraint__plan0_optimized.sql
plypaul: Sure, but I think following the previous comments, it seems like we need to settle on a rule for when it helps to add these CTEs.

PR #1530, opened 2024-11-14
Cleanup: Remove `extraneous_linkable_specs`

courtneyholcomb:
The difference between `queried_linkable_specs`, `required_linkable_specs`, and `extraneous_linkable_specs` has often been overlooked and therefore been the source of many bugs. In [this PR](https://github.com/dbt-labs/metricflow/pull/1506), we removed the unneeded `FilterElementsNodes` that were the main reason we needed those different spec variables. So now we can remove the `extraneous_linkable_specs` variables altogether, and hopefully reduce confusion here.
There is also some related cleanup in here. It's not a huge PR but I still recommend reviewing by commit.

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -260,7 +260,7 @@ class DataflowPlanBuilder:
         )
 
         # Build measure recipes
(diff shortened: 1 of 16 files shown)


courtneyholcomb:
Investigating the Trino error.

plypaul (approved):
The separate commits were helpful - thanks!

PR #1539, opened 2024-11-20
Minimum grain clean up

courtneyholcomb:
Clean up.
1. Rename default grain -> min grain. This was causing me some confusion because we recently introduced the concept of default grain for metrics, which is defined in the YAML spec, but this code was not referring to that.
2. Also moves some related logic to helper functions to improve readability.

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -183,7 +183,6 @@ class DataflowPlanBuilder:
             where_filter_specs=(),
             pushdown_enabled_types=frozenset({PredicateInputType.TIME_RANGE_CONSTRAINT}),
         )
-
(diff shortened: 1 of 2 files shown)


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -1613,10 +1611,6 @@ def _build_aggregated_measure_from_measure_source_node(
 
         # If querying an offset metric, join to time spine before aggregation.
         if before_aggregation_time_spine_join_description is not None:
-            assert queried_agg_time_dimension_specs, (
```
courtneyholcomb, line 1616: Removed this because the `JoinToTimeSpineNode` will raise the same error in `__post_init__` on instantiation, so we don't need both.

plypaul (approved):

PR #1540, opened 2024-11-20
Simplify dataflow to SQL logic for `JoinOverTimeRangeNode`

courtneyholcomb:
There should be no functional changes in this commit, only cleanup and readability improvements. Mostly involves moving complex logic to helper functions.

### DIFF
--- a/metricflow/dataset/sql_dataset.py
+++ b/metricflow/dataset/sql_dataset.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from dataclasses import dataclass
 from typing import List, Optional, Sequence, Tuple
 
 from dbt_semantic_interfaces.references import SemanticModelReference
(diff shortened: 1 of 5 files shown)


metricflow/dataset/sql_dataset.py
```diff
@@ -154,3 +155,25 @@ def column_association_for_time_dimension(self, time_dimension_spec: TimeDimensi
...
         return None
+
+    def annotate(self, alias: str, metric_time_spec: TimeDimensionSpec) -> AnnotatedSqlDataSet:
+        """Convert to an AnnotatedSqlDataSet with specified metadata."""
+        metric_time_column_name = self.column_association_for_time_dimension(metric_time_spec).column_name
+        return AnnotatedSqlDataSet(data_set=self, alias=alias, _metric_time_column_name=metric_time_column_name)
+
+
+@dataclass(frozen=True)
+class AnnotatedSqlDataSet:
```
courtneyholcomb, line 166: This class should be unchanged. I needed to move it to resolve circular imports.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1390,6 +1361,21 @@ def visit_semi_additive_join_node(self, node: SemiAdditiveJoinNode) -> SqlDataSe
             ),
         )
 
+    def _choose_instance_for_time_spine_join(
```
courtneyholcomb, line 1364: This helper function is only used once at this point, but it will be used again for the `JoinToTimeSpineNode` farther up the stack.

metricflow/plan_conversion/instance_converters.py
```diff
@@ -803,6 +803,7 @@ def transform(self, instance_set: InstanceSet) -> SelectColumnSet:  # noqa: D102
         )
 
 
+# TODO: delete this class & all uses. It doesn't do anything.
```
courtneyholcomb, line 806: This class is supposed to change the column names, but if the specs didn't change, then the column names shouldn't either, so it seems like it doesn't do anything. LMK if I'm overlooking something here!

plypaul (approved):
Forgot to approve earlier.

PR #1541, opened 2024-11-20
`JoinToTimeSpineNode` bug fix

courtneyholcomb:
We had previously made a product decision about the behavior of the `JoinToTimeSpineNode` that we later decided was not correct. This is specifically related to what happens if you query with both `metric_time` and the `agg_time_dimension` (e.g., `booking__ds`). If you query with just one or the other, we would use the time spine to fulfill that dimension. But if you queried with both, there was different behavior.

Previous behavior when querying with both:
- Treat `metric_time` as the time spine dimension and fulfill those values with a time spine column
- Treat `agg_time_dimension` as any other dimension, and fulfill those values from the semantic model source table

This behavior felt inconsistent, since you might get different values when querying the `agg_time_dimension` depending on if you included `metric_time` in the query or not. This could be especially confusing if you included `agg_time_dimension` only in the group by, and `metric_time` in the where filter.

New behavior when querying with both:
- Treat `metric_time` and `agg_time_dimension` the same. Fulfill both dimensions with values from the time spine.

PR notes:
- I recommend reviewing by commit.
- I tried to isolate the changes related to the bug fix, but it was not as easy as I'd hoped. Hopefully it isn't too difficult to understand the changes, but let me know if you want to sync to discuss.
- There are a bunch of SQL changes here, but only a few are in optimized snapshots. The optimized ones more clearly demonstrate the actual behavior change.

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -14,7 +14,7 @@ from dbt_semantic_interfaces.protocols.metric import (
     MetricTimeWindow,
     MetricType,
 )
-from dbt_semantic_interfaces.references import MetricReference, TimeDimensionReference
+from dbt_semantic_interfaces.references import MeasureReference, MetricReference, TimeDimensionReference
 from dbt_semantic_interfaces.type_enums.time_granularity import TimeGranularity
 from dbt_semantic_interfaces.validations.unique_valid_name import MetricFlowReservedKeywords
 from metricflow_semantics.dag.id_prefix import StaticIdPrefix
@@ -653,9 +653,6 @@ class DataflowPlanBuilder:
             queried_agg_time_dimension_specs = queried_linkable_specs.included_agg_time_dimension_specs_for_metric(
                 metric_reference=metric_spec.reference, metric_lookup=self._metric_lookup
             )
-            assert (
-                queried_agg_time_dimension_specs
-            ), "Joining to time spine requires querying with metric_time or the appropriate agg_time_dimension."
             output_node = JoinToTimeSpineNode.create(
                 parent_node=output_node,
                 requested_agg_time_dimension_specs=queried_agg_time_dimension_specs,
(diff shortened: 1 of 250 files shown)


tests_metricflow/snapshots/test_cumulative_metric_rendering.py/SqlQueryPlan/DuckDB/test_window_metric_with_non_default_grains__plan0_optimized.sql
```diff
@@ -8,28 +8,28 @@ sql_engine: DuckDB
 ---
 -- Re-aggregate Metric via Group By
 SELECT
-  booking__ds__month
-  , metric_time__week
+  metric_time__week
```
courtneyholcomb, line 11: just order change

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -646,9 +646,6 @@ def _build_derived_metric_output_node(
             queried_agg_time_dimension_specs = queried_linkable_specs.included_agg_time_dimension_specs_for_metric(
                 metric_reference=metric_spec.reference, metric_lookup=self._metric_lookup
             )
-            assert (
```
courtneyholcomb, line 649: This error is handled in the `__post_init__` for `JoinToTimeSpineNode`, so I removed the duplication.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -1380,33 +1378,30 @@ def visit_join_to_time_spine_node(self, node: JoinToTimeSpineNode) -> SqlDataSet
...
-        agg_time_dimension_instances: List[TimeDimensionInstance] = []
-        for instance in parent_data_set.instance_set.time_dimension_instances:
-            if (
-                instance.spec.date_part is None  # Ensure we don't join using an instance with date part
-                and instance.spec.element_name == agg_time_element_name
-                and instance.spec.entity_links == agg_time_entity_links
-            ):
-                agg_time_dimension_instances.append(instance)
+        # Select the dimension for the join from the parent node because it may not have been included in the request.
+        # Default to using metric_time for the join if it was requested, otherwise use the agg_time_dimension.
```
courtneyholcomb, line 1386: This is the one place we still have some differentiation between `metric_time` and `agg_time`. If you request both, we have to choose one of them to use as the join column, and we default to using `metric_time`.

tests_metricflow/snapshots/test_derived_metric_rendering.py/SqlQueryPlan/DuckDB/test_offset_to_grain_metric_filter_and_query_have_different_granularities__plan0_optimized.sql
```diff
@@ -19,8 +19,8 @@ FROM (
   FROM (
     -- Join to Time Spine Dataset
     SELECT
-      subq_12.ds AS metric_time__day
-      , DATE_TRUNC('month', subq_12.ds) AS metric_time__month
+      DATE_TRUNC('month', subq_12.ds) AS metric_time__month
+      , subq_10.metric_time__day AS metric_time__day
```
courtneyholcomb, line 23: You might notice on this line we're selecting `metric_time__day` from the parent instead of the time spine. This is fine because the column is never used beyond the join, and should be equivalent to the time spine column since it's the one used in the join.
This is true in a couple of snapshots. It's not a problem functionally but might be confusing to some users reading the SQL. This gets fixed farther up the stack.

plypaul (approved):

PR #1562, opened 2024-12-10
Create `SqlAddTimeExpression`

courtneyholcomb:
Adds a new SQL expression that adds a time interval to a date/time expression. This differs from the existing time delta expression because it 1) adds instead of subtracts, and 2) accepts a SQL expression for the count instead of an integer. This will be used to build the SQL for custom granularity offset windows.

### DIFF
--- a/metricflow/sql/render/duckdb_renderer.py
+++ b/metricflow/sql/render/duckdb_renderer.py
@@ -37,7 +37,7 @@ class DuckDbSqlExpressionRenderer(DefaultSqlExpressionRenderer):
         }
 
     @override
-    def visit_time_delta_expr(self, node: SqlSubtractTimeIntervalExpression) -> SqlExpressionRenderResult:
+    def visit_subtract_time_interval_expr(self, node: SqlSubtractTimeIntervalExpression) -> SqlExpressionRenderResult:
         """Render time delta expression for DuckDB, which requires slightly different syntax from other engines."""
         arg_rendered = node.arg.accept(self)
 
--- a/metricflow/sql/render/duckdb_renderer.py
+++ b/metricflow/sql/render/duckdb_renderer.py
@@ -15,6 +15,7 @@ from metricflow.sql.render.expr_renderer import (
 )
 from metricflow.sql.render.sql_plan_renderer import DefaultSqlQueryPlanRenderer
 from metricflow.sql.sql_exprs import (
+    SqlAddTimeExpression,
     SqlGenerateUuidExpression,
     SqlPercentileExpression,
     SqlPercentileFunctionType,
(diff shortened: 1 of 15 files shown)


metricflow/sql/render/trino.py
```diff
@@ -45,7 +46,7 @@ def visit_generate_uuid_expr(self, node: SqlGenerateUuidExpression) -> SqlExpres
         )
 
     @override
-    def visit_time_delta_expr(self, node: SqlSubtractTimeIntervalExpression) -> SqlExpressionRenderResult:
+    def visit_subtract_time_interval_expr(self, node: SqlSubtractTimeIntervalExpression) -> SqlExpressionRenderResult:
```
WilliamDee, line 49: Would it have been more intuitive if we just made it so that we only had `visit_time_delta` and the direction just can be done via supplying it a positive or negative count?

metricflow/sql/sql_exprs.py
```diff
@@ -1314,6 +1320,65 @@ def matches(self, other: SqlExpressionNode) -> bool:  # noqa: D102
         return self.count == other.count and self.granularity == other.granularity and self._parents_match(other)
 
 
+@dataclass(frozen=True, eq=False)
+class SqlAddTimeExpression(SqlExpressionNode):
```
WilliamDee, line 1324: Would be good if we added some tests for this with the different engine rendering

WilliamDee (approved):
Nice! Noted down a few things

metricflow/sql/sql_exprs.py
courtneyholcomb: I'll have some up the stack! But it felt very annoying to add those tests without having the logic to render them in a full SQL query yet.

metricflow/sql/render/trino.py
WilliamDee: ooo gotcha!

metricflow/sql/sql_exprs.py
WilliamDee: Ahh i see. Though, you should be able to test this node in isolation here https://github.com/dbt-labs/metricflow/blob/main/tests_metricflow/sql/test_engine_specific_rendering.py, but if you already have these tests up the stack on this, totally fine too!

metricflow/sql/render/duckdb_renderer.py
```diff
@@ -52,6 +53,22 @@ def visit_subtract_time_interval_expr(self, node: SqlSubtractTimeIntervalExpress
...
         )
 
+    @override
+    def visit_add_time_expr(self, node: SqlAddTimeExpression) -> SqlExpressionRenderResult:
+        """Render time delta expression for DuckDB, which requires slightly different syntax from other engines."""
+        arg_rendered = node.arg.accept(self)
+        count_rendered = node.count_expr.accept(self).sql
+
+        granularity = node.granularity
+        if granularity == TimeGranularity.QUARTER:
```
plypaul, line 63: Nit: compare using `is` as per PEP 8. Also for other cases in the PR.

metricflow/sql/render/duckdb_renderer.py
```diff
@@ -52,6 +53,22 @@ def visit_subtract_time_interval_expr(self, node: SqlSubtractTimeIntervalExpress
...
+            granularity = TimeGranularity.MONTH
+            count_rendered = f"{count_rendered} * 3"
```
plypaul, line 65: In case `count_rendered` is an expression, we would need `()` for it. Also for other cases in the PR.

plypaul (approved):
LGTM but please see comments.

metricflow/sql/sql_exprs.py
courtneyholcomb: Good idea I'll add those.

PR #1565, opened 2024-12-11
Add marker for DuckDb-only tests

plypaul:
This PR adds a `pytest` marker that allows tests to only be run when testing using DuckDB. This is useful in integration test cases where engine-specific SQL rendering is not being tested. This can help to reduce test suite runtimes and to reduce the number of snapshots.


WilliamDee (commented):
should we put a backlog to go through the tests to add this marker whereever necessary as a clean up?

plypaul:
Yeah, that would be ideal, but it would be a lot to go through all the tests. I figure we start by updating as we go at first.

PR #1566, opened 2024-12-11
Add improved test for the SQL optimization level in the request

plypaul:
This PR adds an improved test to show that the SQL optimization level in the request is handled correctly in the MF engine API. The previous test made a more general check, but a snapshot helps to check for specific features. Considering adding an assertion for the snapshot as well.

### DIFF
--- a/metricflow/sql/optimizer/optimization_levels.py
+++ b/metricflow/sql/optimizer/optimization_levels.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import functools
 from dataclasses import dataclass
 from enum import Enum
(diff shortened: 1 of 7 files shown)


metricflow/sql/optimizer/optimization_levels.py
```diff
@@ -13,6 +14,7 @@
 from metricflow.sql.optimizer.table_alias_simplifier import SqlTableAliasSimplifier
 
 
+@functools.total_ordering
```
WilliamDee, line 17: wow, didn't know about this

WilliamDee (approved):

PR #1573, opened 2024-12-13
Add support for querying with metric alias

serramatutu:
## Summary

This PR introduces support for querying with aliased metrics. This is done by adding an `AliasSpecsNode` after the order by step in `DataflowPlanBuilder`. To make the alias information reach that point, I also had to modify some of the query interface classes to allow aliases in metrics.

You can review commit by commit.


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -893,6 +897,16 @@ def build_sink_node(
                 parent_node=pre_result_node or parent_node, include_specs=output_selection_specs
             )
 
+        alias_specs = tuple(
+            SpecToAlias(MetricSpec(metric.element_name), MetricSpec(metric.alias))
```
courtneyholcomb, line 901: For the spec with the alias, you should be able to just leave it as `metric` instead of creating a new metric spec because it will still be rendered with the alias.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -161,7 +161,6 @@ def _build_query_output_node(
                 len(metric_spec.filter_spec_set.all_filter_specs) > 0
                 or metric_spec.offset_to_grain is not None
                 or metric_spec.offset_window is not None
-                or metric_spec.alias is not None
```
courtneyholcomb, line 164: We do need to keep this line in - adding some other tests would demonstrate why. The SQL rendering tests are most important here. Can you add a few more tests?
1. A SQL rendering test for the simple case (similar to the integration test you added)
2. A SQL rendering test with an order by and where filter on the aliased metric (using the original metric name)
3. An integration test for the same scenario as 2
4. A SQL rendering test for a derived metric query with a metric alias
5. An integration test for the same scenario as 4

I think the new integration tests added above will fail because we'll render the alias in the subqueries, not just the outer query. So we need to remove the alias before passing it any of the inner nodes.

metricflow-semantics/metricflow_semantics/query/query_resolver.py
```diff
@@ -196,6 +197,18 @@ def _resolve_metric_inputs(
                     )
                 )
             else:
```
courtneyholcomb, line 199: Would be good to reverse this conditional for readability. i.e.,
```
if len(matching_specs) == 1:
  <your code>
else:
  <error code>
```
You can also pull out the one matching spec from the list instead of using a for loop unnecessarily, since we know at this point there's only one matching spec.

metricflow-semantics/metricflow_semantics/query/query_resolver.py
```diff
@@ -196,6 +197,18 @@ def _resolve_metric_inputs(
...
+                # Add the desired alias to the spec if it's provided in the input object
+                if (
+                    isinstance(metric_input.input_obj, MetricQueryParameter)
```
courtneyholcomb, line 202: So, we have generally tried to avoid putting any `isinstance` checks in MF. This is mostly a Paul thing, he tends to think they indicate a problem with the classes themselves if we need to add a check like this. Instead we would want to do something here that's similar to how `GroupByParamaters` are handled. It's a bit more complicated (I'm reading the code and haven't totally figured out the solution yet). I definitely don't think it's worth you trying to figure this out before you leave for vacation so I can pick up this PR from you!

courtneyholcomb (commented):
Thanks for handling this!! Dataflow plan looks good. There are some changes that need to be made to the other parts of the code, and I don't think you should try to make them before you leave for vacation, so I can pick this up!

tests_metricflow/snapshots/test_engine_specific_rendering.py/SqlPlan/BigQuery/test_add_time_expr__plan0.sql
```diff
@@ -0,0 +1,10 @@
+test_name: test_add_time_expr
```
plypaul, line 1: Resolved this in https://github.com/dbt-labs/metricflow/pull/1580, so these should go away if you regenerate the snapshots.

tests_metricflow/integration/query_output/test_query_output.py
```diff
@@ -0,0 +1,63 @@
...
+@pytest.mark.sql_engine_snapshot
+def test_derived_metric_alias(  # noqa: D103
+    request: FixtureRequest,
+    mf_test_configuration: MetricFlowTestConfiguration,
+    sql_client: SqlClient,
+    it_helpers: IntegrationTestHelpers,
+) -> None:
+    query_result = it_helpers.mf_engine.query(
+        MetricFlowQueryRequest.create_with_random_request_id(
+            metrics=(MetricParameter(name="booking_fees", alias="bookings_alias"),),
```
plypaul, line 49: To avoid confusion with the previous test case, how about a different alias? e.g. `booking_fees_alias`

tests_metricflow/query_rendering/test_query_rendering.py
```diff
@@ -603,3 +607,61 @@ def test_non_additive_dimension_with_non_default_grain(
         dataflow_plan_builder=dataflow_plan_builder,
         query_spec=query_spec,
     )
+
+
+@pytest.mark.sql_engine_snapshot
```
plypaul, line 612: Not for this PR, but maybe we should consolidate `tests_metricflow/integration/query_output/test_query_output.py` and `tests_metricflow/query_rendering/test_query_rendering.py`.

tests_metricflow/integration/query_output/test_query_output.py
```diff
@@ -0,0 +1,63 @@
...
+from metricflow_semantics.specs.query_param_implementations import MetricParameter
+from metricflow_semantics.test_helpers.config_helpers import MetricFlowTestConfiguration
+
+from metricflow.engine.metricflow_engine import MetricFlowQueryRequest
+from metricflow.protocols.sql_client import SqlClient
+from tests_metricflow.integration.conftest import IntegrationTestHelpers
+from tests_metricflow.snapshot_utils import assert_str_snapshot_equal
+
+
+@pytest.mark.sql_engine_snapshot
```
plypaul, line 14: Since aliases are consistent across SQL-engines, you can add `@pytest.mark.duckdb_only` to reduce the number of snapshots. Also applies to other tests.

plypaul (commented):
* Since the alias in this case is describing user inputs, can it be added as a field to `ResolverInputForMetric`?
* If not, need to have a think about the placement as the spec pattern is more about identifying items in the semantic manifest.
* Can you add a test case for when the user specifies two aliases that are the same in a query?

courtneyholcomb (approved):
LGTM!

PR #1575, opened 2024-12-18
Add SQL exprs needed for custom offset window

courtneyholcomb:
This includes a `CASE` expression, an integer expression, an arithmetic expression, and some updates to the add time expression & window function expressions. These are all needed to build the following SQL column:
```
CASE
  WHEN ds__martian_day__first_value__offset + INTERVAL (ds__day__row_number - 1) day <= ds__martian_day__last_value__offset
    THEN ds__martian_day__first_value__offset + INTERVAL (ds__day__row_number - 1) day
  ELSE ds__martian_day__last_value__offset
  END AS metric_time__day
```

Note that the first commit is moving the `sql_exprs` file into `metricflow-semantics`, which is needed for a commit further up the stack.


metricflow-semantics/metricflow_semantics/sql/sql_exprs.py
```diff
@@ -1719,3 +1793,158 @@ def lineage(self) -> SqlExpressionTreeLineage:  # noqa: D102
...
+    """Arithmetic operator used to do math in a SQL expression."""
+
+    ADD = "+"
+    SUBTRACT = "-"
+    MULTIPLY = "*"
+    DIVIDE = "/"
+
+
+@dataclass(frozen=True, eq=False)
+class SqlArithmeticExpression(SqlExpressionNode):
```
WilliamDee, line 1898: For potentially cleanup, seems like this can replace `SqlRatioComputationExpression` unless we want to explicitly have a node for that. But if we wanted to do that, i don't see why we wouldn't want to do it for other operations.

metricflow-semantics/metricflow_semantics/sql/sql_exprs.py
```diff
@@ -1719,3 +1793,158 @@ def lineage(self) -> SqlExpressionTreeLineage:  # noqa: D102
...
+    else_expr: Optional[SqlExpressionNode]
+
+    @staticmethod
+    def create(  # noqa: D102
+        when_to_then_exprs: Dict[SqlExpressionNode, SqlExpressionNode], else_expr: Optional[SqlExpressionNode] = None
+    ) -> SqlCaseExpression:
+        parent_nodes: Tuple[SqlExpressionNode, ...] = ()
+        for when, then in when_to_then_exprs.items():
+            parent_nodes += (when,)
+            parent_nodes += (then,)
```
WilliamDee, line 1826: nit: don't have to do it 
```py
parent_nodes += (when, then)
```

WilliamDee (approved):
Looks good! I'm not super familiar with using case when exprs, so I want to defer to @plypaul for that. But the other 2 looks good!

plypaul (approved):

metricflow-semantics/metricflow_semantics/sql/sql_exprs.py
courtneyholcomb: I considered that but decided not to because the ratio expression has a bunch of extra logic related to casting & `NULLIF`, and I wasn't sure we would always want that for this node!

PR #1577, opened 2024-12-18
Fix bug in `JoinToTimeSpine` dataflow plans

courtneyholcomb:
We weren't tracking the parent nodes properly, which resulted in improper optimization and nodes missing when displaying the plan. This should not impact the output data, but will hopefully improve query efficiency now that more CTEs are enabled.

### DIFF
--- a/tests_metricflow/dataflow/optimizer/test_predicate_pushdown_optimizer.py
+++ b/tests_metricflow/dataflow/optimizer/test_predicate_pushdown_optimizer.py
@@ -325,6 +325,7 @@ def test_aggregate_output_join_metric_predicate_pushdown(
     )
 
 
+@pytest.mark.skip("Predicate pushdown is not implemented for some of the nodes in this plan")
 def test_offset_metric_predicate_pushdown(
     request: FixtureRequest,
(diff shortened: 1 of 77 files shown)


tests_metricflow/dataflow/optimizer/test_predicate_pushdown_optimizer.py
```diff
@@ -325,6 +325,7 @@ def test_aggregate_output_join_metric_predicate_pushdown(
     )
 
 
+@pytest.mark.skip("Predicate pushdown is not implemented for some of the nodes in this plan")
```
WilliamDee, line 328: were these passing before this change 🤔

WilliamDee (approved):
nice!

tests_metricflow/dataflow/optimizer/test_predicate_pushdown_optimizer.py
courtneyholcomb: Yes! The optimizers weren't correctly going through all the parent nodes, so this was never hitting the `AliasSpecsNode`. That node is new and doesn't have an implementation for the pushdown optimizer.

PR #1579, opened 2024-12-18
Add new dataflow plan node for custom offset windows

courtneyholcomb:
Adds a new node called `OffsetByCustomGranularityNode`. This will be used to build the SQL for custom offset windows.

### DIFF
--- a/metricflow/dataflow/nodes/offset_by_custom_granularity.py
+++ b/metricflow/dataflow/nodes/offset_by_custom_granularity.py
@@ -1,0 +1,75 @@
+from __future__ import annotations
+
+from abc import ABC
+from dataclasses import dataclass
+from typing import Sequence
+
+from dbt_semantic_interfaces.protocols.metric import MetricTimeWindow
+from metricflow_semantics.dag.id_prefix import IdPrefix, StaticIdPrefix
+from metricflow_semantics.dag.mf_dag import DisplayedProperty
+from metricflow_semantics.specs.time_dimension_spec import TimeDimensionSpec
+from metricflow_semantics.visitor import VisitorOutputT
+
+from metricflow.dataflow.dataflow_plan import DataflowPlanNode
+from metricflow.dataflow.dataflow_plan_visitor import DataflowPlanNodeVisitor
+
+
+@dataclass(frozen=True, eq=False)
+class OffsetByCustomGranularityNode(DataflowPlanNode, ABC):
+    """For a given custom grain, offset its base grain by the requested number of custom grain periods."""
+
+    offset_window: MetricTimeWindow
+    required_time_spine_specs: Sequence[TimeDimensionSpec]
+    time_spine_node: DataflowPlanNode
+
+    def __post_init__(self) -> None:  # noqa: D105
+        super().__post_init__()
+
+    @staticmethod
+    def create(  # noqa: D102
+        time_spine_node: DataflowPlanNode,
+        offset_window: MetricTimeWindow,
+        required_time_spine_specs: Sequence[TimeDimensionSpec],
+    ) -> OffsetByCustomGranularityNode:
+        return OffsetByCustomGranularityNode(
+            parent_nodes=(time_spine_node,),
+            time_spine_node=time_spine_node,
+            offset_window=offset_window,
+            required_time_spine_specs=required_time_spine_specs,
+        )
+
+    @classmethod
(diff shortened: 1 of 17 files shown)


metricflow/dataflow/nodes/offset_by_custom_granularity.py
```diff
@@ -0,0 +1,95 @@
...
+            and other_node.required_time_spine_specs == self.required_time_spine_specs
+        )
+
+    def with_new_parents(  # noqa: D102
+        self, new_parent_nodes: Sequence[DataflowPlanNode]
+    ) -> OffsetByCustomGranularityNode:
+        custom_granularity_bounds_node: Optional[CustomGranularityBoundsNode] = None
+        filter_elements_node: Optional[FilterElementsNode] = None
+        for parent_node in new_parent_nodes:
+            if isinstance(parent_node, CustomGranularityBoundsNode):
```
plypaul, line 80: The combined node should allow this to be removed.

plypaul (commented):
Summarizing a chat:

* Combining the two nodes into a single node would be more useful and allow for a more straightforward functional description.
* However, that is blocked on some updates to the optimizer, which I'll make. @courtneyholcomb is planning to refactor once that change is in.
* Rows are not duplicated in the join with the `CASE` statement when computing the offset. However, if the time interval has a different number of days, (e.g. quarters with different days in each), the last day could be an aggregation of multiple days. @courtneyholcomb is verifying that this is inline with product expectations.

Other notes:

What helped during review was an example of the SQL that the nodes would generate along with a few rows that show the respective values, especially for a boundary.

Could you include a test that includes those using a minimal DAG? `pytest.mark.duckdb_only` makes sense to use given that more extensive tests are present later in the stack.

metricflow/dataset/sql_dataset.py
```diff
@@ -181,18 +183,30 @@ def instance_for_column_name(self, column_name: str) -> MdoInstance:
...
-                and time_dimension_instance.spec.date_part == time_dimension_spec.date_part
+                time_dimension_instance.spec.time_granularity_name == time_granularity_name
+                and time_dimension_instance.spec.date_part == date_part
+                and not time_dimension_instance.spec.window_functions
             ):
                 return time_dimension_instance
 
         raise RuntimeError(
-            f"Did not find a time dimension instance with matching grain and date part for spec: {time_dimension_spec}\n"
+            f"Did not find a time dimension instance with grain '{time_granularity_name}' and date part {date_part}\n"
```
plypaul, line 198: This error would be easier to read via `str(LazyFormat(...))` or `mf_pformat_dict(...)`. Same for other cases.

metricflow/sql/sql_plan.py
```diff
@@ -102,6 +102,25 @@ class SqlSelectColumn:
     # Always require a column alias for simplicity.
     column_alias: str
 
+    @staticmethod
+    def from_table_and_column_names(table_alias: str, column_name: str) -> SqlSelectColumn:
```
plypaul, line 106: To avoid an `and` in the name, how about `from_column_reference`?

metricflow/sql/sql_plan.py
```diff
@@ -102,6 +102,25 @@ class SqlSelectColumn:
...
+        """Create a column that selects a column from a table by name."""
+        return SqlSelectColumn(
+            expr=SqlColumnReferenceExpression.from_table_and_column_names(
+                column_name=column_name, table_alias=table_alias
+            ),
+            column_alias=column_name,
+        )
+
+    def ref_with_new_table_alias(self, new_table_alias: str) -> SqlColumnReferenceExpression:
```
plypaul, line 115: It was a little unclear what this meant initially. How about something like:

```
def reference_from(self, source_table_alias: str) -> SqlColumnReferenceExpression:
```

Or `source_column_reference`

plypaul (commented):
Generally LGTM, but checking if you saw the comment about the minimal DAG test?

courtneyholcomb:
> Generally LGTM, but checking if you saw the comment about the minimal DAG test?

@plypaul ah sorry missed that! Will add

courtneyholcomb:
@plypaul implemented your feedback above.

For the dataflow to SQL test, I ended up adding it in the next PR up the stack so that I could use the helper method there. See that commit [here](https://github.com/dbt-labs/metricflow/pull/1584/commits/0938ede776886367cc4e42fe8faeafff670b74f7).

plypaul (approved):

PR #1584, opened 2024-12-18
Build DataflowPlan for custom offset window

courtneyholcomb:
This is the dataflow plan that will be used if the custom grain is queried with any grains that aren't the same as the grain used in the offset window. There will be a simpler dataflow plan used for queries using only the same grain as what's used in the offset window.

### DIFF
--- a/metricflow/dataflow/builder/dataflow_plan_builder.py
+++ b/metricflow/dataflow/builder/dataflow_plan_builder.py
@@ -92,6 +92,7 @@ from metricflow.dataflow.nodes.join_to_custom_granularity import JoinToCustomGra
 from metricflow.dataflow.nodes.join_to_time_spine import JoinToTimeSpineNode
 from metricflow.dataflow.nodes.metric_time_transform import MetricTimeDimensionTransformNode
 from metricflow.dataflow.nodes.min_max import MinMaxNode
+from metricflow.dataflow.nodes.offset_by_custom_granularity import OffsetByCustomGranularityNode
 from metricflow.dataflow.nodes.order_by_limit import OrderByLimitNode
 from metricflow.dataflow.nodes.read_sql_source import ReadSqlSourceNode
 from metricflow.dataflow.nodes.semi_additive_join import SemiAdditiveJoinNode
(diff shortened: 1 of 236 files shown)


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -1935,3 +1997,9 @@ def _determine_time_spine_join_spec(
                 time_granularity=join_spec_grain, date_part=None
             )
         return join_on_time_dimension_spec
+
+    def _offset_window_is_custom(self, offset_window: Optional[MetricTimeWindow]) -> bool:
+        return (
+            offset_window is not None
+            and offset_window.granularity in self._semantic_model_lookup.custom_granularity_names
```
plypaul, line 2004: What happens when the granularity is not in `custom_granularity_names`? Is this handled through a validation?

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -658,13 +661,18 @@ def _build_derived_metric_output_node(
...
+                queried_time_spine_specs=queried_agg_time_dimension_specs,
+                offset_window=metric_spec.offset_window,
+            )
             output_node = JoinToTimeSpineNode.create(
                 metric_source_node=output_node,
                 time_spine_node=time_spine_node,
                 requested_agg_time_dimension_specs=queried_agg_time_dimension_specs,
                 join_on_time_dimension_spec=self._sort_by_base_granularity(queried_agg_time_dimension_specs)[0],
-                offset_window=metric_spec.offset_window,
+                offset_window=(
```
plypaul, line 673: Are there cases where you can use a `JoinToTimeSpineNode` with a custom offset window? If not, an assertion in the class would help.

plypaul (approved):

metricflow/dataflow/builder/dataflow_plan_builder.py
courtneyholcomb: In that case it would be a standard granularity! If neither were true, it would raise an error in validations.

metricflow/dataflow/builder/dataflow_plan_builder.py
courtneyholcomb: Sure, I'll add an assertion to the the `__post_init__` for `JoinToTimeSpineNode`!

PR #1585, opened 2024-12-18
Tests for custom offset window

courtneyholcomb:
Tests for custom offset window. The snapshots in this PR should give you a better idea of what the SQL and output should look like.

### DIFF
--- a/tests_metricflow/snapshots/test_custom_granularity.py/SqlPlan/DuckDB/test_custom_offset_window__plan0_optimized.sql
+++ b/tests_metricflow/snapshots/test_custom_granularity.py/SqlPlan/DuckDB/test_custom_offset_window__plan0_optimized.sql
@@ -1,0 +1,84 @@
+test_name: test_custom_offset_window
+test_filename: test_custom_granularity.py
+sql_engine: DuckDB
+---
+-- Compute Metrics via Expressions
+SELECT
+  metric_time__day
+  , bookings AS bookings_offset_one_martian_day
+FROM (
+  -- Join to Time Spine Dataset
+  -- Pass Only Elements: ['bookings', 'metric_time__day']
+  -- Aggregate Measures
+  -- Compute Metrics via Expressions
+  SELECT
+    subq_17.ds__day__lead AS metric_time__day
+    , SUM(subq_13.bookings) AS bookings
+  FROM (
+    -- Offset Base Granularity By Custom Granularity Period(s)
+    WITH cte_6 AS (
+      -- Read From Time Spine 'mf_time_spine'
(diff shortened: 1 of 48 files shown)


courtneyholcomb:
Note if reviewing now - I need to update the check query tests one more time to get these tests passing.

tests_metricflow/snapshots/test_offset_metrics.py/str/DuckDB/test_custom_offset_window_with_base_grain__query_output.txt
```diff
@@ -0,0 +1,21 @@
+test_name: test_custom_offset_window_with_base_grain
+test_filename: test_offset_metrics.py
+docstring:
+  Gives a side by side comparison of bookings and bookings_offset_one_martian_day.
+---
+metric_time__martian_day    metric_time__day     bookings    bookings_offset_one_martian_day
```
plypaul, line 6: * With how `martian_day` is defined, it's a little hard to check these snapshots for correctness. 
* A more straightforward mapping would help e.g. martian days are always 10 years before.
* Also time on Mars is represented differently externally, so a different name would help.

plypaul (approved):
LGTM though please see comment about `martian_time`.

tests_metricflow/snapshots/test_offset_metrics.py/str/DuckDB/test_custom_offset_window_with_base_grain__query_output.txt
courtneyholcomb: Since it's not totally relevant to this PR, I'm going to update this at the top of the stack if you don't mind!

tests_metricflow/snapshots/test_custom_granularity.py/SqlPlan/DuckDB/test_custom_offset_window__plan0_optimized.sql
```diff
@@ -0,0 +1,84 @@
...
+  -- Join to Time Spine Dataset
+  -- Pass Only Elements: ['bookings', 'metric_time__day']
+  -- Aggregate Measures
+  -- Compute Metrics via Expressions
+  SELECT
+    subq_17.ds__day__lead AS metric_time__day
+    , SUM(subq_13.bookings) AS bookings
+  FROM (
+    -- Offset Base Granularity By Custom Granularity Period(s)
+    WITH cte_6 AS (
```
courtneyholcomb, line 19: @plypaul is there a way to get this CTE to the top level even though it was built in a subquery?

PR #1586, opened 2024-12-19
Don't reduce verbose SQL expressions

courtneyholcomb:
Prevents the optimizer from reducing expressions that can be very hard to read when collapsed into a second expression. They are especially hard to read when they are repeated in multiple columns in the child expression.


WilliamDee (approved):
any of these logics needs to be added to this file too? https://github.com/dbt-labs/metricflow/blob/main/metricflow/sql/optimizer/sub_query_reducer.py

PR #1592, opened 2025-01-13
Poc clickhouse

thiagosalvatore:


metricflow/sql/render/clickhouse.py
```diff
@@ -0,0 +1,239 @@
...
+    def expr_renderer(self) -> SqlExpressionRenderer:
+        return self.EXPR_RENDERER
+
+    @override
+    def _render_adapter_specific_flags(self) -> Optional[SqlPlanRenderResult]:
+        """Add ClickHouse-specific query settings."""
+        settings = ["allow_experimental_join_condition = 1", "allow_experimental_analyzer = 1", "join_use_nulls = 0"]
+        return SqlPlanRenderResult(sql=f"SETTINGS {', '.join(settings)}", bind_parameter_set=SqlBindParameterSet())
+
+    def _render_joins_section(self, join_descriptions: Sequence[SqlJoinDescription]) -> Optional[SqlPlanRenderResult]:
```
thiagosalvatore, line 180: I don't link this at all.

The problem is that clickhouse doesn't support inequality INNER JOINS

https://clickhouse.com/docs/en/sql-reference/statements/select/join#join-with-inequality-conditions-for-columns-from-different-tables.

The suggested approach is to use CROSS JOIN instead, but it forces a change on the rendering logic of any join that uses inequality.

The tests are passing, but I'm open for suggestions here

metricflow/sql/render/sql_plan_renderer.py
```diff
@@ -322,6 +322,9 @@ def _render_limit_section(self, limit_value: Optional[int]) -> Optional[SqlPlanR
             return None
         return SqlPlanRenderResult(sql=f"LIMIT {limit_value}", bind_parameter_set=SqlBindParameterSet())
 
+    def _render_adapter_specific_flags(self) -> Optional[SqlPlanRenderResult]:
```
thiagosalvatore, line 325: clickhouse has some flags that can be enabled on a query level. Instead of adding a bunch of workarounds to the statement execution I decided to add it here so if we have other adapters with similar features we can just use it.

tests_metricflow/fixtures/sql_clients/adapter_backed_ddl_client.py
```diff
@@ -46,11 +46,19 @@ def create_table_from_data_table(
                 # Format as "column_name column_type"
                 columns_to_insert.append(f"{column_description.column_name} {self._get_sql_type(column_description)}")
 
+            create_table_statement = f"CREATE TABLE IF NOT EXISTS {sql_table.sql} ({', '.join(columns_to_insert)})"
+
+            if self.sql_engine_type == SqlEngine.CLICKHOUSE:
+                create_table_statement = (
+                    f"{create_table_statement} ENGINE = MergeTree ORDER BY ({columns_to_insert[0].split(" ")[0]})"
```
thiagosalvatore: clickhouse enforces that every table being created needs to specify what is the engine

tests_metricflow/fixtures/sql_clients/adapter_backed_ddl_client.py
```diff
@@ -132,8 +140,14 @@ def _quote_escape_value(self, value: str) -> str:
...
-        self.execute(f"CREATE SCHEMA IF NOT EXISTS {schema_name}")
+        if self.sql_engine_type is SqlEngine.CLICKHOUSE:
+            self.execute(f"CREATE DATABASE IF NOT EXISTS {schema_name}")
+        else:
+            self.execute(f"CREATE SCHEMA IF NOT EXISTS {schema_name}")
 
     def drop_schema(self, schema_name: str, cascade: bool = True) -> None:
         """Drop the given schema from the data warehouse. Only used in tests."""
-        self.execute(f"DROP SCHEMA IF EXISTS {schema_name}{' CASCADE' if cascade else ''}")
+        if self.sql_engine_type is SqlEngine.CLICKHOUSE:
```
thiagosalvatore, line 150: there's no concept of schema on clickhouse. A schema is a database

metricflow/sql/render/clickhouse.py
```diff
@@ -0,0 +1,239 @@
...
+    @override
+    def supported_percentile_function_types(self) -> Collection[SqlPercentileFunctionType]:
+        return {
+            SqlPercentileFunctionType.CONTINUOUS,
+            SqlPercentileFunctionType.DISCRETE,
+            SqlPercentileFunctionType.APPROXIMATE_CONTINUOUS,
+        }
+
+    @override
+    def render_date_part(self, date_part: DatePart) -> str:
```
courtneyholcomb, line 57: I'm guessing this will need to be moved to `render_extract()`?

metricflow/sql/render/clickhouse.py
courtneyholcomb: We'll definitely want to discuss this at the MF sync tomorrow. The problem is that a SQL plan theoretically should be engine-agnostic. In this case, it isn't, since the `INNER JOIN` in the SQL plan is invalid for Clickhouse.

A couple of other potential options:
- We change the SQL plan to use `CROSS JOIN`, which I'm assuming would work for all engines. Then, we could implement an engine-specific optimizer that swaps in an `INNER JOIN` if the engine supports it. In practice, if that optimizer errors for some reason, the user might be surprised to see a `CROSS JOIN` here when it's not necessary. We might be ok with that tradeoff.
- Another option would be for us to just create an internal concept of a generic "inequality join" or something similar to use in the SQL plan. And that would get translated into either `INNER JOIN` or `CROSS JOIN` by the SQL renderer.
Both of those options might have issues integrating with the `WHERE` clause, though, since we would need to include that when `CROSS JOIN` is used 🤔 

Let's hold off on making any changes here for now - I want to see if Paul has any better ideas for how to work around this.

tests_metricflow/fixtures/sql_clients/adapter_backed_ddl_client.py
```diff
@@ -132,8 +140,14 @@ def _quote_escape_value(self, value: str) -> str:
 
     def create_schema(self, schema_name: str) -> None:
         """Create the given schema in a data warehouse. Only used in tutorials and tests."""
-        self.execute(f"CREATE SCHEMA IF NOT EXISTS {schema_name}")
+        if self.sql_engine_type is SqlEngine.CLICKHOUSE:
```
courtneyholcomb, line 143: We should add this to the renderers so that we don't need any logic here to check the engine type (especially since it's used twice). You can do similar to what we do on line 125 above to get the string needed for `timestamp_data_type`. You can add a property to the renderer called like `schema_str` or something like that that just returns `"SCHEMA"` or `"DATABASE"`.

tests_metricflow/generate_snapshots.py
```diff
@@ -32,6 +32,10 @@
         "engine_url": trino://...",
         "engine_password": "..."
     },
+    "clickhouse": {
```
courtneyholcomb, line 35: There's a 1pass entry that stores these creds for easy access, so it would be great if we could update that too!

courtneyholcomb (commented):
It's awesome how quickly you were able to figure all this out with so little context! 🙌

tests_metricflow/fixtures/sql_clients/adapter_backed_ddl_client.py
thiagosalvatore: I'm wondering if it makes sense to move this to the renderer given that we only create/drop schemas on tests.

Maybe we can turn this into a property inside the renderer that looks like:

`has_schema_support` that by default is True and then when this value is false we use the create database instead. WDYT?

courtneyholcomb: Sure that works too!

PR #1593, opened 2025-01-13
Bug fix: don't erase order by for metric alias query

courtneyholcomb:
Previously, when applying the `SqlRewritingSubQueryReducer`, we would allow the child order bys to overwrite the parent order bys. This was never a problem because order bys were only present in the outermost SQL query (due to our existing dataflow plans). With the recent change to allow metric aliases, order bys may now be in a parent subquery. This revealed the bug in which they were being erased. This PR fixes that issue.


WilliamDee (approved):
lgtm!

PR #1597, opened 2025-01-14
Move metric alias details from `MetricSpecPattern` to `ResolverInputForMetric`

courtneyholcomb:
This addresses [some feedback](https://github.com/dbt-labs/metricflow/pull/1573#pullrequestreview-2518180663) that did not get addressed before merging [this PR](https://github.com/dbt-labs/metricflow/pull/1573) - specifically:

> Since the alias in this case is describing user inputs, can it be added as a field to ResolverInputForMetric?

### DIFF
--- a/metricflow-semantics/metricflow_semantics/naming/metric_scheme.py
+++ b/metricflow-semantics/metricflow_semantics/naming/metric_scheme.py
@@ -2,12 +2,12 @@ from __future__ import annotations
 
 from typing import Optional
 
+from dbt_semantic_interfaces.references import MetricReference
 from typing_extensions import override
 
 from metricflow_semantics.model.semantic_manifest_lookup import SemanticManifestLookup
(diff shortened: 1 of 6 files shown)


serramatutu:
Beginner question: why is it better for the alias to live in `ResolverInputForMetric` instead of `MetricSpecPattern`? Is it simply a convention that all user input should live there and we want to keep this clear separation? Or is there something more?

courtneyholcomb:
> Beginner question: why is it better for the alias to live in `ResolverInputForMetric` instead of `MetricSpecPattern`? Is it simply a convention that all user input should live there (in resolver input classes in general) and we want to keep this clear separation? Or is there something more?

@serramatutu It's just a logical separation to be more consistent about what lives where. Spec patterns are meant to match user inputs to available specs. Since the alias is irrelevant to what specs metrics are available for a given query, it doesn't make much sense for it to live there.

WilliamDee (approved):

PR #1599, opened 2025-01-15
Bug fix: decouple grain from date part

courtneyholcomb:
When date part is requested for a time dimension, the requested granularity will be ignored in the output SQL. Essentially, granularity does not matter for a dimension that uses date part.
Internally, we have not been treating it that way. We require a time granularity for every time dimension requested, even if date part is included, too. This PR updates the logic to require one or the other - granularity _or_ date part.

There was also a related inconsistency in the code base:
- In the group by resolver, we were supporting all combinations of valid granularities x valid date parts for a given dimension.
- In the dataflow plan, we were supporting valid date parts combined only with the dimension's minimum granularity.

This led to a bug when users queried date part with a different granularity than the minimum grain, which would bypass the user-friendly errors in the group by resolver and show them an error like `Unable to join all items in request`. This is not a common query pattern, but the bug is also triggered when you query a date part (and no granularity) for a sub-daily time dimension, since the default metric_time granularity is typically `DAY` while the dimension's minimum granularity is < `DAY`.

In order to resolve this issue, date part and granularity are no longer supported together. This removes the above inconsistency and is more aligned with how date part behaves in the rendered SQL.

In order to avoid a breaking change, requesting granularity with date part will not raise an error. Instead we will just ignore the requested granularity.


tests_metricflow/query_rendering/test_granularity_date_part_rendering.py
```diff
@@ -437,3 +439,53 @@ def test_subdaily_granularity_overrides_metric_default_granularity(  # noqa: D10
...
+    dataflow_to_sql_converter: DataflowToSqlQueryPlanConverter,
+    sql_client: SqlClient,
+    query_parser: MetricFlowQueryParser,
+) -> None:
+    query_spec = query_parser.parse_and_validate_query(
+        metric_names=["archived_users"],
+        group_by=(
+            TimeDimensionParameter(name="metric_time", grain=TimeGranularity.HOUR.value, date_part=DatePart.YEAR),
+        ),
+        where_constraint_strs="{{ TimeDimension('metric_time', 'month', date_part_name='day') = '2020-01-01' }}",
```
WilliamDee, line 458: need to wrap this in a list, otherwise you get https://github.com/dbt-labs/metricflow/pull/1599/commits/dad3755d1754d2604a048034858b6b1a4c67731d#diff-5d8fe0debedfef559a1f628e0cc669d322f1a1d1484ad11706e3980fe78b1b8bR20 😆

WilliamDee (approved):
just need to fix the one test!

tests_metricflow/query_rendering/test_granularity_date_part_rendering.py
courtneyholcomb: Oooops thank you!!

PR #1602, opened 2025-01-16
Add Github Action workflow to create a release PR

plypaul:
Props to @serramatutu for providing a reference for a similar setup. Ran commands locally, but the action will need to be merged to test fully.


.github/workflows/cd-create-release-pr.yaml
```diff
@@ -0,0 +1,90 @@
...
+        uses: miniscruff/changie-action@v2
+        with:
+          version: latest
+          args: latest
+
+      - name: Save the New Version Into Context
+        id: get-new-package-version
+        run: |
+          VERSION_WITH_V_PREFIX="${{ steps.changie-latest.outputs.output }}"
+          VERSION=$(echo "VERSION_WITH_V_PREFIX" | cut -c 2-)
```
serramatutu, line 47: I think there might be a typo here, you forgot the `$` before `VERSION_WITH_V_PREFIX`.

image

.github/workflows/cd-create-release-pr.yaml
```diff
@@ -0,0 +1,90 @@
...
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+
+      - name: Update Project Version in Source Code via Hatch
+        id: update-package-version
+        run: |
+          hatch version "${{ steps.get-new-package-version.outputs.version }}"
```
serramatutu, line 53: Should we add a sanity check here that ensures `"$(hatch version)" = "${{ steps.changie-latest.outputs.output }}"`?

This _shouldn't_ ever happen, but in case something somehow gets borked, we make sure we never open a PR with mismatched versions in `CHANGELOG.md`, `pyproject.toml` and `__about__.py`.

.github/workflows/cd-create-release-pr.yaml
```diff
@@ -0,0 +1,90 @@
...
+        run: "hatch -v run dev-env:pre-commit run --color=always --all-files"
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          branch: release-${{ steps.get-new-package-version.outputs.version }}
+          title: "Release PR for MetricFlow ${{ steps.get-new-package-version.outputs.version }}"
+          body: >
+            Update changelog, version, and attribution for release version
+            `${{ steps.package-version.outputs.version }}`
```
serramatutu, line 83: `steps.package-version` doesn't exist. Did you want `steps.get-new-package-version`?

.github/workflows/cd-create-release-pr.yaml
```diff
@@ -0,0 +1,90 @@
...
+          labels: "Skip Changelog"
+          commit-message: |
+            Release version: ${{ steps.package-version.outputs.version }}
```
serramatutu, line 86: same as above.

serramatutu:
Suggestion for another time: add [actionlint](https://github.com/rhysd/actionlint/) to run on pre-commit whenever anything in `.github` has changed. It picks up a lot of common errors like the YAML being out of spec, or referencing steps that don't exist etc.

serramatutu:
I made a PR to add a static checker to Github Actions files. It should catch some of this stuff :) https://github.com/dbt-labs/metricflow/pull/1605

plypaul:
Good catches with those renames. Never liked working with GHA because I can't test locally and catch issues. I was almost inclined to move steps to a script for that reason.

.github/workflows/cd-create-release-pr.yaml
plypaul: Updated.

.github/workflows/cd-create-release-pr.yaml
plypaul: Updated.

.github/workflows/cd-create-release-pr.yaml
plypaul: Updated.

serramatutu (approved):

PR #1605, opened 2025-01-16
Add `actionlint` to pre-commit hooks

serramatutu:
## Description

This PR adds [actionlint](https://github.com/rhysd/actionlint/) to our pre-commit hooks. It will run whenever something has changed under `.github/workflows/`.

As the name suggests, actionlint is a Github Actions file linter, and it checks for things such as non-conformance to the YAML schema, referencing non-existing steps/outputs or running some security audits. It also runs [shellcheck](https://github.com/koalaman/shellcheck) to ensure the shell scripts in our actions files have no glaring issues.

After adding actionlint, I ran it over the entire repo to fix the issues it found with our actions.

You can review by commit.

### DIFF
--- a/.github/workflows/cd-push-dbt-metricflow-to-pypi.yaml
+++ b/.github/workflows/cd-push-dbt-metricflow-to-pypi.yaml
@@ -19,7 +19,7 @@ jobs:
       id-token: write
     steps:
       - name: Check-out the repo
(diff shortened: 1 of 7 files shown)


courtneyholcomb (commented):
Nice!! Thank you!

WilliamDee (approved):

PR #1606, opened 2025-01-16
Add config to get_measures_for_metrics response

DevonFulcher:
This method constructs its response with `__ini__` where others use `from_pydantic`. So, this field has to be added manually so that it MFS can gain access to it.


courtneyholcomb (approved):
LGTM!!

PR #1612, opened 2025-01-18
Add `SqlCteAliasMapping` to replace `dict`

plypaul:
This PR adds the `SqlCteAliasMapping` to replace the plain `dict` that is currently used to map a CTE alias to a CTE node. This allows for easier manipulation / type hinting when improved CTE support is added to the optimizers.

### DIFF
--- a/metricflow/sql/optimizer/required_column_aliases.py
+++ b/metricflow/sql/optimizer/required_column_aliases.py
@@ -11,6 +11,7 @@ from typing_extensions import override
 from metricflow.sql.optimizer.tag_column_aliases import NodeToColumnAliasMapping
 from metricflow.sql.sql_plan import (
     SqlCreateTableAsNode,
+    SqlCteAliasMapping,
     SqlCteNode,
     SqlPlanNode,
     SqlPlanNodeVisitor,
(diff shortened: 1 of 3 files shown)


metricflow/sql/optimizer/required_column_aliases.py
```diff
@@ -70,11 +71,14 @@ def __init__(self, start_node: SqlPlanNode, required_column_aliases_in_start_nod
         # Helps lookup the CTE node associated with a given CTE alias. A member variable is needed as any node in the
         # SQL DAG can reference a CTE.
         start_node_as_select_node = start_node.as_select_node
-        self._cte_alias_to_cte_node: Dict[str, SqlCteNode] = (
-            {cte_source.cte_alias: cte_source for cte_source in start_node_as_select_node.cte_sources}
-            if start_node_as_select_node is not None
-            else {}
-        )
+
+        self._current_cte_alias_mapping = SqlCteAliasMapping()
```
courtneyholcomb, line 75: Does "current" add meaning here or just verbosity? Is there another mapping that makes clarification necessary?

courtneyholcomb (commented):
LGTM!

metricflow/sql/optimizer/required_column_aliases.py
courtneyholcomb: Found the answer to this in another PR - yes, we also have a `previous_cte_alias_mapping` variable sometimes

courtneyholcomb (approved):
Whoops, meant to approve before.

PR #1613, opened 2025-01-18
For `SqlColumnPrunerOptimizer` - add support for CTEs in sub-queries

plypaul:
Previously, the `SqlColumnPrunerOptimizer` only accounted for CTEs defined at the top level as in current use, the `DataflowToSqlPlanConverter` only generated SQL plans in that way. To allow for generation of SQL plans with CTEs defined in sub-queries (at any depth), this PR updates `SqlColumnPrunerOptimizer` to keep track of the CTE alias mapping / CTE context when traversing the SQL plan.

### DIFF
--- a/metricflow/sql/optimizer/cte_alias_to_cte_node_mapping.py
+++ b/metricflow/sql/optimizer/cte_alias_to_cte_node_mapping.py
@@ -1,0 +1,82 @@
+from __future__ import annotations
+
+import logging
+from typing import Dict
+
+from metricflow_semantics.mf_logging.lazy_formattable import LazyFormat
+
+from metricflow.sql.sql_plan import SqlCteAliasMapping, SqlSelectStatementNode
+
+logger = logging.getLogger(__name__)
+
+
+class SqlCteAliasMappingLookup:
+    """A mutable lookup that stores the CTE-alias mapping at a given node.
+
+    In cases with nested CTEs in a SELECT, it's possible that a CTE defined in an inner SELECT has an alias that is the
+    same as a CTE defined in an outer SELECT. e.g.
+
+        # outer_cte
+        WITH cte_0 AS (
+            SELECT 1 AS col_0
+        )
+
+        # outer_select
+        SELECT col_0
+        FROM (
+
+            # inner_cte
+            WITH cte_0 AS (
+                SELECT 2 AS col_0
+            )
+            # inner_select
+            SELECT col_0 FROM cte_0
+        )
+        ...
+
+    In this case, `outer_cte` and `inner_cte` both have the same alias `cte_0`. When `cte_0` is referenced from
+    `inner_select`, it is referring to the `inner_cte`. For column pruning, it is necessary to figure out which CTE
+    a given alias is referencing, so this class helps to keep track of that mapping.
+    """
+
+    def __init__(self) -> None:  # noqa: D107
+        self._select_node_to_cte_alias_mapping: Dict[SqlSelectStatementNode, SqlCteAliasMapping] = {}
+
+    def cte_alias_mapping_exists(self, select_node: SqlSelectStatementNode) -> bool:
+        """Returns true if the CTE-alias mapping for the given node has been recorded."""
+        return select_node in self._select_node_to_cte_alias_mapping
+
+    def add_cte_alias_mapping(
+        self,
+        select_node: SqlSelectStatementNode,
+        cte_alias_mapping: SqlCteAliasMapping,
+    ) -> None:
+        """Associate the given CTE-alias mapping with the given node.
+
+        Raises an exception if a mapping already exists.
+        """
+        if select_node in self._select_node_to_cte_alias_mapping:
+            raise RuntimeError(
+                str(
+                    LazyFormat(
+                        "`select_node` node has already been added,",
+                        # child_select_node=child_select_node,
+                        select_node=select_node,
+                        current_mapping=self._select_node_to_cte_alias_mapping,
+                    )
+                )
+            )
+
+        self._select_node_to_cte_alias_mapping[select_node] = cte_alias_mapping
+
+    def get_cte_alias_mapping(self, select_node: SqlSelectStatementNode) -> SqlCteAliasMapping:
+        """Return the CTE-alias mapping for the given node.
+
+        Raises an exception if a mapping was not previously added for the given node.
+        """
+        cte_alias_mapping = self._select_node_to_cte_alias_mapping.get(select_node)
+        if cte_alias_mapping is None:
+            raise RuntimeError(
+                str(LazyFormat("CTE alias mapping does not exist for the given `select_node`", select_node=select_node))
+            )
+        return cte_alias_mapping
--- a/metricflow/sql/optimizer/required_column_aliases.py
+++ b/metricflow/sql/optimizer/required_column_aliases.py
@@ -57,7 +57,12 @@ class SqlMapRequiredColumnAliasesVisitor(SqlPlanNodeVisitor[None]):
         ) source_0
     """
 
-    def __init__(self, start_node: SqlPlanNode, required_column_aliases_in_start_node: FrozenSet[str]) -> None:
+    def __init__(
+        self,
+        start_node: SqlPlanNode,
+        required_column_aliases_in_start_node: FrozenSet[str],
+        cte_alias_mapping_lookup: SqlCteAliasMappingLookup,
+    ) -> None:
         """Initializer.
 
         Args:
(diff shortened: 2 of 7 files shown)


metricflow/sql/optimizer/cte_alias_to_cte_node_mapping.py
```diff
@@ -0,0 +1,82 @@
...
+
+from metricflow.sql.sql_plan import SqlCteAliasMapping, SqlSelectStatementNode
+
+logger = logging.getLogger(__name__)
+
+
+class SqlCteAliasMappingLookup:
+    """A mutable lookup that stores the CTE-alias mapping at a given node.
+
+    In cases with nested CTEs in a SELECT, it's possible that a CTE defined in an inner SELECT has an alias that is the
```
courtneyholcomb, line 16: Why would they be defined with the same alias? Shouldn't we use something similar to `_next_unique_table_alias()` to build CTE aliases? Or is the problem that both the `DataflowToSqlQueryPlanConverter` and the optimizer won't have access to that function to ensure unique aliases?

metricflow/sql/optimizer/cte_alias_to_cte_node_mapping.py
```diff
@@ -0,0 +1,82 @@
...
+    def add_cte_alias_mapping(
+        self,
+        select_node: SqlSelectStatementNode,
+        cte_alias_mapping: SqlCteAliasMapping,
+    ) -> None:
+        """Associate the given CTE-alias mapping with the given node.
+
+        Raises an exception if a mapping already exists.
+        """
+        if select_node in self._select_node_to_cte_alias_mapping:
```
courtneyholcomb, line 58: Is there any scenario where the `select_node` won't be unique in the dataflow plan at this point, and a valid plan will error?

metricflow/sql/optimizer/required_column_aliases.py
```diff
@@ -127,13 +131,34 @@ def _visit_parents(self, node: SqlPlanNode) -> None:
...
         cte_node = self._current_cte_alias_mapping.get_cte_node_for_alias(table_name)
+
         if cte_node is not None:
             self._current_required_column_alias_mapping.add_aliases(cte_node, column_aliases)
             # `visit_cte_node` will handle propagating the required aliases to all CTEs that this CTE node depends on.
             cte_node.accept(self)
 
     def visit_select_statement_node(self, node: SqlSelectStatementNode) -> None:
         """Based on required column aliases for this SELECT, figure out required column aliases in parents."""
+        # If this SELECT node defines any CTEs, it should override ones that were defined in the outer SELECT in case
```
courtneyholcomb, line 142: Similar question here - why do we expect collisions?

metricflow/sql/optimizer/required_column_aliases.py
```diff
@@ -127,13 +131,34 @@ def _visit_parents(self, node: SqlPlanNode) -> None:
...
+        # of CTE alias collisions.
+        if not self._cte_node_lookup.cte_alias_mapping_exists(node):
```
courtneyholcomb, line 144: Just checking my understanding - this works because the CTE nodes from the inner query will be added to the mapping before the nodes from the outer query?

courtneyholcomb (approved):
Looks good assuming we do expect alias collisions in valid plans - but I'm wondering if the collision-handling is necessary if we can build unique aliases appropriately.

metricflow/sql/optimizer/required_column_aliases.py
plypaul: Currently, we do not. But once we start adding CTE nodes within `DataflowToSqlQueryPlanConverter`, there's the potential if unique aliases aren't enforced.

metricflow/sql/optimizer/required_column_aliases.py
plypaul: It's actually the other way around  e.g. for

```
WITH cte_in_outer_query AS (
  SELECT 1
)

SELECT * FROM (
  WITH cte_in_inner_query AS (
    SELECT 1
  )
  SELECT * FROM cte_in_inner_query JOIN cte_in_outer_query
)
```

We add the mapping `{"cte_in_outer_query": ..}` then `{cte_in_outer_query": ..., "cte_in_inner_query": ...}`

courtneyholcomb: I'm confused then 🤔 With this conditional, won't we only add the mapping if the node isn't in the mapping yet? In which case, the outer node will override the inner node, which is the opposite of what the comment says?

metricflow/sql/optimizer/required_column_aliases.py
courtneyholcomb: Would it be simpler to just enforce unique aliases? Or do you think think it would be hard to do that?

metricflow/sql/optimizer/cte_alias_to_cte_node_mapping.py
courtneyholcomb: Gotcha. I definitely was planning to use something similar to `_next_unique_table_alias()` in `DataflowToSqlQueryPlanConverter` for CTE aliases. If you have a strong opinion about handling this defensively, that's fine and we can leave it here. Just might be nice to simplify the logic if it's not necessary!

plypaul: My aim is to try to maintain the contract: if the optimizer is fed in valid SQL, then it should output valid SQL. Otherwise, it becomes harder to use / makes a trap.

courtneyholcomb: Ohhh ok I think I was misunderstanding. I was imagining the CTEs would always be bubbled up to the top level, even if there were defined when visiting a subquery. I didn't realize that you could have a CTE that lives inside a subquery. In that case, your logic makes sense.

metricflow/sql/optimizer/required_column_aliases.py
plypaul: To clarify, the mapping refers to the whole `{"cte_in_outer_query": ..., "cte_in_inner_query": ...}` where as the lookup is like `{"inner_select": {"cte_in_outer_query": ..., "cte_in_inner_query": ...}}`. The conditional check is seeing if we have figured out the CTE-alias mapping for this SELECT statement. So for:

```
WITH cte_in_outer_query AS (
  SELECT 1
)

# outer_select
SELECT * FROM (
  WITH cte_in_inner_query AS (
    SELECT 1
  )
  # inner_select
  SELECT * FROM cte_in_inner_query JOIN cte_in_outer_query
)
```

We first visit `outer_select`. `cte_alias_mapping_exists(outer_select) == False` so we add the mapping to the lookup which then looks like:
```
{
  "outer_select": {"cte_in_outer_query": ..}`
}
```
Then we visit `inner_select`. `cte_alias_mapping_exists(inner_select) == False` as well so we add the mapping ` "outer_select": {"cte_in_outer_query": ..}`. The lookup now looks like:

```
{
  "outer_select": {"cte_in_outer_query": ..},
  "inner_select": {"cte_in_outer_query": ..., "cte_in_inner_query": ...}
}
```

* To make this easier to follow, I've parted out the logic for building a `SqlCteAliasMappingLookup` to a separate class (see last commit).
* There is a issue with the way that a cache in `DataflowNodeToSqlCteVisitor` works right now that is related to this, but it doesn't cause errors (will have to address in a separate PR).

metricflow/sql/optimizer/cte_alias_to_cte_node_mapping.py
plypaul: Yeah, I was thinking about your use case and thought `CTE that lives inside a subquery` was going to be a better fit since it was only locally needed? There's flexibility though, so let me merge this so that you can try it out.

metricflow/sql/optimizer/cte_alias_to_cte_node_mapping.py
plypaul: Actually, looks like some caching in `DataflowNodeToSqlCteVisitor` might cause `select_node won't be unique`. It doesn't cause issues at the moment, but let me taking a look on how to avoid that.

metricflow/sql/optimizer/cte_alias_to_cte_node_mapping.py
courtneyholcomb: Yeah, either should be fine from an execution standpoint, but keeping the CTEs at the top level might be most readable!

metricflow/sql/optimizer/required_column_aliases.py
courtneyholcomb: Ohh helpful, thank you!!

PR #1618, opened 2025-01-22
Fix `SelectStatementNode` Duplication Issue in `DataflowNodeToSqlCteVisitor`

plypaul:
This PR fixes an issue where a `SelectStatementNode` was duplicated / referenced multiple times when generating a `SqlPlan` with CTEs. This was due to a cache in `DataflowNodeToSqlCteVisitor`, so this updates the cache so that different elements are stored. The new elements can be used to create a new `SelectStatementNode` on a cache hit.

In cases where a `SelectStatementNode` is referenced multiple times, the column pruner will be more conservative and not prune as many columns as possible. With the fix, the snapshot changes reflect additional columns that can be pruned.

### DIFF
--- a/metricflow/plan_conversion/dataflow_to_sql.py
+++ b/metricflow/plan_conversion/dataflow_to_sql.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import datetime as dt
 import logging
 from collections import OrderedDict, defaultdict
+from dataclasses import dataclass
 from typing import Callable, Dict, FrozenSet, List, Optional, Sequence, Set, Tuple, TypeVar
 
 from dbt_semantic_interfaces.enum_extension import assert_values_exhausted
@@ -14,7 +15,7 @@ from dbt_semantic_interfaces.type_enums.period_agg import PeriodAggregation
 from dbt_semantic_interfaces.validations.unique_valid_name import MetricFlowReservedKeywords
 from metricflow_semantics.aggregation_properties import AggregationState
 from metricflow_semantics.dag.id_prefix import StaticIdPrefix
-from metricflow_semantics.dag.mf_dag import DagId
+from metricflow_semantics.dag.mf_dag import DagId, NodeId
 from metricflow_semantics.dag.sequential_id import SequentialIdGenerator
 from metricflow_semantics.filters.time_constraint import TimeRangeConstraint
 from metricflow_semantics.instances import (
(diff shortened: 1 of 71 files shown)


metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -2019,6 +2020,29 @@ def strip_time_from_dt(ts: dt.datetime) -> dt.datetime:
         )
 
 
+@dataclass(frozen=True)
+class CteGenerationResult:
+    """This stores parameters for creating a dataset from a CTE."""
+
+    source_dataflow_plan_node_id: NodeId
+    cte_node: SqlCteNode
+    select_columns: Tuple[SqlSelectColumn, ...]
```
courtneyholcomb, line 2029: Wondering if this attribute is redundant since you could get the select columns from the `SqlCteNode`. Might require making `SqlCteNode.select_statement` a `SqlSelectStatementNode` instead of a generic `SqlPlanNode` so we get the type benefits. Not sure if there are any limitations that prevent that, though.

metricflow/plan_conversion/dataflow_to_sql.py
```diff
@@ -2019,6 +2020,29 @@ def strip_time_from_dt(ts: dt.datetime) -> dt.datetime:
...
+    instance_set: InstanceSet
+
+    def get_sql_data_set(self) -> SqlDataSet:
+        """Return a dataset that represents reading from the CTE."""
+        cte_alias = self.cte_node.cte_alias
+        return SqlDataSet(
+            instance_set=self.instance_set,
+            sql_select_node=SqlSelectStatementNode.create(
```
courtneyholcomb, line 2037: The change I mentioned above would also allow you to access the `SqlSelectStatementNode` via `self.cte_node.select_statement` instead of building a new one here.

courtneyholcomb (approved):

metricflow/plan_conversion/dataflow_to_sql.py
plypaul: So the columns from the CTE can't be used in that way right now because the select statement in the CTE might have columns that are expressions. e.g.

```
WITH cte_0 AS (
  SELECT col_0 + col_1 AS col_2
  ...
)
```

I had plans to fix the signature of `SqlDataSet` to make it possible to use the `SqlCteNode`directly. The `SqlDataSet` initializer currently looks like

```
    def __init__(
        self,
        instance_set: InstanceSet,
        sql_select_node: Optional[SqlSelectStatementNode] = None,
        sql_node: Optional[SqlPlanNode] = None,
    ) -> None:
```

 which is a little odd. Once that's addressed, the SELECT could be eliminated.

courtneyholcomb: Sounds good! Not crucial for this PR.

PR #1619, opened 2025-01-22
Fix reuse of `SelectStatementNode` for semi-additive join SQL

plypaul:
This PR fixes a case of `SelectStatementNode` reuse when generating the SQL for a semi-additive join node. There are a few renames in here that increase the diff, but the actual change is small. Please view by commit.


courtneyholcomb (approved):
LGTM!

PR #1620, opened 2025-01-22
Support for custom offset window queried with only the window's custom grain

courtneyholcomb:
Support for querying a custom offset window metric with only the same grain as is used in the offset window. For this query, we can use simpler SQL than when querying with other grains.


metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -1769,19 +1772,22 @@ def _build_aggregated_measure_from_measure_source_node(
...
+        if before_aggregation_time_spine_join_description and queried_agg_time_dimension_specs:
             unaggregated_measure_node = self._build_time_spine_join_node_for_offset(
                 join_description=before_aggregation_time_spine_join_description,
                 measure_properties=measure_properties,
-                base_queried_agg_time_dimension_specs=base_queried_agg_time_dimension_specs,
+                queried_agg_time_dimension_specs=tuple(queried_agg_time_dimension_specs),
                 metric_source_node=unaggregated_measure_node,
             )
 
+        curent_node_time_dimension_specs = self._node_data_set_resolver.get_output_data_set(
```
courtneyholcomb, line 1783: TODO: separate this into its own commit if keeping it. Maybe consider using some alt logic though that doesn't require traversing the whole DAG

metricflow/plan_conversion/to_sql_plan/dataflow_to_subquery.py
```diff
@@ -1284,7 +1284,15 @@ def visit_join_to_time_spine_node(self, node: JoinToTimeSpineNode) -> SqlDataSet
         # Build new instances and columns.
         time_spine_required_spec_set = InstanceSpecSet(time_dimension_specs=required_agg_time_dimension_specs)
         output_parent_instance_set = parent_data_set.instance_set.transform(
-            FilterElements(exclude_specs=time_spine_required_spec_set)
+            FilterElements(
+                exclude_specs=InstanceSpecSet.create_from_specs(
+                    tuple(
+                        spec
+                        for spec in time_spine_required_spec_set.time_dimension_specs
+                        if spec in parent_data_set.instance_set.spec_set.time_dimension_specs
```
courtneyholcomb, line 1292: This filter is necessary because with the new time spine node, the specs we want to output from `JoinToTimeSpineNode` may not always be present in the parent node (the metric source node). This is because the new time spine node will have custom grains satisfied already, while the parent node will only have their base grains.

tests_metricflow/snapshots/test_custom_granularity.py/SqlPlan/DuckDB/test_join_to_time_spine_metric_grouped_by_custom_grain__plan0.sql
```diff
@@ -18,7 +18,7 @@ FROM (
     FROM (
       -- Change Column Aliases
       SELECT
-        subq_5.ds__day AS metric_time__day
+        subq_5.ds__day
```
courtneyholcomb, line 21: This change is due to the `join_on_time_dimension` being not longer required in the output of JoinToTimeSpineNode. It does not impact optimized SQL.

plypaul:
Can you elaborate on the difference between `OffsetBaseGrainByCustomGrainNode` and `OffsetCustomGranularityNode`?

courtneyholcomb:
> Can you elaborate on the difference between `OffsetBaseGrainByCustomGrainNode` and `OffsetCustomGranularityNode`?

@plypaul I think a good way to see the difference it to look at the optimized output SQL for each node side by side:
[OffsetCustomGranularityNode](https://github.com/dbt-labs/metricflow/blob/dbeded9411037fd33b176c85b37ef9dae62309a3/tests_metricflow/snapshots/test_dataflow_to_sql_plan.py/SqlPlan/DuckDB/test_offset_custom_granularity_node__plan0_optimized.sql) vs. [OffsetBaseGrainByCustomGrainNode](https://github.com/dbt-labs/metricflow/blob/dbeded9411037fd33b176c85b37ef9dae62309a3/tests_metricflow/snapshots/test_dataflow_to_sql_plan.py/SqlPlan/DuckDB/test_offset_by_custom_granularity_node__plan0_optimized.sql)

If you group by just the custom grain used in the window, we can use a much simpler plan - basically just get unique rows for the custom grain and offset them. But if you group by any other grain, we have to offset the base grain by the custom grain - that means figuring out the start and end rows within the custom grain, the row number of each date within the custom grain, and then offsetting accordingly. It takes some extra subqueries.

courtneyholcomb:
@plypaul also just a note on the node names - I'm not attached to these names at all! Just had a hard time thinking of anything better to differentiate them since they are quite similar!

plypaul:
So the two nodes are similar in what they do at a high level - given some time dimension, offset it by some amount.  If that's the case, can a single node be used to represent that in the dataflow plan by having conditional logic in the `DataflowNodeToSqlSubqueryVisitor`?

courtneyholcomb:
> So the two nodes are similar in what they do at a high level - given some time dimension, offset it by some amount.  If that's the case, can a single node be used to represent that in the dataflow plan by having conditional logic in the `DataflowNodeToSqlSubqueryVisitor`?

@plypaul In my opinion, these nodes should remain separate. The SQL is quite different between the two nodes. There are also several other things that need to be handled differently for the nodes within the dataflow plan, not just the SQL plan - whether to include the base grains or custom grains, what join spec to use, etc. - so it was much cleaner to implement two separate nodes.

plypaul (approved):
LGTM

PR #1628, opened 2025-01-24
Fix `mf tutorial` and add self-contained dbt project

plypaul:
The `mf tutorial` command was broken due to recent updates to DSI / MF, so this PR makes updates to fix them. Also, this adds an option to create a sample dbt project that can run the tutorial in a self-contained fashion.

Tutorial output looks like:

image

### DIFF
--- a/dbt-metricflow/dbt_metricflow/cli/cli_configuration.py
+++ b/dbt-metricflow/dbt_metricflow/cli/cli_configuration.py
@@ -29,13 +29,12 @@ class CLIConfiguration:
         self._semantic_manifest: Optional[SemanticManifest] = None
         self._semantic_manifest_lookup: Optional[SemanticManifestLookup] = None
 
-    def setup(self, dbt_project_path: Optional[pathlib.Path] = None) -> None:
+    def setup(self) -> None:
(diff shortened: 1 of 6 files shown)


dbt-metricflow/dbt_metricflow/cli/cli_configuration.py
```diff
@@ -17,25 +17,37 @@
...
+        The dbt_artifacts construct must be loaded in order for logging configuration to work correctly.
+        """
+        dbt_project_path = pathlib.Path.cwd()
+        self._dbt_project_metadata = dbtProjectMetadata.load_from_project_path(dbt_project_path)
+
         # self.log_file_path invokes the dbtRunner. If this is done after the configure_logging call all of the
         # dbt CLI logging configuration could be overridden, resulting in lots of things printing to console
         self._configure_logging(log_file_path=self.log_file_path)
 
+    def _get_dbt_project_metadata(self) -> dbtProjectMetadata:
```
WilliamDee, line 44: do we need this when we already have the `@property.dbt_project_metadata`? Doesn't seem like it's used

WilliamDee (approved):

PR #1638, opened 2025-01-25
Update Pypi publish workflow to use Python 3.9

plypaul:
Got this error while trying to publish in the workflow:

```
ERROR    InvalidDistribution: Invalid distribution metadata: license-expression 
         introduced in metadata version 2.4, not 2.1     
```

Might be caused by something else, but fixing the Python version to rule it out.


serramatutu (approved):

plypaul:
No dice.

PR #1641, opened 2025-01-25
Use cache when resolving data sets

courtneyholcomb:
Moves the node to data set cache to `DataflowNodeToSqlSubqueryVisitor`. This allows us to use the cache every time we build a data set for a node, including during dataflow to SQL conversion. You can see from the snapshot changes that this generally decreases the number of subquery aliases created, which indicates the cache is being hit more frequently. Hopefully this adds a marginal performance improvement.
Reviewing by commit is recommended. All snapshot changes should be subquery alias changes.


tests_metricflow/snapshots/test_cumulative_metric_rendering.py/SqlPlan/DuckDB/test_cumulative_metric_no_window_with_time_constraint__plan0_optimized.sql
```diff
@@ -10,15 +10,15 @@ sql_engine: DuckDB
 -- Aggregate Measures
 -- Compute Metrics via Expressions
 SELECT
-  subq_12.metric_time__day AS metric_time__day
-  , SUM(subq_11.txn_revenue) AS revenue_all_time
+  subq_13.metric_time__day AS metric_time__day
```
courtneyholcomb, line 13: increase?

tests_metricflow/snapshots/test_dataflow_to_execution.py/ExecutionPlan/DuckDB/test_multihop_joined_plan__ep_0.xml
```diff
@@ -7,40 +7,40 @@ docstring:
...
-        <!--    'GROUP BY\n'                                                                                        -->
-        <!--    '  subq_8.customer_id__customer_name')                                                              -->
+        <!-- sql =                                                                                                   -->
+        <!--   ('-- Join Standard Outputs\n'                                                                         -->
+        <!--    "-- Pass Only Elements: ['txn_count', 'account_id__customer_id__customer_name']\n"                   -->
+        <!--    '-- Aggregate Measures\n'                                                                            -->
+        <!--    '-- Compute Metrics via Expressions\n'                                                               -->
+        <!--    'SELECT\n'                                                                                           -->
+        <!--    '  subq_11.customer_id__customer_name AS account_id__customer_id__customer_name\n'                   -->
```
courtneyholcomb, line 16: increase?

tests_metricflow/snapshots/test_metric_filter_rendering.py/SqlPlan/DuckDB/test_distinct_values_query_with_metric_filter__plan0.sql
```diff
@@ -6,150 +6,150 @@ sql_engine: DuckDB
 ---
 -- Pass Only Elements: ['listing',]
 SELECT
-  subq_8.listing
+  subq_11.listing
```
courtneyholcomb, line 9: increase?

tests_metricflow/snapshots/test_time_spine_join_rendering.py/SqlPlan/DuckDB/test_join_to_time_spine_with_queried_time_constraint__plan0.sql
```diff
@@ -6,52 +6,52 @@ sql_engine: DuckDB
 ---
 -- Compute Metrics via Expressions
 SELECT
-  subq_10.metric_time__day
-  , COALESCE(subq_10.bookings, 0) AS bookings_fill_nulls_with_0
+  subq_11.metric_time__day
```
courtneyholcomb, line 9: increase?

metricflow/plan_conversion/to_sql_plan/dataflow_to_subquery.py
```diff
@@ -163,6 +167,30 @@ def _next_unique_cte_alias(self) -> str:
...
 
+    def get_output_data_set(self, node: DataflowPlanNode) -> SqlDataSet:
+        """Cached since this will be called repeatedly during the computation of multiple metrics.
+
+        # TODO: The cache needs to be pruned, but has not yet been an issue.
+        """
+        if node not in self._node_to_output_data_set:
+            self._node_to_output_data_set[node] = node.accept(self)
+
+        return self._node_to_output_data_set[node]
```
plypaul, line 178: If the cache is used in the conversion to SQL, this will will need to return a `.copy()` of the SQL node to avoid duplication of nodes in the generated plan.

courtneyholcomb: When would you expect duplicate nodes to be generated? Also, do you mean dataflow plan nodes or SQL nodes?

courtneyholcomb: That should only happen if we have the exact same dataflow plan node, right? Same node ID and everything? I'm just worried if we return a copy, we won't actually hit the cache as often (but maybe I'm wrong about that).

plypaul: > That should only happen if we have the exact same dataflow plan node, right? Same node ID and everything?
Yes.

>  I'm just worried if we return a copy, we won't actually hit the cache as often (but maybe I'm wrong about that).

So hitting the cache will be determined by `node`, which shouldn't change. To clarify, the copy refers to copying the SQL node in `SqlDataSet`.

courtneyholcomb: Ok makes sense! Updated. With the snapshot changes in the latest commit now I see the problem I had overlooked in the prior snapshots.
Should be ready for review again!

plypaul (approved):
LGTM - thanks!

PR #1663, opened 2025-02-06
Custom offset window: support grains smaller than base grain

courtneyholcomb:
Support for querying a custom offset window metric with a grain that's smaller than the custom grain's base grain. Previously this would render invalid SQL.


plypaul (commented):
Still getting familiar with the changes - can you elaborate on why joining the 2nd time spine is needed for the smaller grain vs. starting out with the smaller grain time spine?

courtneyholcomb:
> Still getting familiar with the changes - can you elaborate on why joining the 2nd time spine is needed for the smaller grain vs. starting out with the smaller grain time spine?

@plypaul yes! So for all _standard_ grains, we can pick the smallest grain and use the time spine that satisfies that, and then just `DATE_TRUNC` to satisfy larger grains. For _custom_ grains, they are defined in one time spine and can only be satisfied by that time spine. A good example is [the query in this test.](https://github.com/dbt-labs/metricflow/blob/3b6485479556ff95649d4bb23f2b25431076024e/tests_metricflow/query_rendering/test_custom_granularity.py#L755) To satisfy `metric_time__hour`, we need to use the hourly time spine. But to satisfy the metric, which uses `offset_window: 1 alien_day`, we need to use the daily spine. That's because the `alien_day` column is only available in that time spine.
The query might also contain other custom grains in the group by, which could mean joining to additional time spines, but we don't need to worry about those in the time spine node logic because they will be joined later in the query (after joining to the metric source).

PR #1666, opened 2025-02-07
Update handling for Pydantic objects in `mf_pformat`

plypaul:
This PR updates the handling for Pydantic objects in `mf_pformat` to check if the `dict()` methods exists instead of the `isinstance(BaseModel)` check. The `BaseModel` is actually from the `dsi_shim` and in some cases with different Pydantic versions, it's possible that the check does not pass.  Since it's not foolproof, the formatter falls back to `pprint` if there is an error.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/mf_logging/pretty_print.py
+++ b/metricflow-semantics/metricflow_semantics/mf_logging/pretty_print.py
@@ -2,12 +2,9 @@ from __future__ import annotations
 
 import logging
 import pprint
-from collections.abc import Mapping
 from dataclasses import fields, is_dataclass
 from enum import Enum
-from typing import Any, List, Optional, Sized, Tuple, Union
-
-from dsi_pydantic_shim import BaseModel
+from typing import Any, List, Mapping, Optional, Sized, Tuple, Union
 
 from metricflow_semantics.mf_logging.formatting import indent
 from metricflow_semantics.mf_logging.pretty_formattable import MetricFlowPrettyFormattable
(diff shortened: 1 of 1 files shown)


metricflow-semantics/metricflow_semantics/mf_logging/pretty_print.py
```diff
@@ -353,16 +345,22 @@ def _handle_any_obj(self, obj: Any, remaining_line_width: Optional[int]) -> str:
...
-            return self._handle_mapping_like_obj(
-                mapping,
-                left_enclose_str=type(obj).__name__ + "(",
-                key_value_seperator="=",
-                right_enclose_str=")",
-                is_dataclass_like_object=True,
-                remaining_line_width=remaining_line_width,
-            )
+        # For Pydantic-like objects with a `dict` method that returns field keys / values.
+        dict_method = getattr(obj, "dict", None)
```
thiagosalvatore, line 349: @plypaul isn't there a chance that it breaks with pydantic v2 where `dict` was deprecated in favor of `model_dump` ?

plypaul: It actually still works, though now I see the deprecation warning. Updated the case handling though.

plypaul: Also to provide context on these formatting methods - in case there's a problem formatting, it falls back to the standard pretty printer or `str(...)`.

courtneyholcomb (approved):

PR #1675, opened 2025-02-18
Revert "Don't error for lack of time spines"

courtneyholcomb:
Reverts dbt-labs/metricflow#1428

We removed this error so that the saved query dependency resolver in mantle wouldn't hit it for a manifest that does not have any semantic models. Later, we decided not to treat that as a valid state for any MF classes, and instead to avoid instantiating the dependency resolver in that case. That means we can put this error back.
Plus, [some customer ran into this error scenario](https://dbt-labs.slack.com/archives/C05K4R7KZ5Z/p1739894465024389), and they got a very unclear error instead. This will give them a better error in the future.


theyostalservice (approved):
Thanks for the super-clear summary!  I couldn't have reviewed this without it :)

PR #1692, opened 2025-03-21
Specify DB file path in tutorial-project profile

plypaul:
This PR updates the `profiles.yml` file in the tutorial project so that the `DuckDB` database file is stored within the project directory instead of the current working directory. This helps to allow for later test cases that run the CLI from directories outside of the project directory.

### DIFF
--- a/dbt-metricflow/dbt_metricflow/cli/mf_tutorial_project/profiles.yml
+++ b/dbt-metricflow/dbt_metricflow/cli/mf_tutorial_project/profiles.yml
@@ -3,4 +3,4 @@ mf_tutorial:
   outputs:
     dev:
       type: duckdb
-      path: 'mf_tutorial.duckdb'
+      path: $db_file_path
--- a/.changes/unreleased/Features-20250328-140506.yaml
+++ b/.changes/unreleased/Features-20250328-140506.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Specify DB file path in tutorial-project profile
+time: 2025-03-28T14:05:06.062055-07:00
+custom:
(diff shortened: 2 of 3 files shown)


dbt-metricflow/dbt_metricflow/cli/mf_tutorial_project/profiles.yml
```diff
@@ -3,4 +3,4 @@ mf_tutorial:
   outputs:
     dev:
       type: duckdb
-      path: 'mf_tutorial.duckdb'
+      path: $db_file_path
```
WilliamDee, line 6: wonder if it's worth doing like,
```
> mf_tutorial_project/
     - templated_profiles.yml
     - profiles.yml
```
Where `profiles.yml` is a default valid one like the current one. Then you just read from `templated_profiles.yml` which contains all the formatting strings, and you overwite it into `profiles.yml`? Just wondering if like the user haven't ran the tutorial yet and just looked at this `profiles.yml` would it cause slight confusion.

WilliamDee (approved):

dbt-metricflow/dbt_metricflow/cli/mf_tutorial_project/profiles.yml
plypaul: > Just wondering if like the user haven't ran the tutorial yet and just looked at this profiles.yml

I thought it only gets created when the tutorial is run?

WilliamDee: Oh yea thats right

PR #1720, opened 2025-04-09
Allow no-metric queries without group by

courtneyholcomb:
Support no-metric queries without a group by. This can be used in specific cases where the user knows that the dimensions / entities in the query are unique, so the group by will only slow down query performance.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/query/query_parser.py
+++ b/metricflow-semantics/metricflow_semantics/query/query_parser.py
@@ -44,7 +44,7 @@ from metricflow_semantics.query.query_resolver import MetricFlowQueryResolver
 from metricflow_semantics.query.resolver_inputs.base_resolver_inputs import MetricFlowQueryResolverInput
 from metricflow_semantics.query.resolver_inputs.query_resolver_inputs import (
     InvalidStringInput,
-    ResolverInputForDedupe,
+    ResolverInputForApplyGroupBy,
     ResolverInputForGroupByItem,
     ResolverInputForLimit,
     ResolverInputForMetric,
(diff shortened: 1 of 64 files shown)


metricflow-semantics/metricflow_semantics/query/query_parser.py
```diff
@@ -92,12 +93,13 @@ def __init__(  # noqa: D107
...
-        time_constraint_end: Optional[datetime.datetime],
-        order_by_names: Optional[Sequence[str]],
-        order_by_parameters: Optional[Sequence[OrderByQueryParameter]],
+        where_filters: Optional[Sequence[WhereFilter]] = None,
+        limit: Optional[int] = None,
+        time_constraint_start: Optional[datetime.datetime] = None,
+        time_constraint_end: Optional[datetime.datetime] = None,
+        order_by_names: Optional[Sequence[str]] = None,
+        order_by_parameters: Optional[Sequence[OrderByQueryParameter]] = None,
+        dedupe: bool = True,
```
WilliamDee, line 102: wonder if this is hard to infer what it does based on the name 🤔 would like `dedupe_with_group_by` be better?

WilliamDee (approved):

metricflow-semantics/metricflow_semantics/query/query_parser.py
courtneyholcomb: Ugh naming is hard. I want to call it something like `apply_group_by` (which would be true for metric queries too). But in a lot of places `group_by` is the name of a parameter already so it feels weird, since if `apply_group_by` is false then the `group_by` param name no longer makes sense. But I guess the `group_by` param name is the issue in that case so I shouldn't let that affect how I name this param...

PR #1727, opened 2025-04-25
Alias dimensions & entities

courtneyholcomb:
Support for aliasing dimensions and entities. I recommend reviewing by commit.


metricflow-semantics/metricflow_semantics/specs/query_param_implementations.py
```diff
@@ -137,13 +136,16 @@ def query_resolver_input(  # noqa: D102
     ) -> ResolverInputForMetric:
         naming_scheme = MetricNamingScheme()
         return ResolverInputForMetric(
-            input_obj=self,
+            input_obj=self.with_alias(None),
```
WilliamDee, line 147: is there a need for calling `with_alias` here?

metricflow-semantics/metricflow_semantics/query/query_resolver.py
```diff
@@ -321,7 +321,7 @@ def _resolve_order_by(
             else:
                 order_by_specs.append(
                     OrderBySpec(
-                        instance_spec=matching_specs[0],
+                        instance_spec=matching_specs[0].with_alias(None),
```
WilliamDee, line 306: same here about needing to call `with_alias`

WilliamDee (approved):

metricflow-semantics/metricflow_semantics/specs/query_param_implementations.py
courtneyholcomb: Oops good call, removing all of those from this file!

metricflow-semantics/metricflow_semantics/query/query_resolver.py
courtneyholcomb: This one we'll keep - adding a comment as to why:
`# Ignore aliases in the order by since we'll render the expression instead of the alias.`

PR #1740, opened 2025-04-28
Fix error message with empty queries

plypaul:
This PR fixes typos / formatting of the error message when a query is empty.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/query/group_by_item/resolution_path.py
+++ b/metricflow-semantics/metricflow_semantics/query/group_by_item/resolution_path.py
@@ -5,6 +5,7 @@ from typing import Tuple
 
 from typing_extensions import override
 
+from metricflow_semantics.helpers.string_helpers import mf_indent
 from metricflow_semantics.query.group_by_item.path_prefixable import PathPrefixable
 from metricflow_semantics.query.group_by_item.resolution_dag.resolution_nodes.base_node import GroupByItemResolutionNode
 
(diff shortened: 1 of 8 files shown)


metricflow-semantics/metricflow_semantics/query/group_by_item/resolution_path.py
```diff
@@ -31,14 +30,30 @@ def last_item(self) -> GroupByItemResolutionNode:
...
-        descriptions = tuple(f"[Resolve {path_node.ui_description}]" for path_node in self.resolution_path_nodes)
-        output = descriptions[0]
+        # TODO: Centralize handling of error message formatting.
+        max_line_length = 80
+        lines = []
+        for i, path_node in enumerate(self.resolution_path_nodes):
+            if i == 0:
+                indent = ""
+            else:
+                indent = "  " * i + "-> "
```
theyostalservice, line 40: this is nitpicky, but I'd much prefer this be a constant (even just "TWO_SPACES" or something) so it's abundantly, readably clear what is happening here.

theyostalservice (approved):

metricflow-semantics/metricflow_semantics/query/group_by_item/resolution_path.py
plypaul: Updated. I put in `mf_indent()` for this kind use case a while ago, so used that instead.

PR #1743, opened 2025-05-01
Generate a `SELECT` statement for sink nodes in a dataflow plan

plypaul:
This PR updates the generation of SQL from the sink nodes in a dataflow plan (`WriteTo*...` nodes) to generate a pass-though `SELECT`. The sink nodes in a query request were mostly symbolic and it previously retuned the SQL from the parent node. However, the sink nodes can be used to generate a `SELECT` statement that reorders columns. As the final SQL will be simplified by the optimizers, this update does not change the complexity of the final query.


metricflow/plan_conversion/to_sql_plan/dataflow_to_subquery.py
```diff
@@ -793,17 +793,45 @@ def visit_order_by_limit_node(self, node: OrderByLimitNode) -> SqlDataSet:  # no
...
 
     def visit_write_to_result_data_table_node(self, node: WriteToResultDataTableNode) -> SqlDataSet:  # noqa: D102
-        # Returning the parent-node SQL as an approximation since you can't write to a data_table via SQL.
-        return self.get_output_data_set(node.parent_node)
+        # Currently, this is a SELECT statement that passes all columns through. This will be updated in a later PR
+        # to support specific column ordering.
+        input_data_set: SqlDataSet = self.get_output_data_set(node.parent_node)
+        input_data_set_alias = self._next_unique_table_alias()
+        output_instance_set = input_data_set.instance_set.transform(
+            ChangeAssociatedColumns(self._column_association_resolver)
```
courtneyholcomb, line 801: This transformer doesn't do anything, right? We should probably delete it. I think you can just leave the instances as is.

metricflow/plan_conversion/to_sql_plan/dataflow_to_subquery.py
```diff
@@ -793,17 +793,45 @@ def visit_order_by_limit_node(self, node: OrderByLimitNode) -> SqlDataSet:  # no
...
+                from_source_alias=input_data_set_alias,
+            ),
+        )
 
     def visit_write_to_result_table_node(self, node: WriteToResultTableNode) -> SqlDataSet:  # noqa: D102
         input_data_set: SqlDataSet = self.get_output_data_set(node.parent_node)
-        input_instance_set: InstanceSet = input_data_set.instance_set
+        input_data_set_alias = self._next_unique_table_alias()
+        output_instance_set = input_data_set.instance_set.transform(
+            ChangeAssociatedColumns(self._column_association_resolver)
```
courtneyholcomb, line 820: same here

courtneyholcomb (approved):

metricflow/plan_conversion/to_sql_plan/dataflow_to_subquery.py
plypaul: Good catch - I'll remove.

PR #1745, opened 2025-05-01
Add `MetricQuerySpec.spec_output_order`

plypaul:
This PR adds `MetricQuerySpec.spec_output_order` that describes the relative ordering of the specs as they were specified as inputs into the query parser.


plypaul:
FYI - `output_column_order_key` might have been from a prior implementation. Checking if that's needed now.

plypaul:
> FYI - output_column_order_key might have been from a prior implementation. Checking if that's needed now.
Removed.

courtneyholcomb (approved):
This would be a breaking change if we released it without the next PR up the stack, right? Just want to make sure we merge them together if so.

plypaul:
> This would be a breaking change if we released it without the next PR up the stack, right? Just want to make sure we merge them together if so.

It shouldn't be breaking as the field would just not be used.

PR #1755, opened 2025-05-16
Add `OrderedEnum`

plypaul:
This PR adds an `Enum` base class that can be ordered for consistency in tests. The `OrderedEnum` from `aenum` was considered but favoring this as the implementation is short and avoids adding a dependency.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/experimental/orderd_enum.py
+++ b/metricflow-semantics/metricflow_semantics/experimental/orderd_enum.py
@@ -1,0 +1,27 @@
+from __future__ import annotations
+
+import functools
+from enum import Enum
+
+from metricflow_semantics.experimental.comparison_helpers import ComparisonOtherType
+
+
+@functools.total_ordering
+class OrderedEnum(Enum):
+    """An `Enum` that can be sorted by definition order.
+
+    * This is useful dataclasses that have enums as fields as the default `Enum` isn't sortable.
+    * Considering using the `OrderedEnum` from `aenum`.
+    """
+
+    @classmethod
+    @functools.cache
+    def _enum_to_index(cls) -> dict[OrderedEnum, int]:
+        return {member: i for i, member in enumerate(cls)}
+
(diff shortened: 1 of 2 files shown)


theyostalservice (changes requested):
This should be called SortableEnum, because it's not actually about order but about sorting.

Further supporting this is the fact that since at least python 3.9, Enum classes are iterable in the order in which the enum is defined.
https://docs.python.org/3.9/library/enum.html
> Enumerations support iteration, in definition order

(As well as by the fact that you're actually using their indices to compare them here, rather than sorting on their values or something.  Your PR is already dependent on the fact that enums have a built-in ordering.)

plypaul:
> This should be called SortableEnum, because it's not actually about order but about sorting.

Can you elaborate here? How are you defining `Ordered` vs `Sortable`?

theyostalservice:
> > This should be called SortableEnum, because it's not actually about order but about sorting.
> 
> Can you elaborate here? How are you defining `Ordered` vs `Sortable`?

An ordered thing has an implicit order.  A sortable thing can be sorted.

The enum already has an implicit order.  You can iterate over it in the same, built-in order as many times as you like.  the fact that there is an order is **literally in the python docs for Enum**, as quoted above.

You are not adding an order to it - you are making it work with the "sort" function.

plypaul:
> An ordered thing has an implicit order.

That statement is vague / sounds redundant like `A tagged thing has a tag`. Also, why is it implicit - shouldn't it be explicit?

> The enum already has an implicit order. You can iterate over it in the same, built-in order as many times as you like. 
> the fact that there is an order is literally in the python docs for Enum

That's conflating iteration order with comparison order. The Python docs specify an iteration order, not comparison order.

WilliamDee (approved):

plypaul:
@theyostalservice Bump on above.

theyostalservice:
Ah sorry there was a miscommunication; I thought this was resolved in our slack conversation.

I still believe that specificity and clarity in naming would be preferable and that this is just less clear than it could be.  However, Will's accepted as is, so I'll bow out.

PR #1759, opened 2025-05-22
Support for saved query order by & limit

courtneyholcomb:
Actually apply the `order_by` and `limit` set in a saved query. This was previously implemented in the YAML spec, but never implemented in MetricFlow.
If the user queries the saved query with a different `order_by` / `limit` in the query params, that will override what's set in the saved query.

Note that this required supporting metric object syntax in saved queries, as well, in order to apply the `.descending(..)` syntax.


metricflow-semantics/metricflow_semantics/query/query_resolver.py
```diff
@@ -425,31 +443,31 @@ def _resolve_query(self, resolver_input_for_query: ResolverInputForQuery) -> Met
         min_max_only_input = resolver_input_for_query.min_max_only
         apply_group_by_input = resolver_input_for_query.apply_group_by
 
+        mappings_to_merge: List[InputToIssueSetMapping] = []
+
+        # Resolve metrics.
```
courtneyholcomb, line 448: Needed to move this up - metrics must be resolved before building the `MetricFlowQueryResolutionPath` used later now that object syntax is supported. Previously, the metric names were already resolved since we only supported passing the name directly.

metricflow-semantics/tests_metricflow_semantics/snapshots/test_messages.py/str/test_long_invalid_metric__result.txt
```diff
@@ -16,7 +16,3 @@ Error #1:
   Query Input:
 
     long_invalid_metric_name__________________________________________________
-
```
courtneyholcomb, line 19: These are removed now since metric names aren't resolved at this point, but the above should be enough

tests_metricflow/validation/test_data_warehouse_tasks.py
```diff
@@ -306,7 +306,7 @@ def test_build_saved_query_tasks(  # noqa: D103
         manifest=simple_semantic_manifest,
         sql_client=sql_client,
     )
-    assert len(tasks) == 4
+    assert len(tasks) == 5
```
courtneyholcomb, line 309: Due to new saved query in the YAML

PR #1760, opened 2025-05-27
Reclassify user related errors

WilliamDee:
## Context

We throw a bunch of different exceptions for different scenarios. In attempts to categorize the errors appropriately, we will introduce a `InformativeError` which denotes that it isn't a system error from MF, but an error that falls under the bucket of things that a user can tweak in their input (either in their manifest config or query formulation) or it's just not supported. 

This isn't a complete solution, but there will be further efforts to classify more and make all the error handling be more verbose.


metricflow-semantics/metricflow_semantics/errors/error_classes.py
```diff
@@ -51,23 +57,19 @@ class ExecutionException(Exception):
...
-    """Exception to represent errors related to inference."""
-
-
-class UnsupportedEngineFeatureError(RuntimeError):
+class UnsupportedEngineFeatureError(InformativeError, RuntimeError):
     """Raised when the user attempts to use a feature that isn't supported by the data platform."""
 
 
-class SqlBindParametersNotSupportedError(Exception):
+class SqlBindParametersNotSupportedError(InformativeError):
```
plypaul, line 64: Seems like this isn't actionable?

metricflow-semantics/metricflow_semantics/errors/error_classes.py
```diff
@@ -4,7 +4,13 @@
 from typing import Dict, Optional
 
 
-class CustomerFacingSemanticException(Exception):
+class InformativeError(Exception):
+    """Raised when the error is actionable by the user or is informational."""
```
plypaul, line 8: What errors aren't `informational`?

plypaul (approved):
LGTM, but please see comments. Also, I was thinking of adding a base exception class like:

https://github.com/dbt-labs/metricflow/blob/p/semantic_graph.pr_1760_comment/metricflow-semantics/metricflow_semantics/experimental/metricflow_exception.py

metricflow-semantics/metricflow_semantics/errors/error_classes.py
WilliamDee: Updated to `InformativeUserError` if that's any better and updated the docstring

plypaul:
To clarify, I haven't thought much about the exception class hierarchy, and the linked file was for reference, not that we need to put that in.

PR #1761, opened 2025-05-28
Replace `Singleton` with `@singleton_dataclass()`

plypaul:
This PR replaces the existing `Singleton` base class with a `singleton_dataclass` decorator to make it easier to use. I would have liked to avoid the use of a custom metaclass, but I wasn't able to find a better approach. If there's a widely-used library implementation found later, we can swap it in.


metricflow-semantics/metricflow_semantics/experimental/singleton_decorator.py
```diff
@@ -0,0 +1,121 @@
...
+                    class_name=class_to_return.__name__,
+                    dataclass_field=dataclass_field,
+                )
+            )
+
+    instance_lock = threading.Lock()
+    # Might change to  `WeakValueDictionary`.
+    instance_args_to_instance: dict[AnyLengthTuple, SingletonInstance] = {}
+
+    dataclass_metaclass = type(class_to_return)
```
courtneyholcomb, line 86: nit: should this use Pascal case like most classes/types?

courtneyholcomb (approved):

metricflow-semantics/metricflow_semantics/experimental/singleton_decorator.py
plypaul: I don't see an exception for variables that store types in PEP 8, so leaving as snake case.

PR #1765, opened 2025-05-28
Use TIMESTAMP_TRUNC instead of DATETIME_TRUNC for BigQuery

courtneyholcomb:
A [user flagged](https://dbt-labs.slack.com/archives/C046L0VTVR6/p1748389759085469) that the `DATETIME_TRUNC` operation was causing a query to not use their BQ date partition properly, resulting in more GB of data processed than expected.
After doing some research, it looks like the `DATETIME_TRUNC` also converts a timestamp object to a datetime object, which removes any timezone info. We don't intend to ever change the timezone of the underlying column, so this is not what we want.
The downside of this change is that, for any BQ users who are using timezone-aware data that's not UTC, they might see a change in the output. This is because `DATETIME_TRUNC` converts to UTC by default before truncating. Regardless, most users using BigQuery should be using UTC as best practice.

This PR updates the BQ SQL renderer to use `TIMESTAMP_TRUNC` instead of `DATETIME_TRUNC`.


WilliamDee (approved):

CallumFriend:
Worth flagging that this PR also fixed a stricter failure mode than the one in the description. On BigQuery projects with +require_partition_filter: true on partitioned marts, the current DATETIME_TRUNC causes the query to get rejected.

Would really like to see this merged!

VassMorozov:
@WilliamDee I am just wondering what it will take to get this in? Callum highlights that this PR will fix another failure mode that we would like to see resolved.

Dom-M-C:
We hit this on a production BigQuery warehouse and can confirm the `require_partition_filter` hard-fail @CallumFriend describes, with some dry-run evidence that may help de-risk the merge.

Our fact table is partitioned on a **DATE** column with `require_partition_filter = true` (~212M rows, 77 GB). Dry-run probes of the four candidate predicates against the partition column:

| Predicate on the partition column | Result |
|---|---|
| `col BETWEEN …` (bare) | prunes — 31 MB scanned |
| `DATE_TRUNC(col, day) BETWEEN …` | prunes — 31 MB scanned |
| `TIMESTAMP_TRUNC(col, day) BETWEEN …` | prunes — 31 MB scanned |
| `DATETIME_TRUNC(col, day) BETWEEN …` | **rejected** — `Cannot query … without a filter over column(s) '…' that can be used for partition elimination` |

So `DATETIME_TRUNC` is the only member of the trunc family that BigQuery's pruner can't reason through here. The switch in this PR restores pruning (77 GB → 31 MB on this table) and, for partition-filter-required tables, turns a hard query failure into a working query — MetricFlow currently can't query such a fact table at all.

One note on the timezone concern holding this up: for a **DATE** partition column the value can't shift, since a DATE has no time-of-day or zone to reinterpret — `DATE_TRUNC`/`TIMESTAMP_TRUNC` and `DATETIME_TRUNC` all resolve to the same calendar day. The tz-aware-vs-naive difference only bites on genuinely timezone-bearing (TIMESTAMP) partition columns, so the DATE-column case looks like a safe subset if a staged rollout helps.

Adding a production +1 — currently blocked on this for any semantic model over a partition-filter-required fact table.

🤖 Sent via Claude Code

plypaul:
@VassMorozov @Dom-M-C To confirm: there would be no semantic changes by replacing `DATETIME_TRUNC()` with `TIMESTAMP_TRUNC()` with the current SQL rendering approach because those methods are overloaded to accept either `DATETIME` or `TIMESTAMP` arguments with the matching return type and both default to UTC if the timezone argument is not specified (which is how it's rendered now).

Let us know if you see any other issues. Given that and the cited partition pruning benefits, we can get this in this week.

PR #1771, opened 2025-06-18
Add search, sort, and pagination to metadata methods

courtneyholcomb:


metricflow/engine/metricflow_engine.py
```diff
@@ -590,73 +628,87 @@ def get_measures_for_metrics(self, metric_names: List[str]) -> List[Measure]:  #
...
+            raise InvalidQueryException("Page size must be at least 1")
+
+        output_items = list(items)
+        if sort:
+            output_items.sort(key=lambda x: x.default_search_and_sort_attribute)
+        if page_size:
+            output_items = output_items[(page_num - 1) * page_size : page_num * page_size]
+        return output_items
+
+    def _build_metric_time_from_linkable_dimension(self, linkable_dimension: LinkableDimension) -> Dimension:
```
courtneyholcomb, line 706: This logic has not changed, I just moved it to a helper function.

metricflow/engine/metricflow_engine.py
```diff
@@ -66,6 +72,17 @@
 _telemetry_reporter = TelemetryReporter(report_levels_higher_or_equal_to=TelemetryLevel.USAGE)
 _telemetry_reporter.add_python_log_handler()
 
+SIMPLE_DIMENSIONS_WITHOUT_ANY_PROPERTIES: Set[LinkableElementProperty] = {
```
courtneyholcomb, line 76: Moved to a constant and added the `METRIC` property. This doesn't create changes currently but would be necessary if we added support for metrics in the group by (not just the where filter).

metricflow/engine/metricflow_engine.py
```diff
@@ -665,87 +717,119 @@ def simple_dimensions_for_metrics(  # noqa: D102
...
+                        semantic_model_name=linkable_entity.defined_in_semantic_model.semantic_model_name,
+                        element_name=linkable_entity.element_name,
                     )
                 )
+                if pydantic_entity:
+                    entity = Entity.from_pydantic(pydantic_entity)
+                    if (
+                        not search_str
+                        or search_str in entity.default_search_and_sort_attribute
+                        # Dedupe. Duplicates currently show up because of local linked entities.
```
courtneyholcomb, line 861: Deduping is new functionality here, so you will see that change in some test snapshots.

WilliamDee (approved):

PR #1777, opened 2025-07-10
Update requirements.txt to allow for new dsi version

colin-rogers-dbt:


metricflow-semantics/requirements-files/requirements.txt
```diff
@@ -1,7 +1,7 @@
 # Always support a range of production DSI versions capped at the next breaking version in metricflow-semantics.
 # This allows us to sync new, non-breaking changes to dbt-core without getting a version mismatch in dbt-mantle,
 # which depends on a specific commit of DSI.
-dbt-semantic-interfaces>=0.8.4, <0.9.0
+dbt-semantic-interfaces>=0.8.4, <0.10.0
```
QMalcolm: We should probably bump the minimum to 0.9.0 as well 🤔

metricflow-semantics/requirements-files/requirements.txt
```diff
@@ -1,7 +1,7 @@
 # Always support a range of production DSI versions capped at the next breaking version in metricflow-semantics.
 # This allows us to sync new, non-breaking changes to dbt-core without getting a version mismatch in dbt-mantle,
 # which depends on a specific commit of DSI.
-dbt-semantic-interfaces>=0.8.4, <0.9.0
```
courtneyholcomb, line 4: I think it would be better to pin to >=0.9.0, <0.10.0 since there's a breaking change between 0.8 and 0.9 - but might not matter in practice since we pin to a more specific version in core?

courtneyholcomb (changes requested):

metricflow-semantics/requirements-files/requirements.txt
courtneyholcomb: Ah you'll be blocked on that without making some changes in this repo. I can make that update after my current meeting

colin-k-rogers: 🤷 to my mind if this code is compatible with earlier versions then it's okay to keep the pin looser but you have more context on this than I

courtneyholcomb: Agreed but it is unfortunately not compatible!

colin-k-rogers: updated!

courtneyholcomb: Sorry meeting ran very long! There are some other changes that need to be included based on the deps situation in this repo, working on them now

courtneyholcomb (approved):

PR #1785, opened 2025-07-18
Add interfaces for `LinkableElementSet` / `LinkableSpecResolver`

plypaul:
As hinted in https://github.com/dbt-labs/metricflow/pull/1784, the migration to the semantic graph involves adding interfaces for `LinkableElementSet` / `LinkableSpecResolver` and then providing different implementations (one using the existing classes, another using the semantic graph). This PR:

* Adds `AnnotatedSpec` that has similar properties as the `Linkable*` classes.
* Adds the `BaseLinkableElementSet` interface and makes `LinkableElementSet` implement that interface.
* Renames `LinkableSpecResolver`  to `LegacyLinkableSpecResolver` and makes `LinkableSpecResolver` an interface.
*  Migrates existing code to use those interfaces.

The lack of parallelism in how the interfaces are named is confusing, but the rename to address that ended up making too many conflicts, so we'll have to do the rename later.

These intermediate / temporary classes will need to be refined post-migration. Please view by commit.

### DIFF
--- a/.changes/unreleased/Under the Hood-20250723-123301.yaml
+++ b/.changes/unreleased/Under the Hood-20250723-123301.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Add interfaces for `LinkableElementSet` / `LinkableSpecResolver`
+time: 2025-07-23T12:33:01.545701-07:00
+custom:
+  Author: plypaul
+  Issue: "1785"


metricflow-semantics/metricflow_semantics/model/semantics/linkable_element_set_base.py
```diff
@@ -0,0 +1,257 @@
...
+    time_grain: Optional[ExpandedTimeGranularity]
+    date_part: Optional[DatePart]
+    metric_subquery_entity_link_names: Tuple[str, ...]
+
+    element_properties: Tuple[LinkableElementProperty, ...]
+    # The semantic model(s) where the element (e.g. the categorical dimension) was defined.
+    # There can be multiple models if it's a metric / derived metric that references multiple measures, and the join
+    # path from the measure to the dimension is different.
+    origin_semantic_model_names: Tuple[str, ...]
+    source_semantic_model_names: Tuple[str, ...]
```
courtneyholcomb: Can we better distinguish these two properties? From the names, they sound like the same thing.

If these are going away shortly anyway, though, may not be worth the effort.

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element_set_base.py
```diff
@@ -0,0 +1,257 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def union(self, *others: Self) -> Self:
+        """Intersect this set with the other set to determine possible group-by items."""
+        raise NotImplementedError
+
+
+@fast_frozen_dataclass()
+class AnnotatedSpec(SerializableDataclass):
```
courtneyholcomb, line 96: Is this basically just a generic `LinkableElement` that has all the possible attributes you might find on a `LinkableElement` for type convenience?

metricflow-semantics/metricflow_semantics/specs/instance_spec.py
```diff
@@ -46,6 +48,11 @@ def merge(*specs: Sequence[InstanceSpec]) -> List[InstanceSpec]:
     @property
     def qualified_name(self) -> str:
         """Return the qualified name of this spec. e.g. "user_id__country"."""
+        return self.structured_name.qualified_name
+
+    @property
+    def structured_name(self) -> StructuredLinkableSpecName:
```
courtneyholcomb: It seems like `StructuredLinkableSpecName` was only intended to support `LinkableSpecs`, not all `InstanceSpecs`. Do we need it for all `InstanceSpecs`? If so, we should rename it. If not, we should move this to the `LinkableSpec` class instead.

metricflow-semantics/metricflow_semantics/specs/metadata_spec.py
```diff
@@ -17,9 +19,16 @@ class MetadataSpec(InstanceSpec):
...
-    def qualified_name(self) -> str:  # noqa: D102
-        return f"{self.element_name}{DUNDER}{self.agg_type.value}" if self.agg_type else self.element_name
+    @override
+    @cached_property
+    def structured_name(self) -> StructuredLinkableSpecName:
+        if self.agg_type is not None:
+            return StructuredLinkableSpecName(
+                entity_link_names=(self.element_name,),
+                element_name=self.agg_type.value,
+            )
```
courtneyholcomb: If there is an `agg_type` present here, then you'll end up with an incorrect `element_name`. If we need this attribute on this class, we should add an `agg_type` attribute to `StructuredLinkableSpecName` and handle the dundering in the `qualified_name` attribute on that class.

metricflow/engine/metricflow_engine.py
```diff
@@ -684,32 +676,59 @@ def simple_dimensions_for_metrics(  # noqa: D102
...
-        for path_key, linkable_dimensions_tuple in linkable_element_set.path_key_to_linkable_dimensions.items():
-            for linkable_dimension in linkable_dimensions_tuple:
-                if LinkableElementProperty.METRIC_TIME in linkable_dimension.properties:
-                    dimension = self._build_metric_time_from_linkable_dimension(linkable_dimension)
+
+        for annotated_spec in linkable_element_set.annotated_specs:
+            properties = annotated_spec.properties
+            element_type = annotated_spec.element_type
+            if element_type is LinkableElementType.TIME_DIMENSION:
+                # Simple dimensions shouldn't show date part items.
```
courtneyholcomb, line 686: If we're adding a restriction that makes this function only applicable to simple dimensions, we should rename it to make that clear (e.g., `_filter_simple_linkable_dimensions()`)

courtneyholcomb (approved):

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element_set_base.py
plypaul: Any ideas? This was the best I got so far.

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element_set_base.py
plypaul: That's about right.

metricflow-semantics/metricflow_semantics/model/semantics/linkable_element_set_base.py
courtneyholcomb: Could keep the names that populated them! So then one of them would be `derived_from_semantic_models`, at least for consistency.
For something more clear but verbose, could be `origin_and_joined_semantic_models`.

plypaul: Updated.

metricflow-semantics/metricflow_semantics/specs/instance_spec.py
plypaul: Actually, this change was from an earlier iteration, so I've removed the commit.

metricflow-semantics/metricflow_semantics/specs/metadata_spec.py
plypaul: Same as https://github.com/dbt-labs/metricflow/pull/1785#discussion_r2226471336

metricflow/engine/metricflow_engine.py
plypaul: Updated.

PR #1791, opened 2025-07-23
Add semantic-graph node classes

plypaul:
This PR adds the set of nodes used in the semantic graph. Docstrings are in the process of refinement, but there should be enough context in there for review.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/experimental/semantic_graph/attribute_resolution/attribute_recipe_step.py
+++ b/metricflow-semantics/metricflow_semantics/experimental/semantic_graph/attribute_resolution/attribute_recipe_step.py
@@ -1,0 +1,108 @@
+from __future__ import annotations
+
+import logging
+from abc import ABC
+from functools import cached_property
+from typing import Optional
+
+from dbt_semantic_interfaces.type_enums import DatePart, TimeGranularity
+from typing_extensions import override
+
+from metricflow_semantics.collection_helpers.mf_type_aliases import AnyLengthTuple
+from metricflow_semantics.dag.mf_dag import DisplayedProperty
+from metricflow_semantics.experimental.dataclass_helpers import fast_frozen_dataclass
+from metricflow_semantics.experimental.mf_graph.comparable import Comparable, ComparisonKey
+from metricflow_semantics.experimental.mf_graph.graph_element import HasDisplayedProperty
+from metricflow_semantics.experimental.semantic_graph.model_id import SemanticModelId
+from metricflow_semantics.model.linkable_element_property import LinkableElementProperty
+from metricflow_semantics.model.semantics.linkable_element import LinkableElementType
+from metricflow_semantics.time.granularity import ExpandedTimeGranularity
+
+logger = logging.getLogger(__name__)
+
+
+@fast_frozen_dataclass(order=False)
+class AttributeRecipeStep(HasDisplayedProperty, Comparable):
+    """A step that should be added to the `AttributeRecipe`.
+
+    A path from a metric / measure node to an attribute node in the semantic graph describes the computation that is
+    required to generate the corresponding query. The computation is represented by a `AttributeRecipe`.
+
+    The nodes / edges in the path specify additions to the recipe. For example, an edge from one entity node to
+    another entity node would specify a step to join to a semantic model.
+
+    Currently, the recipe stores aggregated properties, and each step generally adds or sets some of those
+    properties. However, the recipe can be updated to include more context so that it can be used in the
+    `DataflowPlanBuilder`.
+    """
+
+    # Adds an element to the dunder name. For example, the edge from the `user` entity node to the
+    # `country_latest` attribute node would add the `country_latest` element.
+    add_dunder_name_element: Optional[str] = None
+    # Adds a property that describes that attribute. For example, the metric-time entity node would add the
+    # `METRIC_TIME` property.
+    add_properties: Optional[AnyLengthTuple[LinkableElementProperty]] = None
+    add_model_join: Optional[SemanticModelId] = None
+    add_entity_link: Optional[str] = None
+
+    # Set the element type. For example. visiting a time-dimension entity node would set this to `TIME_DIMENSION`.
+    set_element_type: Optional[LinkableElementType] = None
+    # To handle the case where time dimensions have a defined grain, the minimum time grain can be set to limit the
+    # attribute nodes accessible later in the path.
+    set_min_time_grain: Optional[TimeGranularity] = None
+    # With `set_min_time_grain`, `set_time_grain_access` can be used to check for valid paths. e.g. the node associated
+    # with a time dimension with a `month` defined grain can block access to edges that specify access to `day`.
+    set_time_grain_access: Optional[ExpandedTimeGranularity] = None
+    set_date_part_access: Optional[DatePart] = None
+    # Some edges prevent access of date part attributes (e.g. cumulative metrics)
+    set_deny_date_part: Optional[bool] = None
+
+    @override
+    @property
+    def comparison_key(self) -> ComparisonKey:
+        return (
+            self.add_dunder_name_element,
+            self.add_properties,
+            self.add_model_join,
+            self.set_min_time_grain.value if self.set_min_time_grain is not None else None,
+            self.set_element_type,
+            self.add_entity_link,
+            self.set_time_grain_access,
+            self.set_date_part_access.to_int() if self.set_date_part_access is not None else None,
+            self.set_deny_date_part,
+        )
+
+    @override
+    @cached_property
+    def displayed_properties(self) -> AnyLengthTuple[DisplayedProperty]:
+        properties = list(super().displayed_properties)
+
+        if self.add_dunder_name_element is not None:
+            properties.append(
+                DisplayedProperty("add_name", self.add_dunder_name_element),
+            )
+        for linkable_element_property_addition in self.add_properties or ():
+            properties.append(DisplayedProperty("add_prop", linkable_element_property_addition.name))
+        if self.add_model_join is not None:
+            properties.append(DisplayedProperty("add_model_join", self.add_model_join.model_name))
+        if self.set_min_time_grain is not None:
+            properties.append(DisplayedProperty("set_min_grain", self.set_min_time_grain.name))
+        if self.set_element_type is not None:
+            properties.append(DisplayedProperty("add_type", self.set_element_type.name))
+        if self.add_entity_link is not None:
+            properties.append(DisplayedProperty("add_entity_link", self.add_entity_link))
+        if self.set_time_grain_access is not None:
+            properties.append(DisplayedProperty("set_time_grain", self.set_time_grain_access.name))
+        if self.set_date_part_access is not None:
+            properties.append(DisplayedProperty("set_date_part", self.set_date_part_access.name))
+        if self.set_deny_date_part is not None:
+            properties.append(DisplayedProperty("set_deny_date_part", self.set_deny_date_part))
+        return tuple(properties)
+
+
+class AttributeRecipeStepProvider(HasDisplayedProperty, ABC):
+    """Interface for a class that provides a step that can be appended to a recipe."""
+
+    @cached_property
+    def recipe_step_to_append(self) -> AttributeRecipeStep:  # noqa: D102
+        return AttributeRecipeStep()
--- a/metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/attribute_nodes.py
+++ b/metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/attribute_nodes.py
@@ -1,0 +1,172 @@
+from __future__ import annotations
+
+from abc import ABC
+from functools import cached_property
+
+from dbt_semantic_interfaces.type_enums import DatePart, TimeGranularity
+from typing_extensions import override
+
+from metricflow_semantics.experimental.mf_graph.comparable import ComparisonKey
+from metricflow_semantics.experimental.mf_graph.formatting.dot_attributes import (
+    DotColor,
+    DotNodeAttributeSet,
+)
+from metricflow_semantics.experimental.mf_graph.graph_labeling import MetricflowGraphLabel
+from metricflow_semantics.experimental.mf_graph.node_descriptor import MetricflowGraphNodeDescriptor
+from metricflow_semantics.experimental.ordered_set import FrozenOrderedSet, OrderedSet
+from metricflow_semantics.experimental.semantic_graph.attribute_resolution.attribute_recipe_step import (
+    AttributeRecipeStep,
+)
+from metricflow_semantics.experimental.semantic_graph.nodes.node_labels import (
+    GroupByAttributeLabel,
+    KeyAttributeLabel,
+    TimeClusterLabel,
+)
+from metricflow_semantics.experimental.semantic_graph.sg_constant import ClusterNameFactory
+from metricflow_semantics.experimental.semantic_graph.sg_interfaces import SemanticGraphNode
+from metricflow_semantics.experimental.singleton_decorator import singleton_dataclass
+from metricflow_semantics.model.linkable_element_property import LinkableElementProperty
+from metricflow_semantics.model.semantics.linkable_element import LinkableElementType
+from metricflow_semantics.naming.linkable_spec_name import StructuredLinkableSpecName
+from metricflow_semantics.time.granularity import ExpandedTimeGranularity
+
+
+@singleton_dataclass(order=False)
+class AttributeNode(SemanticGraphNode, ABC):
+    """ABC for attribute nodes. See `SemanticGraph` for additional context on usage."""
+
+    attribute_name: str
+
+    @cached_property
+    @override
+    def comparison_key(self) -> ComparisonKey:
+        return (self.attribute_name,)
+
+    @cached_property
+    @override
+    def node_descriptor(self) -> MetricflowGraphNodeDescriptor:
+        return MetricflowGraphNodeDescriptor.get_instance(node_name=f"{self.__class__.__name__}({self.attribute_name})")
+
+    @override
+    def as_dot_node(self, include_graphical_attributes: bool) -> DotNodeAttributeSet:
+        dot_node = super(AttributeNode, self).as_dot_node(include_graphical_attributes)
+        if include_graphical_attributes:
+            dot_node = dot_node.with_attributes(color=DotColor.SALMON_PINK)
+        return dot_node
+
+    @cached_property
+    @override
+    def labels(self) -> OrderedSet[MetricflowGraphLabel]:
+        return FrozenOrderedSet((GroupByAttributeLabel(),))
+
+    @cached_property
+    @override
+    def recipe_step_to_append(self) -> AttributeRecipeStep:
+        return AttributeRecipeStep(
+            add_dunder_name_element=self.attribute_name,
+        )
+
+
+@singleton_dataclass(order=False)
+class TimeAttributeNode(AttributeNode):
+    """An attribute node that corresponds to the different time grains available for querying time dimensions.
+
+    e.g. the graph would contain instances for `day`, `dow`, `month`...
+    """
+
+    element_property_additions: FrozenOrderedSet[LinkableElementProperty]
+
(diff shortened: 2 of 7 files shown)


plypaul:
As a heads up, the `get_instance()` methods are in place as I'm in the process of deprecating the `@singleton_dataclass` decorator.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/attribute_resolution/attribute_recipe_step.py
```diff
@@ -0,0 +1,108 @@
...
+from metricflow_semantics.experimental.semantic_graph.model_id import SemanticModelId
+from metricflow_semantics.model.linkable_element_property import LinkableElementProperty
+from metricflow_semantics.model.semantics.linkable_element import LinkableElementType
+from metricflow_semantics.time.granularity import ExpandedTimeGranularity
+
+logger = logging.getLogger(__name__)
+
+
+@fast_frozen_dataclass(order=False)
+class AttributeRecipeStep(HasDisplayedProperty, Comparable):
```
courtneyholcomb, line 25: How would you explain the difference between an "attribute" and an "entity" in the semantic graph?

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/attribute_resolution/attribute_recipe_step.py
```diff
@@ -0,0 +1,108 @@
...
+    add_entity_link: Optional[str] = None
+
+    # Set the element type. For example. visiting a time-dimension entity node would set this to `TIME_DIMENSION`.
+    set_element_type: Optional[LinkableElementType] = None
+    # To handle the case where time dimensions have a defined grain, the minimum time grain can be set to limit the
+    # attribute nodes accessible later in the path.
+    set_min_time_grain: Optional[TimeGranularity] = None
+    # With `set_min_time_grain`, `set_time_grain_access` can be used to check for valid paths. e.g. the node associated
+    # with a time dimension with a `month` defined grain can block access to edges that specify access to `day`.
+    set_time_grain_access: Optional[ExpandedTimeGranularity] = None
```
courtneyholcomb, line 55: It's not clear to me yet why both `set_min_time_grain` and `set_time_grain_access` are necessary, but maybe it will become clear up the stack?

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/sg_interfaces.py
```diff
@@ -0,0 +1,177 @@
...
+logger = logging.getLogger(__name__)
+
+
+class SemanticGraphNode(MetricflowGraphNode, AttributeRecipeStepProvider, MetricFlowPrettyFormattable, ABC):
+    """A node in the semantic graph.
+
+    At the top level, the semantic graph has two types of nodes: entity nodes and attribute nodes. Entity nodes have
+    other entity nodes or attribute nodes as successors while attribute nodes do not have any successors (leaf nodes).
+
+    The entity nodes in the semantic graph are not directly mapped from entities configured in the semantic manifest.
```
courtneyholcomb, line 38: So it seems like we're changing the meaning of "entity" in MF. What does it mean now? Something that can be joined?
Would it be better to give it a different name, since the name entity is already exposed in the YAML and therefore fairly permanent in its current usage?

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/sg_interfaces.py
```diff
@@ -0,0 +1,177 @@
...
+    The one that corresponds to an entity in the manifest are instances of `ConfiguredEntityNode`. However, there
+    are other entity nodes that model other relationships. For example, `metric_time` is represented by the
+    metric-time entity node, and there is a time-entity node that relates time dimensions and metric time. to the
```
courtneyholcomb: nit: extra period

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/sg_interfaces.py
```diff
@@ -0,0 +1,177 @@
...
+from metricflow_semantics.experimental.semantic_graph.attribute_resolution.attribute_recipe_step import (
+    AttributeRecipeStepProvider,
+)
+from metricflow_semantics.mf_logging.pretty_formattable import MetricFlowPrettyFormattable
+from metricflow_semantics.mf_logging.pretty_formatter import PrettyFormatContext
+
+logger = logging.getLogger(__name__)
+
+
+class SemanticGraphNode(MetricflowGraphNode, AttributeRecipeStepProvider, MetricFlowPrettyFormattable, ABC):
```
courtneyholcomb, line 32: nit: just noticing that we have mixed casing for the word MetricFlow in these class names. We should try to stick to one or the other (Metricflow or MetricFlow), and I think we've generally aligned on MetricFlow in the past.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/sg_interfaces.py
```diff
@@ -0,0 +1,177 @@
...
+
+    @override
+    def inverse(self) -> MutableSemanticGraph:
+        inverse_graph = MutableSemanticGraph.create()
+        for edge in self.edges:
+            inverse_graph.add_edge(edge.inverse)
+        return inverse_graph
+
+    @override
+    def as_sorted(self) -> MutableSemanticGraph:
```
courtneyholcomb, line 169: How does this change the graph? What was the original graph sorted by and what is it sorted by now?

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/attribute_nodes.py
```diff
@@ -0,0 +1,172 @@
...
+from metricflow_semantics.experimental.semantic_graph.sg_interfaces import SemanticGraphNode
+from metricflow_semantics.experimental.singleton_decorator import singleton_dataclass
+from metricflow_semantics.model.linkable_element_property import LinkableElementProperty
+from metricflow_semantics.model.semantics.linkable_element import LinkableElementType
+from metricflow_semantics.naming.linkable_spec_name import StructuredLinkableSpecName
+from metricflow_semantics.time.granularity import ExpandedTimeGranularity
+
+
+@singleton_dataclass(order=False)
+class AttributeNode(SemanticGraphNode, ABC):
```
courtneyholcomb, line 35: Are attributes always group by attributes? If so, would be helpful to include that in the name to make this feel more tangible.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/attribute_nodes.py
```diff
@@ -0,0 +1,172 @@
...
+
+    @staticmethod
+    def get_instance_for_date_part(date_part: DatePart) -> TimeAttributeNode:  # noqa: D102
+        return TimeAttributeNode(
+            attribute_name=StructuredLinkableSpecName.date_part_suffix(date_part),
+            element_property_additions=FrozenOrderedSet((LinkableElementProperty.DATE_PART,)),
+        )
+
+    @staticmethod
+    def get_instance_for_expanded_time_grain(  # noqa: D102
```
courtneyholcomb, line 94: Curious why we need both `TimeGranularity` and `ExpandedTimeGranularity` attributes? Can't `ExpandedTimeGranularity` encompass both? Are you only using `ExpandedTimeGranularity` for custom grains?

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/attribute_nodes.py
```diff
@@ -0,0 +1,172 @@
...
+        expanded_time_grain: ExpandedTimeGranularity,
+    ) -> TimeAttributeNode:
+        return TimeAttributeNode(
+            attribute_name=expanded_time_grain.name,
+            element_property_additions=FrozenOrderedSet((LinkableElementProperty.DERIVED_TIME_GRANULARITY,)),
```
courtneyholcomb, line 99: Relatedly, wondering why `get_instance_for_expanded_time_grain()` assigns the `DERIVED_TIME_GRANULARITY` property but `get_instance_for_time_grain()` does not.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/attribute_nodes.py
```diff
@@ -0,0 +1,172 @@
...
+        )
+
+    @cached_property
+    @override
+    def labels(self) -> OrderedSet[MetricflowGraphLabel]:
+        return super(TimeAttributeNode, self).labels.union((TimeClusterLabel.get_instance(),))
+
+
+@singleton_dataclass(order=False)
+class KeyAttributeNode(AttributeNode):
```
courtneyholcomb, line 117: Can we include `Entity` in the class name for a little bit more clarity?

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/attribute_nodes.py
```diff
@@ -0,0 +1,172 @@
...
+    """Represents the attribute corresponding to the value of a configured entity.
+
+    In the current MF interface, a "query for an entity" means a query for the column values associated with the
+    configured entity in a semantic model. To avoid confusion between entities in the semantic graph and entities
+    configured in semantic models, "query for an entity" is phrased as "query for an entity key" in the semantic graph
+    module.
+
+    Consider using the phrase `entity value` instead.
```
courtneyholcomb, line 125: I do think "entity value" would be more clear. Or even "configured entity value". "Key" points more to the join relationship than to the actual column value.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/entity_nodes.py
```diff
@@ -0,0 +1,383 @@
...
+        return FrozenOrderedSet((LocalModelLabel.get_instance(),))
+
+    @cached_property
+    @override
+    def recipe_step_to_append(self) -> AttributeRecipeStep:
+        return AttributeRecipeStep(add_model_join=self.model_id)
+
+
+@singleton_dataclass(order=False)
+class TimeDimensionNode(SemanticGraphNode):
```
courtneyholcomb, line 172: Why are time dimensions considered entities but categorical dimensions are not? Is it because you can join one time dimension to another?

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/entity_nodes.py
```diff
@@ -0,0 +1,383 @@
...
+    """An entity representing a time dimension configured in a semantic model.
+
+    In the semantic graph, time dimensions are represented as entities that are related to the time entity. The time
+    entity has edges to the different queryable grain-related attributes (e.g. `day`, `month`).
+    """
+
+    dimension_name: str
```
courtneyholcomb, line 179: Does this represent the element name(`country_latest`) or the full name (`listing__country_latest`)?

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/entity_nodes.py
```diff
@@ -0,0 +1,383 @@
...
+        return AttributeRecipeStep(
+            add_dunder_name_element=self.metric_name,
+        )
+
+
+@singleton_dataclass(order=False)
+class BaseMetricNode(MetricNode):
+    """Represents a metric without successors to other metric nodes.
+
+    All non-derived metrics are represented by this node.
```
courtneyholcomb, line 301: What about ratio & conversion metrics?

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/entity_nodes.py
```diff
@@ -0,0 +1,383 @@
...
+        )
+
+    @cached_property
+    @override
+    def labels(self) -> OrderedSet[MetricflowGraphLabel]:
+        return super(DerivedMetricNode, self).labels.union((DerivedMetricLabel.get_instance(),))
+
+
+@singleton_dataclass(order=False)
+class MeasureNode(SemanticGraphNode):
```
courtneyholcomb, line 343: Not really related to this PR, but curious how things will change as we move to a world without measures.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/sg_constant.py
```diff
@@ -0,0 +1,42 @@
...
+    TIME_DIMENSION: Final[str] = "time_dimension"
+
+    @override
+    def __init__(self) -> None:
+        """Can't override `__new__` as that results in a type-check error."""
+        raise TypeError(LazyFormat("Class should not be instantiated", cls=self.__class__))
+
+    @staticmethod
+    def get_name_for_configured_entity(entity_name: str, model_id: SemanticModelId) -> str:  # noqa: D102
+        return f"{model_id}.{entity_name}"
```
courtneyholcomb, line 37: I like that we're including the semantic model in the name representing the entity since we haven't done a always differentiated well between entities of the same name in the past - excited to see how this is used

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/attribute_resolution/attribute_recipe_step.py
plypaul: So if you look at the successors (edges going away from a node), entities can have successors that are attributes or other entities. Attributes do not have successors. In the context of a group-by item, attributes represent the last element in the dunder name.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/attribute_resolution/attribute_recipe_step.py
plypaul: It's one way of reducing the number of nodes / edges in the graph, but the example case is that a time dimension in the semantic model is represented by an entity that relates to the time entity (time spine). A time dimension with a given name is represented by the same time-dimension entity node. However, since time dimensions in a semantic model specify a grain, not all time attributes should be accessible, so `set_min_time_grain` is used to limit access to some time attributes during graph traversal.

Other ways to handle that case would have been to have a different time-dimension node for every possible time-dimension name / time grain combination, or add a composite / associative entity node.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/attribute_nodes.py
plypaul: Currently, metrics / measures fit into the model as attributes of an entity as well. But for this first pass, the group-by items are the main attributes of interest.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/attribute_nodes.py
plypaul: So in the graph, both `TimeGranularity` and `ExpandedTimeGranularity` are modeled the same way. These are just helpers that are useful in building the graph.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/attribute_nodes.py
plypaul: Sure, but let me handle that in the PR that does the `singleton_dataclass` migration.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/entity_nodes.py
plypaul: Should be straightforward - that can be handled by replacing this node with a `BaseMetric` node.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/sg_interfaces.py
plypaul: Good point - let me add that to the end.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/sg_interfaces.py
plypaul: I'm up for a different name - let's chat.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/sg_interfaces.py
plypaul: Updated.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/sg_interfaces.py
plypaul: The nodes and edges in the graph are in insertion order by default. Sorting is only used in tests to keep snapshots consistent.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/attribute_nodes.py
plypaul: That is based on the output of the current resolver - `DERIVED_TIME_GRANULARITY` is always set for the custom grains.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/entity_nodes.py
plypaul: As it is modeled now - time dimensions are entities that join to the time spine. The entity key is the date itself.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/entity_nodes.py
plypaul: Since the inputs are measures, ratio and conversion metrics are considered base metrics here.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/entity_nodes.py
plypaul: It is the element name. I've been using `dunder_name` when it is not the element name.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/sg_interfaces.py
courtneyholcomb: Sounds good - doesn't need to block this PR! Can change names at the top of the stack.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/sg_interfaces.py
courtneyholcomb: But what is the output sorting? Alphabetical?

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/entity_nodes.py
courtneyholcomb: https://docs.getdbt.com/docs/build/ratio
Are these docs wrong? It looks like ratio metrics take metric inputs but I know that wasn't always the case.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/sg_interfaces.py
plypaul: The order is determined by implementing classes (override `comparison_key`). Right now, the key that I've put in makes it so that if you sorted a list of nodes that are of heterogenous types, they are first ordered alphabetically by the class name, then ordered by the comparison order of the attributes / fields of that node.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/nodes/entity_nodes.py
plypaul: Seems like the rules have changed ,and ratio metrics now have input metrics. However, that doesn't change things much for the implementation.

PR #1797, opened 2025-07-25
Fix ambiguous entity path resolution for PRIMARY/FOREIGN entity relationships

### LINKED ISSUE #1780: [Bug] MetricFlow fails to resolve equivalent dimension paths across PRIMARY and FOREIGN entity relationships

### Current Behavior

MetricFlow fails to resolve queries that use the same dimension across PRIMARY and FOREIGN entity metrics. For example, `businessunit__businessunit_name` (from a PRIMARY job entity metric) and `job__businessunit__businessunit_name` (from a FOREIGN job entity metric) are treated as incompatible, even though they are the same dimension. Interestingly, MetricFlow has no problem resolving the same dimension when accessed by multiple different FOREIGN entity metrics - it only fails when one is PRIMARY. Therefore, a possible workaround is to never use PRIMARY entity measures/metrics, and to move all measures into a boilerplate "companion" entity with a foreign relationship. But this would cause us to double the number of entities in our semantic layer.

## Why This Is A Bug

1. **Logical Equivalence**: The dimension paths are logically equivalent - they reference the same underlying data through the same relationships.

2. **Inconsistent Behavior**: MetricFlow can resolve the paths when both metrics use **different** FOREIGN relationships but fails when mixing PRIMARY and FOREIGN, despite the semantic equivalence.

3. **Unnecessary Boilerplate**: The current behavior forces users to create redundant "companion entities" for every entity just to enable basic row counting alongside other metrics.

4. **Path Resolution Logic**: If MetricFlow can successfully join and coalesce `job__businessunit__businessunit_name` from two different FOREIGN relationships (as shown in the working SQL), it should be able to recognize that `businessunit__businessunit_name` (from PRIMARY) and `job__businessunit__businessunit_name` (from FOREIGN) are equivalent.

## Generated SQL from Workaround

When using the companion entity pattern, MetricFlow successfully generates SQL that shows it understands how to join these dimensions:

```sql
SELECT
  COALESCE(subq_12.job__businessunit__businessunit_name, subq_23.job__businessunit__businessunit_name) 
    AS job__businessunit__businessunit_name,
  MAX(subq_12.job_count) AS job_count,
  MAX(subq_23.material_costs) AS material_costs
FROM (
  -- Job count subquery using businessunit__businessunit_name
  SELECT
    subq_8.businessunit__businessunit_name AS job__businessunit__businessunit_name,
    SUM(job_companion_src.companion_count) AS job_count
      FROM SCHEMA.DATABASE.staging_job_companion_table job_companion_src
  LEFT OUTER JOIN (
    SELECT
      job_cte.job AS job,
      businessunit_cte.businessunit_name AS businessunit__businessunit_name
    FROM job_cte job_cte
    LEFT OUTER JOIN businessunit_cte businessunit_cte
      ON job_cte.businessunit = businessunit_cte.businessunit
  ) subq_8
  ON job_companion_src.job_ref = subq_8.job
  GROUP BY subq_8.businessunit__businessunit_name
) subq_12
FULL OUTER JOIN (
  -- Material costs subquery using same dimension
  SELECT
    subq_19.businessunit__businessunit_name AS job__businessunit__businessunit_name,
    SUM(subq_14.material_total_cost) AS material_costs
  FROM (
    SELECT
      CONCAT(_tenant_id, '|', job_id) AS job,
      total_cost AS material_total_cost
    FROM SCHEMA.DATABASE.staging_material_costs_table material_costs_src
  ) subq_14
  LEFT OUTER JOIN (
    SELECT
      job_cte.job AS job,
      businessunit_cte.businessunit_name AS businessunit__businessunit_name
    FROM job_cte job_cte
    LEFT OUTER JOIN businessunit_cte businessunit_cte
      ON job_cte.businessunit = businessunit_cte.businessunit
  ) subq_19
  ON subq_14.job = subq_19.job
  GROUP BY subq_19.businessunit__businessunit_name
) subq_23
ON subq_12.job__businessunit__businessunit_name = subq_23.job__businessunit__businessunit_name
GROUP BY COALESCE(subq_12.job__businessunit__businessunit_name, subq_23.job__businessunit__businessunit_name)
```

### Expected Behavior

MetricFlow should recognize that `businessunit__businessunit_name` and `job__businessunit__businessunit_name` refer to the same dimension and successfully resolve the query. The paths are logically equivalent:
- From PRIMARY job entity: `businessunit__businessunit_name`
- From FOREIGN job entity: `job__businessunit__businessunit_name`

Both ultimately reference the same `businessunit_name` field through the job's relationship to businessunit.

### Semantic Model Configuration

```yaml
# models/semantic_models/jobs_v2.yml
semantic_models:
  - name: jobs_v2
    model: ref('staging_job_table')
    entities:
      - name: job
        type: primary
        expr: job_id
      - name: businessunit
        type: foreign
        expr: businessunit_id
    measures:
      - name: job_count
        agg: count_distinct
        expr: job_id

# models/semantic_models/material_costs.yml
semantic_models:
  - name: material_costs
    model: ref('staging_material_costs_table')
    entities:
      - name: job
        type: foreign
        expr: job_id
    measures:
      - name: material_costs
        agg: sum
        expr: total_cost
```

### Failing Query

```bash
mf query --metrics job_count,material_costs \
  --group-by businessunit__businessunit_name \
  --explain
```

## Actual Behavior

The query fails with the following error:

```
ERROR: Got error(s) during query resolution.

Error #1:
  Message:

    Query(['job_count', 'material_costs']) is built from:

      Metric('job_count'), Metric('material_costs').
    However, the given input does not match to a common item that is available to those parents:

      {
        "Matching items for: Metric('job_count')": [
          'businessunit__businessunit_name',
        ],
        "Matching items for: Metric('material_costs')": [
          'job__businessunit__businessunit_name',
        ],
      }

    For time dimension inputs, please specify a time grain as ambiguous resolution only allows resolution 
    when the parents have the same defined time gain.

  Query Input:

    businessunit__businessunit_name

  Issue Location:

    [Resolve Query(['job_count', 'material_costs'])]
```

## Current Workaround

Creating a "companion entity" pattern where we add a trivial foreign entity relationship to job:

```yaml
# models/semantic_models/job_companion.yml
semantic_models:
  - name: job_companion
    model: ref('staging_job_companion_table')  # Simple SELECT *, 1 as companion_count FROM job
    entities:
      - name: job
        type: foreign  # Forces consistent path resolution
        expr: job_id
    measures:
      - name: job_companion_count
        agg: sum
        expr: companion_count

# Updated metrics to use companion
metrics:
  - name: job_count
    type: simple
    measure: job_companion_count  # Now uses FOREIGN relationship
```

With this pattern, the following query works:

```bash
mf query --metrics job_count,material_costs \
  --group-by job__businessunit__businessunit_name \
  --explain
```

This succeeds because both metrics now traverse through FOREIGN relationships, giving them consistent paths.

### Relevant log output

```shell

```

### Environment

```markdown
- MetricFlow version: 0.8.2
- dbt-core version: 1.9.4
- dbt-snowflake: 1.8.4
- Database: Snowflake
- Python version: 3.11
```

snowflake

## Impact

**We cannot use measures from primary entities in queries that combine metrics from different semantic models.** This forces us to create artificial "companion" entities with foreign relationships for every entity, effectively doubling our semantic model count. What should be a simple `COUNT(DISTINCT job_id)` now requires an entire separate model and staging table.

## Suggested Solution

MetricFlow should enhance its dimension path resolution to recognize when paths are semantically equivalent, particularly when:
1. A PRIMARY entity's direct dimension path (e.g., `businessunit__businessunit_name`)
2. Is equivalent to a FOREIGN entity's path through the same entity (e.g., `job__businessunit__businessunit_name`)

The resolution logic should treat these as the same dimension for query planning purposes, similar to how it already handles multiple FOREIGN relationships to the same dimension.

## Additional Context

- Derived metrics that depend on measures from primary entities also require this workaround

dstein-st:
[ai-assisted]

see: https://github.com/dbt-labs/metricflow/issues/1780

# Fix MetricFlow Issue #1780: Enable PRIMARY/FOREIGN Entity Path Resolution

## Description

This PR fixes Issue #1780 where MetricFlow fails to resolve queries that use the same dimension accessed through different entity paths when one metric uses a PRIMARY entity and another uses a FOREIGN entity.

## Problem

Previously, MetricFlow would fail with "the given input does not match to a common item" when querying:
```bash
mf query --metrics job_count_primary,material_costs --group-by job__businessunit__businessunit_name
```

This occurred because:
- `job_count_primary` metric (using PRIMARY entity) accesses the dimension as `businessunit__businessunit_name`
- `material_costs` metric (using FOREIGN entity) accesses it as `job__businessunit__businessunit_name`
- The system didn't recognize these as semantically equivalent paths to the same dimension

## Solution

The fix implements semantic equivalence recognition between PRIMARY and FOREIGN entity paths while preserving the user's requested dimension naming in output columns. This is achieved through three key changes:

1. **Enhanced Pattern Matching** (`entity_link_pattern.py`):
   - Added suffix matching for DimensionSpecs when the pattern has exactly one more entity link than the candidate
   - This allows `job__businessunit__businessunit_name` to match `businessunit__businessunit_name`
   - Applied only to dimensions (not entities) to maintain precise semantics

2. **Semantic Equivalence Recognition** (`linkable_element_set.py`):
   - Added `_path_keys_are_semantically_equivalent` to recognize equivalence when paths differ by exactly one entity link
   - Enhanced `intersection_by_path_key` to properly group semantically equivalent elements during intersection operations
   - Selects the longer (FOREIGN) path as the representative to ensure universal accessibility

3. **Output Column Name Preservation** (`query_resolver.py`):
   - Added logic in `_resolve_group_by_item_input` to preserve the original requested entity links
   - When resolution uses semantic equivalence, creates a spec with the user's requested entity links for output naming
   - Ensures users get column names matching their query (e.g., `job__businessunit__businessunit_name`)

## Technical Details

1. **Conservative Approach**: The suffix matching only activates when there's exactly a 1-level difference in entity links, preventing overly broad matching that could break existing behavior.

2. **Time Dimension Handling**: Fixed `_match_time_granularities` to properly pass through non-time dimension specs, resolving issues with the new matching logic.

3. **Test Impact**: 
   - Updated `test_non_resolvable_ambiguous_entity_path_due_to_mismatch` which now succeeds (positive change)
   - Fixed SCD integration tests that were affected by column naming changes
   - Updated dimension listing test snapshot to reflect newly accessible dimensions
   - Added new test specifically for PRIMARY/FOREIGN suffix matching

## Testing

- ✅ Original failing query now succeeds
- ✅ All 770 existing tests pass
- ✅ Added new test for PRIMARY/FOREIGN suffix matching
- ✅ Linting passes
- ✅ Output column names match user's requested format

## Breaking Changes

No breaking changes. Queries that previously failed due to PRIMARY/FOREIGN entity path differences will now succeed. All existing working queries continue to work as before.

## Example

Before this fix:
```bash
# This would fail
mf query --metrics job_count_primary,material_costs --group-by job__businessunit__businessunit_name
# Error: "the given input does not match to a common item"
```

After this fix:
```bash
# This now succeeds and preserves the requested column name
mf query --metrics job_count_primary,material_costs --group-by job__businessunit__businessunit_name
# Successfully generates SQL with output column: job__businessunit__businessunit_name
```

## Files Changed

- `metricflow-semantics/metricflow_semantics/specs/patterns/entity_link_pattern.py`
  - Modified `_match_entity_links` to add PRIMARY/FOREIGN suffix matching for dimensions
  - Fixed `_match_time_granularities` to properly handle non-time dimension specs
  
- `metricflow-semantics/metricflow_semantics/model/semantics/linkable_element_set.py`
  - Added `_path_keys_are_semantically_equivalent` for recognizing equivalent paths
  - Rewrote `intersection_by_path_key` to group semantically equivalent elements
  
- `metricflow-semantics/metricflow_semantics/query/query_resolver.py`
  - Added logic to preserve original entity links for output column naming
  
- Test updates:
  - Updated `test_ambiguous_entity_path.py` to reflect more flexible behavior
  - Added `test_primary_foreign_suffix_matching.py` for new functionality
  - Updated dimension listing snapshot with newly accessible dimensions

Fixes #1780

### DIFF
--- a/.changes/unreleased/Fixes-20250728-125748.yaml
+++ b/.changes/unreleased/Fixes-20250728-125748.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Fix queries failing when the same dimension has PRIMARY relationship to one metric and FOREIGN to another.
+time: 2025-07-28T12:57:48.250577-04:00
+custom:
+    Author: dstein-st
+    Issue: "1780"


plypaul:
Thanks for contributing the pull request! There is a bit of a backlog, so we'll update when it has cleared.

PR #1801, opened 2025-07-28
Add subgraph generator for time-related elements

plypaul:
This PR adds 2 subgraph generators for time-related elements in the semantic manifest: one for time dimensions and another for the time entity (similar / related to the time spine).


metricflow-semantics/metricflow_semantics/experimental/semantic_graph/builder/time_dimension_subgraph.py
```diff
@@ -0,0 +1,95 @@
...
+            min_referenced_time_grain: tuple(
+                time_grain
+                for time_grain in TimeGranularity
+                if time_grain.to_int() >= min_referenced_time_grain.to_int()
+            )
+            for min_referenced_time_grain in TimeGranularity
+        }
+
+    @cached_property
+    def _time_grain_to_applicable_date_parts(self) -> Mapping[TimeGranularity, AnyLengthTuple[DatePart]]:
```
courtneyholcomb, line 91: There is an [existing property for this](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/99a38b6caf1605d8728c684be67e3e34a06a5dc2/dbt_semantic_interfaces/type_enums/date_part.py#L55-L56) if that's helpful! Would be good to dedupe logic

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/builder/time_entity_subgraph.py
```diff
@@ -0,0 +1,137 @@
...
+logger = logging.getLogger(__name__)
+
+
+class TimeEntitySubgraphGenerator(SemanticSubgraphGenerator):
+    """Generates the time-entity subgraph.
+
+    * The time-entity subgraph consists of the time-entity node with edges to various  attribute
+      nodes. (e.g. `day`, `dow`, `custom_year`).
+    * An alternative approach would be to use a virtual semantic-model and have time be a configured entity.
+    * There is a bug in the current resolver that shows time gain of the defined time dimension as available to
```
courtneyholcomb, line 43: Should we consider this a bug? They'll get a warning in the parser if they don't have a time spine configured for their smallest defined time dimension grain, and an appropriate error at query time if they try to use a smaller time spine than they have available. They can still query dimensions with smaller grains than the time spines configured, they just won't be able to use any functionality that requires joining to time spine.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/builder/time_entity_subgraph.py
```diff
@@ -0,0 +1,137 @@
...
+                for time_grain in TimeGranularity
+                if time_grain.to_int() >= min_referenced_time_grain.to_int()
+            )
+            for min_referenced_time_grain in TimeGranularity
+        }
+
+    @cached_property
+    def _time_grain_to_applicable_date_parts(self) -> Mapping[TimeGranularity, AnyLengthTuple[DatePart]]:
+        return {
+            time_grain: tuple(date_part for date_part in DatePart if date_part.to_int() >= time_grain.to_int())
```
courtneyholcomb, line 130: same here for the DatePart helper

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/builder/time_entity_subgraph.py
```diff
@@ -0,0 +1,137 @@
...
+                        head_node=attribute_node,
+                        recipe_step=AttributeRecipeStep(
+                            add_properties=(LinkableElementProperty.DERIVED_TIME_GRANULARITY,),
+                            set_time_grain_access=expanded_time_grain,
+                        ),
+                    )
+                )
+
+    @cached_property
+    def _time_grain_to_queryable_time_grains(self) -> Mapping[TimeGranularity, AnyLengthTuple[TimeGranularity]]:
```
courtneyholcomb, line 117: do we need one of these helpers for custom grains?

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/builder/time_dimension_subgraph.py
plypaul: Thanks for the heads up, but the one there actually has perf issues because it's not cached. Normally, this is not an issue, but the semantic graph implementation is fast enough that small things like that make a difference.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/builder/time_entity_subgraph.py
plypaul: The issue I saw was that a dimension like `metric__day` would show up in the API as an available dimension, but you'd get an error when you try to query for it.

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/builder/time_entity_subgraph.py
plypaul: So this is for translating what's in the YAML config to the appropriate representation in the semantic graph. Can custom grains be specified in the YAML?

metricflow-semantics/metricflow_semantics/experimental/semantic_graph/builder/time_entity_subgraph.py
plypaul: Yeah, I think I need to do a pass after this stack to consolidate these lookups.

PR #1826, opened 2025-09-04
Update offset implementation

courtneyholcomb:


courtneyholcomb:
> I need to get a little more familiar with the feature, but it seems odd to have conditional behavior with how a time offset is applied depending on the query.

@plypaul the offset is only applied once (unless there is a nested offset). I think what you may be looking at is that the time spine join might be applied more than once. This is because you could have an offset metric that also uses the `join_to_timespine` key in the YAML. The `join_to_timespine` join should always happen post-aggregation, so if there is also an offset, we need to join both before and after aggregation. But if the offset should be applied post-agg, we can combine those into one join.

PR #1829, opened 2025-09-09
Bump dsi version to 0.9.1

theyostalservice:
This does not bump the minimum version for
metric-semantics since core hasn't been bumped yet.

I'm just bumping this first since 0.9.1 is purely
additive (it should be fully backward compatible), so this should allow for a safer roll out.

Several measure input fields are made optional in the spec, but we don't actually support them yet (@plypaul is working on that), so for now, I added assertions that the measures are provided in order to keep our runtime safe and our typechecker happy.  These assertions can be removed as new behaviors are added.

For a sanity check, when I ran some of the make test options, I saw this line among all the output:

```Successfully installed dbt-adapters-1.16.6 dbt-common-1.29.0 dbt-core-1.10.11 dbt-protos-1.0.348 dbt-semantic-interfaces-0.9.1 jinja2-3.1.6 protobuf-6.32.0 pympler-1.1 termcolor-3.1.0```

### DIFF
--- a/metricflow-semantics/requirements-files/requirements.txt
+++ b/metricflow-semantics/requirements-files/requirements.txt
@@ -1,7 +1,7 @@
 # Always support a range of production DSI versions capped at the next breaking version in metricflow-semantics.
 # This allows us to sync new, non-breaking changes to dbt-core without getting a version mismatch in dbt-mantle,
 # which depends on a specific commit of DSI.
-dbt-semantic-interfaces>=0.9.0, <0.10.0
+dbt-semantic-interfaces>=0.9.1, <0.10.0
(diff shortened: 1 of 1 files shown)


courtneyholcomb (approved):
Looks good - I would recommend bumping the minimum version for `metricflow-semantics` in this same PR, though. In mantle we have `metricflow-semantics` pinned to a specific commit so you won't be breaking anything if you update it here. In mantle the version will actually be set by core since core has a more specific pinned version. But in MF, you might get some weird behavior for `metricflow-semantics` tests if they have a different version of DSI than `metricflow` proper.

theyostalservice:
alright, I'll do that!  thank you!

PR #1835, opened 2025-09-09
Improve engine initialization speed via semantic graph

plypaul:
This PR removes the option to use the legacy resolver when initializing the `MetricFlowEngine` / `SemanticManifestLookup` in favor of the semantic-graph-based resolver. For complex semantic manifests, this improves initialization times by >100x for complex manifests.

The new resolver includes a fix for detecting ambiguous join paths. For such detection, a semantic model with a `UNIQUE` entity should be considered as a candidate model for joining on that entity. Previously, it was not, and the generated query may use any of the possible join paths and give inconsistent results. The fix detects the ambiguous path in such queries and raises an error.

As the SG resolver returns a different implementation for result objects, there are a number of snapshot updates.

### DIFF
--- a/.changes/unreleased/Features-20250910-114456.yaml
+++ b/.changes/unreleased/Features-20250910-114456.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Improve engine initialization speed via semantic graph
+time: 2025-09-10T11:44:56.51024-07:00
+custom:
+  Author: plypaul
+  Issue: "1835"


.changes/unreleased/Fixes-20250909-140537.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Fixes
```
courtneyholcomb, line 1: Good call to add this

PR #1848, opened 2025-09-16
Bug fix: don't error if there are no time dims in manifest

courtneyholcomb:
Seeing some [error logs](https://dbtlabsmt.datadoghq.com/logs?query=trace_id%3A68c91dce0000000098a793d8afe08253&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&event=AwAAAZlRnG__fSBcoAAAABhBWmxSbklLT0FBQUsweGo5Z1Y1dVJRQXIAAAAkZjE5OTUxYTAtMzU5OC00M2Q0LWJmZmQtMDQxNzNkNjAzODczAAE5ww&fromUser=true&messageDisplay=inline&refresh_mode=paused&storage=hot&stream_sort=desc&viz=stream&from_ts=1758009830199&to_ts=1758011830854&live=false) where this min() is called on an empty list. This removes the error.

A manifest without any time dimensions isn't likely very useful (probably in a partially complete state - measures require time dimensions so there can't be any metrics in this case), but this is not where it should error. Instead there are just no metrics to query.

### DIFF
--- a/.changes/unreleased/Fixes-20250916-121517.yaml
+++ b/.changes/unreleased/Fixes-20250916-121517.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Don't error for lack of time dimensions in manfest
+time: 2025-09-16T12:15:17.278238-07:00
+custom:
+  Author: courtneyholcomb
+  Issue: "1848"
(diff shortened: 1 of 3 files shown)


plypaul (approved):
Thanks!

PR #1854, opened 2025-09-18
Pin to a non-dev version of DSI for release

courtneyholcomb:
DSI was pinned to a dev version in the last release, which is a problem for some users who use python package managers that ignore dev versions. This pins it to the prior valid version so we can cut a new release pointing to that.

Also pins click due to a bad release that went out today and caused issues building hatch with python 3.12.

### DIFF
--- a/.github/actions/setup-python-env/action.yaml
+++ b/.github/actions/setup-python-env/action.yaml
@@ -67,7 +67,10 @@ runs:
 
   - name: Install Hatch
     shell: bash
-    run: pip3 install hatch
+    # Pin click to avoid bug in 8.3.0 - can remove when a fix is released
+    run: |
+      pip3 install "click<8.3.0"
+      pip3 install hatch
 
   # Running any command will install the dependencies for the project. Add this step so that the `hatch` environment is
   # cached.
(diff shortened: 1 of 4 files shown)


.github/actions/setup-python-env/action.yaml
```diff
@@ -67,7 +67,10 @@ runs:
 
   - name: Install Hatch
     shell: bash
-    run: pip3 install hatch
+    # Pin click to avoid bug in 8.3.0 - can remove when a fix is released
+    run: |
+      pip3 install "click<8.3.0"
```
plypaul, line 72: Do you have more context as to why this needs to be installed separately?

plypaul (approved):

.github/actions/setup-python-env/action.yaml
courtneyholcomb: It started blocking CI. There was a new version of click that was released today that broke our hatch install with python 3.12. Super random, took forever to debug.

courtneyholcomb: The bug was related to the `--filter` flag in hatch, which we aren't using, but it populated that param with a default value of a `Sentinel` object (whatever that is) that was not valid in JSON parsing.

courtneyholcomb: Here's a link to the CI failure if you're interested: https://github.com/dbt-labs/metricflow/actions/runs/17837131036/job/50717482593

PR #1868, opened 2025-09-24
Consolidate measure-related operations into `MeasureLookup`

plypaul:
This PR consolidates measure-related operations into `MeasureLookup` so that the switch from `measure -> simple metric` is more centralized.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/model/semantics/measure_lookup.py
+++ b/metricflow-semantics/metricflow_semantics/model/semantics/measure_lookup.py
@@ -11,11 +11,14 @@ from dbt_semantic_interfaces.references import (
     SemanticModelReference,
     TimeDimensionReference,
 )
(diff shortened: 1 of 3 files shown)


metricflow-semantics/metricflow_semantics/model/semantics/measure_lookup.py
```diff
@@ -95,6 +105,37 @@ def __init__(  # noqa: D107
                     ),
                 )
 
+                # Populate `_model_reference_to_aggregation_time_dimensions`.
+                agg_time_dimension_reference = semantic_model.checked_agg_time_dimension_for_measure(measure.reference)
+                # Ensure agg_time_dimension is lowercased - this transformation was not enforced on earlier manifests
```
courtneyholcomb, line 110: good call out

PR #1871, opened 2025-09-26
Consolidate measure / metric methods of `GroupByItemSetResolver`

plypaul:
This PR consolidates methods that handle measures and metrics separately in `GroupByItemSetResolver` to aid the `measure -> simple metric` migration.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
+++ b/metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
@@ -130,7 +130,7 @@ class MetricLookup:
         if result is not None:
             return result
 
-        result = self._group_by_item_set_resolver.get_linkable_elements_for_distinct_values_query(element_set_filter)
+        result = self._group_by_item_set_resolver.get_set_for_distinct_values_query(element_set_filter)
         self._group_by_items_for_no_metrics_query_cache.set(cache_key, result)
         return result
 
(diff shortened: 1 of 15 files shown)


metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
```diff
@@ -135,16 +135,20 @@ def linkable_elements_for_no_metrics_query(
...
         """Retrieve the matching set of linkable elements common to all metrics requested (intersection)."""
-        cache_key = (metric_references, element_set_filter)
+        cache_key = (metric_references, set_filter)
         result = self._group_by_items_for_metrics_cache.get(cache_key)
         if result is not None:
             return result
 
-        result = self._group_by_item_set_resolver.get_linkable_elements_for_metrics(
-            metric_references=metric_references, element_filter=element_set_filter
+        # This method has never returned group-by metrics, so ensure via filter.
```
courtneyholcomb, line 146: Note about this in case you didn't already have this context - previously, `linkable_elements_for_metrics` was only used in the APIs for listing metadata. We keep group by metrics out of the list since that can be an extremely long list of options, and it's also not supported in the group by (only filters).
I don't think you did this anywhere, but in case I overlook anything, just want to make sure we're not using this method to resolve supported group bys anywhere that group by metrics actually should be supported.

courtneyholcomb (approved):
LGTM!
Also, not sure if this is handled up the stack, but should we rename the files, too?

metricflow-semantics/metricflow_semantics/model/semantics/metric_lookup.py
plypaul: Noted, thanks!

plypaul (commented):
Yes, I'm saving renames for the end to reduce conflicts. There are some other renames that are pending in addition to the ones in this stack.

PR #1876, opened 2025-10-02
Bump DSI version and include new transformations

theyostalservice:

PR #1877, opened 2025-10-02
Get MF using the new dsi transformations

theyostalservice:
Depends on #1876 

Upstream PR bumps the dsi version, implicitly updating tests to using the default ruleset from dsi.  We can't do that for production MF rn because MF uses an extra transformation internally, so this PR adds all the newer built in PRs from dsi to metricflow-semantics/metricflow_semantics/model/dbt_manifest_parser.py to be used in real circumstances.

(If time was less limited before coalesce, I'd prefer to move the rule upstream if possible, but I don't want to pick up extra work before code freeze.)

Edit: There's also a rule (AddInputMeasuresToMetrics) that only works in dsi.  It probably merits a refactor/rewrite later to clean up the usage code here, but I'd rather prioritize getting this updated here for now.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/model/dbt_manifest_parser.py
+++ b/metricflow-semantics/metricflow_semantics/model/dbt_manifest_parser.py
@@ -3,8 +3,15 @@ from __future__ import annotations
 from dbt_semantic_interfaces.implementations.semantic_manifest import (
     PydanticSemanticManifest,
 )
+from dbt_semantic_interfaces.transformations.flatten_simple_metrics_with_measure_inputs import (
+    FlattenSimpleMetricsWithMeasureInputsRule,
+)
 from dbt_semantic_interfaces.transformations.names import LowerCaseNamesRule
+from dbt_semantic_interfaces.transformations.proxy_measure import CreateProxyMeasureRule
 from dbt_semantic_interfaces.transformations.pydantic_rule_set import PydanticSemanticManifestTransformRuleSet
+from dbt_semantic_interfaces.transformations.replace_input_measures_with_simple_metrics_transformation import (
+    ReplaceInputMeasuresWithSimpleMetricsTransformationRule,
+)
 from dbt_semantic_interfaces.transformations.semantic_manifest_transformer import (
     PydanticSemanticManifestTransformer,
 )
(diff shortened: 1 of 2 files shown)


theyostalservice:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1877?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1877** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1877?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1877?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1876** <a href="https://app.graphite.dev/github/pr/dbt-labs/metricflow/1876?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

metricflow-semantics/metricflow_semantics/model/dbt_manifest_parser.py
```diff
@@ -27,17 +26,21 @@ def parse_manifest_from_dbt_generated_manifest(manifest_json_string: str) -> Pyd
...
-            ConvertMedianToPercentileRule(),
+            *rule_set.legacy_measure_update_rules,
             DedupeMetricInputMeasuresRule(),  # Remove once fix is in core
-            SetCumulativeTypeParamsRule(),
+            # These individual rules come from rule_set.convert_legacy_measures_to_metrics_rules, but
+            # dsi requires AddInputMetricMeasuresRule, and metricflow requires that we do NOT run that rule
+            # as it is incompatible with a parser like dbt-core that pre-populates input measures.
+            CreateProxyMeasureRule(),
+            FlattenSimpleMetricsWithMeasureInputsRule(),
+            ReplaceInputMeasuresWithSimpleMetricsTransformationRule(),
```
theyostalservice, line 42: if we're properly motivated, I can refactor this in dsi to make this cleaner here, but for now, I'd really like to get this in and working today.

courtneyholcomb (approved):
curious what happened with `AddInputMetricMeasuresRule` / if that's an issue? But this looks good

theyostalservice:
It took me a long time to figure this out, but here's what I know about the current situation:
* In DSI, it does what it says, but if the name is hard to parse for you (it was for me!), it looks at all the measures a metric relies on and then fills up its `input_measures` parameter recursively, i.e. it will look at a derived metric and get the measures for the metrics that belong to that metric.
  * If at metric has any values for `input_metrics` when this rule starts to run, it fails and throws exceptions.  (This is what completely breaks things if we keep it in.)
* In metricflow, this was already disincluded because in core, we already [fill the input_metrics fields](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/parser/manifest.py#L1875).

So, imo, the most reasonable clean-up outcomes we could pursue here:
1. Make `AddInputMetricMeasuresRule` a bit smarter, so it only throws if the node has measures that are wrong.  This is the ultimate compromise choice, avoiding any disruption in expected dsi behavior around measures.
2. Remove `AddInputMetricMeasuresRule` from the pydantic rules list.  This seems nice because it gets rid of cruft we just don't need at all.

PR #1889, opened 2025-10-08
Update `metricflow` to migrate `measures -> simple metrics`

plypaul:
This PR contains the updates required for the `measure -> simple metric` migration in `metricflow`. With the elimination of measures, `MeasureSpec` was replaced by `SimpleMetricInputSpec` to represent the state of a column before a simple metric is computed (e.g. with a time spine join).

Significant changes were made to the dataflow plan builder to allow for simple metrics to be used as building blocks for cumulative / conversion metrics, which previously used measures as inputs.

There is currently a difference in filtering behavior (see `test_join_to_time_spine_with_filter_not_in_group_by`) that needs resolution (the filters to apply to the time spine before the join) and currently produces a snapshot diff.

### DIFF
--- a/.changes/unreleased/Under the Hood-20251008-120225.yaml
+++ b/.changes/unreleased/Under the Hood-20251008-120225.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Update `metricflow` to migrate `measures -> simple metrics`
+time: 2025-10-08T12:02:25.007138-07:00
+custom:
+  Author: plypaul
+  Issue: "1889"


tests_metricflow/snapshots/test_fill_nulls_with_rendering.py/SqlPlan/DuckDB/test_join_to_time_spine_with_filter_not_in_group_by__plan0_optimized.sql
```diff
@@ -21,22 +21,22 @@ FROM (
       , DATE_TRUNC('month', ds) AS metric_time__month
     FROM ***************************.mf_time_spine time_spine_src_28006
   ) subq_17
-  WHERE (metric_time__day <= '2020-01-02') AND (metric_time__month > '2020-01-01')
+  WHERE metric_time__month > '2020-01-01'
```
plypaul: @courtneyholcomb This is the difference in filtering behavior. Are the metric's filters supposed to apply to the time spine?

courtneyholcomb: Yes I think they should to make this as non-breaking as possible! I'm confused on this one, how do we even know how to split out the measure & metric filters at this point?

courtneyholcomb: But only the filters on agg_time or metric_time should apply to the time spine, since those are the columns available in the time spine.

plypaul: >  I'm confused on this one, how do we even know how to split out the measure & metric filters at this point?

So the measure filter and metric filter get added to a list, so we can't tell where it came from, but we know that they are different filters.

Let me take a look at applying those filters to the time spine.

tests_metricflow/snapshots/test_fill_nulls_with_0.py/str/DuckDB/test_join_to_time_spine_with_filter_not_in_group_by__query_output.txt
```diff
@@ -5,3 +5,32 @@ metric_time__day       bookings_join_to_time_spine_with_tiered_filters
 -------------------  -------------------------------------------------
 2020-01-01T00:00:00
 2020-01-02T00:00:00                                                  9
+2020-01-03T00:00:00
```
courtneyholcomb: I'm assuming this will change with the filter fix discussed [here](https://github.com/dbt-labs/metricflow/pull/1889#discussion_r2414257461)!

plypaul: Addressed in the latest commits. The row for `2020-01-01T00:00:00` was removed because we now apply both measure-level filters and metric-level filters (as defined in the simple metric) to the timespine.

courtneyholcomb: Yes that I also expected!

PR #1895, opened 2025-10-09
Sync DSI into MF

courtneyholcomb:
We are consolidating `dbt-semantic-interfaces` into `metricflow`. This is step one of that process. Note that this commit does not remove the DSI dependency in MF or change the import paths to use the synced files, that change will come later.
- Adds a script to copy DSI into MF, and update everything needed to pass CI. This may need to be run multiple times before we are able to deprecate the DSI repo.
- I ran the script, so this PR also includes the changes copied over from DSI

I highly recommend reviewing by commit!

### DIFF
--- a/metricflow-semantics/requirements-files/dev-env-requirements.txt
+++ b/metricflow-semantics/requirements-files/dev-env-requirements.txt
@@ -11,7 +11,6 @@ termcolor>=3.0, <4.0
 types-PyYAML
 types-python-dateutil
 types-tabulate
-ruff==0.3.3
 httpx>=0.24,<0.25
 hypothesis>=6.87,<7
 isort>=5.12,<6
--- a/metricflow-semantic-interfaces/metricflow_semantic_interfaces/dataclass_serialization.py
+++ b/metricflow-semantic-interfaces/metricflow_semantic_interfaces/dataclass_serialization.py
@@ -174,7 +174,7 @@ class SerializableDataclass(ABC):
             cls._concrete_subclass_registry or (), key=lambda class_type: (class_type.__module__, class_type.__name__)
         )
 
-    def __init_subclass__(cls, **kwargs) -> None:
+    def __init_subclass__(cls, **kwargs: Any) -> None:  # type: ignore[misc]
         """Adds the implementing class to the registry and check for non-concrete fields.
 
         It would be helpful to check that the fields of the dataclass are concrete fields, but that would need to be
--- a/metricflow-semantic-interfaces/metricflow_semantic_interfaces/implementations/saved_query.py
+++ b/metricflow-semantic-interfaces/metricflow_semantic_interfaces/implementations/saved_query.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from copy import deepcopy
-from typing import Any, List, Optional, Union
+from typing import List, Optional, Union
 
 from msi_pydantic_shim import Field
 from typing_extensions import Self, override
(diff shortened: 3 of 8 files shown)


metricflow-semantics/requirements-files/dev-env-requirements.txt
```diff
@@ -11,3 +11,4 @@ termcolor>=3.0, <4.0
 types-PyYAML
 types-python-dateutil
 types-tabulate
+ruff==0.3.3
```
plypaul: What's the context on needing `ruff` in the `hatch` environment? We've been running it in `pre-commit`, so it's configured in `.pre-commit-config.yaml`.

courtneyholcomb: Maybe it wasn't necessary 🤔 I was seeing some diffs between CI and local and thought this fixed it but will test removing this!

scripts/sync_dsi.py
```diff
@@ -0,0 +1,801 @@
...
+    # Get current branch
+    current_branch = run_git_command(["git", "branch", "--show-current"], dsi_repo_path)
+    print(f"Current branch: {current_branch}")
+
+    # Check if there are uncommitted changes
+    status_output = run_git_command(["git", "status", "--porcelain"], dsi_repo_path)
+    if status_output:
+        print("WARNING: There are uncommitted changes in the dbt-semantic-interfaces repository:")
+        print(status_output)
+        response = input("Continue anyway? (y/N): ").strip().lower()
```
theyostalservice, line 58: <3

scripts/sync_dsi.py
```diff
@@ -0,0 +1,801 @@
...
+        run_git_command(["git", "checkout", "main"], dsi_repo_path)
+
+    # Pull latest changes
+    print("Pulling latest changes from origin/main...")
+    run_git_command(["git", "pull", "origin", "main"], dsi_repo_path)
+
+    print("✅ Successfully updated dbt-semantic-interfaces repository")
+
+
+def update_noqa_comments_in_directory(directory: Path) -> int:
```
theyostalservice, line 75: thought: do we need to check on noqa 501 (line length)?
(this might already exist further in; I'm just writing comments as I go.)

scripts/sync_dsi.py
```diff
@@ -0,0 +1,801 @@
...
+                print("Linter output:")
+                print(result.stdout)
+        else:
+            print(f"⚠️  Linter found issues (exit code: {result.returncode})")
+            if result.stdout.strip():
+                print("Linter stdout:")
+                print(result.stdout)
+            if result.stderr.strip():
+                print("Linter stderr:")
+                print(result.stderr)
```
theyostalservice, line 606: nit - it seems like this code would be a lot simpler if you didn't capture the output, tbh.

metricflow-semantics/requirements-files/dev-env-requirements.txt
```diff
@@ -12,3 +12,10 @@ types-PyYAML
 types-python-dateutil
 types-tabulate
 ruff==0.3.3
+httpx>=0.24,<0.25
+hypothesis>=6.87,<7
+isort>=5.12,<6
+black>=23.3,<24
+pydantic>=1.10,<2
+types-Jinja2>=2.11,<3
+types-jsonschema>=4.17,<5
```
theyostalservice, line 20: I am sure this is a dumb, migraine-induced question, but it seems weird that we need to add these dependencies to metricflow-semantics.  Why would it need these new things?

scripts/sync_dsi.py
theyostalservice: oh sorry this was resolved elsewhere and I just forgot to remove the comment.

metricflow-semantics/requirements-files/dev-env-requirements.txt
courtneyholcomb: Just dev requirements that were required in DSI and not in MF (and only in dev), so the linter failed without some of them and the rest would fail if you tried to run DSI tests!
Adding them to MF semantics means that they're in both MF semantics and MF.

scripts/sync_dsi.py
courtneyholcomb: Cursor ❤️🙂

PR #1907, opened 2025-10-14
Add `dimension_lookup` to `ModelObjectLookup`

plypaul:
This PR adds the `dimension_lookup` to `ModelObjectLookup` as a small step towards migrating `SemanticModelLookup` to `ModelObjectLookup`. The goal of the migration was to simplify the lookup classes and to move more lookups to the semantic graph for performance reasons.

### DIFF
--- a/requirements-files/dev-env-requirements.txt
+++ b/requirements-files/dev-env-requirements.txt
@@ -6,6 +6,7 @@ duckdb-engine>=0.13.0, <0.14.0
 # https://github.com/duckdb/duckdb/issues/16617
 # Version 1.4.0 seems to have issues as well, so pinning to <1.4.0 until those are resolved.
 duckdb !=1.2.1, <1.4.0
+graphviz>=0.18.2, <0.21
 tomli >=2.3.0, <3
 varname >=0.15.1, < 1
 # Developer tools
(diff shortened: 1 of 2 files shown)


metricflow-semantics/metricflow_semantics/experimental/dsi/model_object_lookup.py
```diff
@@ -62,6 +63,27 @@ def time_dimension_name_to_grain(self) -> Mapping[str, TimeGranularity]:  # noqa
     def entity_lookup(self) -> EntityLookup:  # noqa: D102
         return EntityLookup(self._semantic_model.entities)
 
+    @cached_property
+    def dimension_lookup(self) -> DimensionLookup:  # noqa: D102
+        return DimensionLookup(self._semantic_model.dimensions)
+
+    @cached_property
+    def resolved_primary_entity_name(self) -> str:
```
courtneyholcomb: Our [primary entity validation](https://github.com/dbt-labs/dbt-semantic-interfaces/blob/main/dbt_semantic_interfaces/validations/primary_entity.py#L29) seems to only guarantee that semantic models with dimensions have a primary entity (i.e., if you have a model with entities only it doesn't need to have a primary entity). Can we update this code to avoid erroring if we encounter a model like that?

courtneyholcomb: We might not have any tests for this type of model, might be good to add one!

plypaul:
Closing as this is outdated.

PR #1918, opened 2025-10-23
Bump version range of DSI in metricflow-semantics 

WilliamDee:
## Context

Upgrading DSI version to new minor version

### DIFF
--- a/.changes/unreleased/Dependencies-20251024-132913.yaml
+++ b/.changes/unreleased/Dependencies-20251024-132913.yaml
@@ -1,0 +1,6 @@
+kind: Dependencies
+body: Bumping upperbound range for dbt-semantic-interfaces in metricflow-semantics
+time: 2025-10-24T13:29:13.372292-04:00
+custom:
+    Author: WilliamDee
+    Issue: None


metricflow-semantics/requirements-files/requirements.txt
```diff
@@ -1,8 +1,7 @@
 # Always support a range of production DSI versions capped at the next breaking version in metricflow-semantics.
 # This allows us to sync new, non-breaking changes to dbt-core without getting a version mismatch in dbt-mantle,
 # which depends on a specific commit of DSI.
-# bumping to 0.9.3 will cause test failures, so pinning to dev2 until fixed
-dbt-semantic-interfaces==0.9.4.dev0
+dbt-semantic-interfaces>=0.10.0, <0.11.0
```
courtneyholcomb: I think we may need to start by loosening this range for MF semantics specifically before we change the range for MF overall, and then we do the core & mantle update, and then we come back and do the MF update

WilliamDee: is it not possible to do it at the same time? Ie., 
1. Upgrade core https://github.com/dbt-labs/dbt-core/pull/12110
2. Update this PR to include the core version bump
3. Upgrade mantle to new mf-semantics
4. release new version of dbt-metricflow + mf

WilliamDee: Actually, does this mean we would need a core release too? Or would we just pin to a branch/commit of core in MF

courtneyholcomb: Good point. We will need an official core release in order to remove the dev DSI dependency from the currently released version of MF. Not top top prio but maybe we can get them to backport this version bump so that we get a core version with it soon!

WilliamDee: @courtneyholcomb i'm just gonna do what you mentioned and just start by 
1. bumping range in `metricflow-semantics`
2. bumping DSI in core
3. release core
4. bumping core + DSI in MF + bump the lowerbound range in `metricflow-semantics`

courtneyholcomb: sounds good!

courtneyholcomb (approved):

PR #1920, opened 2025-10-28
Update Cursor rules file to include linting + loading of additional rules

plypaul:
This PR updates the Cursor rules file to include a command for linting and a directive to load additional rules in `git_ignored/AGENTS.md`.

### DIFF
--- a/.changes/unreleased/Under the Hood-20251029-140033.yaml
+++ b/.changes/unreleased/Under the Hood-20251029-140033.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Update Cursor rules file to include linting + loading of additional rules
+time: 2025-10-29T14:00:33.96369-07:00
+custom:
+  Author: plypaul
+  Issue: "1920"


AGENTS.md
```diff
@@ -5,7 +5,11 @@ This project uses **hatch** for dependency management and testing:
 - Dependencies: Use `hatch` commands, not `pip install` directly
 - Tests: Use `hatch run dev-env:pytest <path>` instead of `pytest` directly
 - Snapshots: Add `--overwrite-snapshots` for new snapshot tests
-- Common: `make test` or `hatch run dev-env:pytest tests_metricflow/`
+
```
courtneyholcomb, line 12: Wondering if this will slow things down a lot if this is running for smaller changes?

plypaul: Like we don't want to run all tests if the change is small?

courtneyholcomb: Yeah, worried it might cause the agent to be very slow to finish responding to each prompt. Have you tried it locally and had success?

plypaul: Yeah, it seemed fine to me as the tests take ~25s to run on. My experience is that it would frequently only run the specific unit tests and break other tests. Did you have thoughts on a different directive?

courtneyholcomb: Hmmm I'm not sure that I would want cursor to run tests on EVERY change it makes, maybe only significant ones. But also, there are plenty of times when I just want it to finish more quickly in case I don't like what it did anyway and I can reject it, instead of waiting for it to also run tests.
It might be a matter of preference. I wonder if there is a way you can add your own user-specific cursor rules that don't get committed to git?

plypaul: Yeah, that works - see the added commits. I've verified that cursor loads any user-specific rules in `git_ignored/AGENTS.md` by running `Reload project rules`. 

In my `git_ignored/AGENTS.md`, I put the following:

```
Always run `make test` after any set of changes.
```

and that gets the behavior I want.

courtneyholcomb: perfect thank you!

AGENTS.md
```diff
@@ -2,14 +2,14 @@
...
 
-- Dependencies: Use `hatch` commands, not `pip install` directly
-- Tests: Use `hatch run dev-env:pytest <path>` instead of `pytest` directly
-- Snapshots: Add `--overwrite-snapshots` for new snapshot tests
+- Dependencies: Use `hatch` commands, not `pip install` directly.
+- Linting: Run `make lint` to detect and fix lint errors.
+- Testing:
+  - Use `hatch run dev-env:pytest <path>` to run tests in a specific file.
+  - Use `make test` to run all tests.
+- Snapshots: Add `--overwrite-snapshots` for new snapshot tests.
```
courtneyholcomb: Did this directive work? wondering if cursor will try to add that after `make test` and it will error, and we should direct it to use `make test-snap` if using the make command

plypaul: Yeah, it knew to add the flag to the `pytest` command. I'm going to rephrase that line to be more explicit though.

PR #1931, opened 2025-11-06
Remove measures from `simple_manifest`

plypaul:
This PR rewrites `simple_manifest` to remove measures from semantic models (via a script). In addition, the definition of simple metrics were moved to the YAML file containing the associated semantic model. Other semantic manifests will be updated in a subsequent PR.

### DIFF
--- a/metricflow-semantics/metricflow_semantics/model/dbt_manifest_parser.py
+++ b/metricflow-semantics/metricflow_semantics/model/dbt_manifest_parser.py
@@ -3,6 +3,9 @@ from __future__ import annotations
 from dbt_semantic_interfaces.implementations.semantic_manifest import (
     PydanticSemanticManifest,
 )
+from dbt_semantic_interfaces.transformations.fix_proxy_metrics import (
+    FixProxyMetricsRule,
+)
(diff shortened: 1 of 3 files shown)


metricflow-semantics/metricflow_semantics/model/dbt_manifest_parser.py
```diff
@@ -39,7 +36,6 @@ def transform_dbt_generated_manifest(manifest: PydanticSemanticManifest) -> Pyda
             CreateProxyMeasureRule(),
             FlattenSimpleMetricsWithMeasureInputsRule(),
             ReplaceInputMeasuresWithSimpleMetricsTransformationRule(),
-            FixProxyMetricsRule(),
```
WilliamDee, line 42: I recently added this in to fix a bug from core's transformation. I think we would need this in here since there will be legacy manifests that still has that issue that we wouldn't want to break.

plypaul: I was using an older SHA of DSI that did not have that rule, so I thought it was removed. Added a commit to use a later version of DSI and reverted this change.

PR #1933, opened 2025-11-10
Port `dbt-metricflow` package CI tests

plypaul:
https://github.com/dbt-labs/metricflow/pull/1653 added a CI test for the `dbt-metricflow` package in a release branch, but it's missing from `main`. This PR adds it.


dbt-metricflow/requirements-files/requirements-metricflow.txt
```diff
@@ -1 +1 @@
-metricflow==0.209.0
+metricflow @ git+https://github.com/dbt-labs/metricflow.git@a430e4c2aeb77fe18d1c876326ef54ccf1cd464f
```
tlento: Per the description, we don't want to merge like this, right? So this would be updated once the 0.210 or whatever release is deployed and then we'd roll this out with 0.211 (although since this is a package test it'll continue to run in the interim)?

tlento (commented):
This seems reasonable but I don't have full context on these package tests. Might be good to chat briefly just so I can get up to speed.

WilliamDee (approved):

dbt-metricflow/requirements-files/requirements-metricflow.txt
plypaul: This has been removed.

PR #1934, opened 2025-11-10
Update `dbt-metricflow` package tests to check log file

plypaul:
As per title.

### DIFF
--- a/dbt-metricflow/scripts/ci_tests/dbt_metricflow_package_test.py
+++ b/dbt-metricflow/scripts/ci_tests/dbt_metricflow_package_test.py
@@ -6,6 +6,8 @@ import textwrap
 from pathlib import Path
 from typing import Optional
 
+from dbt_metricflow.cli.cli_configuration import CLIConfiguration
+
 
 def _run_shell_command(command: str, cwd: Optional[Path] = None) -> None:
     if cwd is None:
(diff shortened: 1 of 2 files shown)


dbt-metricflow/scripts/ci_tests/dbt_metricflow_package_test.py
```diff
@@ -37,3 +39,16 @@ def _run_shell_command(command: str, cwd: Optional[Path] = None) -> None:
         "mf query --metrics transactions --group-by metric_time --order metric_time",
         cwd=tutorial_directory,
     )
+
+    # Check that log messages are written as spected to the log file.
+    log_file_path = tutorial_directory.joinpath("logs", CLIConfiguration.LOG_FILE_NAME)
```
tlento, line 49: "logs" is a magic string. Is there no constant for that path location name?

tlento (approved):
It'd be good to clean up the magic string before merge, or if that's an annoying thing requiring package updates and deployment then a follow up is fine.

dbt-metricflow/scripts/ci_tests/dbt_metricflow_package_test.py
plypaul: The logs directory is configured by the dbt project, so there isn't a code reference for it.

QMalcolm: Of note "logs" _might_ fail. The logging directory in dbt-core is settable via a CLI flag `--log-path` or environment variable `DBT_ENGINE_LOG_PATH`. However, that path is not recorded as part of the manifest, run results, or anywhere else. Thus, though "logs" might fail, there also isn't a good alternative

QMalcolm (approved):
Looks good to me!

dbt-metricflow/scripts/ci_tests/dbt_metricflow_package_test.py
plypaul: Appreciate the context!

PR #1935, opened 2025-11-10
Fix optimizer bug for combining simple metrics

WilliamDee:
## Context
There was a bug where 2 simple metrics were being joined together in a derived metric and when both metrics has an alias, the optimizer ends up not rendering the expr for one of the metric. This seemed to expose a bug in the optimizer and this PR fixes that.

NOTE: Check first commit test case to see the bug

### DIFF
--- a/metricflow-semantics/metricflow_semantics/test_helpers/semantic_manifest_yamls/simple_manifest/metrics.yaml
+++ b/metricflow-semantics/metricflow_semantics/test_helpers/semantic_manifest_yamls/simple_manifest/metrics.yaml
@@ -1013,9 +1013,13 @@ metric:
   description: "test simple derived metric"
   type: derived
   type_params:
-    expr: alias_1 + alias_2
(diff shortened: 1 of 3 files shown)


metricflow-semantics/metricflow_semantics/test_helpers/semantic_manifest_yamls/simple_manifest/metrics.yaml
```diff
@@ -20,6 +27,18 @@ metric:
   type_params:
     measure: instant_bookings
 ---
+metric:
+  name: "test_simple_derived_metric"
+  description: "test simple derived metric"
+  type: derived
+  type_params:
+    expr: bookings - instant_bookings
```
plypaul: The expression should be something like `bookings_alias - total_bookings_alias` given the aliases for the metrics below.

WilliamDee: updated

PR #1938, opened 2025-11-13
Add dunder prefix to column aliases for simple-metric inputs

plypaul:
To avoid alias conflicts in the generated SQL when a semantic model has a dimension and a simple metric with the same name, this PR updates the naming scheme for SQL column aliases. The longer term fix is to add a validation so that semantic models are required to have unique metric / measure / entity names. 

Due to the large snapshot changes, please view by commit.


courtneyholcomb (approved):
That was a lot simpler than I expected! Thank you!

PR #1941, opened 2025-11-20
Render simple-metric inputs without a dunder prefix for the `WHERE` filter

plypaul:
This PR updates the SQL rendering for the `WhereConstraintNode` to use a subquery that generates column aliases without a dunder prefix for simple metric inputs e.g. (`bookings` instead of `__bookings`).

This is added to handle a migration case - some customers have been writing SQL strings in the filters that reference internal column names, which is an unsupported use case. However, to reduce errors during migration, this PR temporarily restores the previous behavior.

Due to the snapshot changes, please view by commit.

### DIFF
--- a/.changes/unreleased/Fixes-20251119-233621.yaml
+++ b/.changes/unreleased/Fixes-20251119-233621.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Render simple-metric inputs without a dunder prefix for the `WHERE` filter
+time: 2025-11-19T23:36:21.068696-08:00
+custom:
+  Author: plypaul
+  Issue: "1941"


tests_metricflow/snapshots/test_conversion_metric_rendering.py/SqlPlan/DuckDB/test_conversion_metric__plan0_optimized.sql
```diff
@@ -31,12 +31,12 @@ FROM (
     -- Aggregate Inputs for Simple Metrics
     SELECT
       metric_time__day
-      , SUM(__visits) AS __visits
+      , SUM(visits) AS __visits
```
courtneyholcomb, line 34: This works but it's definitely going to look confusing to users. I think we can release as-is but I wonder if we can improve this by rendering only the non-dundered name anytime after the initial source node. But @plypaul maybe you already looked into that and it didn't work?

PR #1944, opened 2025-12-04
Filter out specs not needed before where filter is applied

courtneyholcomb:
This is needed to ensure the optimizer does not run into bugs related to name conflicts relating to elements that were not requested in the query.

### DIFF
--- a/tests_metricflow/snapshots/test_derived_metric_rendering.py/SqlPlan/BigQuery/test_nested_derived_metric_offset_with_joined_where_constraint_not_selected__plan0.sql
+++ b/tests_metricflow/snapshots/test_derived_metric_rendering.py/SqlPlan/BigQuery/test_nested_derived_metric_offset_with_joined_where_constraint_not_selected__plan0.sql
@@ -4,387 +4,380 @@ sql_engine: BigQuery
 ---
 -- Write to DataTable
 SELECT
-  subq_20.metric_time__day
-  , subq_20.bookings_offset_twice
+  subq_19.metric_time__day
+  , subq_19.bookings_offset_twice
 FROM (
   -- Compute Metrics via Expressions
   SELECT
-    subq_19.metric_time__day
+    subq_18.metric_time__day
     , 2 * bookings_offset_once AS bookings_offset_twice
   FROM (
     -- Pass Only Elements: ['metric_time__day', 'bookings_offset_once']
     SELECT
-      subq_18.metric_time__day
-      , subq_18.bookings_offset_once
+      subq_17.metric_time__day
+      , subq_17.bookings_offset_once
     FROM (
       -- Constrain Output with WHERE
       SELECT
-        subq_17.booking__is_instant
-        , subq_17.metric_time__day
-        , subq_17.bookings_offset_once
+        subq_16.booking__is_instant
+        , subq_16.metric_time__day
+        , subq_16.bookings_offset_once
       FROM (
-        -- Pass Only Elements: ['booking__is_instant', 'metric_time__day', 'bookings_offset_once']
+        -- Join to Time Spine Dataset
         SELECT
-          subq_16.metric_time__day
-          , subq_16.booking__is_instant
-          , subq_16.bookings_offset_once
+          subq_15.metric_time__day AS metric_time__day
+          , subq_11.booking__is_instant AS booking__is_instant
+          , subq_11.bookings_offset_once AS bookings_offset_once
         FROM (
-          -- Join to Time Spine Dataset
+          -- Pass Only Elements: ['metric_time__day']
           SELECT
-            subq_15.metric_time__day AS metric_time__day
-            , subq_11.booking__is_instant AS booking__is_instant
-            , subq_11.bookings_offset_once AS bookings_offset_once
+            subq_14.metric_time__day
           FROM (
             -- Pass Only Elements: ['metric_time__day']
             SELECT
-              subq_14.metric_time__day
+              subq_13.metric_time__day
             FROM (
-              -- Pass Only Elements: ['metric_time__day']
+              -- Change Column Aliases
               SELECT
-                subq_13.metric_time__day
+                subq_12.ds__day AS metric_time__day
+                , subq_12.ds__week
+                , subq_12.ds__month
+                , subq_12.ds__quarter
+                , subq_12.ds__year
+                , subq_12.ds__extract_year
+                , subq_12.ds__extract_quarter
+                , subq_12.ds__extract_month
+                , subq_12.ds__extract_day
+                , subq_12.ds__extract_dow
+                , subq_12.ds__extract_doy
+                , subq_12.ds__alien_day
               FROM (
-                -- Change Column Aliases
+                -- Read From Time Spine 'mf_time_spine'
                 SELECT
-                  subq_12.ds__day AS metric_time__day
-                  , subq_12.ds__week
-                  , subq_12.ds__month
-                  , subq_12.ds__quarter
-                  , subq_12.ds__year
-                  , subq_12.ds__extract_year
-                  , subq_12.ds__extract_quarter
-                  , subq_12.ds__extract_month
-                  , subq_12.ds__extract_day
-                  , subq_12.ds__extract_dow
-                  , subq_12.ds__extract_doy
-                  , subq_12.ds__alien_day
-                FROM (
-                  -- Read From Time Spine 'mf_time_spine'
-                  SELECT
-                    time_spine_src_28006.ds AS ds__day
-                    , DATETIME_TRUNC(time_spine_src_28006.ds, isoweek) AS ds__week
-                    , DATETIME_TRUNC(time_spine_src_28006.ds, month) AS ds__month
-                    , DATETIME_TRUNC(time_spine_src_28006.ds, quarter) AS ds__quarter
-                    , DATETIME_TRUNC(time_spine_src_28006.ds, year) AS ds__year
-                    , EXTRACT(year FROM time_spine_src_28006.ds) AS ds__extract_year
-                    , EXTRACT(quarter FROM time_spine_src_28006.ds) AS ds__extract_quarter
-                    , EXTRACT(month FROM time_spine_src_28006.ds) AS ds__extract_month
-                    , EXTRACT(day FROM time_spine_src_28006.ds) AS ds__extract_day
-                    , IF(EXTRACT(dayofweek FROM time_spine_src_28006.ds) = 1, 7, EXTRACT(dayofweek FROM time_spine_src_28006.ds) - 1) AS ds__extract_dow
-                    , EXTRACT(dayofyear FROM time_spine_src_28006.ds) AS ds__extract_doy
-                    , time_spine_src_28006.alien_day AS ds__alien_day
-                  FROM ***************************.mf_time_spine time_spine_src_28006
-                ) subq_12
-              ) subq_13
-            ) subq_14
-          ) subq_15
-          INNER JOIN (
+                  time_spine_src_28006.ds AS ds__day
+                  , DATETIME_TRUNC(time_spine_src_28006.ds, isoweek) AS ds__week
+                  , DATETIME_TRUNC(time_spine_src_28006.ds, month) AS ds__month
+                  , DATETIME_TRUNC(time_spine_src_28006.ds, quarter) AS ds__quarter
+                  , DATETIME_TRUNC(time_spine_src_28006.ds, year) AS ds__year
+                  , EXTRACT(year FROM time_spine_src_28006.ds) AS ds__extract_year
+                  , EXTRACT(quarter FROM time_spine_src_28006.ds) AS ds__extract_quarter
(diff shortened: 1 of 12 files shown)


metricflow/plan_conversion/to_sql_plan/dataflow_to_subquery.py
```diff
@@ -1276,7 +1276,6 @@ def visit_semi_additive_join_node(self, node: SemiAdditiveJoinNode) -> SqlDataSe
 
         # Get the output_instance_set of the parent_node
         output_instance_set = from_data_set.instance_set
```
plypaul, line 1265: Removing this can create a bug - an instance in `from_data_set` could map something like `col -> MetricSpec("bookings")` instead of `bookings -> MetricSpec("bookings")`.

metricflow/dataflow/builder/dataflow_plan_builder.py
```diff
@@ -1988,6 +1990,26 @@ def _build_pre_aggregation_plan(
                 parent_node=output_node, time_dimension_spec=custom_granularity_spec
             )
 
+        # Filter to specs needed for the rest of the query. This is needed to remove the potential for column name
+        # conflicts for elements not needed in the query.
+        if filter_to_specs:
```
plypaul: Don't we need to do this selection whenever there are `where_filter_specs` instead of just when `filter_to_specs` is set?

tests_metricflow/snapshots/test_name_edge_caes.py/SqlPlan/DuckDB/test_filter_by_metric_name_with_2_measures_from_same_source_node__plan0.sql
```diff
@@ -6,369 +6,283 @@ sql_engine: DuckDB
...
-      , subq_9.__homonymous_metric_and_dimension AS homonymous_metric_and_dimension
+      subq_11.metric_time__day
+      , subq_11.__homonymous_metric_and_dimension AS homonymous_metric_and_dimension
     FROM (
       -- Aggregate Inputs for Simple Metrics
       SELECT
-        subq_8.metric_time__day
-        , SUM(subq_8.__homonymous_metric_and_dimension) AS __homonymous_metric_and_dimension
+        subq_10.metric_time__day
+        , SUM(subq_10.__homonymous_metric_and_dimension) AS __homonymous_metric_and_dimension
```
plypaul, line 160: The filter `WHERE homonymous_metric_and_entity IS NOT NULL` is still invalid, but the optimized query works as kind of a fluke.

courtneyholcomb: I think that is how it was working before, right?

metricflow/plan_conversion/to_sql_plan/dataflow_to_subquery.py
courtneyholcomb: I can remove this commit, thought I was going to need changes to this file but it turned out to be unnecessary.

metricflow/dataflow/builder/dataflow_plan_builder.py
courtneyholcomb: This will only ever be empty in the case of a nested aggregation (i.e., nested derived offset metrics) so we've already removed all the unnecessary specs in those cases.
There's also a totally unrelated bug in that code that might necessitate a change here but this retains the current behavior.

tests_metricflow/snapshots/test_name_edge_caes.py/SqlPlan/DuckDB/test_filter_by_metric_name_with_2_measures_from_same_source_node__plan0.sql
plypaul: Yeah, seems like it. The filtering behavior is odd and probably not what the user is expecting. It's acting as a filter for a measure that depends on the value of another measure.

metricflow/dataflow/builder/dataflow_plan_builder.py
plypaul: Something still seems a little off about the argument - to clarify how this is supposed to be used for other cases, can you add a docstring for the argument? Also, given that the element filtering happens at the end anyway, is the filtering supposed to applied even when there are no where filters?

courtneyholcomb: Sure I can!
Fair question - I worry that we still might run into column name conflicts if we don't apply it for all queries. And I do still think this is a useful filter for all queries to avoid unexpected optimizer issues. But I think we differ on that point so open to your thoughts here.

metricflow-semantics/metricflow_semantics/test_helpers/semantic_manifest_yamls/name_edge_case_manifest/metrics.yaml
```diff
@@ -1,8 +1,16 @@
 ---
 metric:
-  name: conflicted_name
+  name: homonymous_metric_and_dimension
```
courtneyholcomb, line 3: Copied some of these tests over from Paul's PR that was reverted.

tests_metricflow/integration/query_output/test_query_output.py
```diff
@@ -291,3 +291,30 @@ def test_saved_query_override_order_by_and_limit(  # noqa: D103
         snapshot_str=query_result.result_df.text_format(),
         sql_engine=sql_client.sql_engine_type,
     )
+
+
+@pytest.mark.sql_engine_snapshot
+@pytest.mark.duckdb_only
+def test_semi_additive_measure_with_where_filter(  # noqa: D103
```
courtneyholcomb, line 298: We previously only had a check query test for semi additive joins with a where filter, so this adds a more explicit check.

tests_metricflow/query_rendering/test_name_edge_caes.py
```diff
@@ -1,33 +0,0 @@
-from __future__ import annotations
```
courtneyholcomb, line 1: Renamed this file due to a typo.

tests_metricflow/snapshots/test_name_edge_caes.py/SqlPlan/DuckDB/test_filter_by_metric_name_with_2_measures_from_same_source_node__plan0_optimized.sql
```diff
@@ -0,0 +1,25 @@
...
+  , SUM(__homonymous_metric_and_entity) AS homonymous_metric_and_entity
+  , SUM(__homonymous_metric_and_dimension) AS homonymous_metric_and_dimension
+FROM (
+  -- Read Elements From Semantic Model 'bookings_source'
+  -- Metric Time Dimension 'ds'
+  SELECT
+    DATE_TRUNC('day', ds) AS metric_time__day
+  FROM ***************************.fct_bookings bookings_source_src_32000
+) subq_13
+WHERE homonymous_metric_and_entity IS NOT NULL
```
courtneyholcomb: This demonstrates the BigQuery issue related to where filters for metric names. This SQL would fail in BQ due to referencing a column alias in the where filter.

tests_metricflow/snapshots/test_name_edge_cases.py/SqlPlan/DuckDB/test_homonymous_metric_and_entity__plan0_optimized.sql
```diff
@@ -0,0 +1,24 @@
...
+  metric_time__day
+  , SUM(__homonymous_metric_and_entity) AS homonymous_metric_and_entity
+FROM (
+  -- Read Elements From Semantic Model 'bookings_source'
+  -- Metric Time Dimension 'ds'
+  SELECT
+    DATE_TRUNC('day', ds) AS metric_time__day
+  FROM ***************************.fct_bookings bookings_source_src_32000
+) subq_7
+WHERE homonymous_metric_and_entity IS NOT NULL
```
courtneyholcomb, line 24: Demonstrates the issue where a column alias is referenced in the where filter.

metricflow-semantics/metricflow_semantics/specs/non_additive_dimension_spec.py
```diff
@@ -85,3 +88,16 @@ def linkable_specs(self, non_additive_dimension_grain: TimeGranularity) -> Seque
                 time_granularity=ExpandedTimeGranularity.from_time_granularity(non_additive_dimension_grain),
             ),
         ) + tuple(LinklessEntitySpec.from_element_name(entity_name) for entity_name in self.window_groupings)
+
+    @property
```
courtneyholcomb, line 92: Added these to ensure we could use as close to the same logic as possible in the dataflow plan and the dataflow to subquery visitor as possible to determine what specs should be used in this node.

tests_metricflow/snapshots/test_name_edge_cases.py/SqlPlan/DuckDB/test_filter_by_metric_name_with_2_measures_from_same_source_node__plan0_optimized.sql
```diff
@@ -11,15 +11,18 @@ sql_engine: DuckDB
 -- Write to DataTable
 SELECT
   metric_time__day
-  , SUM(__homonymous_metric_and_entity) AS homonymous_metric_and_entity
-  , SUM(__homonymous_metric_and_dimension) AS homonymous_metric_and_dimension
+  , SUM(homonymous_metric_and_entity) AS homonymous_metric_and_entity
```
courtneyholcomb, line 14: Fixes the issue of referencing an alias in the where filter.

tests_metricflow/snapshots/test_name_edge_cases.py/str/DuckDB/test_filter_by_metric_name_with_2_measures_from_same_source_node__query_output.txt
```diff
@@ -0,0 +1,127 @@
+test_name: test_filter_by_metric_name_with_2_measures_from_same_source_node
```
courtneyholcomb, line 1: This snapshot is new in this commit because the test would error before the prior commit.

tests_metricflow/snapshots/test_name_edge_cases.py/str/DuckDB/test_homonymous_metric_and_entity__query_output.txt
```diff
@@ -0,0 +1,127 @@
+test_name: test_homonymous_metric_and_entity
```
courtneyholcomb, line 1: Same for this new snapshot.

tests_metricflow/snapshots/test_common_dataflow_branches.py/str/test_shared_metric_query__result.txt
```diff
@@ -24,24 +24,35 @@ dataflow_plan:
                       <!-- null_fill_value_mapping = {} -->
                       <FilterElementsNode>
                           <!-- description = "Pass Only Elements: ['__bookings', 'metric_time__day']" -->
-                          <!-- node_id = NodeId(id_str='pfe_0') -->
```
courtneyholcomb, line 27: Looked through all the optimized snapshots in this commit and the only things changed are:
- subquery aliases
- column order

metricflow/dataflow/builder/dataflow_plan_builder.py
plypaul: The way I'm thinking about the logic here is: would someone reading the docstring of this function expect this logic to be here? Or if you gave someone just the method signature / docstring and ask them to implement it or rewrite it, would they do this as well? So it depends on how `filter_to_specs` is defined.

metricflow/dataset/dataset_classes.py
```diff
@@ -34,6 +34,24 @@ def metric_time_dimension_instances(self) -> Sequence[TimeDimensionInstance]:
...
+        metric_time_instances = sorted(
+            (
+                instance
+                for instance in self.metric_time_dimension_instances
+                if not instance.spec.has_custom_grain and not instance.spec.date_part
+            ),
+            key=lambda x: x.spec.base_granularity_sort_key,
+        )
+
+        assert (
```
plypaul, line 49: It seems a little odd to have a property that can raise an exception without having a way to know if it's safe to access the property. How about returning an optional value?

metricflow/dataflow/builder/dataflow_plan_builder.py
courtneyholcomb: Note for posterity - we synced offline about this and came up with a path forward. @plypaul I updated this logic in the last 3 commits for easier review!
This should be ready for re-review, just updating SQL engine snapshots now.

plypaul (approved):
Much more clear - thanks! 🚢

PR #1957, opened 2026-01-07
Fix time period end boundaries for time granularities finer than day

akuttig-block:
## Summary

Fixes a **critical bug in time period boundary calculations** that caused generated SQL queries to exclude most of the end date when filtering data at DAY granularity and above.

## Problem

The `adjust_to_end_of_period` method in `DateutilTimePeriodAdjuster` was incorrectly truncating end-of-period timestamps to `00:00:00` (midnight) for requested granularities coarser than `DAY`. This resulted in SQL queries that only captured the first portion of the end date rather than the entire day when the underlying dimension had a granularity finer than day.

### Example of Incorrect Behavior

For a query requesting data grouped by day for 01/01/2025 through 01/31/2025 on a second-granularity time dimension, the generated SQL would be:

```sql
WHERE DATE_TRUNC('second', occurred_at) BETWEEN '2025-01-01 00:00:00' AND '2025-01-31 00:00:00'
```

This would only include records at exactly midnight on January 31st, **excluding the remaining 23:59:59 of that day**.

## Solution

Updated the end-of-period calculation to set the time to `23:59:59` when requesting DAY granularity and above, ensuring the full period is captured:

```sql
WHERE DATE_TRUNC('second', occurred_at) BETWEEN '2025-01-01 00:00:00' AND '2025-01-31 23:59:59'
```

### Core Fix
- **`dateutil_adjuster.py`**: Changed end-of-period time from `hour=0, minute=0, second=0` to `hour=23, minute=59, second=59` for granularities >= DAY

### Test Improvements
- Added comprehensive test coverage for sub-day granularities (SECOND, MINUTE, HOUR)
- Added test cases with various times throughout the day (midnight, noon, 23:59:59, etc.)
- Added validation assertions to ensure:
  - End-of-period timestamps are always >= start-of-period
  - Correct time components for each granularity:
    - MINUTE+: `second=59`
    - HOUR+: `minute=59, second=59`
    - DAY+: `hour=23, minute=59, second=59`

## Impact

This fix ensures that:
- ✅ Time range constraints properly capture the full period when filtering data
- ✅ Queries no longer have off-by-one errors that exclude most of the end date
- ✅ Data analysis results are accurate and complete for the requested time ranges

## Testing

All existing tests pass with updated snapshots. The new validation assertions provide ongoing protection against regression of this issue.

### DIFF
--- a/.changes/unreleased/Fixes-20260107-163537.yaml
+++ b/.changes/unreleased/Fixes-20260107-163537.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Fix time period end boundaries to include the full end date
+time: 2026-01-07T16:35:37.657489-07:00
+custom:
+    Author: akuttig-block
+    Issue: "1957"


metricflow-semantics/metricflow_semantics/time/dateutil_adjuster.py
```diff
@@ -127,7 +127,7 @@ def adjust_to_end_of_period(
         # Strip time data that is irrelevant to avoid rendering an unnecessarily complex time constraint.
         date_to_adjust = date_to_adjust.replace(microsecond=0)
         if time_granularity.to_int() >= TimeGranularity.DAY.to_int():
-            date_to_adjust = date_to_adjust.replace(hour=0, minute=0, second=0)
+            date_to_adjust = date_to_adjust.replace(hour=23, minute=59, second=59)
```
tlento, line 130: The addition of these extra values is causing the rendered SQL to change in cases where the output expression was correct (i.e., for anything >= DAY grain). We'll need to consider the best approach here. This may be it, although I'll need to recreate a bunch of snapshots and run further tests before we can get this merged, but there are some other options which might be less sensitive to the underlying column granularity (i.e., in the warehouse rather than in the MetricFlow model spec).

PR #1963, opened 2026-01-21
Fix optimization of queries containing metrics with`fill_nulls_with` set

plypaul:
This PR addresses the case in https://github.com/dbt-labs/metricflow/pull/1962.

### DIFF
--- a/.changes/unreleased/Fixes-20260203-115108.yaml
+++ b/.changes/unreleased/Fixes-20260203-115108.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Fix optimization of queries containing metrics with`fill_nulls_with` set
+time: 2026-02-03T11:51:08.877673-08:00
+custom:
+  Author: plypaul
+  Issue: "1963"


metricflow/dataflow/optimizer/source_scan/cm_branch_combiner.py
```diff
@@ -251,7 +251,7 @@ def visit_aggregate_simple_metric_inputs_node(  # noqa: D102
         assert len(combined_parent_nodes) == 1
         combined_parent_node = combined_parent_nodes[0]
 
-        if self._current_left_node.null_fill_value_mapping != current_right_node.null_fill_value_mapping:
+        if self._current_left_node.null_fill_value_mapping.has_conflict(current_right_node.null_fill_value_mapping):
```
tlento, line 291: This is incredibly subtle. I'm assuming the behavior change here is the fix. If that's wrong then I think we've got more explaining to do.

If that's right, we need better documentation of why it is this way. An outsider (or even someone like me, and probably you as well when you first wrote this) would assume that the original equality check is what we should be doing here, because we don't want to merge metric input nodes that do null fills with metric input nodes that do not.

Except it seems like we actually DO want to do this. Why? Is this because null fill happens after the point where these branches are merging? And is that always true?

Assuming this is the fix, then at a minimum I'd want a docstring on this method explaining what the function is meant to do. Ideally we'd have some kind of test case illustrating the behavior as well - the snapshot cases aren't adequate, because they're too high level for this kind of detail behavior, but an errant change to this behavior will just break everything again (or newly, in different subtle but annoying ways).

tlento (commented):
See inlines. My big questions are:  


1. Are we sure this fix doesn't cause other issues where maybe we fill nulls when we shouldn't?
2. If we are sure, can we add a simple direct test case somehow and document the expected behavior via test enforcement? I know visitors are almost impossible to test directly in this way, but maybe we can do something. Absent that a docstring will suffice.

metricflow/dataflow/optimizer/source_scan/cm_branch_combiner.py
plypaul: @tlento Added updated comments and test cases.

metricflow/dataflow/optimizer/source_scan/cm_branch_combiner.py
```diff
@@ -230,9 +230,44 @@ def visit_join_on_entities_node(  # noqa: D102
...
+
+        * The left and right branches have a null-fill value mapping that contains no common simple-metric inputs.
+
+        e.g.
+            left:
+                null_fill_value_mapping = {"bookings": 0}
+            right:
+                null_fill_value_mapping = {"booking_value": 0}
+
+        Since different instances are represented with different columns, there are no issues using different null-fill
```
tlento, line 256: Ah, thanks, this helped me understand. It's ok because essentially you have different branches that are filling nulls for each of these columns, so now we combine those branches and fill nulls for both at once.

metricflow/dataflow/builder/aggregation_helper.py
```diff
@@ -80,11 +81,11 @@ def has_conflict(self, other: InstanceAliasMapping) -> bool:  # noqa: D102
 class NullFillValueMapping(Mergeable, SerializableDataclass):
     """Stores the mapping for which instances should have null values set to a configured value."""
 
-    _element_name_and_null_fill_value_items: Tuple[Tuple[str, int], ...]
+    _element_name_and_null_fill_value_items: Tuple[Tuple[str, Optional[int]], ...]
```
tlento, line 86: Huh, I thought `int | None` is officially preferred over `Optional` and I swear I hit some kind of typechecker errors about this not too long ago (although maybe that was with pyright?).

Anyway, we use Optional in the codebase already so eh, do whatever, but the union expression is more explicit.

metricflow/dataflow/builder/aggregation_helper.py
```diff
@@ -124,24 +128,51 @@ def null_fill_value_spec(self, spec: SimpleMetricInputSpec) -> Optional[SimpleMe
...
-                        current_null_fill_value=current_null_fill_value,
-                        other_null_fill_value=other_null_fill_value,
-                    )
+        conflicting_element_names = self._conflicting_element_names(other)
+        if len(conflicting_element_names) != 0:
+            raise RuntimeError(
+                LazyFormat(
+                    "Can't merge fill value mappings with conflicting items."
+                    " Conflicts should have been checked before merging.",
+                    conflicting_element_names=conflicting_element_names,
```
tlento: nit: I think this is marginally cleaner, although if there is an error we call the inner function twice

```suggestion
        if self.has_conflict(other):
            raise RuntimeError(
                LazyFormat(
                    "Can't merge fill value mappings with conflicting items."
                    " Conflicts should have been checked before merging.",
                    conflicting_element_names=self._conflicting_element_names(other),
```

metricflow/dataflow/builder/aggregation_helper.py
plypaul: `|` was added in Python `3.10` (https://peps.python.org/pep-0604/), so we'll have to wait until we remove support for `3.9`.

metricflow/dataflow/builder/aggregation_helper.py
plypaul: Updated.

PR #1964, opened 2026-01-27
Add memoization to resolve_available_items

wiggzz:
## Why

When resolving group-by items for queries with many metrics (e.g., 260+ metrics in a PowerBI metadata discovery query), `resolve_available_items()` was being called repeatedly (23+ times in observed traces) with the same arguments. Each call performs a full DAG traversal taking 9-20 seconds, leading to query resolution times of 40+ minutes.

This was causing:
- CPU pegging in metricflow-server (MFS)
- Kubernetes liveness probe failures
- Service outages

The root cause is in `push_down_visitor.py` - when a measure node has no matching items and suggestion_generator is set, it calls `resolve_available_items()` to generate error suggestions. With many metrics, this happens for each measure node, and without caching, each call does a full expensive DAG traversal.

## What

Add a simple instance-level cache to `GroupByItemResolver.resolve_available_items()` that caches results by `(resolution_node_id, pattern_ids)`. Since the same `GroupByItemResolver` instance is used throughout a single query resolution, and the same patterns are passed each time, subsequent calls return cached results instead of re-traversing the DAG.

This reduces query resolution from O(N × DAG_traversal_time) to O(1 × DAG_traversal_time) for repeated calls with the same arguments.

## Notes

- The cache uses `id()` for the resolution node and patterns since they're immutable during resolution
- The cache is instance-scoped, so it naturally gets cleared when a new `GroupByItemResolver` is created
- Existing tests pass without modification

Drafted by Claude Opus 4.5 under the direction of @wiggzz

### DIFF
--- a/.changes/unreleased/Fixes-20260128-155328.yaml
+++ b/.changes/unreleased/Fixes-20260128-155328.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Add memoization to `resolve_available_items` to prevent repeated expensive DAG traversals during query resolution with invalid group-by items
+time: 2026-01-28T15:53:28.000000-00:00
+custom:
+  Author: wiggzz
+  Issue: "1964"


metricflow-semantics/metricflow_semantics/query/group_by_item/group_by_item_resolver.py
```diff
@@ -90,6 +90,8 @@ def __init__(  # noqa: D107
     ) -> None:
         self._manifest_lookup = manifest_lookup
         self._resolution_dag = resolution_dag
+        # Cache for resolve_available_items to avoid repeated expensive DAG traversals
+        self._available_items_cache: dict[Tuple[int, Tuple[int, ...]], AvailableGroupByItemsResolution] = {}
```
plypaul: We've been trying out a wrapper for caches like this so that we can instrument cache hit rates / sizes at some point: 

https://github.com/dbt-labs/metricflow/blob/main/metricflow-semantics/metricflow_semantics/semantic_graph/attribute_resolution/sg_linkable_spec_resolver.py#L74

wiggzz: @plypaul updated for this. Once this is merged, will a release be automatically made? Then we'll need to bump metricflow in MFS.

PR #1966, opened 2026-02-02
Add SQLAlchemy-based SqlClient implementations for metricflow tests

tlento:
Add SQLAlchemy-based SqlClient implementations for metricflow tests

The engine tests currently execute via calls through dbt adapters
maanged by an AdapterBackedSqlClient instance (and its corresponding
DDL-enabled class).

In order to allow for ongoing development and testing in a world where
dbt-core depends on metricflow we need to remove the pass-through
dependency on dbt-core that our reliance on dbt-adapters imposes on
us.

This is the first step in removing metricflow's test package dependencies
on dbt adapters - moving test execution to a SQLAlchemy-based client
that has many of the same test configuration advantages of the 
AdapterBackedSqlClient without the dbt dependencies.

The full changeset associated with this (PR #1966) includes the addition
of the client plus independent changes to cut over each engine, as follows:

1. Cut DuckDB tests over to SqlAlchemySqlClient

This includes one necessary update to error messaging snapshots, as the dbt
adapters use a custom exception wrapper that does dbt-specific formatting.

2. Cut postgres tests over to SqlAlchemySqlClient

No special changes required.

3. Cut Trino over to SqlAlchemy SqlClient

Trino uses a catalog rather than a database, and in their
sqlalchemy dialect they override the standard URL class to always
be a string url. Since that was a nuisance I chose to simply overload
the database element in the standard SqlAlchemy URL format and URL class
to be the catalog for the Trino case. This works in our tests, and probably
works in general, but it might not be robust in all scenarios.

In addition, semi-colons in sql expressions were causing syntax
errors with the sqlalchemy Trino client, so these were removed from the
affected test cases.

4. Cut Databricks tests over to SqlAlchemySqlClient

This includes a bug fix for extracting the httppath parameter
from the engine URL. The original issue never emerged because it was in
a dead code branch, but the SqlAlchemy client uses it.

5. Cut Snowflake tests over to SqlAlchemySqlClient

Snowflake's SQLAlchemy connector hews to the SQLAlchemy standard
of rendering case-insensitive object names in lower-case, so this
change updates our snapshots accordingly.

6. Cut Redshift tests over to SqlAlchemySqlClient

Making Redshift work with SqlAlchemy was a bit trickier than expected,
as the "official" SqlAlchemy client for Redshift hasn't been updated
in 2 years and does not work with SqlAlchemy 2.x.

As a hack around this problem, this change adds a bare-bones custom
dialect override for  psycopg2 that clobbers some class property to
allow for backwards compatibility with Redshift. Since we only ever
pass in sql query strings as sa_text() and always qualify our
relation names with the schema inline in the queries this works, but
if we ever need to relax those constraints we'll either need to find
a more full-featured dialect for Redshift or else switch to using the
redshift-connector package directly.

7. Cut BigQuery tests over to SqlAlchemySqlClient

BigQuery has some special connection requirements, and does not support
the EXPLAIN keyword, so the dry run configuration is also special.

This updates the client accordingly.

### DIFF
--- a/tests_metricflow/fixtures/sql_client_fixtures.py
+++ b/tests_metricflow/fixtures/sql_client_fixtures.py
@@ -160,9 +160,8 @@ def make_sqlalchemy_test_sql_client(url: str, password: str, schema: str) -> Sql
     sqlalchemy_url = SqlAlchemyUrlBuilder.build_url(connection_params, password, schema)
 
     if dialect is SqlDialect.BIGQUERY:
-        bq_credential_string = password.replace("'", "")
         # `strict=False` required to work with BQ password characters.
-        bq_credentials = json.loads(bq_credential_string, strict=False)
+        bq_credentials = json.loads(password, strict=False)
         engine = create_engine(
             sqlalchemy_url,
             pool_pre_ping=True,  # Verify connections before using
--- a/tests_metricflow/fixtures/sql_clients/sqlalchemy_client.py
+++ b/tests_metricflow/fixtures/sql_clients/sqlalchemy_client.py
@@ -19,7 +19,7 @@ from metricflow.sql.render.sql_plan_renderer import SqlPlanRenderer
 logger = logging.getLogger(__name__)
 
 
-class MFRedshiftDialect(PGDialect_psycopg2):
+class MetricFlowRedshiftDialect(PGDialect_psycopg2):
     """Custom dialect for Redshift.
 
     The sqlalchemy-redshift package does not support SqlAlchemy 2.x natively, so we include this simple custom
@@ -33,20 +33,21 @@ class MFRedshiftDialect(PGDialect_psycopg2):
     Note - none of the upstream base classe are typed, so we just override everything here for mypy.
     """
 
+    # These properties are set at class level in the built in dialect classes, so we follow that standard here.
     name = "mf_redshift_psycopg2"
-    supports_statement_cache: bool = False  # type: ignore  # noqa
+    supports_statement_cache = False
 
-    def _set_backslash_escapes(self, connection):  # type: ignore  # noqa
+    def _set_backslash_escapes(self, connection):  # type: ignore
         """Override for problematic method in SqlAlchemy 2.x.
 
         This is a method returning a value that overrides a boolean property in the base class.
         Without this override the default Postgres dialect will execute a statement that redshift does not support.
         """
-        return False  # type: ignore  # noqa
+        return False
 
 
 # Make our custom redshift dialect available to SqlAlchemy.
-registry.register("mf_redshift_psycopg2", __name__, "MFRedshiftDialect")
+registry.register("mf_redshift_psycopg2", __name__, "MetricFlowRedshiftDialect")
 
 
 class SqlAlchemyBasedSqlClient:
@@ -68,7 +69,7 @@ class SqlAlchemyBasedSqlClient:
         self._dry_run_engine = engine if dry_run_engine is None else dry_run_engine
         self._sql_engine_type = sql_engine_type
         self._sql_plan_renderer = sql_plan_renderer
-        logger.debug(f"Initialized SqlAlchemyBasedSqlClient with engine type `{sql_engine_type.value}`")
+        logger.debug(LazyFormat("Initialized SqlAlchemyBasedSqlClient.", engine_type=f"{sql_engine_type.value}"))
 
     @property
     def sql_engine_type(self) -> SqlEngine:
@@ -126,7 +127,7 @@ class SqlAlchemyBasedSqlClient:
                 # (appropriate for SELECT queries)
 
         except SQLAlchemyError as e:
-            logger.error(f"Query failed: {e}")
+            logger.error(LazyFormat("Query failed:", error=f"{e}"))
             raise
 
         stop = time.perf_counter()
@@ -172,11 +173,11 @@ class SqlAlchemyBasedSqlClient:
                 conn.commit()
 
         except SQLAlchemyError as e:
-            logger.error(f"Execute failed: {e}")
+            logger.error(LazyFormat("Execute failed:", error=f"{e}"))
             raise
 
         stop = time.perf_counter()
-        logger.info(f"Finished execute() in {stop - start:.2f}s")
+        logger.info(LazyFormat("Finished execute()", runtime=f"{stop - start:.2f}s"))
 
     def dry_run(
         self,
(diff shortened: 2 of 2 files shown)


tests_metricflow/fixtures/sql_clients/sqlalchemy_client.py
```diff
@@ -17,6 +19,36 @@
 logger = logging.getLogger(__name__)
 
 
+class MFRedshiftDialect(PGDialect_psycopg2):
```
plypaul: We haven't been using `MF` in class names (though maybe we should) to shorten `MetricFlow`. For consistency, how about `MetricFlowRedshiftDialect`? Or just `RedshiftDialect`?

tests_metricflow/fixtures/sql_clients/sqlalchemy_client.py
```diff
@@ -17,6 +19,36 @@
...
+    this is sufficient for our purposes.
+
+    A full-featured redshift SqlClient would need to instead use something like redshift-connector, or fall back
+    to sqlalchemy-redshift with sqlalchemy < 2.0.
+
+    Note - none of the upstream base classe are typed, so we just override everything here for mypy.
+    """
+
+    name = "mf_redshift_psycopg2"
+    supports_statement_cache: bool = False  # type: ignore  # noqa
```
plypaul: What's the context for setting the field this way?

tests_metricflow/fixtures/sql_client_fixtures.py
```diff
@@ -150,21 +148,43 @@ def __initialize_dbt() -> None:
...
     sqlalchemy_url = SqlAlchemyUrlBuilder.build_url(connection_params, password, schema)
 
-    # Create engine with connection pooling
-    engine = create_engine(
-        sqlalchemy_url,
-        pool_pre_ping=True,  # Verify connections before using
-        echo=False,  # Set to True for SQL logging
-    )
+    if dialect is SqlDialect.BIGQUERY:
+        bq_credential_string = password.replace("'", "")
```
plypaul: Is it supposed to remove all instances of `'` or just at the ends?

tests_metricflow/fixtures/sql_clients/sqlalchemy_client.py
```diff
@@ -0,0 +1,218 @@
...
+
+        Args:
+            engine: SqlAlchemy Engine instance (already configured)
+            sql_engine_type: MetricFlow SqlEngine enum value
+            sql_plan_renderer: Dialect-specific SQL plan renderer
+        """
+        self._engine = engine
+        self._sql_engine_type = sql_engine_type
+        self._sql_plan_renderer = sql_plan_renderer
+        logger.debug(f"Initialized SqlAlchemyBasedSqlClient with engine type `{sql_engine_type.value}`")
```
plypaul: Assuming this was from prior code so we can leave as is, but we've been moving log messages to use `LazyFormat`.

tests_metricflow/fixtures/sql_clients/sqlalchemy_client.py
```diff
@@ -0,0 +1,218 @@
...
+        sql_bind_parameter_set: SqlBindParameterSet = SqlBindParameterSet(),
+    ) -> MetricFlowDataTable:
+        """Execute a query and return results as MetricFlowDataTable.
+
+        Uses SqlAlchemy Connection with automatic transaction management.
+        For SELECT queries, transaction is rolled back after fetching results.
+        """
+        start = time.perf_counter()
+
+        if sql_bind_parameter_set.param_dict:
```
plypaul, line 96: We actually received a request to support bind parameters again. Outside the scope of the PR, but would it be straightforward to enable this case?

tests_metricflow/fixtures/sql_client_fixtures.py
tlento: I think it's all but I don't remember why this was here. I copied it from the existing logic.

I can add a comment indicating what I think is supposed to happen and why, but ultimately all the BigQuery token handling logic here is going to have to be "keep futzing around with it until it works" because BigQuery is so obnoxious.

tests_metricflow/fixtures/sql_clients/sqlalchemy_client.py
tlento: Will do!

tests_metricflow/fixtures/sql_clients/sqlalchemy_client.py
tlento: I can switch to LazyFormat, wasn't sure if that was preferred or if it was only for cases where it would help.

tests_metricflow/fixtures/sql_clients/sqlalchemy_client.py
tlento: Nothing about bind parameters is straightforward. 😛 But given that, this client shouldn't be a major source of problems for you.

We use sa_text() everywhere, so you can pass bind parameters through to the execute() call. Within a test context this shouldn't be too difficult to get working. The issue is we have to render the inline parameter using the SqlAlchemy format, which works for some-but-not-all dialects even within SqlAlchemy. Databricks, for example, caused a bunch of problems back in the day, but we were able to work around that.

More broadly, you'll have to be cautious about how those bind parameters get specified and rendered in the final query text. The rendering is probably not the same for every dialect within SqlAlchemy (and our Redshift shim might cause weird issues, although I'd expect it to work). It also may not be the same for SqlAlchemy vs JDBC/DBAPI/ArrowFlightSQL/etc. - engines do have divergence in input and output formats between their SqlAlchemy implementations and their direct-access APIs.

tests_metricflow/fixtures/sql_clients/sqlalchemy_client.py
tlento: SqlAlchemy does a lot of magical things under the hood, and their built-in dialects (and the handful of other dialects I've looked at) [set these fields at class level](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/dialects/postgresql/psycopg2.py#L605-L607), so I figured I wouldn't mess with the format.

If you're asking about the weird typing and stuff I think that's AI slop I didn't notice - I asked Cursor to suppress errors and I think it got a little over-eager. I can clean that bit up, at least.

tests_metricflow/fixtures/sql_client_fixtures.py
tlento: Looks like we don't need the replace at all, so I took it out!

PR #1967, opened 2026-02-02
Remove all dbt dependencies from metricflow test environments

tlento:
We no longer use dbt at all inside of Metricflow prpoer, so
this removes all dependencies.

The main changes of interest involve the updates to the pre-commit
configuration. The original config was not package-specific since
all critical dependencies were available from the metricflow
environment, so we could run linters without any custom installations.

Now that metricflow has no dbt dependency in its dev environment,
mypy will no longer typecheck correctly via pre-commit in the
dbt-metricflow package. Since mypy is the only affected item we work
around this by invoking it from the makefile for dbt-metricflow,
and otherwise excluding dbt-metricflow from the pre-commit execution.


tlento:
### Merge activity

* **Feb 5, 1:39 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/dbt-labs/metricflow/1967).
* **Feb 5, 1:42 AM UTC**: [Graphite](https://app.graphite.com/github/pr/dbt-labs/metricflow/1967) rebased this pull request as part of a merge.

PR #2012, opened 2026-04-03
Fix check for aggregation time dimensions in filters

plypaul:
This PR fixes one of the bugs that is causing incorrect filter placement. The check to see if the filter is only on aggregation time dimensions

```
if len(included_agg_time_specs) == len(filter_spec.linkable_specs):
```

has an edge case when both are empty.


theyostalservice (approved):
ok, this makes sense.

PR #2013, opened 2026-04-03
Store filter placement context in `SimpleMetricRecipe`

plypaul:
The logic for how filters are applied for time-offset metrics was scattered in several places in the dataflow plan builder and was challenging to understand. To make later bug fixes easier to apply / review, this PR consolidates the logic for filter placement and stores placement information into `SimpleMetricRecipe`. The recipe becomes the source of truth for filter placement and can be passed around.

There should be no logic changes (no changes in the test snapshots) and some of the code comments were copied without revision.

Later, we can break down the build method, move it to a separate class, revise comments, etc.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -16,6 +16,12 @@ This project uses **hatch** for dependency management and testing:
   previously reflected the type, rename the variable appropriately to 
   reflect the new type.
 - If `git_ignored/AGENTS.md` exists, append the rules in that file.
+- If a function's return type is not mutable, the returned object does
+  not need to be a copy.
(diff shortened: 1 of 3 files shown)


theyostalservice (approved):
Ok, after our conversation, this makes sense to me.  Everything looks correct, but I'm also glad there are a lot of tests to back up the idea that this isn't changing anything.

In the future, I think it would be worth the time investment to build a document that covers all the different forces determing how a join to timespine and metric time aggregation filter are going to collide into this exponential stretch of madness.  We really need to make it easier to on board people to the hard things in this codebase. :/

PR #2014, opened 2026-04-03
Clickhouse support

ubiquitousbyte:
Hey everyone, 

I've decided to revive @britton-stamper's initial effort (https://github.com/dbt-labs/metricflow/pull/1961) to add ClickHouse support to MetricFlow.

I've had to disable some subquery optimizations in the reducer **only** for the ClickHouse engine. For this reason, each engine now has dedicated `test_reduce_sub_query__<before|after>_reducing.sql` files.

Try it out! :)

How to run all the tests:
```
make clickhouse
hatch run clickhouse-env:pytest tests_metricflow/ --no-header -q
```


local-data-warehouses/README.md
```diff
@@ -26,3 +26,32 @@ export MF_SQL_ENGINE_PASSWORD="metricflowing"
...
+
+```sh
+make clickhouse
+```
+
+Then, when running `pytest`, ensure that `MF_SQL_ENGINE_URL` and `MF_SQL_ENGINE_PASSWORD` are setup
+to access the ClickHouse instance.
+
+```sh
+export MF_SQL_ENGINE_URL="clickhouse://metricflow@localhost:8123/metricflow"
```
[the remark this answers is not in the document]
ubiquitousbyte: Change your suggestion to use `clickhousedb`, without the +http portion.

metricflow/sql/optimizer/rewriting_sub_query_reducer.py
```diff
@@ -300,6 +303,18 @@ def _current_node_can_be_reduced(self, node: SqlSelectStatementNode) -> bool:
...
+        # analyzer (enable_analyzer=1) resolves the WHERE column reference to the SELECT aggregate alias instead of the
+        # source column when the names match, causing Code 184 ILLEGAL_AGGREGATION. This guard is opt-in and only
+        # enabled for ClickHouse via the prevent_where_hoist_with_aggregates flag.
+        if (
+            self._prevent_where_hoist_with_aggregates
+            and from_source_node_as_select_node.where is not None
+            and any(col.expr.lineage.contains_aggregate_exprs for col in node.select_columns)
+        ):
+            return False
+
```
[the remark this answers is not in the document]
ubiquitousbyte: I would rather rely on ClickHouse's internal query optimizer to detect such cases, rather than fill up the reducer with ClickHouse-specific logic for the time being. The PR is large enough, let's revisit this at a later point.

requirements-files/requirements-clickhouse.txt
```diff
@@ -0,0 +1 @@
+clickhouse-connect
```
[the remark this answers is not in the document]
ubiquitousbyte: It's not gated behind an extra. But there are newer versions. Check them out and suggest a more up-to-date version constraint.

metricflow/sql/optimizer/rewriting_sub_query_reducer.py
```diff
@@ -89,6 +89,9 @@ class SqlRewritingSubQueryReducerVisitor(SqlPlanNodeVisitor[SqlPlanNode]):
     Unlike SqlSubQueryReducerVisitor, this will re-write expressions to realize more reductions.
     """
 
+    def __init__(self, prevent_where_hoist_with_aggregates: bool = False) -> None:  # noqa: D107
```
plypaul, line 92: Could you add an example of SQL that is generated with and with out this set?

plypaul (commented):
Thanks for picking this up! While we can help get this in, we'll have to classify it as a preliminary / experimental feature until we get more deployment experience.

metricflow/sql/optimizer/rewriting_sub_query_reducer.py
ubiquitousbyte: Yes, of course! 


Here's a query with this set (`prevent_where_hoist_with_aggregates = True`).
```sql
SELECT <--- Level 1
  metric_time__day
  , SUM(__homonymous_metric_and_entity) AS homonymous_metric_and_entity
FROM (
  SELECT <--- Level 2
    metric_time__day
    , homonymous_metric_and_entity AS __homonymous_metric_and_entity
  FROM (
    SELECT <--- Level 3
      toStartOfDay(ds) AS metric_time__day
      , booking_value AS homonymous_metric_and_entity
    FROM mf_test.fct_bookings bookings_source_src_32000
  ) subq_32010
  WHERE homonymous_metric_and_entity IS NOT NULL
) subq_32012
GROUP BY
  metric_time__day
```
In this query, the Level 2 check `WHERE homonymous_metric_and_entity IS NOT NULL` refers to `booking_value` in Level 3. ClickHouse knows how to resolve the alias, everything works.

The reducer detects the following optimization opportunities: 
1. The `toStartOfDay(ds)` can be hoisted up one level, from Level 3 to Level 2.
2. The `WHERE homonymous_metric_and_entity IS NOT NULL` can be hoisted up from Level 2 to Level 1.

In combination, these two changes completely eliminate the Level 3 query, leading to the following (`prevent_where_hoist_with_aggregates = False`):
```sql
SELECT
  metric_time__day
  , SUM(homonymous_metric_and_entity) AS homonymous_metric_and_entity
FROM (
  SELECT
    toStartOfDay(ds) AS metric_time__day
    , booking_value AS homonymous_metric_and_entity
  FROM mf_test.fct_bookings bookings_source_src_32000
) subq_32010
WHERE homonymous_metric_and_entity IS NOT NULL
GROUP BY
  metric_time__day
```
The problem with this query is that the `homonymous_metric_and_entity` reference in the `WHERE` clause is ambiguous, because it can refer to one of two things:
1. The `booking_value` field from the nested query.
2. The `SUM(homonymous_metric_and_entity)` expression that it aliases. 

Most engines assume 1. and resolve the alias correctly to the inner query. Unfortunately, ClickHouse assumes 2, which produces an invalid query because aggregate expressions like `SUM(homonymous_metric_and_entity`) are not allowed in `WHERE` clauses. 


## Other fixes considered
1. ClickHouse has a setting `prefer_column_name_to_alias=1` that technically fixes this issue. Unfortunately, that setting causes other ambiguities, e.g ClickHouse starts resolving aliases in `ORDER BY` clauses to raw source columns rather than aggregate expressions.

ubiquitousbyte:
@plypaul, thanks for the review.

> Thanks for picking this up! While we can help get this in, we'll have to classify it as a preliminary / experimental feature until we get more deployment experience.

That makes sense, I'd also prefer marking this an experimental feature for now.

metricflow/sql/optimizer/rewriting_sub_query_reducer.py
plypaul: Thanks for the example - that was very helpful. To clarify Clickhouse behavior, is the issue specific to aggregation functions? To modify your example, in the query below:

```sql
SELECT
  metric_time__day
  , coalesce(homonymous_metric_and_entity, 0) AS homonymous_metric_and_entity
FROM (
  SELECT
    toStartOfDay(ds) AS metric_time__day
    , booking_value AS homonymous_metric_and_entity
  FROM mf_test.fct_bookings bookings_source_src_32000
) subq_32010
WHERE homonymous_metric_and_entity IS NOT NULL
```

Does the filter `WHERE homonymous_metric_and_entity IS NOT NULL` apply to `booking_value` from the inner query or `coalesce(homonymous_metric_and_entity, 0)`?

If it applies to `coalesce(homonymous_metric_and_entity, 0)`, it seems like the order of evaluations is different for Clickhouse. Instead of evaluating the `WHERE` clause and then the `SELECT` clause, it evaluates in reverse.

ubiquitousbyte: Thank you for the quick followup and the great question, @plypaul. 

The filter `WHERE homonymous_metric_and_entity IS NOT NULL` in your example is applied to `coalesce(homonymous_metric_and_entity, 0)`. The reason it only blows up visibly with aggregates is because ClickHouse separately rejects aggregate functions inside the `WHERE` clause. With `coalesce`, the resolution still works but you get semantically different (and incorrect) filtering.

I did some further research to figure out why that is. Here are my findings.

In ClickHouse, aliases can be defined and used in **any part** of a query, as stated in https://github.com/ClickHouse/ClickHouse/issues/23194. This means that the order of evaluations is indeed different for ClickHouse compared to a standard SQL engine. Aliases in the `SELECT` clause are visible in the `WHERE` clause.

Hence, `WHERE homonymous_metric_and_entity IS NOT NULL` sees `coalesce(homonymous_metric_and_entity, 0) AS homonymous_metric_and_entity ` in ClickHouse.

In standard SQL, aliases defined in `SELECT` are only visible in `GROUP BY`/`HAVING`/`ORDER BY`, because `WHERE` is logically evaluated before `SELECT`. 

Hence, the evaluation of the alias **happens after** the evaluation of the `WHERE` clause and the coalescing is never visible during filtering.

With this information, I am now certain that the check `prevent_where_hoist_with_aggregates` is insufficient and only handles a subset of potential issues that may occur for ClickHouse.

So how can we handle this?

My initial thinking is make all references fully table-qualified. A table-qualified reference cannot be resolved to a `SELECT`-list alias by any SQL engine — it unambiguously refers to the `FROM` source — so the collision is impossible. 

Your example would turn into:
```sql
SELECT
  metric_time__day
  , coalesce(homonymous_metric_and_entity, 0) AS homonymous_metric_and_entity
FROM (
  SELECT
    toStartOfDay(ds) AS metric_time__day
    , booking_value AS homonymous_metric_and_entity
  FROM mf_test.fct_bookings bookings_source_src_32000
) subq_32010
WHERE subq_32010.homonymous_metric_and_entity IS NOT NULL <-- Now table-qualified
```

I am not sure whether this is possible at the moment, nor where to do it - I'll look into that. 

In the meantime, do you agree with the approach? Any concerns?

plypaul: Actually, I'm thinking about another approach. How about updating `SqlRewritingSubQueryReducerVisitor._current_node_can_be_reduced` so that it return false if the engine evaluates aliases early (i.e. Clickhouse) and the current node contains any expression that is a `SqlStringExpression`? That should be the only way that a column is referenced without a table alias as `SqlColumnReferenceExpression` requires a table alias.

I haven't fully worked out the details, so there might be some issues. Let me know what you think and if you run into any difficulties. The SQL rewriting is a tricky part of the system.

roarking:
@ubiquitousbyte @plypaul, following up on @plypaul’s note about wanting more deployment experience before merge.

We’re running this PR’s adapter at Kustomer as the engine inside a metric-service PoC. The fork is v0.211.0 with the substantive commits from this PR squashed on top (renderer, reducer guard, dbt-clickhouse adapter wiring). No behavior changes from the PR itself.

**Setup**
Warehouse: ClickHouse Cloud (current major).
dbt adapter: dbt-clickhouse against curated models in our warehouse.
Manifest: 14 semantic models, 56 metrics from our dbt project,
hot-reloaded via a small wrapper.
Query shape mix (10 corpus shapes covering our exit criteria):
simple count, no slicing
simple count rolled up by metric_time__day
simple count sliced by a categorical dimension
derived ratio, no slicing
derived ratio sliced by a categorical dimension
multi-metric one query (two counts + one ratio)
limit + descending order on a metric
cross-domain query against a different semantic model
derived ratio with a denominator-can-be-zero shape:
(measure_a / nullif(measure_b, 0)) * 100
cumulative metric, 30-day rolling window
All 10 return correct values. For each one we re-execute the compiled
SQL directly against ClickHouse Cloud and compare; values match.

**One quirk in production**
The renderer logs this on first compile, then continues: WARNING metricflow.sql.render.clickhouse: The ClickHouse SQLAlchemy dialect loses column metadata (result.keys()) for zero-row results when the SQL query begins with leading -- comment lines. Comments are suppressed to avoid this bug.
It hits because the renderer strips leading SQL comments that mf query --compile normally includes. Output is correct, but the served SQL is not byte-identical to mf query --compile from the same manifest. Worth a line in the experimental-feature docs so deployers diffing served vs CLI SQL don’t spend an afternoon chasing it.

The cleaner fix is upstream in clickhouse-sqlalchemy; the renderer’s workaround is the right call until that exists.

**On @plypaul’s _current_node_can_be_reduced suggestion**
For our workload mix above we haven’t seen optimization regressions from the current prevent_where_hoist_with_aggregates guard. We haven’t benchmarked the alternative formulation, so we can’t compare in the general case, but we wouldn’t block merge on the guard’s exact shape. Tracking it post-merge as a refinement works for us.

**Our ask**
Classify as experimental (as agreed) and merge. The deployment-experience signal we can add: an in-product service calling the engine compiles every metric shape in our corpus and returns correct values against ClickHouse Cloud, not just local docker.

**What Kustomer can contribute back**
Our wrapper-fidelity test pattern: re-execute the compiled SQL, compare values. Quick smoke check that the renderer isn’t lying about what it ran. Can open as a separate small contribution if useful.
Integration-test scenarios for the 10 corpus shapes, as ClickHouse snapshot fixtures, if the harness in tests_metricflow/snapshots/.../ClickHouse/ can absorb them. Real-traffic latency numbers from our PoC’s drilldown shape, in progress; we’ll follow up in this thread when they’re ready.

metricflow/sql/optimizer/rewriting_sub_query_reducer.py
ubiquitousbyte: So if my understanding is correct, `SqlStringExpression` is the only expression type that introduces unqualified column names into the SQL tree, making it the sole vector for ClickHouse's early alias resolution bug.

I've made the necessary changes to reflect your proposal. Please, give it another read when you can, @plypaul.

plypaul:
@ubiquitousbyte Changes look good, but checking to see if you saw the CI failures? Also, could you add a job similar to Trino in `.github/workflows/cd-sql-engine-tests.yaml`?

metricflow/sql/optimizer/rewriting_sub_query_reducer.py
```diff
@@ -870,11 +890,20 @@ class SqlRewritingSubQueryReducer(SqlPlanOptimizer):
     GROUP BY foo
     """
 
-    def __init__(self, use_column_alias_in_group_bys: bool = False) -> None:  # noqa: D107
+    def __init__(
+        self,
+        use_column_alias_in_group_bys: bool = False,
+        has_ambiguous_alias_resolution_in_where: bool = False,
```
plypaul, line 896: With Clickhouse, the alias resolution behavior is not specific to the `WHERE` clause right? If so, should we just rename this to `has_ambiguous_alias_resolution`? No logic changes seem necessary.

roarking:
@ubiquitousbyte @plypaul Following up from our June deployment report: we've resynced our fork to the current PR head (`da5c424c`) to pick up the reworked `has_ambiguous_alias_resolution_in_where` guard. The full engine test suite passes on v0.211.0 + this PR's delta. The more targeted formulation is a real improvement for our workload — qualified column references now hoist freely, and only raw string expressions in WHERE block reduction, which matches the actual failure vector we observed.

On the failing lint job: it's ruff D107 on `SqlRewritingSubQueryReducer.__init__` — the `# noqa: D107` sits on the closing line of the multi-line signature, where ruff doesn't honor it. Moving it to the `def` line fixes it. @ubiquitousbyte, I've opened ubiquitousbyte/metricflow#2 against your `clickhouse-support` branch with that one-line change and merging it should turn the lint job green here.

Still happy to contribute the wrapper-fidelity test pattern and ClickHouse snapshot fixtures for our corpus shapes once the `cd-sql-engine-tests` job is done.

dave-connors-3:
hey @plypaul! long time no see! 

@ubiquitousbyte are you able to accept the PR from @roarking to update the formatting? would love to help get this over the line if we can!

dave-connors-3:
@plypaul -- is https://github.com/dbt-labs/metricflow/pull/2124 the latest PR for clickhouse support? still very eager to add this feature, but wondering where to centralize conversation

PR #2016, opened 2026-04-04
Fix bug with filters applied before aggregation in time-offset metrics

plypaul:
This PR fixes an issue for time-offset metrics where filters on aggregation time dimensions are applied before the time spine join. Applying the filter before the time-offset results in incorrect filtering of the source as the date specified in the filter refers to the value after the time offset.

### DIFF
--- a/tests_metricflow/snapshots/test_offset_metrics_with_filters.py/SqlPlan/DuckDB/test_offset_metric_with_separate_metric_time_and_dimension_filters__plan0_optimized.sql
+++ b/tests_metricflow/snapshots/test_offset_metrics_with_filters.py/SqlPlan/DuckDB/test_offset_metric_with_separate_metric_time_and_dimension_filters__plan0_optimized.sql
@@ -1,0 +1,73 @@
+test_name: test_offset_metric_with_separate_metric_time_and_dimension_filters
+test_filename: test_offset_metrics_with_filters.py
+docstring:
+  Test querying a time-offset metric with separate filters that allow for different filter placement.
+sql_engine: DuckDB
+expectation_description:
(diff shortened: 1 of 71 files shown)


theyostalservice (approved):
Thanks for walking through the code with me.

This seems correct and the tests that are updated look great.  Conditionally approving assuming you will:
1. update the postgres snapshots and get a green test run :-P
2. Find or add a test that shows a case where there are BOTH a filter that should only be applied to the time spine and also a filter that remains applied to the base table.  (None of the ones we found left any filter in place on the source table, and that is functionality we should be sure is working and stays working, too. :-) )

tests_metricflow/snapshots/test_offset_metrics_with_filters.py/SqlPlan/DuckDB/test_offset_metric_with_separate_metric_time_and_dimension_filters__plan0_optimized.sql
```diff
@@ -0,0 +1,73 @@
...
+          , listing_id AS listing
+          , 1 AS __bookings
+        FROM ***************************.fct_bookings bookings_source_src_28000
+      ) subq_19
+      LEFT OUTER JOIN
+        ***************************.dim_listings_latest listings_latest_src_28000
+      ON
+        subq_19.listing = listings_latest_src_28000.listing_id
+    ) subq_24
+    WHERE listing__country_latest == 'us'
```
plypaul, line 67: @theyostalservice FYI here is the result of the separate filters case (one on the aggregation time dimension and another on a dimension). The one for the aggregation time dimension is applied to the time spine and the one for the dimension is applied to the source.

PR #2017, opened 2026-04-06
OSI Converters

QMalcolm:
## Summary                                                                                                                      
  - Adds Pydantic data models for OSI v0.1.1 (`OSIDocument`, `OSISemanticModel`, `OSIDataset`, `OSIField`, `OSIMetric`, `OSIRelationship`, and supporting types)                                                                                              
  - Implements `MSIToOSIConverter` — converts a `SemanticManifest` to an `OSIDocument`, including dataset/field conversion, entity-based relationship inference, and recursive metric expression inlining for SIMPLE, CUMULATIVE, RATIO, and DERIVED metric types
  - Implements `OSIToMSIConverter` — reverse direction, converting an `OSIDocument` back to a `PydanticSemanticManifest` with field classification (entity/dimension/measure) inferred from key declarations and metric expressions                                       
  - Flattens MSI `WhereFilter` Jinja templates into SQL `CASE WHEN` expressions inside aggregations so metric-level, measure-level, and input-metric-level filters are preserved in OSI output                                                                                
  - CONVERSION metrics are skipped (no OSI equivalent); cumulative window/grain semantics are dropped with the base aggregation preserved                                                                                                                             
                                         
## Known limitations                                                                                        
  - Filter column references render as MetricFlow-qualified names (`entity__dimension`) rather than resolved SQL column aliases — full resolution via `WhereFilterSpecFactory` is deferred                                                                                   
  - CONVERSION metrics have no OSI representation and are omitted from output
  - Cumulative window/grain-to-date semantics are not expressible in OSI and are dropped


metricflow/converters/msi_to_osi.py
```diff
@@ -181,5 +323,34 @@ def _build_relationships(
                 )
         return relationships
 
+    @staticmethod
+    def _build_agg_expression(
+        agg: AggregationType,
+        col: str,
+        agg_params: Optional[MeasureAggregationParameters],
+    ) -> str:
+        if agg == AggregationType.SUM:
```
plypaul, line 410: For better type safety, you can do something like 

https://github.com/dbt-labs/metricflow/blob/2c0a38888af33c35b2156500421a30de9e087727/metricflow_semantics/sql/sql_exprs.py#L766

to allow the type checker to verify that all cases have been handled. This applies for other times that a switch is done on an enum.

metricflow/converters/osi/converter.py
```diff
@@ -256,9 +255,8 @@ def _resolve_cumulative(
 
         # cumulative_type_params.metric path: recurse into the referenced metric
         cumulative_params = metric.type_params.cumulative_type_params
-        assert (
-            cumulative_params is not None and cumulative_params.metric is not None
-        ), f"CUMULATIVE metric '{metric.name}' has no resolvable measure or sub-metric"
+        if cumulative_params is None or cumulative_params.metric is None:
+            raise ValueError(f"CUMULATIVE metric '{metric.name}' has no resolvable measure or sub-metric")
```
plypaul: We've been consolidating exception and log message formatting to use `LazyFormat`. e.g.

```
raise ValueError(
    LazyFormat(
        "A CUMULATIVE metric has no resolvable measure or sub-metric",
        metric_name=metric.name
    )
)
```

 It handles escaping (e.g. what if `metric.name` has a `'`), automatic object to string formatting (e.g. you could just log the whole metric object), and does not format stings until needed. Not so relevant for this case, but we previously found performance bottlenecks in string formatting.

metricflow/converters/osi/converter.py
```diff
@@ -691,101 +691,31 @@ def _convert_dataset(
...
+                field,
+                expr,
+                expr_or_none,
+                dataset.name,
+                primary_key_cols,
+                unique_key_cols,
+                foreign_key_cols,
+                measure_index,
+            )
+            if isinstance(element, PydanticEntity):
```
plypaul: General repo practice is to avoid `isinstance` checks except for cases where it's needed like `__eq__`. Could we do this another way?

plypaul (commented):
Some high level comments:

* Can the Pydantic models be made immutable? There have been some cases where it would be helpful to use those objects as keys.
* For many of the tests, it seems like they could use snapshots instead. This would make them easier to update as the OSI datasets / fields change. If so, I can move / update the snapshot utilities to make them available.
* Instead of parsing SQL via string operations, how about using `sqlglot`?

Happy to chat to provide additional context.

QMalcolm:
@plypaul I think only feedback I haven't gotten to yet is the tests, I'll do that next when I'm back at it tomorrow 🙂

tests_metricflow_semantic_interfaces/osi/helpers.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+def _measure(
+    name: str,
+    agg: AggregationType = AggregationType.SUM,
+    expr: str | None = None,
+    description: str | None = None,
+    label: str | None = None,
+) -> PydanticMeasure:
+    return PydanticMeasure(
```
plypaul, line 109: We've deprecated measures and changed them to simple metrics instead.

QMalcolm: I've updated the converts so that OSI -> MSI creates no measures. However, on MSI -> OSI we have to handle the situation where a simple metric exists that references a measure. Thus, I've left the relevant code in related to that. Now when converting a SimpleMetric MSI -> OSI, we check whether it has a measure. If it does, it uses that information for generating the expression. If it doesn't have a measure, then we assume it follows the new definition with the details being in the AggregationParams.

plypaul: Given that we have the transformation that creates simple metrics for measures, do we need to handle the case where a simple metric references a measure?

metricflow/converters/expression_utils.py
```diff
@@ -1,100 +1,136 @@
...
+    if isinstance(tree, (exp.PercentileCont, exp.PercentileDisc)):
+        order = tree.args.get("expression")
+        if isinstance(order, exp.Order) and order.expressions:
+            ordered = order.expressions[0]
+            col_node = ordered.this if isinstance(ordered, exp.Ordered) else ordered
+            col = _col_name(col_node)
+            if isinstance(tree, exp.PercentileCont):
+                try:
+                    p = float(tree.this.name)
+                except (AttributeError, ValueError):
```
plypaul: What's the context on handling the errors with `p = 0.5`?

tests_metricflow_semantic_interfaces/osi/helpers.py
QMalcolm: Hmmmmm. I assume you are referring to this transformation, [FixProxyMetricRule](https://github.com/dbt-labs/metricflow/blob/849411149b418dd59fcb47b246e5d4e96d999b47/metricflow_semantic_interfaces/transformations/fix_proxy_metrics.py#L19). Currently, as things stand, I don't think guarantee that `FixProxyMetricRule` has been run. That is `FixProxyMetricRule` is not run on the dbt-core side. `FixProxyMetricRule` is _only_ run (currently) when MetricFlow loads a semantic manifest.

To make such a change, we have to guarantee that the PydanticSemanticManifest being run on (converting MSI -> OSI), has been run through the `FixProxyMetricRule`. I think there are three possible paths forward:
1. Have dbt-core run the `FixProxyMetricRule`as part of it's parsing process
  a. Technically this should be a guarantee, however if there is a semantic_manifest.json produced by another system, then the guarantee is lost
2. Have the `MSIToOSIConverter` run the given semantic_manifest through the `FixProxyMetricRule`
3. Assume that the semantic_manifest given to `MSIToOSIConverter` has not been run through the `FixProxyMetricRule` and handle metrics that that still have a `measure` appropriately.

I think my vote is (2)? It has perhaps the worst time complexity of the three, but it gives the strongest guarantee while eliminating the measure related code in the converter.

plypaul: I'm not too familiar with the code that does the measure -> simple metric transformation, but I believe it's handled by `ReplaceInputMeasuresWithSimpleMetricsTransformationRule` and `FlattenSimpleMetricsWithMeasureInputsRule`.

The idea in option 2 sounds good though. Last I recall, the transformations are idempotent - could we just run the entire set?

QMalcolm: Yea we should be okay to run the entire set. I can also verify that by running it three times 😅

metricflow/converters/expression_utils.py
QMalcolm: This was actually masking an underlying issue, that the converter wasn't properly producing percentile metrics. It was building percentile metrics with `agg_params=None`. This try/except was catching when that happened and was defaulting to `p=0.5` as the "safest" option. the solution was to fix how we create percentile metrics 🙂

metricflow/converters/filter_utils.py
```diff
@@ -0,0 +1,141 @@
...
+        if f is None:
+            continue
+        for wf in f.where_filters:
+            rendered = _render_filter_template(wf.where_sql_template).strip()
+            if rendered:
+                parts.append(rendered)
+    return _merge_filter_sqls(*parts)
+
+
+def _merge_filter_sqls(*parts: Optional[str]) -> Optional[str]:
```
plypaul, line 134: In what cases would `None` be passed in?

metricflow/converters/models.py
```diff
@@ -0,0 +1,146 @@
...
+class OSIAIContextObject(FrozenBaseModel):
+    """Structured AI context with instructions, synonyms, and examples."""
+
+    instructions: Optional[str] = None
+    synonyms: Optional[List[str]] = None
+    examples: Optional[List[str]] = None
+
+    class Config:  # noqa: D106
+        extra = "allow"
+        allow_mutation = False
```
plypaul: This is a duplicate of `FrozenBaseModel`, but you need it because we need `extra = "allow"`? Would this work instead?

```
class Config(FrozenBaseModel.Config):
...
```

metricflow/converters/models.py
```diff
@@ -0,0 +1,146 @@
...
+    MDX = "MDX"
+    TABLEAU = "TABLEAU"
+    DATABRICKS = "DATABRICKS"
+
+
+class OSIAIContextObject(FrozenBaseModel):
+    """Structured AI context with instructions, synonyms, and examples."""
+
+    instructions: Optional[str] = None
+    synonyms: Optional[List[str]] = None
```
plypaul: Since these classes are frozen, the fields should also be immutable. An alternative constructor like

```
def create(instructions: Optional[Iterable[str]] = None, ...) ->  OSIAIContextObject:
```

can be handy if there are a lot of call sites that use mutable variables.

metricflow/converters/msi_to_osi.py
```diff
@@ -0,0 +1,424 @@
...
+        )
+
+    @staticmethod
+    def _extract_keys(entities: Sequence[Entity]) -> Tuple[Optional[List[str]], List[List[str]]]:
+        primary_key: Optional[List[str]] = None
+        unique_keys: List[List[str]] = []
+
+        for entity in entities:
+            col = entity.expr if entity.expr is not None else entity.name
+            if entity.type == EntityType.PRIMARY:
```
plypaul: For enum comparisons, it's preferable to use `is`. Applies to other enum comparisons in the PR/

metricflow/converters/msi_to_osi.py
```diff
@@ -0,0 +1,424 @@
...
+        """Map metric name to Metric for recursive resolution."""
+        return {metric.name: metric for metric in metrics}
+
+    @staticmethod
+    def _lookup_metric(metric_index: Dict[str, Metric], name: str, context: str) -> Metric:
+        """Look up a metric by name, raising a clear ValueError if not found."""
+        try:
+            return metric_index[name]
+        except KeyError:
+            raise ValueError(LazyFormat("references unknown metric", context=context, metric_name=name))
```
plypaul: Nit: maybe `Unknown metric referenced`?

metricflow/converters/msi_to_osi.py
```diff
@@ -0,0 +1,424 @@
...
+                if entity.type == EntityType.NATURAL:
+                    continue
+                col = entity.expr if entity.expr is not None else entity.name
+                index[entity.name].append((sm.name, col, entity.type))
+        return dict(index)
+
+    @staticmethod
+    def _relationship_direction(
+        ds_a: str, col_a: str, type_a: EntityType, ds_b: str, col_b: str, type_b: EntityType
+    ) -> Tuple[str, str, str, str]:
```
plypaul: Oof, lots of items in the tuple. How about a dataclass return type?

metricflow/converters/msi_to_osi.py
```diff
@@ -0,0 +1,424 @@
...
+        """Return (from_ds, to_ds, from_col, to_col) obeying OSI directionality.
+
+        OSI spec: ``from`` is the many-side (FK holder), ``to`` is the one-side (PK holder).
```
plypaul: Did you mean to use double backticks?

tests_metricflow_semantic_interfaces/osi/test_osi_to_msi.py
```diff
@@ -0,0 +1,382 @@
...
+            ]
+        )
+        sm = OSIToMSIConverter().convert(doc).semantic_models[0]
+
+        dim = sm.dimensions[0]
+        assert dim.description == "Order status"
+        assert dim.label == "Status"
+
+
+class TestOSIToMSIMetricConversion:  # noqa: D101
```
plypaul, line 204: Curious why a test class is used?

metricflow/converters/msi_to_osi.py
```diff
@@ -0,0 +1,424 @@
...
+
+
+class MSIToOSIConverter:
+    """Converts an MSI SemanticManifest into an OSI Document."""
+
+    def __init__(self, dialect: OSIDialect = OSIDialect.ANSI_SQL) -> None:  # noqa: D107
+        self._dialect = dialect
+
+    def convert(  # noqa: D102
+        self, manifest: PydanticSemanticManifest, model_name: str = "semantic_model"
```
plypaul: Nit: since models are used in MSI and OSI, maybe `osi_model_name`?

metricflow/converters/msi_to_osi.py
```diff
@@ -0,0 +1,424 @@
...
+                )
+            )
+        col = metric.type_params.expr if metric.type_params.expr is not None else metric.name
+        col = self._qualify_col(col, agg_params_obj.semantic_model)
+        return self._build_agg_expression(agg_params_obj.agg, col, agg_params_obj.agg_params, filter_sql)
+
+    # ConvertCountToSumRule rewrites COUNT measure exprs to this form before FlattenSimpleMetrics
+    # copies them to metric.type_params.expr, so the inner column arrives pre-wrapped and needs
+    # its own qualification pass.
+    _COUNT_CONVERSION_RE = re.compile(r"^CASE WHEN ([A-Za-z_][A-Za-z0-9_]*) IS NOT NULL THEN 1 ELSE 0 END$")
```
plypaul: Since this depends on `ConvertCountMetricToSumRule`, how about moving this to a constant in that class?

metricflow/converters/filter_utils.py
```diff
@@ -0,0 +1,141 @@
...
+    """Join non-None SQL filter strings with AND, wrapping each in parens when multiple."""
+    active = [p for p in parts if p]
+    if not active:
+        return None
+    if len(active) == 1:
+        return active[0]
+    return " AND ".join(f"({p})" for p in active)
```
plypaul, line 141: In case there are many conditions, how about using `\nAND `to join?

metricflow/converters/msi_to_osi.py
```diff
@@ -0,0 +1,424 @@
...
+
+    @staticmethod
+    def _qualify_col(col: str, semantic_model: str) -> str:
+        """Qualify col with semantic_model if it is an unqualified identifier or a COUNT-converted expr."""
+        if re.match(r"^[A-Za-z_][A-Za-z0-9_]*$", col):
```
plypaul, line 244: Can `col` be escaped? Like if they had a column that uses a special character so they specify it with
```
`special.col`
```

plypaul (approved):
Looks good, just some minor comments. What do you think about checking in a OSI manifest JSON file that is relatively comprehensive and then adding a test that snapshots the conversion result? Maybe the other direction too. Kind of like an integration test. Could be done in a follow up since this is getting large.

metricflow/converters/msi_to_osi.py
QMalcolm: No I did not 🤦🏻

metricflow/converters/msi_to_osi.py
QMalcolm: Can do!

metricflow/converters/filter_utils.py
QMalcolm: I think I'm going to leave this one as is for now. The merged filter SQL is never emitted as a standalone WHERE clause always ends up embedded inside a CASE WHEN expression (via _build_agg_expression). With `\nAND`, you'd get:
```sql
  SUM(CASE WHEN (status = 'paid')
  AND (region = 'intl') THEN orders.amount END)
```                                                                                                                       
I feel like the line break landing in the middle of the CASE WHEN ends up reading more like a broken WHERE clause than clean inline SQL. Of note this is for filters, OSI is currently working on some additions to the spec to include filters. Once that lands, the need for this and-ing will go away 🙂

metricflow/converters/msi_to_osi.py
QMalcolm: Hmmmm I might leave this as is for now. You're absolutely right that quoted identifiers like `"my column"` or `` `my column` `` would fall through unqualified. I'm not sure that we have a great way to handle this at this time. Different data warehouses could have different escaping specifications. A better option at this time is perhaps for other converters of OSI to detect when escaping is needed for their use case. I've added a comment near the regex that documents the limitation so it's visible as a known gap rather than an accidental oversight

tests_metricflow_semantic_interfaces/osi/test_osi_to_msi.py
QMalcolm: This is on me. It's a convention we have in core that most of our tests go into test classes for organizational purposes. I'll get it flattened as it's not adding anything here 🙂

metricflow/converters/filter_utils.py
QMalcolm: `None` can come from two places currently. First `parent_filter` starts as `None` when a metric has no inherited filter, and this gets passed down through each recursive resolver call. Second `_collect_filter_sql` returns `None` when a metric carries no filter. Accepting `None` as a valid type means every call to the function can pass these values directly instead of guarding with `if x is not None` first. I've updated the docstring to make this explicit 🙂

PR #2021, opened 2026-04-15
Fix error when querying the same dimension with a different alias

plypaul:


plypaul:
Closing as this needs to be reworked.

PR #2038, opened 2026-04-28
Remove `graphviz` dependency from `dbt-metricflow`

plypaul:
This PR removes the `graphviz` dependency from `dbt-metricflow` as it should be handled as an optional one.


QMalcolm:
I'm not sure I fully understand how requirements in the different subpackages yet, but it appears that we only specify graphviz as a dep here and in dev requirements. Is it safe to say that this `import graphviz` is only ever reached during dev? https://github.com/dbt-labs/metricflow/blob/9aec0ab27037b1fd7dadf182bd4a5d5537e681fd/metricflow_semantics/toolkit/mf_graph/formatting/mf_to_dot.py#L12

plypaul:
Yes, that should be only be used in tests / dev-env.

QMalcolm (approved):
Looks good to me!

PR #2043, opened 2026-05-06
decouple CI and make metricflow package 3.14 compat

ash2shukla:
Decouples CI for metricflow and dbt-metricflow package. Required for independently bumping metricflow to support 3.14


ash2shukla:
This is first in a 2 part PR, we first need to bump metricflow because dbt-core depends on metricflow semantic interfaces, then we will add 3.14 support in dbt-core and then we can add support for dbt-metricfow.

QMalcolm (approved):
Looks good to me!

QMalcolm:
@plypaul This might affect some of the release workflow work that's being done

PR #2050, opened 2026-05-11
Allow multiple `--where` arguments in the CLI

plypaul:
This PR updates the CLI to allow multiple `--where` arguments. This provides a UX improvement and can be used to push different predicates to different levels.


QMalcolm (approved):
Looks good to me! 🚀

PR #2051, opened 2026-05-11
Update DuckDB dev dependency to `>= 1.5`

plypaul:
This PR updates DuckDB dependency in the development environment to `>= 1.5` as the Python 3.14 wheels for some older versions are not available. Without the wheel, CI checks for Python 3.14 are slow as the package needs to be built locally.


QMalcolm (approved):
Looks good to me, good job catching this!

PR #2060, opened 2026-05-24
Add StarRocks SQL engine support

ndrluis:
Some Notes:

* This PR normalizes a few SQL check-query aliases from uppercase to lowercase. The aliases are declared lowercase in the query, and StarRocks is stricter about resolving alias case than some existing engines, so the rendered check SQL needs to reference the alias exactly as declared.
* A few check queries now use `cast_expr_to_ts()` / `cast_to_ts()` instead of hardcoding `CAST(... AS TIMESTAMP)`. This keeps expected SQL engine-aware. StarRocks uses `DATETIME` as its timestamp-like type in the renderer, so hardcoded `TIMESTAMP` casts can diverge from generated SQL or fail on StarRocks.
* Tests that require millisecond `DATE_TRUNC` now use explicit capability checks because StarRocks does not support sub-second truncation.
* Percentile support is limited to approximate continuous percentile rendering for StarRocks.


plypaul:
Thanks for the contribution! While we can help get this in, we'll have to classify StarRocks support as a preliminary / experimental feature until we get more deployment experience.

PR #2072, opened 2026-06-16
feat(ipc): Create a binary for sidecars using Nuitka

QMalcolm:
## Summary

This PR adds `sidecar/mf_entry.py`, the production inter-process communication (IPC) entry point that other applications can spawn as a subprocess and communicate with over stdin/stdout to compile metric queries to SQL. It implements the mf-ipc v1 protocol, wrapping `MetricFlowEngine.explain()` across all seven supported SQL engines.

The server is designed to run as a long-lived process. Engines are cached by `(manifest_path, mtime, sql_engine)` so repeated `explain` calls against the same manifest are cheap, with automatic invalidation when the manifest file changes. `sys.stdout` is replaced with `sys.stderr` at startup (before any library code runs) so incidental `print()` calls from MetricFlow internals can't corrupt the JSON framing when reading from pipes.

Nine subprocess integration tests cover the full protocol surface: ping, explain (happy path and error cases), unknown method, protocol version mismatch, malformed JSON, shutdown, and EOF. The test fixture spawns a real `mf_entry.py` process over stdin/stdout, no mocking of the IPC layer or MetricFlowEngine is required. The `sidecar/validate_mf_entry.py` script is a local validation helper that sends the same `explain` request to both the Python interpreter and the compiled Nuitka binary and diffs the SQL output; it's the basis for the CI validation gate that will land in a follow-on PR.

## How to try it yourself

**Run the integration tests (no compilation needed):**
```bash
hatch run dev-env:pytest sidecar/tests/test_mf_entry.py -v
```

**Interactive smoke test:**
```bash
# Ping
echo '{"id":"1","method":"ping","v":1}' | hatch run dev-env:python sidecar/mf_entry.py

# Explain (bookings metric)
MANIFEST=metricflow_semantics/test_helpers/semantic_manifest_yamls/sg_00_minimal_manifest
printf '{"id":"2","method":"explain","v":1,"params":{"manifest_path":"%s","metric_names":["bookings"],"group_by_names":["metric_time"],"sql_engine":"DUCKDB"}}\n' "$MANIFEST" \
  | hatch run dev-env:python sidecar/mf_entry.py
```

**Compile and validate the Nuitka binary** (takes ~10 min on first run):
```bash
hatch run nuitka-build:build-and-validate
```

## Protocol (mf-ipc v1)

```
ready:    {"status":"ready","metricflow_version":"X","python_version":"X","protocol_version":1}
explain:  {"id":"...","method":"explain","v":1,"params":{"manifest_path":"...","metric_names":[...],"group_by_names":[...],"sql_engine":"DUCKDB"}} → {"id":"...","ok":true,"sql":"..."}
ping:     {"id":"...","method":"ping","v":1} → {"id":"...","ok":true}
shutdown: {"id":"...","method":"shutdown","v":1} → {"id":"...","ok":true}
error:    {"id":"...","ok":false,"error":{"type":"ExceptionClass","message":"..."}}
```

`manifest_path` may be a YAML directory (dev/testing) or a `manifest.json` file (production). `sql_engine` must match a `SqlEngine` enum name: `DUCKDB`, `BIGQUERY`, `DATABRICKS`, `POSTGRES`, `REDSHIFT`, `SNOWFLAKE`, or `TRINO`.

### DIFF
--- a/sidecar/mf_entry.py
+++ b/sidecar/mf_entry.py
@@ -30,7 +30,7 @@ import sys
 import traceback as _traceback
 import types
 from pathlib import Path
-from typing import IO
+from typing import IO, Literal
 
 from metricflow_semantics.model.semantic_manifest_lookup import SemanticManifestLookup
 from metricflow_semantics.sql.sql_bind_parameters import SqlBindParameterSet
@@ -212,7 +212,7 @@ def _dispatch(req: object) -> ExplainResponse | OkResponse | ErrorResponse:
     )
 
 
-def main(argv: list[str]) -> int:  # noqa: D103
+def main(argv: list[str]) -> Literal[0, 1]:  # noqa: D103
     global _ipc, _debug
 
     parser = argparse.ArgumentParser(description="MetricFlow IPC entry point (mf-ipc v1)")
--- a/sidecar/mf_entry.py
+++ b/sidecar/mf_entry.py
@@ -22,7 +22,6 @@ from __future__ import annotations
 
 import argparse
 import importlib.metadata
-import json
 import logging
 import os
 import signal
@@ -188,15 +187,6 @@ def _handle_explain(req_id: RequestId, raw_params: dict) -> ExplainResponse | Er
         return _err(req_id, e)
 
 
-def _parse_envelope(req: object) -> RequestEnvelope | ErrorResponse:
-    """Validate a decoded JSON value into a RequestEnvelope, or a structured error."""
-    try:
-        return RequestEnvelope.model_validate(req)
-    except ValidationError as e:
-        fallback_id = req.get("id") if isinstance(req, dict) else None
-        return _err(fallback_id, e)
-
-
 def _dispatch(envelope: RequestEnvelope) -> ExplainResponse | OkResponse | ErrorResponse:
     """Route a validated envelope to its method handler.
 
@@ -280,16 +270,9 @@ def main(argv: list[str]) -> Literal[0, 1]:  # noqa: D103
             if not line:
                 continue
             try:
-                # Parsed as raw JSON, then separately validated into a RequestEnvelope,
-                # so malformed JSON (JSONDecodeError) and schema-invalid JSON
-                # (ValidationError) remain distinguishable error types for the caller.
-                req = json.loads(line)
-            except json.JSONDecodeError as e:
-                _write(ErrorResponse(id=None, error=ErrorDetail(type="JSONDecodeError", message=str(e))))
-                continue
-            envelope = _parse_envelope(req)
-            if isinstance(envelope, ErrorResponse):
-                _write(envelope)
+                envelope = RequestEnvelope.model_validate_json(line)
+            except ValidationError as e:
+                _write(_err(None, e))
                 continue
             if envelope.method == Method.SHUTDOWN:
                 _write(OkResponse(id=envelope.id))
--- a/sidecar/mf_entry.py
+++ b/sidecar/mf_entry.py
@@ -38,6 +38,18 @@ from metricflow_semantics.test_helpers.manifest_helpers import (
     mf_load_manifest_from_json_file,
     mf_load_manifest_from_yaml_directory,
 )
+from mf_ipc_protocol import (
+    ErrorDetail,
+    ErrorResponse,
+    ExplainParams,
+    ExplainResponse,
+    OkResponse,
+    ReadyMessage,
+    RequestEnvelope,
+    RequestId,
+    StartupErrorMessage,
+)
+from pydantic import BaseModel, ValidationError
 
 from metricflow.data_table.mf_table import MetricFlowDataTable
 from metricflow.engine.metricflow_engine import MetricFlowEngine, MetricFlowQueryRequest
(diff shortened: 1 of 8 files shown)


plypaul:
Makes sense, but IPC with stronger typing would be preferred. Given that this would be called from a different programming language, how about using protobuff / gRPC?

QMalcolm:
> Makes sense, but IPC with stronger typing would be preferred. Given that this would be called from a different programming language, how about using protobuff / gRPC?

@plypaul Hmmmm I'm unsure of my opinion on this at the present 🤔 

I think at some point this _will_ be necessary, but I'm unsure if the amount of surface area right now is worth it. In my head, any given version of Core V2 will pin to a specific MF Nuitka binary build (not allow for a range). Assuming the testing in Core V2 covers it's use cases of MF, then any time the pinned MF binary changes, breakages should be apparent immediately. However, as the entry point for the binary expands what interfaces are available, the guarantee becomes weaker.

So in my head
Necessary eventually = more than likely
Necessary now = perhaps not?

When we start defining a strict contract (whether that is now or later), I'd lean the protobuf direction. I believe proto buf is more standard at dbt (at least in my exposure). Additionally, grpcio might complicate the nuitka binary in that it does more dynamic linking which might break the build. I'm doing quick spike on that now though to validate the risk.

plypaul:
In that case, how about defining Pydantic objects on the messages to get an initial typed schema?

QMalcolm:
> In that case, how about defining Pydantic objects on the messages to get an initial typed schema?

I believe the relevant structures have now been added 🙂

sidecar/tests/test_mf_entry.py
```diff
@@ -0,0 +1,161 @@
...
+import subprocess
+import sys
+from collections.abc import Iterator
+from pathlib import Path
+
+import pytest
+
+_REPO_ROOT = Path(__file__).parent.parent.parent
+_MF_ENTRY = _REPO_ROOT / "sidecar" / "mf_entry.py"
+_MANIFEST_DIR = (
```
plypaul: If available, `SG_00_MINIMAL_MANIFEST.directory` would be the same

sidecar/tests/test_mf_entry.py
```diff
@@ -0,0 +1,161 @@
...
+import json
+import subprocess
+import sys
+from collections.abc import Iterator
+from pathlib import Path
+
+import pytest
+
+_REPO_ROOT = Path(__file__).parent.parent.parent
+_MF_ENTRY = _REPO_ROOT / "sidecar" / "mf_entry.py"
```
plypaul: If imports work, `mf_entry.py` could expose `PATH(__NAME__)` and then be used here.

sidecar/mf_ipc_protocol.py
```diff
@@ -0,0 +1,102 @@
...
+from __future__ import annotations
+
+from typing import Literal
+
+from pydantic import BaseModel
+
+RequestId = str | int | None
+
+
+class ReadyMessage(BaseModel):
```
plypaul: Can these classes be immutable with appropriate field types like tuple or sequence?

sidecar/mf_ipc_protocol.py
```diff
@@ -0,0 +1,102 @@
...
+work (evaluating gRPC/protobuf, or a JSON Schema shared with Fusion) is
+tracked separately in DI-4709.
+"""
+from __future__ import annotations
+
+from typing import Literal
+
+from pydantic import BaseModel
+
+RequestId = str | int | None
```
plypaul, line 21: In what cases is `None` used?

sidecar/mf_ipc_protocol.py
```diff
@@ -0,0 +1,102 @@
...
+
+    `method` and `v` are intentionally left unconstrained (not Literal) here:
+    an unknown method or a protocol version other than 1 is a valid, expected
+    input that must reach _dispatch's own UnknownMethod / ProtocolVersionError
+    handling, not fail at the envelope-parsing stage.
+    """
+
+    id: RequestId = None
+    method: str | None = None
+    v: int = 1
```
plypaul: Maybe `api_version` or similar for improved readability?

sidecar/tests/test_mf_entry.py
```diff
@@ -0,0 +1,161 @@
+"""Subprocess integration tests for sidecar/mf_entry.py (mf-ipc v1).
```
plypaul, line 1: Comment that applies to many of the tests: Now that the Pydantic objects are available, we can use them instead of manually constructing the dicts.

sidecar/mf_entry.py
```diff
@@ -0,0 +1,286 @@
...
+    except ValidationError as e:
+        fallback_id = req.get("id") if isinstance(req, dict) else None
+        return _err(fallback_id, e)
+
+    if envelope.v != 1:
+        return ErrorResponse(
+            id=envelope.id,
+            error=ErrorDetail(type="ProtocolVersionError", message=f"Expected v=1, got v={envelope.v!r}"),
+        )
+    if envelope.method == "ping":
```
plypaul: How about a enum for these method names? It would also make it less error-prone to generate the error message later in the function.

sidecar/mf_entry.py
```diff
@@ -0,0 +1,286 @@
...
+        )
+    )
+
+    try:
+        for line in sys.stdin:
+            line = line.strip()
+            if not line:
+                continue
+            try:
+                req = json.loads(line)
```
plypaul: Any particular reason for loading as JSON vs. loading using `RequestEnvelope`?

sidecar/mf_entry.py
```diff
@@ -0,0 +1,286 @@
...
+            except json.JSONDecodeError as e:
+                _write(ErrorResponse(id=None, error=ErrorDetail(type="JSONDecodeError", message=str(e))))
+                continue
+            resp = _dispatch(req)
+            _write(resp)
+            if isinstance(req, dict) and req.get("method") == "shutdown":
```
plypaul: This case seems a little odd due to split handling. Maybe `_dispatch` can handle all messages except shutdown, and shutdown is handled in this loop instead?

sidecar/mf_ipc_protocol.py
QMalcolm: Originally, it could be None if the requester didn't provide one, and that wasn't considered malformed 😅  Now it is considered malformed, in which case the `RequestId` on the error _response_ will be `None` if one wasn't found in the request.

sidecar/mf_entry.py
QMalcolm: It was a two stage process allowing us to delineate between "this json was invalid for our spec" and "this was not json at all". However, delineating those two doesn't actually matter that much given this is built to be used with trusted applications. I've changed it to load directly now 🫡

sidecar/mf_entry.py
QMalcolm: Shutdown is now handled in the main loop, and `_dispatch` has been split into two parts: `_parse_envelope` (validation) and `_dispatch` (routing) 🙂

plypaul (approved):

PR #2079, opened 2026-07-08
DI-4631: CI Nuitka binary build and artifact publication

QMalcolm:
## Summary

Implements DI-4631 — a CI pipeline that compiles `sidecar/mf_entry.py` with Nuitka
across macOS arm64, macOS x86_64, Linux x86_64, Linux arm64, and Windows x64, and
publishes the resulting artifacts to GitHub Releases for Fusion to consume via
`rust-embed`.

## What's implemented

- Pinned Nuitka to `4.1.2` (`pyproject.toml`) for reproducible builds
- Fixed `setup-python-env` to work on non-Linux runners (guarded `lsb_release`,
  OS-scoped cache key)
- 4-leg active build matrix (`cd-build-sidecar-binaries.yaml`): macos-14/aarch64-apple-darwin,
  ubuntu-22.04/x86_64-unknown-linux-gnu, ubuntu-24.04-arm/aarch64-unknown-linux-gnu,
  windows-latest/x86_64-pc-windows-msvc — each leg compiles and validates (Python
  vs. binary SQL diff) before uploading `sidecar/mf_entry.dist/` as an artifact.
  (macos-13/x86_64-apple-darwin is commented out; see "Out of scope" below.) The
  triple list matches what `fs` and `dbt-core` v2 already build for.
- Publish job (gated to `refs/tags/v*`): downloads all platform artifacts, packages
  them (`.tar.gz` / `.zip` for Windows), generates `SHA256SUMS.txt`, and attaches
  everything to the tag's GitHub Release
- `.github/actionlint.yaml`, allowlisting `ubuntu-24.04-arm` — GitHub's native
  hosted Linux ARM64 runner label is newer than what the pinned actionlint version
  recognizes
- Documented the artifact naming/versioning contract in `sidecar/README.md`, since
  Fusion's build tooling will hard-code against it

## Validated

- macOS arm64, Linux x86_64, and Windows x64 legs compiled and validated successfully
  on real GitHub runners (Windows ~42 min)

## Not yet validated

- The publish job itself has never fired — it's gated to a real release tag and
  can't be exercised from a PR or `workflow_dispatch` (this workflow doesn't exist
  on `main` yet, so `workflow_dispatch` isn't invocable). It'll run for the first
  time on the next real MetricFlow release tag.
- The full matrix hasn't been re-run since this branch was rebased onto `main`
  (post PR #2072's squash-merge) and since the Linux arm64 leg was added. The new
  `ubuntu-24.04-arm` leg specifically has never executed yet — it should behave
  identically to the proven `ubuntu-22.04` leg (same steps, native GitHub-hosted
  runner, no cross-compilation), but that's an expectation, not an observation.

## Out of scope (tracked separately under epic DI-4630)

- `macos-13` (Intel) has never successfully scheduled a runner in validation runs
  (queued 2+ hours, zero steps executed, both attempts) — GitHub-hosted runner
  scarcity, not a config issue. Since `publish` needs the whole `build` matrix, a
  hung leg here would block release assets for every platform, so the leg is now
  commented out of the matrix rather than left to hang on a real release tag.
  `fs`/`dbt-core` avoid this by routing macOS through Depot instead of GitHub's
  native runner pool; `metricflow` has no Depot access configured yet. Filed as
  [DI-4819](https://dbtlabs.atlassian.net/browse/DI-4819) — deliberately kept out of
  this PR since it depends on an external access grant with its own timeline;
  re-enabling this leg is the follow-up once access lands.
- Full snapshot-suite validation across all dialects/metric types — DI-4632
- musl/Alpine Linux build — DI-4633
- Code signing/notarization (macOS DI-4634, Windows EV DI-4635)

Refs DI-4631

[DI-4819]: https://dbtlabs.atlassian.net/browse/DI-4819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/cd-build-sidecar-binaries.yaml
+++ b/.github/workflows/cd-build-sidecar-binaries.yaml
@@ -23,6 +23,8 @@ jobs:
             target-triple: x86_64-apple-darwin
           - runner: ubuntu-22.04
             target-triple: x86_64-unknown-linux-gnu
+          - runner: ubuntu-24.04-arm
+            target-triple: aarch64-unknown-linux-gnu
           - runner: windows-latest
             target-triple: x86_64-pc-windows-msvc
     steps:
--- a/.github/workflows/cd-build-sidecar-binaries.yaml
+++ b/.github/workflows/cd-build-sidecar-binaries.yaml
@@ -1,7 +1,5 @@
 # Compiles sidecar/mf_entry.py into standalone Nuitka binaries for every platform
-# Fusion needs to embed via rust-embed (DI-4631). Windows x64 is intentionally not
-# included yet — see ci-sidecar-windows-nuitka-spike.yaml, which is de-risking that
-# leg in isolation before it's added here.
+# Fusion needs to embed via rust-embed (DI-4631).
 name: Build Sidecar Binaries
 
 on:
@@ -25,6 +23,8 @@ jobs:
             target-triple: x86_64-apple-darwin
           - runner: ubuntu-22.04
             target-triple: x86_64-unknown-linux-gnu
+          - runner: windows-latest
+            target-triple: x86_64-pc-windows-msvc
     steps:
       - name: Check-out the repo
         uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # actions/checkout@v4
--- a/.github/workflows/cd-build-sidecar-binaries.yaml
+++ b/.github/workflows/cd-build-sidecar-binaries.yaml
@@ -52,3 +52,42 @@ jobs:
           name: mf_entry-${{ matrix.target-triple }}
           path: sidecar/mf_entry.dist/
           if-no-files-found: error
+
+  publish:
+    name: Publish release assets
+    needs: build
+    # Only publish on an actual release tag. workflow_dispatch and pull_request runs exist
+    # to validate the build matrix on real runners and have no tag to attach assets to.
+    # `needs: build` already means this only runs if every matrix leg succeeded.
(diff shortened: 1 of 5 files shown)


.github/workflows/cd-build-sidecar-binaries.yaml
```diff
@@ -0,0 +1,101 @@
...
+      - name: Compile mf_entry with Nuitka
+        shell: bash
+        run: hatch run nuitka-build:compile
+
+      - name: Validate compiled binary against Python interpreter
+        shell: bash
+        run: hatch run nuitka-build:validate
+
+      - name: Upload mf_entry.dist for ${{ matrix.target-triple }}
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # actions/upload-artifact@v4
```
plypaul, line 58: Is there an issue with v4? Also for other cases in the workflow.

QMalcolm: I don't believe there is an issue with v4. The pin actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 is a specific commit of actions/upload-artifact v4. My understanding is that this is an org wide standard, we pin to specific action commits. We do the exact same pin in core https://github.com/dbt-labs/dbt-core/blob/6d5a0e95b9b0f72bc865bd59aa1c45c3cd4c8376/.github/workflows/release-v2.yml#L348 🙂

plypaul (approved):

PR #2106, opened 2026-07-27
DI-4871: decouple sidecar release versioning from MetricFlow release tags

QMalcolm:
## Summary

Sidecar binary releases were 1:1 tied to MetricFlow's own `v<version>` tags, so there was no way to publish a new sidecar binary (e.g. a new `mf_entry.py` entry point) without a MetricFlow version bump — and re-publishing under an existing tag silently clobbered assets (`softprops/action-gh-release` defaults to `overwrite_files: true`).

This introduces a `mf-entry-bin/v<mf_version>+<YYMMDDHHmm>.<sha8>` tag namespace, decoupled from MetricFlow's release cadence. The design went through a round of review (see below) before landing here:

- `cd-build-sidecar-binaries.yaml` now triggers on and publishes for `mf-entry-bin/v*` tags exclusively (the bare `v<version>` tag's only remaining job is triggering the PyPI publish workflow — triggering on both would double-build and double-publish for every release).
- `scripts/release_tool/release_step_3.py` automatically pushes a matching sidecar tag alongside every real MetricFlow release, using the version/commit it already knows — no separate manual step, no git lookups needed.
- A new `workflow_dispatch` GitHub Action (`Cut Ad Hoc Sidecar Release`, restricted to `main`) lets a sidecar-only change publish a release between MetricFlow releases — self-service. It resolves `<mf_version>` directly from `metricflow/__about__.py` (often a `.devN` pre-release version between releases — see below for why that's the right call, not a shortcut).
- Both of the above build the *exact same* tag format via one shared function (`scripts/sidecar_release/tag.py`) — there's no more special-cased "baseline" tag vs. "ad hoc" tag, since the timestamp+sha disambiguator needs no stateful counter to compute.
- `overwrite_files: false` so an accidental re-push of an existing tag fails loudly instead of silently overwriting a published release.
- A centralized `build-info.json` records the embedded MetricFlow version, source commit, and the Nuitka/Python versions actually used.
- `sidecar/README.md` documents all of the above.

**Changes from review (Paul):**
1. *Tag naming* — the namespace was originally `sidecar/`. `mf-ipc` was considered and rejected (that name is already taken by the wire protocol, versioned independently as `mf-ipc v1` — conflating the two would tie unrelated axes together). Landed on `mf-entry-bin/`, matching the existing `mf_entry-*` archive-naming convention.
2. *Scriptify the release process* — the archive-packaging, `build-info.json` generation, and ad hoc-release tag resolution were originally inline bash/python embedded directly in the workflow YAML. All three are now real, stdlib-only scripts under `scripts/sidecar_release/`, runnable locally with a bare `python3` (no hatch env) — the workflows now mostly orchestrate these scripts rather than containing the logic themselves.
3. *Version format* — switched from a stateful `+<counter>` (reset per MetricFlow version, computed by scanning existing tags) to `+<YYMMDDHHmm>.<sha8>` (UTC timestamp + short commit SHA). This needs zero git-tag lookups and can't race on concurrent triggers. Not PEP 440-constrained, since unlike MetricFlow's own tags this one never goes through PyPI/pip/twine.

**Further refinement (post-review):** the ad hoc-release script originally resolved `<mf_version>` via `git describe --tags` against the nearest real release tag, specifically to avoid embedding `metricflow/__about__.py`'s `.devN` pre-release version. On reconsideration, that was backwards — between releases, `__about__.py`'s version is exactly as static as the last release tag (nothing bumps either except the release process itself), so the real difference is what each one *claims*. Labeling an ad hoc build with the last released tag asserts "this embeds exactly what that release shipped," which this design already accepts can be false (ad hoc releases don't guard against `metricflow`/`metricflow_semantics`/`metricflow_semantic_interfaces` having changed since). `.devN` makes no such claim — it honestly signals "unreleased, in-development state" either way. Switched to reading `__about__.py` directly, which also removes a failure mode (no reachable release tag) and let the workflow's checkout drop `fetch-depth: 0`/`fetch-tags: true` entirely.

Full design rationale (12 resolved decision points from the original `/grill-me` interview, plus the changes above) is in [DI-4871](https://dbtlabs.atlassian.net/browse/DI-4871).

## Test plan

- [x] `hatch run dev-env:pytest tests_metricflow/sidecar_release/ tests_metricflow/release_tools/` — 32 tests pass, covering: tag building/parsing (`tag.py`), archive packaging incl. tar vs. zip and checksum generation (`package_archives.py`), `build-info.json` assembly (`generate_build_info.py`), and ad hoc-release version/commit resolution against real throwaway git repos incl. missing-`__about__.py` and missing-version-assignment failure cases (`cut_adhoc_release.py`) — plus the release-tool snapshot suite, updated for the new tag format via an injectable clock (`FakeNow`, mirroring the existing `FakeSleep` pattern) so the snapshot doesn't embed real wall-clock time
- [x] `mypy`, `ruff`, `black` clean on all changed Python files
- [x] `actionlint` clean on all workflow files
- [x] Every command in `sidecar/README.md`'s local-usage examples was actually run, not just described — including `package_archives.py`, which needs a build staged into the `mf_entry-<triple>/` layout the script expects before it'll find anything to package (an earlier draft of that example pointed straight at `sidecar/`, which would have silently matched nothing)
- [x] Live-ran `cut_adhoc_release.py --dry-run` against this actual repo/branch to confirm it resolves a real tag end-to-end
- [ ] Not yet exercised in real CI — worth a deliberate watch on the first real use of either the automated release_step_3.py tag (next MetricFlow release) or the ad hoc-release Action

[DI-4871]: https://dbtlabs.atlassian.net/browse/DI-4871

### DIFF
--- a/.changes/unreleased/Under the Hood-20260730-164420.yaml
+++ b/.changes/unreleased/Under the Hood-20260730-164420.yaml
@@ -1,0 +1,8 @@
+kind: Under the Hood
+body: Add CI pipeline to compile MetricFlow into standalone Nuitka binaries for macOS,
+  Linux, and Windows, and publish them as versioned GitHub Release artifacts for embedding
+  in dbt-core v2 (Fusion)
+time: 2026-07-30T16:44:20.063502-05:00
+custom:
+  Author: QMalcolm
+  Issue: "2106"
--- a/.github/workflows/cd-build-sidecar-binaries.yaml
+++ b/.github/workflows/cd-build-sidecar-binaries.yaml
@@ -6,15 +6,15 @@ on:
   push:
     tags:
       # DI-4871: every sidecar binary release -- including the one cut alongside a new
-      # MetricFlow version -- goes through the mf-entry-bin/v<mf_version>+<timestamp>.<sha8>
+      # MetricFlow version -- goes through the mf-stdio-sidecar/v<mf_version>+<timestamp>.<sha8>
       # namespace. release_step_3.py automatically pushes one of these tags alongside
       # MetricFlow's own `v<version>` tag on every release, so this is the sole trigger for
       # this workflow; the bare `v*` tag's only remaining job is triggering
-      # cd-push-metricflow-to-pypi.yaml. (Triggering on both `v*` and `mf-entry-bin/v*`
+      # cd-push-metricflow-to-pypi.yaml. (Triggering on both `v*` and `mf-stdio-sidecar/v*`
       # would double-build and double-publish for every release, once per tag.) Ad hoc
       # sidecar-only releases are pushed the same way via the ad hoc-release Action's
       # re-dispatch.
-      - "mf-entry-bin/v[0-9]+.[0-9]+.[0-9]+*"
+      - "mf-stdio-sidecar/v[0-9]+.[0-9]+.[0-9]+*"
   workflow_dispatch:
 
 jobs:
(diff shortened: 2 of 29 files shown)


plypaul (approved):
LGTM though now the sidecar name seems appealing when qualified. `mf-ipc-sidecar`?

plypaul:
`mf-stdio-sidecar` too.

PR #2110, opened 2026-08-05
Require metrics for metric-scoped list commands

### LINKED ISSUE #2109: [Bug] Metric-scoped list commands do not mark --metrics as required

### Current Behavior

Running `mf list dimensions` or `mf list entities` without `--metrics` fails with an internal sequence-validation message, even though the CLI help presents the option as optional:

```text
Error: Invalid value for '--metrics': Input is empty, when the minimum expected number of input is 1
```

Both commands configure `SequenceParamType(min_length=1)` with `default=""`, but do not mark the option as required. Their command descriptions also say that they list all unique dimensions or entities, while the implementation returns the intersection for specified metrics.

This is a CLI contract and error-message issue; metric query behavior is not affected.

### Expected Behavior

Because both commands require at least one metric, `--metrics` should be displayed as required. Omitting it should produce Click’s standard missing-option error:

```text
Error: Missing option '--metrics'.
```

The command descriptions should state that the returned dimensions or entities are common to the specified metrics.

### Steps To Reproduce

1. Install `dbt-metricflow==0.13.0`.
2. Run `mf list entities`.
3. Observe the internal empty-sequence validation error.
4. Run `mf list dimensions`.
5. Observe the same error.
6. Run either command with `--help` and observe that `--metrics` is not marked required.

### Relevant log output

```shell
Usage: mf list entities [OPTIONS]
Try 'mf list entities --help' for help.

Error: Invalid value for '--metrics': Input is empty, when the minimum expected number of input is 1
```

### Environment

- OS: Linux
- Python: 3.12.13
- dbt: 1.11.12
- dbt-metricflow: 0.13.0
- metricflow: 0.211.0

### Additional Context

The commands were introduced to return objects common to the supplied metric inputs. A focused fix is to replace the empty defaults with `required=True`, preserve the existing option help, clarify the command-level descriptions, and add regression coverage for both commands.

fanruan-tuco:
## Summary

- mark `--metrics` as required for `mf list dimensions` and `mf list entities`
- replace the misleading all-object command descriptions with metric-scoped wording
- add shared regression coverage for help output and missing-option errors
- add a changelog entry for the CLI fix

Both commands already require at least one metric through `SequenceParamType(min_length=1)`. Replacing the empty-string defaults with `required=True` makes the Click contract match the existing behavior and produces the standard missing-option error.

## Testing

- `hatch run dev-env:pytest tests_dbt_metricflow/cli/test_cli.py` (16 passed)
- dbt-metricflow pre-commit: ruff, black, mypy, and file checks passed
- `changie batch --dry-run 999.999.999` passed

Fixes #2109

### DIFF
--- a/.changes/unreleased/Fixes-20260805-200603.yaml
+++ b/.changes/unreleased/Fixes-20260805-200603.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Require --metrics for metric-scoped dimension and entity list commands
+time: 2026-08-05T20:06:03.842688283+08:00
+custom:
+    Author: fanruan-tuco
+    Issue: "2109"
--- a/dbt-metricflow/dbt_metricflow/cli/main.py
+++ b/dbt-metricflow/dbt_metricflow/cli/main.py
@@ -346,14 +346,14 @@ def metrics(cfg: CLIConfiguration, show_all_dimensions: bool = False, search: Op
 @click.option(
     "--metrics",
     type=click_custom.SequenceParamType(min_length=1),
-    default="",
+    required=True,
     help="List dimensions by given metrics (intersection). Ex. --metrics bookings,messages",
 )
 @pass_config
 @exception_handler
 @log_call(module_name=__name__, telemetry_reporter=_telemetry_reporter)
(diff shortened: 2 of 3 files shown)


fanruan-tuco:
CLA submitted.

PR #2114, opened 2026-08-19
Keep non-deterministic expressions materialized in the sub-query reducer

### LINKED ISSUE #2111: [Bug] Conversion metric is overcounted when querying its conversion input metric together
## Is this a new bug in metricflow?

## Current Behavior

The following self-contained DuckDB project reproduces an incorrect result when a Conversion Metric and its conversion input Simple Metric are queried together.

Create these files:

\`dbt_project.yml\`

\`\`\`yaml
name: mf_conversion_repro
version: '1.0'
config-version: 2
profile: mf_conversion_repro
model-paths: [models]
\`\`\`

\`profiles.yml\`

\`\`\`yaml
mf_conversion_repro:
  target: dev
  outputs:
    dev:
      type: duckdb
      path: repro.duckdb
      threads: 1
\`\`\`

\`models/sessions.sql\`

\`\`\`sql
select * from (values
  ('S001', 'C001', timestamp '2026-07-01 08:00:00', 'organic'),
  ('S002', 'C001', timestamp '2026-07-02 09:00:00', 'organic')
) as t(session_id, customer_id, session_started_at, acquisition_channel)
\`\`\`

\`models/completed_events.sql\`

\`\`\`sql
select * from (values
  (1001, 'C001', timestamp '2026-07-02 10:00:00', 'organic')
) as t(conversion_order_id, customer_id, converted_at, order_channel)
\`\`\`

\`models/metricflow_time_spine.sql\`

\`\`\`sql
select date_day
from generate_series(
  date '2026-06-01', date '2026-07-31', interval 1 day
) as t(date_day)
\`\`\`

\`models/schema.yml\`

\`\`\`yaml
version: 2
models:
  - name: metricflow_time_spine
    time_spine:
      standard_granularity_column: date_day
    columns:
      - name: date_day
        granularity: day
\`\`\`

\`models/semantic.yml\`

\`\`\`yaml
version: 2

semantic_models:
  - name: sessions
    model: ref('sessions')
    defaults:
      agg_time_dimension: session_started_at
    entities:
      - name: session
        type: primary
        expr: session_id
      - name: customer
        type: foreign
        expr: customer_id
    dimensions:
      - name: session_started_at
        type: time
        type_params:
          time_granularity: day
      - name: acquisition_channel
        type: categorical
    measures:
      - name: session_events
        expr: '1'
        agg: count

  - name: completed_events
    model: ref('completed_events')
    defaults:
      agg_time_dimension: converted_at
    entities:
      - name: conversion_order
        type: primary
        expr: conversion_order_id
      - name: customer
        type: foreign
        expr: customer_id
    dimensions:
      - name: converted_at
        type: time
        type_params:
          time_granularity: day
      - name: order_channel
        type: categorical
    measures:
      - name: completed_order_events
        expr: '1'
        agg: count

metrics:
  - name: session_count
    label: Session Count
    description: Session count
    type: simple
    type_params:
      measure: session_events

  - name: completed_order_event_count
    label: Completed Order Event Count
    description: Completed order event count
    type: simple
    type_params:
      measure:
        name: completed_order_events
        fill_nulls_with: 0

  - name: session_to_order_conversion_rate
    label: Session to Order Conversion Rate
    description: Seven day conversion rate
    type: conversion
    type_params:
      conversion_type_params:
        base_measure: session_events
        conversion_measure:
          name: completed_order_events
          fill_nulls_with: 0
        entity: customer
        window: 7 days
        calculation: conversion_rate
        constant_properties:
          - base_property: acquisition_channel
            conversion_property: order_channel
\`\`\`

Run:

\`\`\`bash
pip install dbt-duckdb==1.10.1 dbt-metricflow==0.13.0
dbt build --profiles-dir .

mf query --metrics session_to_order_conversion_rate --decimals 4
# 0.5000

mf query \
  --metrics completed_order_event_count,session_to_order_conversion_rate \
  --decimals 4
# 1.0000
\`\`\`

The expected result for both queries is \`1 / 2 = 0.5000\`. The combined query overcounts the single conversion event as two.

## Expected Behavior

The combined query should return \`0.5000\`, matching the standalone Conversion Metric query.

## Relevant SQL

The standalone query partitions by the stable generated key:

\`\`\`sql
PARTITION BY ..., subq_10.mf_internal_uuid, ...
\`\`\`

The combined query renders a fresh random value inside the partition:

\`\`\`sql
PARTITION BY ..., GEN_RANDOM_UUID(), ...
\`\`\`

This prevents \`DISTINCT\` from removing conversion fan-out duplicates.

## Environment

\`\`\`text
dbt-core: 1.11.12
dbt-metricflow: 0.13.0
metricflow: 0.211.0
dbt-duckdb: 1.10.1
DuckDB: 1.10.1
Python: 3.12.3
\`\`\`

## Which database are you using?

DuckDB

## Additional Context

The issue only reproduces when the conversion input metric and Conversion Metric are queried together. Querying the Conversion Metric alone is correct. The combined query changes the conversion deduplication key from the stable \`mf_internal_uuid\` column to a new \`GEN_RANDOM_UUID()\` expression.

arthi-arumugam-git:
Resolves #2111

## What the bug looks like

Query a conversion metric on its own and the rate is right. Query it together with its conversion input metric and the rate reads high. On the fixture data in this repo:

```
metrics: visit_buy_conversion_rate_7days, buys_fill_nulls_with_0
standalone conversion rate: 0.5
combined query (before this change): 0.75
combined query (after this change): 0.5
```

Nothing raises. The combined number is plausible and just wrong, in the flattering direction.

## Why it happens

The conversion flow de-duplicates fan-out rows with a window that partitions by a generated key, selected once in a sub-query as `GEN_RANDOM_UUID() AS mf_internal_uuid`. In the combined query the plan wraps that sub-query in a shape the rewriting sub-query reducer is allowed to collapse. The reducer substitutes each referenced column with its defining expression, and it had no notion of determinism, so the substitution turned one materialized UUID into a fresh `GEN_RANDOM_UUID()` at every reference site, including inside the `PARTITION BY`. Every row then lands in its own partition, the de-duplication keeps the duplicates, and conversions are overcounted. This matches the SQL quoted in #2111, where the standalone query partitions by `subq_10.mf_internal_uuid` and the combined query partitions by `GEN_RANDOM_UUID()`.

I could reproduce the wrong number on v0.211.0 (the version in the issue) and on current main, both through `MetricFlowEngine` against DuckDB.

## The change

- `SqlExpressionNode` gains an `is_deterministic` property, true unless a node in the tree says otherwise. `SqlGenerateUuidExpression`, whose docstring already calls itself non-deterministic, reports false.
- Both reduction paths in `SqlRewritingSubQueryReducer` leave a sub-query alone when one of its select columns is non-deterministic, next to the existing `is_verbose` checks.

The column pruner and table alias simplifier do not substitute expressions, so they are untouched.

## Tests

- Two reducer-level tests (plain sub-query and joined sub-query) assert that a select materializing a generated UUID survives reduction. Both fail on current main.
- An integration test runs the #2111 query shape end to end on DuckDB and snapshots the output, using a new `buys_fill_nulls_with_0` metric added to the simple manifest fixture mirroring the issue's input metric (`fill_nulls_with: 0`, joined to the time spine, which is part of the trigger). It fails on main with 0.75 and passes here with 0.5.
- Conversion-metric SQL snapshots change as expected: the de-duplication sub-query stays materialized. I regenerated DuckDB and engine-agnostic snapshots; the credentialed engines (BigQuery, Snowflake, Redshift, Databricks, Trino, Postgres) need a `generate_snapshots.py` run I cannot do from a fork, and I am happy to follow up on review feedback there.

## Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/metricflow/blob/main/CONTRIBUTING.md)
- [x] I have signed the CLA
- [x] This PR includes tests
- [x] I have added a changelog entry via changie


arthi-arumugam-git:
CLA form submitted, in case the check needs a nudge to re-run.

metricflow_semantics/test_helpers/semantic_manifest_yamls/simple_manifest/metrics.yaml
```diff
@@ -1046,3 +1046,13 @@ metric:
       window: 7 days
       entity: user
       calculation: conversion_rate
+---
+metric:
+  name: buys_fill_nulls_with_0
```
plypaul: This metric name results in internal column names with `...0_1` in the generated SQL. That doesn't affect the correctness, but for improved aesthetics in the snapshots, how about naming this `buys_null_filled` or similar?

plypaul (commented):
Good catch! I kicked off the CI checks, and there are a couple of failures. Mind taking a look?

arthi-arumugam-git:
Thanks for kicking off CI, and good call on the name.

Renamed it to `buys_null_filled`. The reason it looked like that is a bit worse than aesthetics: `buys_fill_nulls_with_0` is already the auto-generated name for that measure, so my metric was colliding with it, which is where the `_1` suffixes were coming from. With the rename the snapshots read cleanly again.

The other two failures are dealt with too. black wanted one change in `rewriting_sub_query_reducer.py`, though I should flag that the assert it reformatted isn't from this PR: the hook runs over the whole file once the file is touched, so it picked up something that was already there. Happy to pull that out into its own commit or drop it if you would rather keep this diff to my own lines.

The two snapshot failures were the semantic-graph sets picking up the new metric, and those are regenerated, along with the DuckDB conversion plans that the rename touched.

One thing still outstanding from my side: I can only regenerate DuckDB and the engine-agnostic snapshots from a fork, so BigQuery, Snowflake, Redshift, Databricks and Trino still need a run. Let me know how you would like that handled.

plypaul:
Since we have access to all SQL engines, it might be easiest if we took care of updating the snapshots for engines other than DuckDB or Postgres.

For now, could you update the snapshots for DuckDB and Postgres and add them to the PR since the CI checks depend on them? You can do that with the following commands (recently updated):

```
make regenerate-test-snapshots ENGINE=duckdb
make regenerate-test-snapshots ENGINE=postgres
```

A local postgres instance can be started with `make postgres`.

arthi-arumugam-git:
Snapshots are in, `85d6a3a4`, DuckDB and Postgres both. Locally that's 1158 passed and 51 skipped
on dev-env, 1052 passed and 157 skipped on postgres-env.

One thing worth flagging on the commands. My branch is five commits behind `db501c34a`, which is
the change that added the `--engine` flag, so `make regenerate-test-snapshots ENGINE=duckdb` fails
on it with `KeyError: 'duck_db'`. I didn't want to fold a rebase into a snapshot commit, so I drove
`scripts.generate_snapshots.run_tests()` directly with the same config the newer script builds. The
pytest invocations come out identical.

Most of the diff is just the rename. 27 of the 28 modified files change nothing except
`__buys_fill_nulls_with_0` to `__buys_null_filled`. Two are not:

- `test_list_metrics__result0.txt` gains one line and removes none. The committed file had no
  `buys` entry at all, so that's the fixture metric this PR adds showing up for the first time.
- One new file, the Postgres snapshot for the regression test. It only had a DuckDB one before.
  It's byte-identical to the DuckDB sibling and reads `0.5` rather than `0.75`, so the fix holds
  on Postgres too.

Separately, and not from this PR:
`..._join_to_timespine_and_fill_nulls_with_0__plan0_optimized.sql` was already stale. It was last
touched in #1973 and neither of my commits regenerated it, so that one predates me.

The workflow runs on the new commit are sitting at `action_required`, so nothing has executed yet.
Could you kick them off when you get a chance?

plypaul (approved):

Issue #2122, opened 2026-08-27
OSI converter: read/write datatype for Ossie round-trip fidelity


QMalcolm:
## Context
[PRD-330: datatypes from core to MF to support Ossie](https://app.notion.com/p/3acbb38ebda780979fb3fcfb9102c1d0)

Depends on #2121 landing the `datatype` field on `metricflow-semantic-interfaces`.

Today, `OSIToMSIConverter` (`metricflow/converters/osi_to_msi.py`) collapses Ossie's specific `datatype` into one of two `DimensionType` buckets:
- `DATE`, `TIME`, `DATE_TIME`, `DATE_TIME_TZ` → `DimensionType.TIME`
- `INTEGER`, `STRING` → `DimensionType.CATEGORICAL`

`MSIToOSIConverter` (`metricflow/converters/msi_to_osi.py`) has no type information to draw from at all, so a `DimensionType.CATEGORICAL` dimension has no principled way to become `string` vs. `integer` vs. `boolean` in Ossie output.

## Proposal
- `osi_to_msi.py`: in addition to computing the `DimensionType` bucket, populate the new `datatype` field on `PydanticDimension` directly from Ossie's `Field.datatype`.
- `msi_to_osi.py`: when `datatype` is present on the source `Dimension`, write it straight to Ossie's `Field.datatype` (and `Metric.datatype`, pending the metric-placement decision in #2121) instead of guessing from `DimensionType`. Fall back to the existing heuristic when `datatype` is unset, for backward compatibility with hand-authored MetricFlow manifests.

## Success criteria (from PRD)
- An Ossie `Field`/`Metric.datatype` value survives an Ossie → MetricFlow → Ossie round trip unchanged.
- No behavior change for manifests that don't set `datatype`.

PR #2123, opened 2026-08-28
Add optional datatype field to Dimension and Metric

### LINKED ISSUE #2121: Add optional datatype field to Dimension (and Metric) in metricflow-semantic-interfaces
## Context
[PRD-330: datatypes from core to MF to support Ossie](https://app.notion.com/p/3acbb38ebda780979fb3fcfb9102c1d0)

Apache Ossie's `Field`/`Metric` spec carries a closed `datatype` enum (`string`, `integer`, `float`, `decimal`, `boolean`, `date`, `time`, `datetime`, `timetz`, `opaque`), added in [ossie#113](https://github.com/apache/ossie/pull/113). `metricflow-semantic-interfaces` only exposes a coarse `DimensionType` (`CATEGORICAL`, `TIME`), so Ossie → MetricFlow conversion is lossy, and MetricFlow → Ossie has no principled way to pick a concrete type back.

This is phase 1 of the PRD's three-phase rollout. This issue covers the spec addition and parsing support only — it does **not** cover populating the field (phase 2, lives in dbt-core/Fusion) or the OSI converter changes (phase 3, tracked separately).

## Proposal
- Add an optional `datatype` field to `Dimension` (protocol in `metricflow_semantic_interfaces/protocols/dimension.py`, implementation in `metricflow_semantic_interfaces/implementations/elements/dimension.py`), defaulting to `None`.
- Field is purely a passthrough annotation — unused by MetricFlow's own validation/query logic, no change to how `DimensionType` drives categorical/time semantics.
- Add YAML parsing + JSON schema support (`metricflow_semantic_interfaces/parsing/generated_json_schemas/`) so the field round-trips through manifest parsing without behavior change when absent.
- Add test coverage confirming existing manifests that don't set `datatype` are completely unaffected.

## Open questions (from PRD, need resolution before/during implementation)
- What type should `datatype` use — mirror Ossie's enum directly, or define an independent, implementation-neutral enum other consumers could also use? (Risk: enum drift if Ossie's enum changes independently, e.g. adds `geometry`.)
- Does `datatype` belong on `Dimension` only, or also on `Metric` (Ossie's `Metric.datatype` needs a home too)?

## Success criteria
- New field parses correctly from YAML and defaults to `None`.
- No behavior change for existing semantic manifests that don't set it.
- `DimensionType` itself and its internal usage are unchanged.

QMalcolm:
## Summary
Add new metadata only datatypes to dimensions and metrics to support round-tripping of data types with Ossie. The new datatypes are optional.

- New `DataType` enum (`string`, `integer`, `decimal`, `float`, `boolean`, `date`, `time`, `datetime`, `datetime_tz`, `opaque`) — defined independently from Ossie's own `DataType` enum (see decision log in the first commit) so MSI's public contract doesn't inherit Ossie's naming/schema evolution.
- `datatype: Optional[DataType] = None` added to `Dimension` and `Metric` protocols + Pydantic implementations, plus matching JSON schema / parsing support.
- Field is unused by MetricFlow's own validation/query logic — purely a passthrough annotation. No behavior change for existing manifests that don't set it.

Converter changes to actually populate/consume this field during OSI ↔ MSI conversion are tracked separately in #2122.

Closes #2121

### DIFF
--- a/metricflow_semantic_interfaces/type_enums/data_type.py
+++ b/metricflow_semantic_interfaces/type_enums/data_type.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
-from metricflow_semantic_interfaces.enum_extension import ExtendedEnum
+from enum import Enum
 
 
-class DataType(ExtendedEnum):
+class DataType(Enum):
     """Logical data type of a dimension or metric value, independent of physical representation.
 
     This is a passthrough annotation, not used by MetricFlow's own validation or query logic. It exists so
--- a/metricflow_semantic_interfaces/type_enums/data_type.py
+++ b/metricflow_semantic_interfaces/type_enums/data_type.py
@@ -1,0 +1,23 @@
+from __future__ import annotations
+
+from metricflow_semantic_interfaces.enum_extension import ExtendedEnum
+
+
+class DataType(ExtendedEnum):
+    """Logical data type of a dimension or metric value, independent of physical representation.
+
+    This is a passthrough annotation, not used by MetricFlow's own validation or query logic. It exists so
+    that specific data types (e.g. from Ossie's `datatype` field) can be preserved through conversions without
+    being collapsed into `DimensionType`'s coarse CATEGORICAL/TIME buckets.
+    """
+
+    STRING = "string"
+    INTEGER = "integer"
+    DECIMAL = "decimal"
+    FLOAT = "float"
+    BOOLEAN = "boolean"
+    DATE = "date"
+    TIME = "time"
+    DATETIME = "datetime"
+    DATETIME_TZ = "datetime_tz"
+    OPAQUE = "opaque"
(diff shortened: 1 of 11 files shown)


tests_metricflow_semantic_interfaces/parsing/test_metric_parsing.py
```diff
@@ -112,6 +113,47 @@ def test_base_metric_parsing() -> None:
...
+
+def test_metric_without_datatype_parsing() -> None:
+    """Test that a metric's datatype defaults to None when not specified, for backward compatibility."""
+    yaml_contents = textwrap.dedent(
+        """\
+        metric:
+          name: base_test
+          type: simple
+          type_params:
+            measure:
```
plypaul: Since measures have been deprecated, mind updating this test case?

metricflow_semantic_interfaces/type_enums/data_type.py
```diff
@@ -0,0 +1,23 @@
+from __future__ import annotations
+
+from metricflow_semantic_interfaces.enum_extension import ExtendedEnum
+
+
+class DataType(ExtendedEnum):
```
plypaul: I was looking to deprecate `ExtendedEnum` - can this be a regular `Enum`?

plypaul (approved):
LGTM aside from minor comments.

PR #2124, opened 2026-08-31
Add experimental ClickHouse SQL engine support

felipempleite:
## Summary

Continuation of [#2014](https://github.com/dbt-labs/metricflow/pull/2014) — the existing maintainer thread for adding ClickHouse as a MetricFlow SQL engine. This is not a rewrite: Athena on `main` is kept, and the renderer / optimizer-guard design from #2014 is the base.

Adds **experimental** ClickHouse support (same bar Paul asked for on #2014). Generated SQL is meant to be correct against typical ClickHouse tables; hosted Semantic Layer / Fusion and `FINAL` for ReplacingMergeTree / CDC are out of scope.

## Why the compiled SQL is correct

ClickHouse fills unmatched LEFT/FULL OUTER JOIN cells with type defaults (`0`, `''`) unless `join_use_nulls = 1`. MetricFlow fill-nulls and ratio metrics assume SQL NULL, so unmatched time-spine days would otherwise look like real zeros. The contract is query-level `SETTINGS join_use_nulls = 1` on compiled SQL — not session `SET` (HTTP often has no session; readonly ClickHouse Cloud users can apply query SETTINGS but not `SET`).

`CAST(NULL AS Float64)` raises `CANNOT_INSERT_NULL_IN_ORDINARY_COLUMN`, so CAST targets are `Nullable(Float64)` / `Nullable(DateTime64(3))`. Session `data_type_default_nullable` does not apply to explicit CAST.

ClickHouse resolves unqualified WHERE names to SELECT aliases ([ClickHouse#23194](https://github.com/ClickHouse/ClickHouse/issues/23194)). The reducer uses the `SqlStringExpression` guard Paul asked for on #2014 (`has_ambiguous_alias_resolution`), so hoisting cannot turn an unqualified string expression into `ILLEGAL_AGGREGATION` or a silent wrong result.

## What changed on top of #2014

- The ClickHouse renderer appends `SETTINGS join_use_nulls = 1` to the complete statement returned by `DefaultSqlPlanRenderer`. This keeps the engine requirement explicit without a custom SQL scanner or adapter-level `SET`.
- ClickHouse CI job on `clickhouse/clickhouse-server:26.3`, same label gate as Trino (`Run Tests With Other SQL Engines`).
- Percentiles: `CONTINUOUS` renders as `quantileExactInclusive` (R-7 interpolation, the `PERCENTILE_CONT` definition) and `APPROXIMATE_CONTINUOUS` as `quantile`. `DISCRETE` / `APPROXIMATE_DISCRETE` are not advertised (same pattern as Trino/Vertica): `quantileExactLow` picks a different element than `PERCENTILE_DISC` whenever p*n is integral, and `quantileTDigest` returns interpolated non-member floats at scale.
- Rebased onto current `main`, retaining the newer Vertica and Athena engine behavior and snapshots.
- Pinned `clickhouse-connect[sqlalchemy]>=0.8.0,<1.0.0` and `dbt-clickhouse>=1.8,<2`.
- Changelog and experimental notes in `local-data-warehouses/README.md`.
- Tests that unmatched LEFT JOIN is `0` without SETTINGS and SQL NULL with renderer SETTINGS; adapter dry-run uses `EXPLAIN QUERY TREE` for SELECT and `EXPLAIN SYNTAX` for DDL.
- Small edits to shared `itest_*.yaml` check queries, all semantics-preserving and valid on every engine: identifier case made consistent (`T.ds` -> `t.ds`; ClickHouse identifiers are case-sensitive) and a few inner aliases de-shadowed (`ds` -> `ds__month` / `ds__day` / `ds__min`) where an inner SELECT alias had the same name as the joined column it was compared to, which ClickHouse resolves to the alias ([ClickHouse#23194](https://github.com/ClickHouse/ClickHouse/issues/23194)).
- `test_reduce_sub_query` is engine-parametrized (`sql_engine_snapshot`) so the ambiguous-alias guard is exercised per engine; that is why one snapshot pair per engine appears in the diff.
- Configured-case check queries get the same `SETTINGS join_use_nulls = 1` on ClickHouse. They are hand-written ANSI SQL, so without it their outer joins return `0` where MetricFlow returns NULL; the test client itself stays on engine defaults.

## Validation

Validated after rebasing onto current `main`:

- Full ClickHouse MetricFlow suite: **1106 passed, 157 skipped**.
- Focused ClickHouse renderer tests: **29 passed**.
- Live ClickHouse `join_use_nulls` coverage: **5 passed**.
- Representative configured-case integration coverage: **20 passed**.
- dbt adapter coverage: **5 passed**.
- Full root and `dbt-metricflow` pre-commit suites pass, including Ruff, Black, mypy, actionlint, and repository hygiene hooks.

The ClickHouse GitHub job uses the `Run Tests With Other SQL Engines` label gate. A previous run exposed that configured-case check queries also require `join_use_nulls`; without it, fill-nulls / join-to-timespine comparisons produce `0` instead of NULL and the harness spends minutes formatting each large diff. Applying the setting to those check queries made the job pass in under three minutes on the fork.

The 22 Trino conversion-metric snapshot failures seen in the earlier CI run are pre-existing: the identical set failed on upstream `main` in a comparison run.

Original engine work by [@ubiquitousbyte](https://github.com/ubiquitousbyte). D107 noqa placement from [@roarking](https://github.com/roarking) (re-authored onto this fork so CLA is a single GitHub identity; credit unchanged). Kustomer ran #2014 against ClickHouse Cloud with matching values.

Please apply **`Run Tests With Other SQL Engines`** when you want the ClickHouse service job; it does not run on every push.


felipempleite:
@cla-bot check

felipempleite:
Hi @plypaul! 

This is a continuation of #2014 . It's my first-time fork, so GitHub is waiting on Approve and run workflows. Could you trigger those when you have a moment? After DuckDB/lint are green, the ClickHouse job is the Trino-style service you asked for on #2014; it needs the Run Tests With Other SQL Engines label.

plypaul:
@felipempleite I just kicked off the CI run and saw that there were some failures.

felipempleite:
Thanks @plypaul, I addressed the failures that showed up.

The 22 Trino conversion-metric snapshot failures reproduce identically on main (fork run), so they look pre-existing rather than from this branch.

Details are in the "CI notes" section of the description. Could you please help run it again here? It should go green apart from the Trino ones.

plypaul:
One thing that I noticed was that `metricflow/sql/render/clickhouse.py` is a bit complicated and potentially brittle with the regex matches. I'm thinking about some alternative solutions.

felipempleite:
> One thing that I noticed was that `metricflow/sql/render/clickhouse.py` is a bit complicated and potentially brittle with the regex matches. I'm thinking about some alternative solutions.

Thanks @plypaul! I removed the regex/scanner in [ed5b6bcfb](https://github.com/dbt-labs/metricflow/commit/ed5b6bcfbdbdb6d4e3d32eafdac925b87cdeb151). The ClickHouse renderer now appends SETTINGS join_use_nulls = 1 after the complete generated statement. The focused tests, full ClickHouse suite, and dbt adapter tests pass locally.

allan-simon:
awesome work, is there any way one can help to get the PR merged ? :)

PR #2125, opened 2026-09-01
OSI converter: read/write datatype for Ossie round-trip fidelity

### LINKED ISSUE #2122: OSI converter: read/write datatype for Ossie round-trip fidelity

QMalcolm:
## Summary
Phase 3 (converter) of [PRD-330](https://app.notion.com/p/3acbb38ebda780979fb3fcfb9102c1d0), building on the `datatype` field added to `Dimension`/`Metric` in #2121.

- `OSIToMSIConverter` previously collapsed Ossie's specific `datatype` into one of `DimensionType`'s two buckets (TIME/CATEGORICAL). `MSIToOSIConverter` had no type information to draw from at all when writing a dimension back out, so a specific Ossie type could never survive a round trip.
- Added `OSIDataType` and optional `datatype` fields to `OSIField`/`OSIMetric` (the local, hand-maintained OSI wire-format models in `metricflow/converters/models.py`, which didn't expose the field at all yet).
- Centralized the OSI ↔ MSI value mapping in a new `metricflow/converters/datatype_mapping.py` — a single source of truth, since the PRD itself flags enum drift between the two type systems as an ongoing maintenance risk.
- Both conversion directions now read/write `datatype` on dimensions and metrics. A RATIO metric's auto-generated numerator/denominator sub-metrics intentionally do *not* inherit the parent's `datatype` (mirroring existing `description` handling for those synthetic sub-metrics).
- Entities and measures are untouched — MSI's `datatype` field only exists on `Dimension`/`Metric` (per #2121), so there's nothing to convert for those element types.

Closes #2122

### DIFF
--- a/.changes/unreleased/Features-20260901-111954.yaml
+++ b/.changes/unreleased/Features-20260901-111954.yaml
@@ -1,0 +1,8 @@
+kind: Features
+body: The Ossie converter now reads and writes the `datatype` field on `Field`/`Metric`,
+  preserving Ossie's specific data type through an Ossie -> MetricFlow -> Ossie round
+  trip instead of collapsing it into the coarse DimensionType bucket
+time: 2026-09-01T11:19:54.741523-05:00
+custom:
+  Author: QMalcolm
+  Issue: "2122"
--- a/metricflow/converters/datatype_mapping.py
+++ b/metricflow/converters/datatype_mapping.py
@@ -1,0 +1,32 @@
+from __future__ import annotations
+
+from typing import Dict, Optional
+
+from metricflow.converters.models import OSIDataType
+from metricflow_semantic_interfaces.type_enums import DataType
+
+# Single source of truth for the OSI <-> MSI datatype mapping. The two enums are semantically
+# equivalent 1:1 today; if Ossie's DataType enum changes (e.g. gains a new value), update here.
+_OSI_TO_MSI_DATATYPE: Dict[OSIDataType, DataType] = {
+    OSIDataType.STRING: DataType.STRING,
+    OSIDataType.INTEGER: DataType.INTEGER,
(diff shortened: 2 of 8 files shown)


plypaul (approved):
LGTM, but as a high-level comment for later, can the tests be simplified via a snapshot?

PR #2132, opened 2026-09-08
Fix blank error from mf list dimension-values on empty result

### LINKED ISSUE #685: [Bug] Empty error message when dimension-values returns an empty set

### Current Behavior

When running tests against databricks I hit an issue where `mf list dimension-values` was attempting to query an empty table, and subsequently failing with a blank error message. Looking into the logs the issue is an assertion in the CLI that dim_vals is not falsey, but that means an empty return set will raise a blank error.

### Expected Behavior

The CLI should return a useful message in this case.

### Steps To Reproduce

1. Load an empty dataset where dimensions are supposed to live
2. Run `mf list dimension-values` targeting a dimension defined against that empty dataset

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt:
- metricflow:
```

_No response_

### Additional Context

_No response_

beratli:
## Summary

`mf list dimension-values` asserts `assert dim_vals` after querying. When a dimension has **no values** (e.g. it's defined over an empty table), `get_dimension_values` returns an empty list, and `assert []` raises an `AssertionError` with no message — so the CLI fails with a blank error instead of telling the user what happened.

## Change

Assert only that the result is not `None` (`assert dim_vals is not None`, which still narrows the `Optional` for type-checking) so an empty result falls through to the normal success path and reports `🌱 We've found 0 dimension values ...`.

## Testing

- `ruff` / `black` / `mypy` clean on the changed file.

Note: the empty-result path requires a dimension over an empty warehouse table, which the existing isolated CLI test harness (which builds a real dbt project) isn't set up to produce, so this is not covered by an automated test.

Fixes #685

### DIFF
--- a/.changes/unreleased/Fixes-20260908-140000.yaml
+++ b/.changes/unreleased/Fixes-20260908-140000.yaml
@@ -1,6 +1,5 @@
 kind: Fixes
-body: Fix `mf list dimension-values` failing with a blank error when a dimension has
-  no values.
+body: '`mf list dimension-values` no longer errors when a dimension has no values.'
 time: 2026-09-08T14:00:00.000000+00:00
 custom:
     Author: beratli
--- a/.changes/unreleased/Fixes-20260908-140000.yaml
+++ b/.changes/unreleased/Fixes-20260908-140000.yaml
@@ -1,0 +1,7 @@
+kind: Fixes
+body: '`mf list dimension-values` no longer fails with a blank error message when a
+  dimension has no values; it now reports that zero values were found.'
+time: 2026-09-08T14:00:00.000000+00:00
+custom:
+    Author: beratli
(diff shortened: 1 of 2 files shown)


.changes/unreleased/Fixes-20260908-140000.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Fixes
+body: '`mf list dimension-values` no longer fails with a blank error message when a
```
plypaul: Nit: Since this goes in the changelog, how about something more concise?

beratli: Good call — shortened it. Thanks!

plypaul (approved):
Thanks @beratli !

